PR #23, opened 2021-10-04
Add onnxruntime type and update readme

echarlaix:


src/optimum/onnxruntime/optimize_model.py
```diff
@@ -30,34 +34,53 @@ def parser_optimize(parser=None):
         type=Path,
         help="Input ONNX model path.",
     )
+    parser.add_argument(
+        "--opt_level",
+        type=int,
+        choices=[0, 1, 2, 99],
+        default=0,
+        help="ONNX Runtime optimization level.",
```
mfuntowicz, line 42: I would describe slightly more what optimization levels mean. Especially because setting `opt_level > 1` will generate hardware dependent computation graphs.

mfuntowicz (changes requested):

PR #25, opened 2021-10-08
Support Intel neural_compressor v1.7

PenghuiCheng:
# What does this PR do?
* Support Intel Neural_compressor v1.7
* Support to pass configure class to quantization of Intel Neural_compressor
* Add IncOptimizedConfig class for quantized configure of the model.


README.md
```diff
@@ -1,5 +1,5 @@
 [![ONNX Runtime](https://github.com/huggingface/optimum/actions/workflows/test-onnxruntime.yml/badge.svg)](https://github.com/huggingface/optimum/actions/workflows/test-onnxruntime.yml)
-[![LPOT](https://github.com/huggingface/optimum/actions/workflows/test-intel.yml/badge.svg)](https://github.com/huggingface/optimum/actions/workflows/test-intel.yml)
+[![neural_compressor](https://github.com/huggingface/optimum/actions/workflows/test-intel.yml/badge.svg)](https://github.com/huggingface/optimum/actions/workflows/test-intel.yml)
```
hshen14, line 2: How about using full name "Intel Neural Compressor"?

README.md
```diff
@@ -28,13 +28,13 @@ Convert a hub model:
 Optimize a hub model:
 `optimum_optimize` 
 
-Apply LPOT quantization:
+Apply neural_compressor quantization:
```
hshen14, line 31: How about using "Apply quantization from Intel Neural Compressor (INC)"?

src/optimum/intel/neural_compressor/config.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(
+            self,
+            config_path: str,
+    ):
+        """
+        Args:
+            config_path (:obj:`str`):
+                Path to the YAML configuration file used to control the tuning behavior.
+        Returns:
+            config: neural_compressorConfig object.
```
echarlaix, line 35: config: IncConfig object.

src/optimum/intel/neural_compressor/config.py
```diff
@@ -0,0 +1,161 @@
...
+    @classmethod
+    def from_pretrained(
+            cls,
+            config_name_or_path: Union[str, os.PathLike],
+            config_file_name: str,
+            cache_dir: Optional[Union[str, os.PathLike]] = None,
+            **config_kwargs
+    ):
+        """
+        Instantiate a neural_compressorConfig object from a configuration file which can either be hosted on
```
echarlaix, line 60: Instantiate an IncConfig object from a configuration file which can either be hosted on huggingface.co or from a local directory path.

examples/pytorch/text-classification/run.sh
```diff
@@ -0,0 +1,66 @@
...
+        --per_device_train_batch_size=${batch_size} \
+        --pad_to_max_length \
+        --do_eval \
+        --provider inc \
+        --quantization_approach dynamic \
+        --int8 \
+        --quantize \
+
+}
+
```
echarlaix, line 65: Not sure about the added value of this bash script, I think it would be best to not include it in the main branch

examples/pytorch/text-classification/run_glue.py
```diff
@@ -503,23 +513,32 @@ def take_train_steps(model, trainer, resume_from_checkpoint, last_checkpoint):
...
 
+    if model_args.benchmark:
+        from optimum.intel.neural_compressor import IncOptimizedConfig
+        if model_args.int8:
+            print("runing int8 mode...")
+            optimized_config = IncOptimizedConfig.from_pretrained(training_args.output_dir)
+            from optimum.intel.neural_compressor.utils import load_inc_config
+            model = load_inc_config(model, optimized_config.path)
+
+        metric = take_eval_steps(model, trainer, metric_name)
```
echarlaix, line 524: What do think about replacing this part with a loading verification at the end of the script ? (such as https://github.com/huggingface/optimum/blob/add-lpot-loading-quant-model/examples/pytorch/text-classification/run_glue.py#L582 ) Otherwise the behaviour is not very clear as the user should provide a directory with a quantized model / optimized config.

src/optimum/intel/neural_compressor/config.py
```diff
@@ -0,0 +1,161 @@
...
+                )
+            except requests.exceptions.RequestException:
+                raise ValueError(f"{config_file_name} NOT FOUND in HuggingFace Hub")
+
+        config = cls(config_file, **config_kwargs)
+        return config
+
+
+class IncOptimizedConfig:
+
```
echarlaix, line 99: Not sure about the added value of IncOptimizedConfig in comparison to IncConfig as the user might want to save the quantized model or config file with a different name for example

echarlaix (changes requested):

examples/pytorch/text-classification/run_glue.py
hshen14: @echarlaix Thanks for your quick comments. The idea here is to introduce a simply simulated deployment environment to run the inference/benchmark. So we assumed people will have a directory with a quantized model/optimized config. Your proposed change is also making a lot of sense, especially for the result checker. Some general questions: does the example under Optimum focus more on the functionality of model quantization? Where do we plan to verify the performance or end-to-end flow? I am thinking whether we may bring some end-to-end examples in transformer repo (link to Optimum) to give people overall idea?

src/optimum/intel/neural_compressor/config.py
echarlaix: Yes I see your point, it would be good to bring this discussion during our meeting you are right

PenghuiCheng: removed

PR #27, opened 2021-10-08
Add LPOT quantized model loading 

echarlaix:


src/optimum/intel/lpot/quantization.py
```diff
@@ -199,11 +203,175 @@ class LpotQuantizerForSequenceClassification(LpotQuantizer):
...
+                raise ValueError("Need batch_size and sequence_length for tracing the model with torch fx.")
+
+            model = symbolic_trace(
+                model,
+                input_names=["input_ids", "attention_mask", "token_type_ids", "labels"],
+                batch_size=batch_size,
+                sequence_length=sequence_length
+            )
+
+        q_model = apply_quantization_from_config(deploy_config.config, model)
```
PenghuiCheng, line 338: How about use neural_compressor.utils.pytorch.load here?

src/optimum/intel/neural_compressor/config.py
```diff
@@ -93,3 +92,28 @@ def from_pretrained(
...
+        self.path = config_path
+        self.config = self._read_config()
+        self.usr_cfg = self.config
+
+    def _read_config(self):
+        with open(self.path, 'r') as f:
+            try:
+                config = yaml.load(f, Loader=yaml.Loader)
+            except yaml.YAMLError as exc:
+                print(exc)
```
michaelbenayoun, line 117: Why you are not using the logger as in run_glue?

echarlaix: Good catch, thanks !

src/optimum/intel/neural_compressor/quantization.py
```diff
@@ -203,11 +204,175 @@ class IncQuantizerForSequenceClassification(IncQuantizer):
...
+
+        if deploy_config.get_config("framework") == "pytorch_fx":
+            from transformers.utils.fx import symbolic_trace
+
+            if batch_size is None or sequence_length is None:
+                raise ValueError("Need batch_size and sequence_length for tracing the model with torch fx.")
+
+            model = symbolic_trace(
+                model,
+                input_names=["input_ids", "attention_mask", "token_type_ids", "labels"],
```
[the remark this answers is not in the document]
xin3he: Sorry for confusing you. What I meant is that the model passed into IncQuantizedModel class maybe a torch.fx.GraphModule object. No need for symbolic_trace again.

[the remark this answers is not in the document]
echarlaix: The model is created inside IncQuantizedModel using the from_pretrained method from the AutoModel classes as you can see [here](https://github.com/huggingface/optimum/blob/add-lpot-loading-quant-model/src/optimum/intel/neural_compressor/quantization.py#L361), before tracing the latter is thus not a torch.fx.GraphModule object

PR #33, opened 2021-10-22
Add question answering

echarlaix:


src/optimum/intel/neural_compressor/utils.py
```diff
@@ -97,3 +102,61 @@ def _get_quantizable_ops_recursively(
...
+    try:
+        gm = deepcopy_graph(gm_original)
+        gm.eval()
+    except Exception as e:
+        gm = gm_original
+        logger.warning(f"Deepcopy failed: {repr(e)}, model is modified inplace.")
+
+    graph = gm.graph
+    output_node = None
+    for node in graph.nodes:
```
michaelbenayoun, line 130: I think you can simply retrieve the last node of the graph.nodes as it is topologically sorted (needs to be checked).

src/optimum/intel/neural_compressor/utils.py
```diff
@@ -97,3 +102,61 @@ def _get_quantizable_ops_recursively(
...
+        if node.name == "output":
+            output_node = node
+            break
+
+    for node in graph.nodes:
+        if node.op == "placeholder" and node.target in inputs_to_remove:
+
+            def remove_users(node, output_node):
```
michaelbenayoun, line 138: I think you should define this function oustide of the loop.

src/optimum/intel/neural_compressor/utils.py
```diff
@@ -97,3 +102,61 @@ def _get_quantizable_ops_recursively(
...
+                output_args = output_node.args[0] if output_node is not None else None
+
+                for user in list(node.users):
```
michaelbenayoun, line 141: Do you need to make node.users a list?

michaelbenayoun (approved):

src/optimum/intel/neural_compressor/utils.py
echarlaix: This is done in order to make a copy of the keys of the dictionnary node.users as the latter is being modified

PR #58, opened 2021-12-21
Add ORTQuantizer handling ONNX Runtime static and dynamic quantization 

echarlaix:
In this PR, we introduce ORTQuantizer, a class which integrates the ONNX Runtime post-training static and dynamic quantization approaches.


michaelbenayoun (approved):

optimum/onnxruntime/quantization.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def get_next(self):
+        return next(self._iter, None)
+
+
+class ORTQuantizer:
+    def __init__(
+        self,
+        model_name_or_path: str,
+        output_dir: Union[str, Path],
```
mfuntowicz, line 71: You can use `os.PathLike` to be more generic 🙂

optimum/onnxruntime/quantization.py
```diff
@@ -0,0 +1,262 @@
...
+        self._iter = iter([{key: data[key].numpy() for key in data} for data in calib_dataloader])
+
+    def get_next(self):
+        return next(self._iter, None)
+
+
+class ORTQuantizer:
+    def __init__(
+        self,
+        model_name_or_path: str,
```
mfuntowicz, line 70: What about `Union[str, os.PathLike]` here? people will get the sense this two paths are distincts

optimum/onnxruntime/quantization.py
```diff
@@ -0,0 +1,262 @@
...
+        config_kwargs_default = [
+            ("cache_dir", None),
+            ("force_download", False),
+            ("resume_download", False),
+            ("revision", None),
+        ]
+        config_kwargs = {name: kwargs.get(name, default_value) for (name, default_value) in config_kwargs_default}
+        self.cache_dir = config_kwargs.get("cache_dir")
+        self.model_name_or_path = model_name_or_path
+        self.output_dir = output_dir if isinstance(output_dir, Path) else Path(output_dir)
```
mfuntowicz, line 128: Open question: Do we want to put the `output_dir` as part of the instance attributes?

We can expose this parameter on methods, to reuse the same quantizer, without being dependent on the model.

For instance: Let's imagine I have a quantization config, I would like to recover the quantizer, potentially with a different output path for the generated model.

optimum/onnxruntime/quantization.py
```diff
@@ -0,0 +1,262 @@
...
+        opset = self.onnx_config.default_onnx_opset if self.ort_config.opset is None else self.ort_config.opset
+        _ = export(self.tokenizer, self.model, self.onnx_config, opset, self.model_path)
+
+    def fit(self) -> None:
+        """
+        Load and export a model to an ONNX Intermediate Representation (IR) after applying the specified quantization
+        approach.
+        """
+        self.export()
+        if self.quantization_approach == ORTQuantizationMode.DYNAMIC.value:
```
mfuntowicz, line 166: I would suggest you rely on enum comparison rather than string comparison (`.value`).

You can do: 

```python
self.quantization_approach = ORTQuantizationMode(quantization_approach)
...
if self.quantization_approach == ORTQuantizationMode.DYNAMIC
```

optimum/onnxruntime/quantization.py
```diff
@@ -0,0 +1,262 @@
...
+
+        if self.ort_config.max_samples is not None and len(calib_dataset) > self.ort_config.max_samples:
+            calib_dataset = calib_dataset.select(range(self.ort_config.max_samples))
+
+        ignored_columns = list(set(calib_dataset.column_names) - set(self.onnx_config.inputs.keys()))
+        calib_dataset = calib_dataset.remove_columns(ignored_columns)
+
+        generator = torch.Generator()
+        generator.manual_seed(int(torch.empty((), dtype=torch.int64).random_().item()))
+        sampler = RandomSampler(calib_dataset, generator=generator)
```
mfuntowicz, line 245: Let's expose the seed as an optional parameter in the method, this would be expected by users to be able to reproduce quantization experiment.

Let's potentially also store the seed within the serialized config 👍🏻

mfuntowicz:
Looks great!! 🔥 

Added few UX comments, good job!

optimum/onnxruntime/quantization.py
echarlaix: It's actually a good point as I feel it may be a more intuitive usage

optimum/onnxruntime/quantization.py
echarlaix: Good point !

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+        "bart": {"num_heads":"encoder_attention_heads", "hidden_size":"d_model"},
+        "gpt2": {"num_heads":"n_head", "hidden_size":"n_embd"},
+        }
+
+    @staticmethod
+    def get_num_heads(model_type:str) -> str:
+        try:
+            return OnnxConfigManager.__conf[model_type]["num_heads"]
+        except KeyError:
+            print(f"{model_type} undefined in the configuration, please define it with `add_model` or it will be set to default value.")
```
echarlaix, line 62: Shouldn't it be `update_model` instead of  `add_model` ?

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+            __conf (:obj:`dict`):
+                Dictionary register the attribute names of models.
+        """
+    __conf = {
+        "bert": {"num_heads":"num_attention_heads", "hidden_size":"hidden_size"},
+        "distilbert": {"num_heads":"n_heads", "hidden_size":"hidden_size"},
+        "roberta": {"num_heads":"num_attention_heads", "hidden_size":"hidden_size"},
+        "bart": {"num_heads":"encoder_attention_heads", "hidden_size":"d_model"},
+        "gpt2": {"num_heads":"n_head", "hidden_size":"n_embd"},
+        }
```
echarlaix, line 55: As there is a redundancy in the dictionnaries keys ("num_heads" and "hidden_size"), we could potentially replace the dictionnaries with tuples

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+            ("disable_bias_gelu", False),
+            ("enable_gelu_approximation", False),
+            ("use_mask_index", False),
+            ("no_attention_mask", False),
+            ("disable_embed_layer_norm", True),
+        ]
+        fusion_config_kwargs = AttrDict({name: kwargs.get(name, default_value) for (name, default_value) in fusion_config_kwargs_default})
+        optimization_options = FusionOptions.parse(fusion_config_kwargs)
+
+        print()
```
echarlaix, line 196: Can be removed

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+                Repository name in the Hugging Face Hub or path to a local directory hosting the model.
+            ort_config (:obj:`Union[ORTConfig, str]`):
+                Configuration file containing all the information related to the model optimization and quantization.
+                Can be either:
+                    - an instance of the class :class:`ORTConfig`,
+                    - a string valid as input to :func:`ORTConfig.from_pretrained`.
+            feature (:obj:`str`):
+                Feature used when exporting the model.
+            data_files (:obj:`str`, `optional`):
+                Path to source data files.
```
echarlaix, line 110: Shouldn't `data_files` be removed ?

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+                identifier allowed by git.
+        """
+        config_kwargs_default = [
+            ("cache_dir", None),
+            ("force_download", False),
+            ("resume_download", False),
+            ("revision", None),
+        ]
+        config_kwargs = {name: kwargs.get(name, default_value) for (name, default_value) in config_kwargs_default}
+        self.cache_dir = config_kwargs.get("cache_dir")
```
echarlaix, line 132: `self.cache_dir` not used so can be removed

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+        Param:
+            output_dir (:obj:`Union[str, os.PathLike]`):
+                The output directory where the optimized model will be saved.
+        """
+        output_dir = output_dir if isinstance(output_dir, Path) else Path(output_dir)
+        self.onnx_model_path = output_dir.joinpath("model.onnx")
+        self.optim_model_path = generate_identified_filename(self.onnx_model_path, "-optimized")
+
+        self.export(self.onnx_model_path)
+        config = self.onnx_config._config
```
echarlaix, line 173: We could use `self.model.config` instead of `self.onnx_config._config`

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+            num_heads,
+            hidden_size,
+            opt_level=self.ort_config.opt_level,
+            optimization_options=optimization_options,
+            use_gpu=self.ort_config.use_gpu,
+            only_onnxruntime=self.ort_config.only_onnxruntime,
+        )
+
+        optimizer.save_model_to_file(self.optim_model_path.as_posix(), self.ort_config.use_external_data_format)
+        print(f"Optimized model saved to: {self.optim_model_path}")
```
echarlaix, line 209: We could use logging instead of print

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+        # Top 5 reduced operations & node details onnx model v.s. optimized model
+        if nodes_details:
+            op_types = []
+            for model in [onnx_model, optimizer]:
+                for node in model.nodes():
+                    if node.op_type not in op_types: op_types.append(node.op_type)
+                        
+            nodes_change = dict(map(lambda op_type: (op_type, get_node_change(op_type)), op_types))
+            sorted_nodes_change = sorted(nodes_change.items(), key=lambda op: abs(op[1]))
+            sorted_nodes_change.reverse()
```
echarlaix, line 272: `-abs(op[1]))` would allow to not perform the reverse operation

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+        self.onnx_config = model_onnx_config(self.model.config)
+        opset = self.onnx_config.default_onnx_opset if self.ort_config.opset is None else self.ort_config.opset
+        _ = export(self.tokenizer, self.model, self.onnx_config, opset, model_path)
+
+    def fit(self, output_dir: Union[str, os.PathLike], **kwargs) -> None:
+        """
+        Load and export a model to an ONNX Intermediate Representation (IR) and apply the graph optimization.
+
+        Param:
```
echarlaix, line 164: Could be nice to add docstring of args related to optimization (`disable_gelu`, `disable_layer_norm`, ...)

optimum/onnxruntime/optimization.py
JingyaHuang: Indeed!

optimum/onnxruntime/optimization.py
JingyaHuang: Good idea, it shall look lighter with tuples.

optimum/onnxruntime/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+        self.per_channel = kwargs.pop("per_channel", False)
+        self.reduce_range = kwargs.pop("reduce_range", False)
+        self.activation_type = kwargs.pop("activation_type", "uint8")
+        self.weight_type = kwargs.pop("weight_type", "uint8")
+        self.quant_format = kwargs.pop("quant_format", "operator")
+        self.calibration_method = kwargs.pop("calibration_method", "minmax")
+        self.split = kwargs.pop("split", "train")
+        self.max_samples = kwargs.pop("max_samples", 80)
+        self.calib_batch_size = kwargs.pop("calib_batch_size", 8)
+        self.seed = kwargs.pop("seed", 42)
```
michaelbenayoun, line 45: I would say that `seed` falls under general, but that is not really important.

optimum/onnxruntime/optimization.py
michaelbenayoun: I am not sure that tuples are better, it is clearer to read `OnnxConfigManager.__conf[model_type]["num_heads"]` than `OnnxConfigManager.__conf[model_type][0]` in my opinion,

optimum/onnxruntime/optimization.py
JingyaHuang: ✅

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+    def __init__(self, *args, **kwargs):
+        super(AttrDict, self).__init__(*args, **kwargs)
+        self.__dict__ = self
+
+
+class OnnxConfigManager:
+    """A class that notes down the attribute names(for `number of heads` and `hidden size`) of models in `transformers/models`. 
+    It is optional for BERT model, but for other model types, you need specify the name of these parameters. It is possible to add
+    customized model information with `update_model()` method.
+        Args:
```
michaelbenayoun, line 45: I think it should be "Attributes" here instead of "Args"

optimum/onnxruntime/optimization.py
JingyaHuang: ✅

optimum/onnxruntime/optimization.py
michaelbenayoun: How is it set to a default value if an exception is being raised?

optimum/onnxruntime/optimization.py
JingyaHuang: ✅

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+    def remove_model(model_type:str):
+        try:
+            OnnxConfigManager.__conf.pop(model_type)
+        except KeyError:
+            print(f"{model_type} undefined in the configuration")
+    
+    @staticmethod
+    def check_supported_model(model_type:str) -> bool:
+        has_model = model_type in OnnxConfigManager.__conf.keys()
+        return has_model
```
michaelbenayoun, line 87: You could directly do : 
```python
return model_type in OnnxConfigManager.__conf
```

optimum/onnxruntime/optimization.py
```diff
@@ -0,0 +1,274 @@
...
+        self.tokenizer = AutoTokenizer.from_pretrained(self.model_name_or_path)
+        self.model = FeaturesManager.get_model_from_feature(self.feature, self.model_name_or_path)
+        self.onnx_model_path = None
+        self.optim_model_path = None
+
+    def export(self, model_path: os.PathLike) -> None:
+        """
+        Load and export a model to an ONNX Intermediate Representation (IR).
+
+        Param:
```
michaelbenayoun, line 148: It should be `Args` instead of `Param`

tests/onnxruntime/test_onnxruntime.py
```diff
@@ -12,45 +12,114 @@
...
+                    )
+                    tokenizer = quantizer.tokenizer
+                    quantizer.fit(output_dir)
+                    validate_model_outputs(
+                        quantizer.onnx_config,
+                        tokenizer,
+                        quantizer.model,
+                        q8_model_path,
+                        list(quantizer.onnx_config.outputs.keys()),
+                        atol=12,
```
michaelbenayoun, line 101: Is such a big atol wanted?

tests/onnxruntime/test_onnxruntime.py
```diff
@@ -12,45 +12,114 @@
...
+    all_suite = unittest.TestSuite(suites_list)
+
+    runner = unittest.TextTestRunner()
+    runner.run(all_suite)
 
 
 if __name__ == "__main__":
-    unittest.main()
+    run(True, False)
+    # unittest.main()
```
michaelbenayoun, line 125: Should remove the comment.

michaelbenayoun (approved):
Great work!

optimum/onnxruntime/optimization.py
JingyaHuang: ✅

optimum/onnxruntime/optimization.py
JingyaHuang: ✅

optimum/onnxruntime/optimization.py
JingyaHuang: Exact

optimum/onnxruntime/optimization.py
JingyaHuang: Exactly, ✅

optimum/onnxruntime/optimization.py
JingyaHuang: > How is it set to a default value if an exception is being raised?

The default value will be "num_heads" which is the most frequent name used in the transformers model to represent the number of heads.

optimum/onnxruntime/optimization.py
JingyaHuang: Done!

mfuntowicz:
Can we merge this one? 😃

PR #45, opened 2021-11-12
Intel OpenVINO backend

dkurt:
# What does this PR do?

This PR introduces Intel OpenVINO backend for inferencing. Port of https://github.com/huggingface/transformers/pull/14203

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

PR #64, opened 2022-01-12
Add a reference to Optimum OpenVINO

dkurt:
Use standalone OpenVINO package alternatively to https://github.com/huggingface/optimum/pull/45


dkurt:
@echarlaix, @michaelbenayoun, please guide if this approach more applicable for adding new Optimum backend.

PR #67, opened 2022-01-17
Fix qa data mismatch

xin3he:
# What does this PR do?

This PR fixed `ValueError: Got [n] predictions and [n+m] features` of `run_qa.py`.
If `training_args.dataloader_drop_last` was set, the length of predictions may not match features, due to dataset didn't reduce last batch.


echarlaix (approved):
Well seen, thanks for the fix @xin3he !

PR #68, opened 2022-01-17
Added a few more quantization parameters required for qranium.

mfuntowicz:
This PR adds supports for explicitly setting nodes to quantization/nodes to exclude while exporting models through the `ORTQuantizer`

### DIFF
--- a/optimum/onnxruntime/configuration.py
+++ b/optimum/onnxruntime/configuration.py
@@ -47,6 +47,6 @@ class ORTConfig(BaseConfig):
         self.split = kwargs.pop("split", "train")
         self.max_samples = kwargs.pop("max_samples", 80)
         self.calib_batch_size = kwargs.pop("calib_batch_size", 8)
-        self.nodes_to_quantize = (kwargs.pop("nodes_to_quantize", []),)
-        self.nodes_to_exclude = (kwargs.pop("nodes_to_exclude", []),)
+        self.nodes_to_quantize = kwargs.pop("nodes_to_quantize", [])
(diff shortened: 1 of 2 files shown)


optimum/onnxruntime/configuration.py
```diff
@@ -43,3 +47,6 @@ def __init__(self, **kwargs):
         self.split = kwargs.pop("split", "train")
         self.max_samples = kwargs.pop("max_samples", 80)
         self.calib_batch_size = kwargs.pop("calib_batch_size", 8)
+        self.nodes_to_quantize = (kwargs.pop("nodes_to_quantize", []),)
+        self.nodes_to_exclude = (kwargs.pop("nodes_to_exclude", []),)
```
echarlaix, line 51: Could we keep `self.nodes_to_quantize` and `self.nodes_to_exclude` as lists ?

mfuntowicz: Thanks for catching, it's a mistake, sorry about that 🙏🏻

optimum/onnxruntime/configuration.py
```diff
@@ -23,14 +23,18 @@ class ORTConfig(BaseConfig):
     FULL_CONFIGURATION_FILE = "ort_config.json"
 
     def __init__(self, **kwargs):
+        super().__init__(**kwargs)
```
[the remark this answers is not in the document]
mfuntowicz: I would say it's the common way to initialize objects with inheritance.

I might be source of issue in the future if we want to initialize super class attribute, these would not be upstream to parent class.

echarlaix (approved):

PR #75, opened 2022-02-02
Add onnx runtime text classification, token classification and question answering examples

echarlaix:
In this PR we add the ONNX Runtime examples for static and dynamic quantization for the following tasks : text classification, token classification and question answering. We also introduce a temporary `ORTModel` class which allows to evaluate an ORT model, this is useful in order to compare the evaluation results of the original full-precision model with the resulting quantized model.

### DIFF
--- a/examples/onnxruntime/pytorch/question-answering/README.md
+++ b/examples/onnxruntime/pytorch/question-answering/README.md
@@ -1,0 +1,39 @@
+<!---
+Copyright 2020 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Question answering
+
+
+The script [`run_qa.py`](https://github.com/huggingface/optimum/blob/main/examples/onnxruntime/pytorch/question-answering/run_qa.py)
+allows us to apply different quantization approaches (such as dynamic and static quantization) as well as graph 
+optimizations using [ONNX Runtime](https://github.com/microsoft/onnxruntime) for question answering tasks.
+
+Note that if your dataset contains samples with no possible answers (like SQuAD version 2), you need to pass along 
+the flag `--version_2_with_negative`.
+
+The following example applies post-training static quantization on a DistilBERT fine-tuned on the SQuAD1.0 dataset.
+
+```bash
+python run_qa.py \
+    --model_name_or_path distilbert-base-uncased-distilled-squad \
+    --dataset_name squad \
+    --quantize \
+    --quantization_approach static \
+    --do_eval \
+    --output_dir /tmp/quantized_distilbert_squad
+```
+
+In order to apply dynamic or static quantization, `quantization_approach` must be set to  respectively `dynamic` or `static`.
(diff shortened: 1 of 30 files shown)


examples/onnxruntime/pytorch/question-answering/run_qa.py
```diff
@@ -0,0 +1,735 @@
...
+        quantizer.fit(
+            model_args.model_name_or_path,
+            output_dir,
+            feature="question-answering",
+            cache_dir=model_args.cache_dir,
+        )
+        onnx_config = quantizer.onnx_config
+        opt_model_path = output_dir.joinpath("model-quantized.onnx")
+
+    elif optim_args.optimize:
```
JingyaHuang, line 707: Is it possible to apply quantization as well as optimization?

examples/onnxruntime/pytorch/question-answering/README.md
```diff
@@ -0,0 +1,39 @@
+<!---
+Copyright 2020 The HuggingFace Team. All rights reserved.
```
JingyaHuang, line 2: Is it necessary to change the year of license copyright?

examples/onnxruntime/pytorch/question-answering/run_qa.py
```diff
@@ -0,0 +1,735 @@
...
+        )
+
+    # Preprocessing the datasets.
+    # Preprocessing is slighlty different for training and evaluation.
+    if training_args.do_train:
+        column_names = raw_datasets["train"].column_names
+    elif training_args.do_eval:
+        column_names = raw_datasets["validation"].column_names
+    else:
+        raise ValueError("--do_train or --do_eval are both set to False")
```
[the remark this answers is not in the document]
JingyaHuang: We might assume that users will normally not come with an already fine-tuned model and just run the script for prediction.

examples/onnxruntime/pytorch/text-classification/run_glue.py
```diff
@@ -0,0 +1,639 @@
...
+        quantizer.fit(
+            model_args.model_name_or_path,
+            output_dir,
+            feature="sequence-classification",
+            cache_dir=model_args.cache_dir,
+        )
+        onnx_config = quantizer.onnx_config
+        opt_model_path = output_dir.joinpath("model-quantized.onnx")
+
+    elif optim_args.optimize:
```
JingyaHuang, line 611: Idem question on the possibility of quantization+optimization

examples/onnxruntime/pytorch/token-classification/run_ner.py
```diff
@@ -0,0 +1,682 @@
...
+        quantizer.fit(
+            model_args.model_name_or_path,
+            output_dir,
+            feature="token-classification",
+            cache_dir=model_args.cache_dir,
+        )
+        onnx_config = quantizer.onnx_config
+        opt_model_path = output_dir.joinpath("model-quantized.onnx")
+
+    elif optim_args.optimize:
```
JingyaHuang, line 655: Idem question on the possibility of quantization+optimization

optimum/onnxruntime/model.py
```diff
@@ -0,0 +1,111 @@
+#  Copyright 2021 The HuggingFace Team. All rights reserved.
```
JingyaHuang, line 1: ORTModel has some overlaps with ORTTrainer, we might go for it in this PR and then merge it with ORTTrainer in PR#77!

JingyaHuang (commented):
Nice work! 
Now we have examples well arranged with quantizer and optimizer, next, we could expand the examples with ORTTrainer implemented in PR#77.

examples/onnxruntime/pytorch/question-answering/run_qa.py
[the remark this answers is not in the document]
echarlaix: You are right, I added the prediction step for all the examples scripts

examples/onnxruntime/pytorch/question-answering/run_qa.py
echarlaix: Optimization can be applied with quantization (when `optimize_model` attribute of `ORTConfig` is set to True), however, `ORTQuantizer` and `ORTOptimizer` are not very compatible (in the sense that if we first use `ORTOptimizer`  and then `ORTQuantizer` , a lot of operators will not be quantized)

PR #76, opened 2022-02-04
Create class AutoOnnxConfig

JingyaHuang:
# What does this PR do?
Add `AutoOnnxConfig` class to `optimum.onnx.auto` to enable automatic instantiation of onnx configuration.

## Changes
-  Add `optimum/onnx/auto/configuartion_onnx_auto.py` with the definition of the `AutoOnnxConfig` class
-  Update `optimum/onnx/auto/__init__.py`
-  Add the corresponding test in `tests/onnx/test_configuration_auto.py` 

## Goal
Wrap the `XXXOnnxConfig` directly on the top of transformers configuration. 

## Reviewers
@echarlaix @michaelbenayoun @mfuntowicz


optimum/onnx/auto/configuration_onnx_auto.py
```diff
@@ -0,0 +1,211 @@
...
+    _LazyConfigMapping,
+)
+from transformers.models.auto.dynamic import get_class_from_dynamic_module
+from transformers.onnx.config import OnnxConfig
+from transformers.utils import logging
+
+
+logger = logging.get_logger(__name__)
+
+ONNX_CONFIG_MAPPING_NAMES = OrderedDict(
```
michaelbenayoun, line 38: Why do we need an `OrderedDict` here?

optimum/onnx/auto/configuration_onnx_auto.py
```diff
@@ -0,0 +1,211 @@
...
+        ("t5", "T5OnnxConfig"),
+        ("xlm-roberta", "XLMRobertaOnnxConfig"),
+        ("xlm-roberta-xl", "XLMRobertaXLOnnxConfig"),
+    ]
+)
+
+
+def config_class_to_model_type(config):
+    """Converts a onnx config class name to the corresponding model type"""
+    for key, cls in ONNX_CONFIG_MAPPING_NAMES.items():
```
michaelbenayoun, line 60: A solution would also to build a "reverse" mapping `ONNX_CONFIG_TO_MODEL_TYPE` or something like that right after `ONNX_CONFIG_MAPPING_NAMES`'s instantation. That would allow to not loop all the time.

optimum/onnx/auto/configuration_onnx_auto.py
```diff
@@ -0,0 +1,211 @@
...
+            config: MODEL_NAMES_MAPPING[model_type] for model_type, config in ONNX_CONFIG_MAPPING_NAMES.items()
+        }
+        lines = [
+            f"{indent}- [`{config_name}`] configuration class: {config_to_name[config_name]} ({config_to_model_name[config_name]} model)"
+            for config_name in sorted(config_to_name.keys())
+        ]
+    return "\n".join(lines)
+
+
+def replace_list_option_in_docstrings(config_to_class=None, use_model_types=True):
```
michaelbenayoun, line 103: The purpose of this decorator is to add the list of supported models in the docstring?

tests/onnx/test_configuration_auto.py
```diff
@@ -0,0 +1,59 @@
...
+    def test_config_from_model_shortcut(self):
+        model_names = {"bert-base-cased", "distilbert-base-uncased", "facebook/bart-base", "gpt2", "roberta-base"}
+        task_names = {"default", "masked-lm", "sequence-classification"}
+
+        for model_name in model_names:
+            for task_name in task_names:
+                with self.subTest(model_name=model_name):
+                    config = AutoOnnxConfig.from_pretrained(model_name, task=task_name)
+                    self.assertIsInstance(config, OnnxConfig)
+                    print(type(config))
```
michaelbenayoun, line 54: I think you forgot to delete those prints.

tests/onnx/test_configuration_auto.py
```diff
@@ -0,0 +1,59 @@
...
+
+from transformers.onnx import OnnxConfig
+
+from optimum.onnx.auto.configuration_onnx_auto import AutoOnnxConfig
+
+
+SAMPLE_ROBERTA_CONFIG = os.path.join(os.path.dirname(os.path.abspath(__file__)), "fixtures/dummy-config.json")
+
+
+class AutoOnnxConfigTest(unittest.TestCase):
```
michaelbenayoun, line 44: What could be interesting is adding also a test for checking that non registered model cannot be loaded with an `AutoOnnxConfig`.

tests/onnx/test_configuration_auto.py
```diff
@@ -0,0 +1,59 @@
...
+    def test_config_from_model_shortcut(self):
+        model_names = {"bert-base-cased", "distilbert-base-uncased", "facebook/bart-base", "gpt2", "roberta-base"}
```
michaelbenayoun, line 46: You could use [this](https://github.com/huggingface/transformers/blob/master/tests/test_onnx_v2.py#L219) function with the mappings defined [here](https://github.com/huggingface/transformers/blob/master/tests/test_onnx_v2.py#L171) to test all the supported ones.

tests/onnx/test_configuration_auto.py
```diff
@@ -0,0 +1,59 @@
...
+import os
+import tempfile
+import unittest
+
+from transformers.onnx import OnnxConfig
+
+from optimum.onnx.auto.configuration_onnx_auto import AutoOnnxConfig
+
+
+SAMPLE_ROBERTA_CONFIG = os.path.join(os.path.dirname(os.path.abspath(__file__)), "fixtures/dummy-config.json")
```
michaelbenayoun, line 41: Are you using this?

optimum/onnx/auto/configuration_onnx_auto.py
JingyaHuang: > Why do we need an `OrderedDict` here?

There are models in transformers with OnnxConfig defined, but some are not. So here I track all those models with their OnnxConfig defined.

tests/onnx/test_configuration_auto.py
JingyaHuang: > I think you forgot to delete those prints.

Indeed... Will delete them!

optimum/onnx/auto/configuration_onnx_auto.py
JingyaHuang: Yes, but here we actually didn't use it for AutoOnnxConfig, I left it there to make it similar to what we have for Autoconfig in [transformers](https://github.com/huggingface/transformers/blob/master/src/transformers/models/auto/configuration_auto.py#L333)

optimum/onnx/auto/configuration_onnx_auto.py
JingyaHuang: Yes, I forgot to add back the docstring, will do!

optimum/onnx/auto/configuration_onnx_auto.py
JingyaHuang: But indeed, it will be better to get it directly from FeaturesManager so that we don't need to change it all the time.

optimum/onnx/auto/configuration_onnx_auto.py
JingyaHuang: But indeed, it will be better to get it directly from FeaturesManager so that we don't need to change it all the time.

tests/onnx/test_configuration_auto.py
JingyaHuang: Nope, in the end, I did not use it, I will delete it!

mfuntowicz:
@JingyaHuang is this PR still needed? I can have a deeper look to merge and close

JingyaHuang:
Pinging @mfuntowicz for a review. A refactoring of `AutoOnnxConfig` is done. Can you take a look? Then it will be ready to merge.
Currently, it can just wraps the corresponding xxxOnnxConfig over the xxxxConfig. As now we have wrappers to include non-default inputs (eg. [`OnnxConfigWithLoss`](https://github.com/huggingface/optimum/blob/main/optimum/onnx/configuration.py#L36)), does it make sense to you to add some flags in `AutoOnnxConfig` like this:
```python
from optimum import AutoOnnxConfig
onnx_config = AutoOnnxConfig("bert-base-uncased", with_loss=True)
```

JingyaHuang:
Close as outdated.

PR #88, opened 2022-03-07
Add verification model type supported by ONNX Runtime

echarlaix:
In this PR we verify the models supported and compatible with ONNX Runtime before applying an optimization or quantization step


JingyaHuang (approved):
Hi @echarlaix, excellent job, it looks good to me!

PR #86, opened 2022-03-03
Add basic docs and GitHub workflows

lewtun:
# What does this PR do?

This PR adds a `docs` folder with basic MDX files to test that we can generate documentation on hf.co. I've also added the GitHub actions to automatically build the docs on each PR / push to `main`

@Pierrci do you happen to know what's needed on the `moon-landing` side to be able to visit the docs once they're generated?

## Remarks

* I've used the `transformers` repository as a guide and just focused on adding landing page & quickstart for now.

PR #92, opened 2022-03-11
Use main branch as default "version" when building docs

lewtun:
# What does this PR do?

In #86 I built the docs with the `optimum` version number, but in general we'll want to have the docs reflect the state on the `main` branch. This PR does that and once we have integrated the rest of the API docs, we can revisit setting the version number with future releases.

cc @mishig25

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -69,9 +69,8 @@ jobs:
 
       - name: Make documentation
         run: |
-          version=$(pip show optimum | grep Version | cut -c 10-20) &&
           cd doc-builder &&
-          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version $version --html &&
+          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version main --html &&
           cd ..
         env:
           NODE_OPTIONS: --max-old-space-size=6656


.github/workflows/build_documentation.yml
```diff
@@ -69,9 +69,8 @@ jobs:
 
       - name: Make documentation
         run: |
-          version=$(pip show optimum | grep Version | cut -c 10-20) &&
           cd doc-builder &&
-          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version $version --html &&
+          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version main --html &&
```
mishig25, line 73:
```suggestion
          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --html &&
```

whether it is is `master` or `main` branch gets detected automatically 
see [here](https://github.com/huggingface/accelerate/blob/main/.github/workflows/build_documentation.yml#L70-L73)

mishig25: @lewtun lets merge it & try if the doc build in correct place? 😊

mishig25: companion [moon-landing pr](https://github.com/huggingface/moon-landing/pull/2329)

lewtun: Actually, for `optimum` we have to specify the branch manually for now because the repo has a different structure than the other HF libraries (no `__init__.py` in the root of the repo where we can store the `__version__` attribute). Without it, one gets the following error:

```python
AttributeError: module 'optimum' has no attribute '__version__'
```

I plan to fix this next week, so let's keep the manual flag for now. Gently pinging @echarlaix for permission to merge this :)

mishig25: oh I see, then in that case ignore my suggested change
the hardcoded  `--version main` works too

mishig25 (approved):

echarlaix (approved):

PR #114, opened 2022-03-24
ONNX weight sharing

michaelbenayoun:
# What does this PR do?

This PR adds a transformation on ONNX models that scans for weights that are shared (or exact duplicates) and prunes all the redundancy. The resulting model ends up smaller, and the computation remains unchanged.

This can be useful for models such as Albert.


echarlaix (approved):
Looks good to me ! I think some additional tests (at least 2 : model with redundant weights and not) should be added before merging though

JingyaHuang:
Just tested with `t5-small`, the number of initializers reduces from 338 to 139, that's quite significant. Remarkable job @michaelbenayoun !!

PR #124, opened 2022-03-30
[ORTTrainer] Fix to be compatible with onnxruntime 1.11.0

JingyaHuang:
# What does this PR do?
-  Fix `fix_atenops_to_gather` broken due to changed ops names.
-  Patch `Seq2SeqORTTrainer` to be compatible with `BART` model.
- Enable inference within PyTorch when ONNX Runtime doesn't work.

### DIFF
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -1239,7 +1239,8 @@ class ORTTrainer(Trainer):
             loss = self.label_smoother(outputs, labels)
         else:
             # We don't use .loss here since the model may return tuples instead of ModelOutput.
-            loss = outputs["loss"] if isinstance(outputs, dict) else outputs[0]
+            # loss = outputs["loss"] if isinstance(outputs, dict) else outputs[0]
+            loss = outputs[0]
 
         return (loss, outputs) if return_outputs else loss
 
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -648,6 +648,20 @@ class ORTTrainer(Trainer):
         ignore_keys: Optional[List[str]] = None,
         metric_key_prefix: str = "eval",
     ) -> Dict[str, float]:
+        try:
+            return self.evaluate_ort(eval_dataset, ignore_keys, metric_key_prefix)
+        except:
+            logger.warning(
+                f"Unable to do inference within ONNX Runtime for {self.model.config.name_or_path} model. Evaluate with PyTorch backend instead."
+            )
+            return super().evaluate(eval_dataset, ignore_keys, metric_key_prefix)
+
+    def evaluate_ort(
+        self,
+        eval_dataset: Optional[Dataset] = None,
+        ignore_keys: Optional[List[str]] = None,
+        metric_key_prefix: str = "eval",
+    ) -> Dict[str, float]:
         """
         Run evaluation within ONNX Runtime backend and returns metrics.(Overriden from `Trainer.evaluate()`)
         """
@@ -691,6 +705,17 @@ class ORTTrainer(Trainer):
         return output.metrics
 
     def predict(
+        self, test_dataset: Dataset, ignore_keys: Optional[List[str]] = None, metric_key_prefix: str = "test"
+    ) -> PredictionOutput:
+        try:
+            return self.predict_ort(test_dataset, ignore_keys, metric_key_prefix)
+        except:
+            logger.warning(
+                f"Unable to do inference within ONNX Runtime for {self.model.config.name_or_path} model. Predict with PyTorch backend instead."
+            )
+            return super().predict(test_dataset, ignore_keys, metric_key_prefix)
+
+    def predict_ort(
         self, test_dataset: Dataset, ignore_keys: Optional[List[str]] = None, metric_key_prefix: str = "test"
     ) -> PredictionOutput:
         """
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -647,6 +647,7 @@ class ORTTrainer(Trainer):
         eval_dataset: Optional[Dataset] = None,
         ignore_keys: Optional[List[str]] = None,
         metric_key_prefix: str = "eval",
+        inference_with_ort: bool = False,
     ) -> Dict[str, float]:
         """
         Run evaluation within ONNX Runtime backend and returns metrics.(Overriden from `Trainer.evaluate()`)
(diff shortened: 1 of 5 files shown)


optimum/onnxruntime/trainer.py
```diff
@@ -1194,7 +1229,7 @@ def compute_loss_ort(self, model, inputs, input_names, output_names, return_outp
 
         input_feed = dict(map(lambda input_name: (input_name, inputs[input_name].cpu().numpy()), input_names))
         outputs = self.infer_sess.run(output_names, input_feed)
-        outputs[0] = torch.Tensor(outputs[0])  # logits: `numpy.ndarray` -> tensor
+        outputs[0] = torch.Tensor(outputs[0])
```
echarlaix, line 1232: Shouldn't you do this on the loss as [here](https://github.com/huggingface/optimum/blob/ort-train-11-compatibility/optimum/onnxruntime/trainer.py#L1242) you leave the possibility of outputs being a dictionnary

tests/onnxruntime/test_onnxruntime_train.py
```diff
@@ -90,7 +86,8 @@ def compute_metrics(eval_pred):
                             return metric.compute(predictions=predictions, references=eval_pred.label_ids)
 
                         training_args = TrainingArguments(
-                            output_dir=tmp_dir,
+                            # output_dir=tmp_dir,
+                            output_dir="./results",
```
echarlaix, line 90: Shouldn't it be `tmp_dir` ?

optimum/onnxruntime/trainer.py
JingyaHuang: Actually, the function `compute_loss_ort` will only be used for the ORT circumstance, otherwise `compute_loss` will be used. Thus the case of `outputs` being a dictionary does not really exist in that case. 
Indeed it is confusing because I didn't remove that cases of  `outputs` being a dictionary just 10 lines below...

tests/onnxruntime/test_onnxruntime_train.py
JingyaHuang: yep! Will correct that!

optimum/onnxruntime/trainer.py
echarlaix: I think it would make sense to remove this case then, what do you think ?

JingyaHuang: Yep, done in the new commit!

optimum/onnxruntime/trainer.py
```diff
@@ -657,16 +658,39 @@ def evaluate(
...
+                    prediction_loss_only=True if self.compute_metrics is None else None,
+                    ignore_keys=ignore_keys,
+                    metric_key_prefix=metric_key_prefix,
+                )
+            except Exception as error:
+                logger.warning(
+                    f"[ERROR!] Evaluation with ONNX Runtime is not available for {self.model.config.name_or_path} model. Remove `inference_with_ort` to evaluate within PyTorch."
+                )
+                logger.error(error)
+                raise
```
echarlaix, line 679: Why not `raise error` ?

optimum/onnxruntime/trainer.py
```diff
@@ -703,10 +731,34 @@ def predict(
...
+                    description="Prediction",
+                    ignore_keys=ignore_keys,
+                    metric_key_prefix=metric_key_prefix,
+                )
+            except Exception as error:
+                logger.warning(
+                    f"[ERROR!] Prediction with ONNX Runtime is not available with {self.model.config.name_or_path} model. Remove `inference_with_ort` to predict within PyTorch."
+                )
+                logger.error(error)
+                raise
```
echarlaix, line 749: same comment

optimum/onnxruntime/trainer_seq2seq.py
```diff
@@ -72,7 +74,17 @@ def evaluate(
         """
         self._max_length = max_length
         self._num_beams = num_beams
-        return super().evaluate(eval_dataset, ignore_keys=ignore_keys, metric_key_prefix=metric_key_prefix)
+        if self.args.predict_with_generate and inference_with_ort:
+            logger.error(
+                f"[ERROR!] Generate method is not available with prediction within ONNX Runtime. Remove `inference_with_ort` or `predict_with_generate`."
+            )
+            raise
```
echarlaix, line 81:
```suggestion
            raise NotImplementedError("Generate method is not available with prediction within ONNX Runtime. Remove `inference_with_ort` or `predict_with_generate`.")
```

optimum/onnxruntime/trainer_seq2seq.py
```diff
@@ -113,7 +126,17 @@ def predict(
         """
         self._max_length = max_length
         self._num_beams = num_beams
-        return super().predict(test_dataset, ignore_keys=ignore_keys, metric_key_prefix=metric_key_prefix)
+        if self.args.predict_with_generate and inference_with_ort:
+            logger.error(
+                f"[ERROR!] Generate method is not available with prediction within ONNX Runtime. Remove `inference_with_ort` or `predict_with_generate`."
+            )
+            raise
```
echarlaix, line 133:
```suggestion
            raise NotImplementedError("Generate method is not available with prediction within ONNX Runtime. Remove `inference_with_ort` or `predict_with_generate`.")
```

optimum/onnxruntime/trainer.py
JingyaHuang: Actually here I will be more likely to put then in a reverse order, like:
 ```
logger.error(error)
logger.warning(
                    f"[ERROR!] Evaluation with ONNX Runtime is not available for {self.model.config.name_or_path} model. Remove `inference_with_ort` to evaluate within PyTorch."
                )
raise            
```
 
I want our warning to be in the last line so that it could be more clear for user to know what to do next. The original error could not normally be helpful for them.

File: optimum/onnxruntime/trainer.py [hunk 7]
```diff
@@ -1194,7 +1226,7 @@ def compute_loss_ort(self, model, inputs, input_names, output_names, return_outp
 
         input_feed = dict(map(lambda input_name: (input_name, inputs[input_name].cpu().numpy()), input_names))
         outputs = self.infer_sess.run(output_names, input_feed)
-        outputs[0] = torch.Tensor(outputs[0])  # logits: `numpy.ndarray` -> tensor
+        outputs[0] = torch.Tensor(outputs[0])
```
echarlaix, line 1229: Should be removed.

optimum/onnxruntime/trainer.py
```diff
@@ -1204,7 +1236,8 @@ def compute_loss_ort(self, model, inputs, input_names, output_names, return_outp
             loss = self.label_smoother(outputs, labels)
         else:
             # We don't use .loss here since the model may return tuples instead of ModelOutput.
-            loss = outputs["loss"] if isinstance(outputs, dict) else outputs[0]
+            # loss = outputs["loss"] if isinstance(outputs, dict) else outputs[0]
+            loss = outputs[0]
```
echarlaix, line 1240: Should be removed as `outputs[0]` is not the loss.

echarlaix (approved):

PR #133, opened 2022-04-06
Add quantization preprocess to include fully connected layer nodes

echarlaix:
In this PR, we add the `IncludeFullyConnectedNodes` class, which allows to find the nodes composing the model's fully connected layers, in order to include those to the list of nodes to quantize. We also force the intersection of the two sets respectively representing the nodes to quantize `global_nodes_to_quantize` an the nodes to exclude `global_nodes_to_exclude` to be an empty set, in order to prevent any overlap.

### DIFF
--- a/optimum/onnxruntime/preprocessors/passes/fully_connected.py
+++ b/optimum/onnxruntime/preprocessors/passes/fully_connected.py
@@ -1,0 +1,33 @@
+#  Copyright 2022 The HuggingFace Team. All rights reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from typing import Optional, Set, Tuple
+
+from onnx import ModelProto
+from onnxruntime.transformers.onnx_model import OnnxModel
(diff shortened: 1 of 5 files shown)


optimum/onnxruntime/preprocessors/passes/fully_connected.py
```diff
@@ -0,0 +1,33 @@
...
+from onnx import ModelProto
+from onnxruntime.transformers.onnx_model import OnnxModel
+from optimum.onnxruntime.preprocessors import PreprocessorPass
+
+
+class IncludeFullyConnectedNodes(PreprocessorPass):
+    def __init__(self):
+        super().__init__()
+
+    def __call__(self, graph: ModelProto, model: OnnxModel) -> Tuple[Optional[Set[str]], Optional[Set[str]]]:
```
michaelbenayoun, line 25: Small detail but the return type seems to be ` Tuple[Optional[Set[str]], Set[str]]`

optimum/onnxruntime/preprocessors/passes/fully_connected.py
```diff
@@ -0,0 +1,33 @@
...
+        fc_subgraphs = []
+        for matmul_node in model.get_nodes_by_op_type("Add"):
+            fc_components = model.match_parent_path(matmul_node, ["MatMul"], [1])
```
michaelbenayoun, line 28: Aren't those supposed to be the potential matmul nodes?

michaelbenayoun (approved):
LGTM!

optimum/onnxruntime/preprocessors/passes/fully_connected.py
echarlaix: Well seen, the variable name should be `add_node` instead of `matmul_node`

optimum/onnxruntime/preprocessors/passes/fully_connected.py
echarlaix: Yes it even should be `Tuple[Set[str], Set[str]]`

PR #139, opened 2022-04-11
Add missing dependency for ONNX Runtime training test

fxmarty:
# What does this PR do?

Add `nltk` in the test dependencies, as it is used https://github.com/huggingface/optimum/blob/main/tests/onnxruntime/test_onnxruntime_train.py#L19 and not installed with the vanilla `datasets`.


setup.py
```diff
@@ -14,7 +14,7 @@
 
 REQUIRED_PKGS = ["coloredlogs", "sympy", "transformers>=4.15.0", "torch>=1.9", "packaging"]
 
-TESTS_REQUIRE = ["pytest"]
+TESTS_REQUIRE = ["pytest", "nltk"]
```
JingyaHuang, line 17: Thanks for noticing that! 
Sorry that I didn't mention that the CI workflow of ONNX Runtime training test makes use of `examples/onnxruntime.training/run.sh` to install the dependencies required for `tests/onnxruntime/test+onnxruntime_train.py` instead of `setup.py`. So in this case, additional lib in `TESTS_REQUIRE` won't impact the test for onnxruntime training. But indeed I haven't seen any CI making use of  `TESTS_REQUIRE`, pinging @echarlaix and @michaelbenayoun for more explanation.

echarlaix: Yes, this will add the nltk package dependency  to dev and tests extras, impacting the workflows `test_optimum_common`, `build_documentation` and `build_dev_documentation`, which do not require it.

fxmarty: @echarlaix @JingyaHuang Thank you, indeed that makes sense not to add every dependencies for [test_onnxruntime_train.py](https://github.com/huggingface/optimum/blob/main/tests/onnxruntime/test_onnxruntime_train.py) in there as there are quite a few not necessary elsewhere, and not trivial to set up.

I think it would be worthwhile to document how this test is supposed to run, naively, users developing on top of optimum may want to do a `pip install -e .[dev]` and expect `pytest` to run without issues, which is not the case right now.

fxmarty: Makes sense, closing.

PR #142, opened 2022-04-13
Create ONNX configurations with loss as output

JingyaHuang:
# What does this PR do?

__Issue__: `ORTTrainer.evaluate()` doesn't have the loss in the output.  

* Add classes `OnnxConfigWithLoss`, `OnnxConfigWithPastAndLoss`,  `OnnxSeq2SeqConfigWithPastAndLoss` as wrappers of `OnnxConfig`, `OnnxConfigWithPast` and `OnnxSeq2SeqConfigWithPast` to include `loss` in the outputs.
* Tests for the created classes

## Before submitting
- [x] This PR fixes a typo or improves the docs 


optimum/onnx/configuration.py
```diff
@@ -0,0 +1,229 @@
...
+import torch
+from transformers.onnx import OnnxConfig, OnnxConfigWithPast, OnnxSeq2SeqConfigWithPast
+
+
+logger = logging.get_logger(__name__)
+
+
+class OnnxConfigWithLoss(OnnxConfig, ABC):
+    """
+    Wrapper for the childern classes of `transformers.onnx.OnnxConfig` to export the model through the ONNX format with loss in outputs.
```
mfuntowicz, line 46: Small typo `childern` -> `children`

mfuntowicz (approved):
LGTM 🔥

optimum/onnx/configuration.py
```diff
@@ -0,0 +1,229 @@
...
+            framework,
+            num_channels,
+            image_width,
+            image_height,
+            tokenizer,
+        )
+        for label, input in self._tasks_to_extra_inputs[self.task].items():
+            if "sequence" in input.values():
+                dummy_inputs[label] = torch.zeros(
+                    self.default_fixed_batch, self.default_fixed_sequence, dtype=torch.long
```
michaelbenayoun, line 161: I think you should use `compute_effective_axis_dimension` to determine the batch size and the sequence length.

optimum/onnx/configuration.py
```diff
@@ -0,0 +1,229 @@
...
+
+    @property
+    def outputs(self) -> Mapping[str, Mapping[int, str]]:
+        common_outputs = super().outputs
+        if self.use_past:
+            self._onnx_config.fill_with_past_key_values_(common_outputs, direction="outputs")
+
+        return common_outputs
+
+    def generate_dummy_inputs(
```
michaelbenayoun, line 196: Do you really need to overwrite it?

michaelbenayoun (commented):
Really good @JingyaHuang !
Left a few comments

optimum/onnx/configuration.py
JingyaHuang: Thanks Morgan, fixing it!

optimum/onnx/configuration.py
JingyaHuang: Indeed I did not consider the case when sequence length and batch size are not dynamic. However, the labels unlike other elements in inputs don't need to consider the tokens to add, so it is either default_fixed_batch/seq(dynamic) or  batch_size/seq. I will use `compute_effective_axis_dimension` by setting `token_to_add=0`, thanks!

JingyaHuang (commented):
Thanks for reviewing! @michaelbenayoun I will soon adjust according to your comments!

optimum/onnx/configuration.py
```diff
@@ -0,0 +1,229 @@
...
+    @property
+    def outputs(self) -> Mapping[str, Mapping[int, str]]:
+        """
+        Mapping containing the axis definition of the output tensors(including loss) to provide to the model
+        Returns:
+            For each output: its name associated to the axes symbolic name and the axis position within the tensor
+        """
+        common_outputs = self._onnx_config.outputs
+        common_outputs.update(self._tasks_to_extra_outputs["default"])
+        if "loss" in common_outputs.keys():
```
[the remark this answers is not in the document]
michaelbenayoun: Yes, but I'm under the impression that you add the loss to `common_outputs` then check right after that it contains it?
I might be missing something here, just wanted to point it out.

optimum/onnx/configuration.py
```diff
@@ -0,0 +1,229 @@
...
+            }
+        ),
+        "image-classification": OrderedDict({"labels": {0: "batch"}}),
+    }
+    _tasks_to_extra_outputs = {
+        "default": OrderedDict({"loss": {}}),
+    }
+
+    def __init__(self, config: OnnxConfig):
+        self.__dict__ = config.__dict__
```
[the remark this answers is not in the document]
michaelbenayoun: Yes, I feel like it would be better to only copy the task then, or at least to work on a deepcopy of the original config.
```suggestion
        self.__dict__ = copy.deepcopy(config.__dict__)
```

optimum/onnx/configuration.py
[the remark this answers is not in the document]
JingyaHuang: Oh I get what you mean now, sorry that I misunderstood it. Yeah in the case that we only have one `default` case, The if statement is not necessary. I've just put things this way in case we call loss differently in some cases. will replace "loss" by `common_outputs.keys` and delete the `if`

optimum/onnx/configuration.py
```diff
@@ -0,0 +1,256 @@
...
+        for label, input in self._tasks_to_extra_inputs[self.task].items():
+            batch_size = compute_effective_axis_dimension(
+                batch_size, fixed_dimension=self.default_fixed_batch, num_token_to_add=0
+            )
+            if "sequence" in input.values():
+                seq_length = compute_effective_axis_dimension(
+                    seq_length, fixed_dimension=self.default_fixed_sequence, num_token_to_add=0
+                )
+                if framework == TensorType.PYTORCH:
+                    import torch
```
michaelbenayoun, line 158: In transformers, we check that the framework is available.

```suggestion
                    if not torch_is_available():
                        raise ValueError("Could not generate dummy inputs to perform the ONNX export through PyTorch because not installation was found.")
                    else:
                        import torch
```

We have a similar `is_tf_available` method for TensorFlow.

michaelbenayoun (approved):

Issue #146, opened 2022-04-14
`ShardedDDP` and `FullyShardedDDP` not imported in `optimum/onnxruntime/trainer.py`


jambayk:
`ShardedDDP` is used here at [this line](https://github.com/huggingface/optimum/blob/dfe330a2888f3e7fd82fff73d7481faa34fe94e5/optimum/onnxruntime/trainer.py#L1300) but it was never imported. Similarly for `FullyShardedDDP`.

echarlaix:
Thanks for reporting it ! cc @JingyaHuang

JingyaHuang:
Close this as it is solved in PR #148.

PR #148, opened 2022-04-15
Patch ORTTrainer's compatibility with DeepSpeed

### LINKED ISSUE #145: `ORTTrainer` doesn't work with distributed training and/or DeepSpeed
`ORTTrainer.train` fails with distributed training and/or deepspeed. 

[The line](https://github.com/jambayk/optimum/blob/dfe330a2888f3e7fd82fff73d7481faa34fe94e5/optimum/onnxruntime/trainer.py#L639) `inference_manager = model._torch_module._execution_manager._inference_manager` assumes that `model` is of type `ORTModule`. However, when deepspeed is enabled, it is of type `DeepSpeedEngine`. Its type is `DistributedDataParallel` during distributed training.

This leads to an AttributeError for such cases.

JingyaHuang:
# What does this PR do?

Fixes #145 and #146 

## Contents

- [ ] Add missing dependencies  
- [ ] Fix compatibility with `deepspeed`
- [ ] Enable `sharded_ddp`

## Checks
- [ ] Fixes typos 
- [ ] Updates the docs.
- [ ] Necessary tests of all above

### DIFF
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -41,7 +41,7 @@ from transformers import PreTrainedModel, __version__
 from transformers.configuration_utils import PretrainedConfig
 from transformers.data.data_collator import DataCollator
 from transformers.debug_utils import DebugOption, DebugUnderflowOverflow
-from transformers.deepspeed import deepspeed_init, deepspeed_reinit
+from transformers.deepspeed import deepspeed_init, deepspeed_reinit, is_deepspeed_zero3_enabled
 from transformers.file_utils import (
     CONFIG_NAME,
     WEIGHTS_NAME,
@@ -95,6 +95,12 @@ if is_torch_tpu_available():
     import torch_xla.core.xla_model as xm
     import torch_xla.debug.metrics as met
     import torch_xla.distributed.parallel_loader as pl
+
+if is_fairscale_available():
+    dep_version_check("fairscale")
+    from fairscale.nn.data_parallel import FullyShardedDataParallel as FullyShardedDDP
+    from fairscale.nn.data_parallel import ShardedDataParallel as ShardedDDP
+    from fairscale.nn.wrap import auto_wrap
 
 if is_sagemaker_dp_enabled():
     import smdistributed.dataparallel.torch.distributed as dist


jambayk:
Hi, I saw that this PR is linked to issue https://github.com/huggingface/optimum/issues/145 I opened. 

I do not think this resolves the issue which came from `model` being used as if it's of type `ORTModule`. But it isn't when deepspeed or any other wrappers such as DDP are used. The inference sessions information needs to be obtained before ORTModule gets wrapped or a reference to the ORTModule needs to be maintained (if training needs to complete before the inference info is obtained). I do the latter in this commit on my private fork https://github.com/jambayk/optimum/commit/c2d53f43c3beaa60b1c1887cac2132c09884393b.

JingyaHuang:
> Hi, I saw that this PR is linked to issue #145 I opened.
> 
> I do not think this resolves the issue which came from `model` being used as if it's of type `ORTModule`. But it isn't when deepspeed or any other wrappers such as DDP are used.
> 
> EDIT: I apologize if this is still work in progress and you had a fix in the works. Please ignore this comment if so. Just wanted to make sure I provided enough context for the issue.

Hi @jambayk , no worries, I opened this PR so that there will be more transparency on the progress. Sorry for the confusion that it might have caused, and thanks a lot for the information on the issue.

JingyaHuang:
The compatibility of `ORTTrainer` and DeepSpeed/Fairscale is implemented. Check the result of the tests [here](https://github.com/pytorch/ort/issues/108).

optimum/onnxruntime/trainer.py
```diff
@@ -310,6 +310,13 @@ def train(
                     "`ORTTrainer` does not support ZeRO stage 3 for the moment. Please use DeepSpeed stage 1 or 2 instead."
                 )
 
+            if args.bf16:
+                warnings.warn(
+                    "ONNX Runtime doesn't support BF16 for executing `Aten` operator. The execution will fail if"
```
lewtun, line 315: nit:
```suggestion
                    "ONNX Runtime doesn't support BF16 when executing `Aten` operators. The execution will fail if"
```

optimum/onnxruntime/trainer.py
```diff
@@ -310,6 +310,13 @@ def train(
...
+                    "there are any `Aten` op in the IR. The implementation is in progress from ONNX Runtime side, stay tuned!",
```
lewtun, line 316:
```suggestion
                    "there are any `Aten` op in the IR. Support for this in ONNX Runtime is currently in progress, stay tuned!",
```

optimum/onnxruntime/trainer.py
```diff
@@ -773,6 +780,13 @@ def evaluation_loop_ort(
             )
 
             logger.info("[INFO] Exporting the model to ONNX...")
+            if args.deepspeed and args.fp16:
+                warnings.warn(
+                    "Make sure that `transformers.onnx.export_pytorch` of the transformers verion supports "
```
lewtun, line 785:
```suggestion
                    "Make sure that `transformers.onnx.export_pytorch` of the transformers version supports "
```

optimum/onnxruntime/trainer.py
```diff
@@ -773,6 +780,13 @@ def evaluation_loop_ort(
...
+                    "exporting ONNX on cuda.",
```
lewtun, line 786:
```suggestion
                    "exporting ONNX on CUDA.",
```

optimum/onnxruntime/trainer.py
```diff
@@ -977,7 +991,15 @@ def prediction_loop_ort(
             onnx_model_path = Path(
                 os.path.join(self.args.output_dir, self.model.config.name_or_path.split("/")[-1] + ".onnx")
             )
+
             logger.info("[INFO] Exporting the model to ONNX...")
+            if args.deepspeed and args.fp16:
+                warnings.warn(
+                    "Make sure that `transformers.onnx.export_pytorch` of the transformers verion supports "
```
lewtun, line 998:
```suggestion
                    "Make sure that `transformers.onnx.export_pytorch` of the transformers version supports "
```

optimum/onnxruntime/trainer.py
```diff
@@ -977,7 +991,15 @@ def prediction_loop_ort(
...
+                    "exporting ONNX on cuda.",
```
lewtun, line 999:
```suggestion
                    "exporting ONNX on CUDA.",
```

optimum/onnxruntime/trainer.py
```diff
@@ -773,6 +780,13 @@ def evaluation_loop_ort(
             )
 
             logger.info("[INFO] Exporting the model to ONNX...")
+            if args.deepspeed and args.fp16:
+                warnings.warn(
```
lewtun, line 784: I wonder if we should check the `transformers` version and then raise a warning if the detected version doesn't match the required one for ONNX export on CUDA?

For now, this warning is OK but we'll probably want to revisit this once we bump `transformers` with your PR :)

lewtun (approved):
Thanks for adding these warnings @JingyaHuang ! LGTM 🚀 !

optimum/onnxruntime/trainer.py
JingyaHuang: @lewtun Yes, exactly. I put it this way since I am not sure in which version of transformers will include it. Will check the transformers version once I have the information.

lewtun: Sounds good!

optimum/onnxruntime/trainer.py
```diff
@@ -850,13 +874,19 @@ def evaluation_loop_ort(
 
             # Update containers on host
             if loss is not None:
+                if self.args.local_rank != -1:
+                    loss = loss.to("cuda")
```
echarlaix, line 878: Why use the rank and not `self.args.device` ?

echarlaix: What if `local_rank != -1` and `args.device="cpu"` ? I think using `args.device` in all cases would solve the issue, what do you think ?

JingyaHuang: Agreed! It would definitely be much better to send the tensors to the device than hard code the device name. Just modified that! Thanks for the suggestion.

echarlaix (approved):

PR #149, opened 2022-04-16
Segment generate dummy inputs to smaller cases

JingyaHuang:
# What does this PR do?

Segment the `generate_dummy_inputs()` method into two cases: PyTorch / TensorFlow for clarity.

### DIFF
--- a/optimum/onnx/configuration.py
+++ b/optimum/onnx/configuration.py
@@ -184,13 +184,19 @@ class OnnxConfigWithLoss(OnnxConfig, ABC):
             seq_length, fixed_dimension=self.default_fixed_sequence, num_token_to_add=0
         )
 
-        if framework == TensorType.PYTORCH and is_torch_available():
-            return self._generate_extra_dummy_inputs_pt(dummy_inputs, label_batch_size, label_seq_length)
-        elif framework == TensorType.TENSORFLOW and is_tf_available():
-            return self._generate_extra_dummy_inputs_tf(dummy_inputs, label_batch_size, label_seq_length)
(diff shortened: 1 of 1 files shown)


optimum/onnx/configuration.py
```diff
@@ -147,52 +177,21 @@ def generate_dummy_inputs(
...
 
-                        dummy_inputs[label] = tf.zeros(batch_size, dtype=tf.int64)
-        return dummy_inputs
+        if framework == TensorType.PYTORCH and is_torch_available():
+            return self._generate_extra_dummy_inputs_pt(dummy_inputs, label_batch_size, label_seq_length)
+        elif framework == TensorType.TENSORFLOW and is_tf_available():
+            return self._generate_extra_dummy_inputs_tf(dummy_inputs, label_batch_size, label_seq_length)
+        else:
+            raise ValueError(
+                f"Could not generate dummy inputs to perform the ONNX export through {framework} because not installation was found."
```
echarlaix, line 193: Something that could be interesting in order to better target the error would be to add some function such [as](https://github.com/huggingface/transformers/blob/main/src/transformers/onnx/features.py#L351)

echarlaix (approved):
It looks great, thanks @JingyaHuang

optimum/onnx/configuration.py
JingyaHuang: Good point! Thanks Ella, more precise framework information is updated with adding the consideration of frameworks other than PyTorch and TensorFlow. @echarlaix

JingyaHuang:
Clean-up done, will merge.

PR #160, opened 2022-04-28
Add a top level init

fxmarty:
# What does this PR do?

A `__init__.py` was I think missing at the `optimum/` level to allow to import the introduced optimum pipeline in https://github.com/huggingface/optimum/commit/a31e59eddafc53f6146a4850ffe91d4cfc00c0a6 as `from optimum import pipeline` as suggested in the doc: https://huggingface.co/docs/optimum/main/en/pipelines#optimum-pipelines-for-inference

@philschmid (I can not add reviewers so I ping you)

## Before submitting


fxmarty:
Woops. @philschmid Is the fact that this line https://github.com/huggingface/optimum/blob/a31e59eddafc53f6146a4850ffe91d4cfc00c0a6/optimum/pipelines.py#L17 is outside of the `if is_onnxruntime_available():` check wanted? Looks like it break things up if optimum is not installed with the dependencies for onnxruntime.

regisss:
Adding a `__init__.py` at the level of `optimum` will probably break the import of the namespace subpackages.

What you want to do is to replace `from optimum import pipeline` by `from optimum.pipelines import pipeline`.

fxmarty:
Indeed what's broken with this approach is that unittest (`python -m unittest discover -s tests -p 'test_*.py'`) includes `tests/onnxruntime` in the submodule search location, and thus [`importlib.util.find_spec("onnxruntime")`](https://github.com/huggingface/optimum/blob/81b85fa656730b3e0fb7b792781ff072d3f702cb/optimum/utils/__init__.py#L19) does not return `None`. So the [`if is_onnxruntime_available():`](https://github.com/huggingface/optimum/blob/81b85fa656730b3e0fb7b792781ff072d3f702cb/optimum/pipelines.py#L23) condition passes, and everything gets broken from there.

Interestingly, `pytest tests/test_*.py` is not unhappy as long as we add an `__init__.py` in the `tests/` folder as well. Looks like to be the approach taken by transformers, although I could be wrong.

In any case, I followed your suggestion. I think this can be merged.

Issue #163, opened 2022-05-05
Message "ORTModelForX is not supported for Y" when loading in pipeline


axsaucedo:
As initially reported in #161, when loading an ORT model in transformer pipeline there is a warning that suggests it's not supported. As proposed, having different message would be more intuitive.

To reproduce:

```python
model_id = "gpt2"
onnx_model = ORTModelForCausalLM.from_pretrained(model_id, from_transformers=True)
tokenizer = AutoTokenizer.from_pretrained(model_id)
pp = pipeline("text-generation", model=onnx_model, tokenizer=tokenizer)
```

output:

```
The model 'ORTModelForCausalLM' is not supported for text-generation. Supported models are ['XGLMForCausalLM', 'PLBartForCausalLM', 'QDQBertLMHeadModel', 'TrOCRForCausalLM', 'GPTJForCausalLM', 'RemBertForCausalLM', 'RoFormerForCausalLM', 'BigBirdPegasusForCausalLM', 'GPTNeoForCausalLM', 'BigBirdForCausalLM', 'CamembertForCausalLM', 'XLMRobertaXLForCausalLM', 'XLMRobertaForCausalLM', 'RobertaForCausalLM', 'BertLMHeadModel', 'OpenAIGPTLMHeadModel', 'GPT2LMHeadModel', 'TransfoXLLMHeadModel', 'XLNetLMHeadModel', 'XLMWithLMHeadModel', 'ElectraForCausalLM', 'CTRLLMHeadModel', 'ReformerModelWithLMHead', 'BertGenerationDecoder', 'XLMProphetNetForCausalLM', 'ProphetNetForCausalLM', 'BartForCausalLM', 'MBartForCausalLM', 'PegasusForCausalLM', 'MarianForCausalLM', 'BlenderbotForCausalLM', 'BlenderbotSmallForCausalLM', 'MegatronBertForCausalLM', 'Speech2Text2ForCausalLM', 'Data2VecTextForCausalLM'].
```

PR #164, opened 2022-05-06
[fix] Warning when loading ORTModels with pipelines of not supported class type

philschmid:
# What does this PR do?

This PR fixes the log warning user received when loading an `ORTModelForXXX` when creating a `pipeline` instance. The `pipeline` has a [model_type_check](https://github.com/huggingface/transformers/blob/cad61b68396a1a387287a8e2e2fef78a25b79383/src/transformers/pipelines/base.py#L863) which prints out a warning when using a non transformers class. 

```bash
The model 'ORTModelForQuestionAnswering' is not supported for question-answering
. Supported models are ['YosoForQuestionAnswering', 'NystromformerForQuestionAns
wering', 'QDQBertForQuestionAnswering', ...].
```
After discussing with @LysandreJik and @Narsil the best/easiest way to solve this is to register the `ORTModelFor` for the `transformers` auto_classes. 

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -5,7 +5,16 @@ from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
 import torch
-from transformers import AutoTokenizer, PretrainedConfig
+from transformers import (
+    AutoConfig,
+    AutoModel,
+    AutoModelForCausalLM,
+    AutoModelForQuestionAnswering,
+    AutoModelForSequenceClassification,
+    AutoModelForTokenClassification,
+    AutoTokenizer,
+    PretrainedConfig,
+)
 from transformers.file_utils import add_start_docstrings, add_start_docstrings_to_model_forward, default_cache_path
 from transformers.generation_utils import GenerationMixin
 from transformers.modeling_outputs import (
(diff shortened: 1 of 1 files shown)


optimum/onnxruntime/modeling_ort.py
```diff
@@ -77,6 +86,11 @@ def __init__(self, model=None, config=None, **kwargs):
         self.model_save_dir = kwargs.get("model_save_dir", None)
         self.latest_model_name = kwargs.get("latest_model_name", "model.onnx")
 
+        # registers the ORTModelForXXX classes into the transformers AutoModel classes
+        # to avoid warnings when create a pipeline https://github.com/huggingface/transformers/blob/cad61b68396a1a387287a8e2e2fef78a25b79383/src/transformers/pipelines/base.py#L863
```
lewtun, line 90: turbo nit:
```suggestion
        # to avoid warnings when creating a pipeline https://github.com/huggingface/transformers/blob/cad61b68396a1a387287a8e2e2fef78a25b79383/src/transformers/pipelines/base.py#L863
```

lewtun (approved):
Thanks for the fix! LGTM 🚀 

Btw this PR fixes #163

PR #168, opened 2022-05-09
Add version number to doc build

lewtun:
# What does this PR do?

This PR extracts the version number of the library from `version.py` and uses it to build the documentation with every push to `main` or `v*-release`.

Note the addition of ` .dev0` to the version number - this ensures all builds on `main` are mapped to the [`main` folder](https://github.com/huggingface/doc-build/tree/main/optimum) on `doc-build`.

When making a release, one simply needs to create a `vx.x.x-release` branch, fix the version number in `version.py` and push to remote. You can see how this is done in `transformers` here: https://github.com/huggingface/transformers/tree/v4.18-release

I've also re-instated the action to delete PR docs once the PR is closed


.github/workflows/build_documentation.yml
```diff
@@ -67,7 +68,7 @@ jobs:
       - name: Make documentation
         run: |
           cd doc-builder &&
-          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version main --html &&
+          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version ${{ env.WRITE }} --html &&
```
echarlaix, line 71: Should it be  `env.VERSION ` ?

echarlaix (approved):
Great work, thanks a lot @lewtun !

.github/workflows/build_documentation.yml
lewtun: Oh nice catch!

PR #176, opened 2022-05-11
Extend python required version to 3.7 and higher

echarlaix:


philschmid (approved):
Thank you for updating! LGTM!

PR #177, opened 2022-05-11
Fix version number shown in doc build

lewtun:
# What does this PR do?

This PR fixes a small bug by including a `v` prefix to the generated builds in the `doc-build` repo: https://github.com/huggingface/doc-build/tree/main/optimum

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Make documentation
         run: |
           cd doc-builder &&
-          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version ${{ env.VERSION }} --html &&
+          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version v${{ env.VERSION }} --html &&
           cd ..
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_documentation.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       - name: Make documentation
         run: |
           cd doc-builder &&
-          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version ${{ env.VERSION }} --html &&
+          doc-builder build optimum ../optimum/docs/source --build_dir ../doc-build --clean --version v${{ env.VERSION }} --html &&
```
lewtun, line 71: A question for @mishig25: currently this builds the docs with either `vX.X.X` (when triggered from a `v*-release` branch) or `vX.X.X.dev0` when triggered from the `main` branch. 

My understanding is that adding `.dev0` would ensure those builds correspond to the ones show on the `main` section of the docs dropdown, but now I'm seeing `1.2.0.dev0`:

Screen Shot 2022-05-11 at 11 41 46

Is there something else that needs to be done to ensure all doc builds on the `main` branch end up in the `main` section of the dropdown?

echarlaix (approved):

PR #192, opened 2022-05-15
Fix logging mistake where "dynamic" is logged if static quantization is performed

regisss:
# What does this PR do?

In `optimum/onnxruntime/quantization.py`, we currently have:
```python
LOGGER.info(
    f"Creating {'dynamic' if quantization_config.is_static else 'static'} quantizer: {quantization_config}"
)
```
'dynamic' and 'static' should be swapped.

## Before submitting


echarlaix:
Good catch, thanks @regisss !

PR #193, opened 2022-05-16
[WIP] Auto-compare optimized models vs. transformers models

### LINKED ISSUE #128: Create benchmarking suite for optimised models
Now that we have tight Hub integration coming via #113, it could be useful to implement a simple benchmarking suite that allows users to:

* Select a dataset on the Hub
* Select a metric on the Hub
* Select N models (could already be optimised models)
* Optimise the models (if needed)
* Report a table of results comparing the gains in latency and impact on the model metric

In a first step, we might simply want to benchmark latency with some dummy input at various sequence lengths etc.

fxmarty:
This is a work in progress PR, feedback welcome, notably for the design, code quality, etc. Documentation is lacking so I will work on it now, but I rather not push it too much till I am sure the design is good.

This PR aims at introducing an unified way to benchmark transformers vs. optimized models, **backend-independent** (in the sense of, any backend can be plugged for inference and evaluation), **code-free** (in the sense of, the user does not need to code to start runs and evaluate them).

The two main contributions are:
* Introduce helper classes, methods for data preprocessing, inference, evaluation. In several files:
	* **`optimum/runs_base.py`**: general methods, this should be backend-agnostic.
	* **`optimum/utils/preprocessing/`**: handle loading and preprocessing datasets, running inference with pipelines, running evaluation. This should be backend-agnostic.
	* **`optimum/onnxruntime/runs/`**: OnnxRuntime specific methods
* Define an evaluation API to compare optimized (for now, optimized in the sense quantized) vs. transformers models. The API calls the utility methods introduced above.
	* **`optimum/utils/quantization-api/`**

The API allows to launch so-called "runs" and "experiences" (the latter being a collection of runs). The main idea is that the API allows to optimize a transformers model, and evaluate it vs. the original model. Data is stored locally in a MongoDB database, and on the Hub (see e.g. https://huggingface.co/fxmarty/20220516-h11m07s05_).

For now, dataset preprocessing and evaluation are task-specific, the supported tasks are:
* `text-classification`
* `token-classification`
* `question-answering`

As for evaluation of transformers models, I believe there is some duplicate work with what exists in the AutoTrain backend and what is being done in https://github.com/huggingface/evaluate. However, my understanding being that it is not a priority to support Optimum-based inference within AutoTrain, it makes sense to me to have a common implementation to evaluate transformers/optimized models for them to be comparable. I hope we can make it such that we minimize duplicate efforts.

I used pipelines for inference for the general metrics, and `ORTModel.forward()` for measuring latency/throughput.

Probably the API itself should not be integrated into Optimum, this is up to discussion.

## Run locally

Into the optimum parent folder, run

```
docker build -f optimum/utils/quantization-api/Dockerfile -t api_quantization . && docker run -v $(pwd) /optimum:/code/optimum/optimum -v $(pwd)/optimum/utils/quantization-api/data/db:/data/db -v $(pwd)/optimum/utils/quantization-api/data/hf_cache:/root/.cache/huggingface --net=host -it api_quantization
```

The huggingface cache of the docker image is saved in `optimum/utils/quantization-api/data/hf_cache`, while the MongoDB database is saved in `optimum/utils/quantization-api/data/db`.

To rerun, don't forget to first get ownership of the data:
```
sudo chown -R $USER:$USER optimum/utils/quantization-api/data/
```

Inspect the data with
```
mongosh
> use quantization
> show collections
> db.runs.find()
```

Run examples from `optimum/quantization-api/examples/`.

## Tasks before (or after) merge
- [ ] Documentation
- [ ] Test on several datasets
- [ ]  See if it would make sense to use `Trainer.evaluate()` instead of an explicit loop for evaluation
- [ ]  Make use of `train-eval-index` metadata from datasets to auto-infer data, label columns (see e.g. https://github.com/huggingface/datasets/pull/4234)
- [ ]  Support multi-column data (2 would be sufficient I guess, see https://github.com/huggingface/transformers/issues/8573)
- [ ] Document node exclusion
- [ ] Support node exclusion for dynamic quantization for OnnxRuntime (implemented in https://github.com/huggingface/optimum/pull/166)
- [ ] Avoid tracking PyTorch to Numpy conversion for time measurements, in https://github.com/huggingface/optimum/blob/cf91bd7276714c7a39324c0bb3d2e57f820b0ad6/optimum/onnxruntime/modeling_ort.py#L323-L327
- [ ] Still some work to distinguish backend-agnostic code vs. backend-specific code, especially in the API part
- [ ] Extend the API (search)
- [ ] Clean all remainings # TODO

This should close #128 .


fxmarty:
test

PR #203, opened 2022-05-31
Align device and ORT provider in ORT models

### LINKED ISSUE #191: Not possible to configure GPU in pipelines nor leveraging batch_size parallelisation
When setting the `device` variable in the `pipeline` function/class to `>= 0`, an error appears `AttributeError: 'ORTModelForCausalLM' object has no attribute 'to' - when running in GPU`. This was initially reported in #161 so opening this issue to encompass supporting the `device` parameter in the ORT classes. This is important as otherwise it won't be possible to allow configuration of CPU/GPU similar to normal transformer libraries.

Is there currently a workaround to ensure that the class is run on GPU? By default it seems this woudl eb set in CPU even when GPU is available:
```python
>>> m = ORTModelForCausalLM.from_pretrained("gpt2", from_transformers=True)
>>> t = AutoTokenizer.from_pretrained("gpt2")
>>> pp = pipeline("text-generation", model=m, tokenizer=t)
>>> pp.device

device(type='cpu')
```

This is still the case even with the `optimum[onnxruntime-gpu]` package. I have validated by testing against a normal transformer with `batch_size=X` (ie `pp = pipeline("text-generation", model=m, tokenizer=t, batch_size=128)`) and it seems there is no optimization with parallel processing with optimum, whereas normal transformer is orders of magnitude faster (which is most likely as it's not utilizing the parallelism)

I can confirm that the model is loaded with GPU correctly:

```python
>>> m.device

device(type='cuda', index=0)
```

And GPU is configured correctly:

```python
>>> from optimum.onnxruntime.utils import _is_gpu_available
>>> _is_gpu_available()

True
```

Is there a way to enable GPU for processing with batching in optimum?

### LINKED ISSUE #213: Passing GPU Device to ORT Pipeline fails
I was trying to run a GPU accelerated optimum pipeline but passing device to the pipeline object fails in the following way

```python
from optimum.pipelines import pipeline as ort_pipeline

ort_pipeline(
  "text-classification", model=quantized_model, tokenizer=tokenizer, return_all_scores=True, device=0
)
```

returns error
`AttributeError: 'ORTModelForSequenceClassification' object has no attribute 'to'`

This works for the standard transformers pipeline object. I am exporting my quantized model which does not seem to support a `to_gpu` flag like the fit command does. Wondering how people export a quantized model then run it on GPU.

lewtun:
# What does this PR do?

This PR implements the following:

* Aligns the torch device used in `transformers` pipelines with the appropriate ONNX Runtime provider.
* Ensures input / output tensors are on the same device as the associated provider.
* Sets the default device / provider to CPU as I think this is more natural.

Partially fixes #191 (the batch parallelization issue requires a separate investigation)

Closes #213


philschmid (approved):
LGTM! But can we also add tests to check the device placement and movement on GPUs? 
For this we could create a new CI action similar to [training](https://github.com/huggingface/optimum/blob/main/.github/workflows/test_onnxruntime_train.yml) to start an AWS EC2 with GPU which installs onnxruntime-gpu? which runs on a reduced schedule?

lewtun:
I've now added a GPU CI workflow that runs daily, along with two sets of tests for ORT models:

* `test_model_on_gpu`
* `test_pipeline_on_gpu`

I'm not sure what's the best way to test the CI workflow will actually work once this is merged - any tips @philschmid?

.github/workflows/test_onnxruntime_gpu.yml
```diff
@@ -0,0 +1,85 @@
...
+    needs: start-runner # required to start the main job when the runner is ready
+    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    env:
+      AWS_REGION: us-east-1
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt -y update && sudo apt install -y libopencv-dev python3-opencv
```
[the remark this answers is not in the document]
philschmid: to build the documentation? Are we building the documentation?

.github/workflows/test_onnxruntime_gpu.yml
```diff
@@ -0,0 +1,85 @@
...
+          sudo pip install --upgrade pip && sudo pip install --upgrade protobuf
+          pip install .[onnxruntime-gpu]
+      - name: Run all tests on GPU
+        run: |
+          python -m pytest tests/onnxruntime/
+
```
philschmid, line 62: we can add a slack message here to get the failing test to some channel to test. And after the usage of optimum grew I think we can move then to an always-up GPU runner. https://github.com/slackapi/slack-github-action

philschmid (approved):
Awesome! Thank you for adding those tests!

tests/onnxruntime/testing_utils.py
```diff
@@ -0,0 +1,15 @@
+import unittest
+
+import torch
+
+from optimum.onnxruntime.utils import get_device_for_provider, get_provider_for_device
+
+
+class TestProviderAndDeviceGetters(unittest.TestCase):
```
echarlaix, line 8: We should rename this file `test_*.py` (https://docs.pytest.org/en/7.1.x/getting-started.html#run-multiple-tests) or adapt the onnxruntime workflows

echarlaix (approved):
Thanks for taking care of this @lewtun, it looks great !

.github/workflows/test_onnxruntime_gpu.yml
lewtun: I don't have bandwidth to setup the Slack webhook etc right now, so will leave this as a future improvement in the interest of getting this "bug" fixed.

I ran all the tests for ORT on a local GPU machine and they pass :)

PR #205, opened 2022-06-01
Bump `huggingface_hub` version to be able to use `evaluate` & `protobuf` fix

philschmid:
# What does this PR do?

Bump `huggingface_hub` version to be able to use evaluate


philschmid:
@echarlaix can we do a patch when this is merged? Since currently all `onnxruntime` installs are broken because of `protobuf`

lewtun (approved):
Thanks for the fix - LGTM!

PR #214, opened 2022-06-07
[Fix] Local Load and save for ORTModel for optimized/quantized models

philschmid:
# What does this PR do?

Currently when you load a locally saved optimized/quantized model (different file name than `model.onnx`) with `from_pretrained` it will correctly load but is not saving the "file" name for saving the model in a later state. Currently it always defaults to `model.onnx`. 

Reproduce:

```python
model = ORTModelForSequenceClassification.from_pretrained(onnx_path,file_name="model-quantized.onnx")
print(model.latest_model_name)
# model.onnx
model.save_pretrained(tmp_store_directory)
# save the model.onnx and not the model-quantized.onnx
```

Current workaround is to set `model.latest_model_name` to `file_name` after loading.


philschmid:
yes, let's add a test for it. I'll try to add it this week.

PR #215, opened 2022-06-07
Remove pydantic for benchmarks & use dataclass instead

fxmarty:
# What does this PR do?

Remove the previous pydantic dependency for the benchmark code. This PR allows to generate the documentation without additional dependencies.

## Before submitting


mfuntowicz (approved):
LGTM

PR #216, opened 2022-06-08
Graph mode quantization

michaelbenayoun:
# What does this PR do?

This PR adds transformers adapted versions of quantization functions present in PyTorch:

- `fuse_fx`: fuses modules before doing calibration / quantization aware training
- `prepare_fx`: prepares a model for calibration (post training static quantization)
- `prepare_qat_fx`: prepares a model for quantization aware training

https://github.com/huggingface/transformers/pull/17625 needs to be merged first.


optimum/fx/quantization.py
```diff
@@ -0,0 +1,232 @@
...
+
+if TYPE_CHECKING:
+    from torch.fx import Graph
+    from transformers import PreTrainedModel
+
+
+class QuantizationTracer(HFTracer):
+    specialized_concrete_args: Optional[Dict[str, Any]] = None
+
+    def __init__(self, skipped_module_names: List[str], skipped_module_classes: List[Callable]):
```
lewtun, line 38: Could we add a docstring here and also for the public methods?

optimum/fx/quantization.py
```diff
@@ -0,0 +1,232 @@
...
+
+@check_if_available
+def fuse_fx(
+    model: Union["PreTrainedModel", GraphModule],
+    fuse_custom_config_dict: Optional[Dict[str, Any]] = None,
+    input_names: Optional[List[str]] = None,
+    check: bool = True,
+) -> GraphModule:
+    """
+    Transformers models compatible version of torch.quantization.quantize_fx.fuse_fx, refer to PyTorch documentation
```
lewtun, line 100: Perhaps we can provide a link to the PyTorch docs to make it a bit easier for users?

optimum/fx/quantization.py
```diff
@@ -0,0 +1,232 @@
...
+    for more details.
+
+    Args:
+        model (`PreTrainedModel` or `torch.fx.GraphModule`):
+            The model to fuse.
+        fuse_custom_config_dict (`Dict[str, Any]`, *optional*):
+            Refer to PyTorch documentation.
```
lewtun, line 107: Could we add a one-liner to explain what this arg is, followed by a link to the PyTorch docs?

optimum/fx/quantization.py
```diff
@@ -0,0 +1,232 @@
...
+        self.skipped_module_classes = skipped_module_classes
+        # NB: initialized the module_type of top level module to None
+        # we are assuming people won't configure the model with the type of top level
+        # module here, since people can use "" for global config
+        # We can change this if there is a use case that configures
+        # qconfig using top level module type
+        self.scope = Scope("", None)
+        self.node_name_to_scope: Dict[str, Tuple[str, type]] = {}
+
+    def is_leaf_module(self, m: torch.nn.Module, module_qualified_name: str) -> bool:
```
lewtun, line 50: nit: I think it would be good to avoid using single-letter variable names in the codebase if possible. would something like the following work?

```suggestion
    def is_leaf_module(self, module: torch.nn.Module, module_qualified_name: str) -> bool:
```

optimum/fx/quantization.py
```diff
@@ -0,0 +1,232 @@
...
+        return (
+            (m.__module__.startswith("torch.nn") and not isinstance(m, torch.nn.Sequential))
+            or module_qualified_name in self.skipped_module_names
+            or type(m) in self.skipped_module_classes
+            or isinstance(m, _FusedModule)
+            or super().is_leaf_module(m, module_qualified_name)
+        )
+
+    def call_module(
+        self, m: torch.nn.Module, forward: Callable[..., Any], args: Tuple[Any, ...], kwargs: Dict[str, Any]
```
lewtun, line 60: Same comment here about using `module` instead of `m`

optimum/fx/quantization.py
```diff
@@ -0,0 +1,232 @@
...
+        `torch.fx.GraphModule`: A GraphModule with the fused modules.
+    """
+    if not isinstance(model, GraphModule):
+        if input_names is None:
+            input_names = model.dummy_inputs.keys()
+        input_names = list(input_names)
+        model = symbolic_trace(model, input_names, disable_check=not check)
+    orig_symbolic_trace = torch.fx.symbolic_trace
+    torch.fx.symbolic_trace = lambda x: x
+    gm = orig_fuse_fx(model, fuse_custom_config_dict=fuse_custom_config_dict)
```
lewtun, line 124: nit:
```suggestion
    graph_module = orig_fuse_fx(model, fuse_custom_config_dict=fuse_custom_config_dict)
```

optimum/fx/quantization.py
```diff
@@ -0,0 +1,232 @@
...
+        check_if_model_is_supported(model)
+    tracer_cls = QuantizationTracer
+    if not isinstance(model, GraphModule):
+        if input_names is None:
+            input_names = model.dummy_inputs.keys()
+        input_names = list(input_names)
+        tracer_cls = specialized_quantization_tracer_creator(get_concrete_args(model, input_names))
+    orig_quantization_tracer = torch.ao.quantization.quantize_fx.QuantizationTracer
+    torch.ao.quantization.quantize_fx.QuantizationTracer = tracer_cls
+    gm = orig_prepare_fx(
```
lewtun, line 173: nit:
```suggestion
    graph_module = orig_prepare_fx(
```

optimum/fx/quantization.py
```diff
@@ -0,0 +1,232 @@
...
+        check_if_model_is_supported(model)
+    tracer_cls = QuantizationTracer
+    if not isinstance(model, GraphModule):
+        if input_names is None:
+            input_names = model.dummy_inputs.keys()
+        input_names = list(input_names)
+        tracer_cls = specialized_quantization_tracer_creator(get_concrete_args(model, input_names))
+    orig_quantization_tracer = torch.ao.quantization.quantize_fx.QuantizationTracer
+    torch.ao.quantization.quantize_fx.QuantizationTracer = tracer_cls
+    gm = orig_prepare_qat_fx(
```
lewtun, line 225: nit:
```suggestion
    graph_module = orig_prepare_qat_fx(
```

optimum/fx/quantization.py
michaelbenayoun: It's basically from PyTorch codebase, and it was to keep the same signature, I can change that.

optimum/fx/quantization.py
michaelbenayoun: Same answer.

optimum/fx/quantization.py
michaelbenayoun: Is it needed since I'm just overriding methods that already have a docstring in the base class?

lewtun: Ah sorry I missed that - indeed we can ignore them here in that case

lewtun:
> About the guide, I am not sure this is needed for this, because it's just basically allowing what PyTorch enables on transformers models. That being said, when we add a higher level API, with a guide explaining how to use it, we can also add a section in it explaining how to do things at a lower level using those functions?

I see, so this is just a first pass at defining the low-level functionality and we don't expect users to use it? If yes, then I think it's fine to proceed without docs if none of this will be public-facing.



> Finally, to install it I guess a user would do `pip install optimum`.

Sounds good!

michaelbenayoun:
> > About the guide, I am not sure this is needed for this, because it's just basically allowing what PyTorch enables on transformers models. That being said, when we add a higher level API, with a guide explaining how to use it, we can also add a section in it explaining how to do things at a lower level using those functions?
> 
> I see, so this is just a first pass at defining the low-level functionality and we don't expect users to use it? If yes, then I think it's fine to proceed without docs if none of this will be public-facing.

It will be public-facing, but provide nothing new from what PyTorch already provides.

lewtun:
> It will be public-facing, but provide nothing new from what PyTorch already provides.

OK, but then my main issue is that without documentation, how will `optimum` users know this feature exists? Sorry if I'm misunderstanding something - could you share a code snippet on how this optimization would work in practice?

PR #218, opened 2022-06-10
Fix documentation dependency

fxmarty:
# What does this PR do?

Install the dependencies on `optuna` and `tqdm` before generating the documentation (necessary for the benchmark part). For some reason the action `build_pr_documentation` passed (see https://github.com/huggingface/optimum/actions/runs/2455809780 or this PR dev doc), but the `build_documentation` fails because of missing dependencies still (760a7e6e4aba963feefd1fae86fa8590d1200ada).

## Before submitting


mfuntowicz (approved):
LGTM.

Thanks for fixing the doc issue ❤️

PR #225, opened 2022-06-20
[ORT] Filter out invalid inputs in ORTModelForXXX forward pass

### LINKED ISSUE #207: Add Support for DeBERTaV2
I would like to use DeBERTaV2 for sequence classification as a quantized model. Please let me know what needs to be done to open a PR to add this support!

JingyaHuang:
# Context

__TL;DR__ 

[Transformers #17617](https://github.com/huggingface/transformers/pull/17617)

__Long story to tell...__
For DeBERTa model, the tokenizer gives out `token_type_ids` by default. However, the exported IR might not contain `token_type_ids`(eg. the case when `config.type_vocab_size=0` if exported by `transformers.onnx.export`). In this situation:
1.  The forward pass will fail if the user takes directly the output as input(as our snippet does). 
2. Otherwise, they need to add another line to filter out invalid input themselves which needs a deeper understanding of the model and its tokenizer. 

Considering the user experience, I think that we shall add this filter directly in the ORTModelForXXX.

# What does this PR do?

* Filter out invalid inputs in ORTModelForXXX.

Fixes #207


philschmid:
>  However, the exported IR might not contain token_type_ids(eg. the case when config.type_vocab_size=0 if exported by transformers.onnx.export). In this situation:

When the tokenizers return `token_type_ids` by default why aren't we exporting them in `transformers.onnx.export` by default too? We are doing similar things for `GPT2` which also always return `attention_mask` in the IE for batching.

JingyaHuang:
> > However, the exported IR might not contain token_type_ids(eg. the case when config.type_vocab_size=0 if exported by transformers.onnx.export). In this situation:
> 
> When the tokenizers return `token_type_ids` by default why aren't we exporting them in `transformers.onnx.export` by default too? We are doing similar things for `GPT2` which also always return `attention_mask` in the IE for batching.

In the implementation of DeBERTa there exists [a control flow to decide whether the embedding will be instantiated](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deberta_v2/modeling_deberta_v2.py#L824), in the case of `config.type_vocab_size=0`, the `token_type_ids` will not be traced even when it exists in the onnx_config.inputs, which is different to the case of GPT-2.

philschmid:
> In the implementation of DeBERTa there exists [a control flow to decide whether the embedding will be instantiated](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deberta_v2/modeling_deberta_v2.py#L824), in the case of config.type_vocab_size=0, the token_type_ids will not be traced even when it exists in the onnx_config.inputs, which is different to the case of GPT-2.

How is transformers and regular pipeline currently handling this situation? Is there a way to commit the `token_type_ids` in the `tokenizer` for example?

JingyaHuang:
> > In the implementation of DeBERTa there exists [a control flow to decide whether the embedding will be instantiated](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deberta_v2/modeling_deberta_v2.py#L824), in the case of config.type_vocab_size=0, the token_type_ids will not be traced even when it exists in the onnx_config.inputs, which is different to the case of GPT-2.
> 
> How is transformers and regular pipeline currently handling this situation? Is there a way to commit the `token_type_ids` in the `tokenizer` for example?

I don't think that the pipeline in transformers handles this case, as `token_type_ids` is truly optional for PyTorch model. But for our case, as `InferenceSession` is not tolerant of invalid inputs, the pipeline might fail. And for the tokenizer, as it is independent of config and the IR whether contains `token_type_ids` depends on `config.type_vocab_size`, I can't see an elegant way to work around it. Any idea?

JingyaHuang:
> > I don't think that the pipeline in transformers handles this case, as token_type_ids is truly optional for PyTorch model. But for our case, as InferenceSession is not tolerant of invalid inputs, the pipeline might fail. And for the tokenizer, as it is independent of config and the IR whether contains token_type_ids depends on config.type_vocab_size, I can't see an elegant way to work around it. Any idea?
> 
> Maybe adjusting/configuring the tokenizer to output `token_type_ids` or not.

It makes sense to me! Though the case is a little tricky for users, you are right that it is hacky to introduce it into ORTModel which has a bad impact on the speed. 

And for the configuration of the tokenizer for the inference, 
* Case 1: users use `ORTModelForXXX` for inference
This case is ok, users can configure `return_token_type_ids` to avoid the problem.

```diff
 tokenizer = {processor_class}.from_pretrained("{checkpoint}")
model = {model_class}.from_pretrained("{checkpoint}")
-inputs = tokenizer("My name is Philipp and I live in Germany.", return_tensors="pt")
+inputs = tokenizer("My name is Philipp and I live in Germany.", return_tensors="pt", return_token_type_ids=False)
outputs = model(**inputs)
```

* Case 2: users use pipeline

In this case, as users can't configure `token_type_ids` when instantiating the tokenizer, and the `preprocess` function [doesn't take `return_token_type_ids` into consideration] (https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/token_classification.py#L195-L200), the pipeline will fail when the model doesn't contain `token_type_ids` as input.

Possible solution:
Add `return_token_type_ids` argument to `preprocess` function of each pipeline in transformers and set to False when `self.model.config.type_vocab_size < 1`. What do you think about that?

JingyaHuang:
Issue opened in transformers to continue the discussion. Thanks @philschmid!

JingyaHuang:
Close as this should be a work in transformers.

PR #226, opened 2022-06-20
Add support for image classification

### LINKED ISSUE #173: Support for computer vision tasks
Thanks for the amazing library! It would be great to add support for computer vision tasks to 🤗 optimum!

In particular, it would be amazing to use the new inference pipelines to take advantage of ONNX for inference when using image classification models. 

This is partially selfishly motivated for use in (https://github.com/davanstrien/flyswot/) which is being used to 'deploy' transformer models via command line (running on CPU). I have previously had good success using onnx to speed things up but found the process of exporting onnx models, getting data ready for inference etc. not very fun. The API for pipelines looks like it will make this task much more enjoyable and flexible so it would be amazing to have support for image classification models in optimum.

regisss:
# What does this PR do?

Fixes #173. This PR enables:
- to use an image-classification pipeline
- to perform quantization/optimization with computer vision models

At the moment, static post-training quantization with percentile/entropy calibration may not work with some models due to an issue with ONNX Runtime.

## Before submitting

### DIFF
--- a/docs/source/onnxruntime/modeling_ort.mdx
+++ b/docs/source/onnxruntime/modeling_ort.mdx
@@ -12,13 +12,13 @@ specific language governing permissions and limitations under the License.
 
 # Optimum Inference with ONNX Runtime
 
-Optimum is a utility package for building and running inference with accelerated runtime like ONNX Runtime. 
-Optimum can be used to load optimized models from the [Hugging Face Hub](hf.co/models) and create pipelines 
+Optimum is a utility package for building and running inference with accelerated runtime like ONNX Runtime.
+Optimum can be used to load optimized models from the [Hugging Face Hub](hf.co/models) and create pipelines
 to run accelerated inference without rewriting your APIs.
 
 ## Switching from Transformers to Optimum Inference
 
-The Optimum Inference models are API compatible with Hugging Face Transformers models. This means you can just replace your `AutoModelForXxx` class with the corresponding `ORTModelForXxx` class in `optimum`. For example, this is how you can use a question answering model in `optimum`: 
+The Optimum Inference models are API compatible with Hugging Face Transformers models. This means you can just replace your `AutoModelForXxx` class with the corresponding `ORTModelForXxx` class in `optimum`. For example, this is how you can use a question answering model in `optimum`:
 
 ```diff
 from transformers import AutoTokenizer, pipeline
@@ -57,8 +57,8 @@ You can find a complete walkhrough Optimum Inference for ONNX Runtime in this [n
 
 ### Working with the Hugging Face Model Hub
 
-The Optimum model classes like [`~onnxruntime.ORTModelForSequenceClassification`] are integrated with the [Hugging Face Model Hub](https://hf.co/models), which means you can not only 
-load model from the Hub, but also push your models to the Hub with `push_to_hub()` method. Below is an example which downloads a vanilla Transformers model 
+The Optimum model classes like [`~onnxruntime.ORTModelForSequenceClassification`] are integrated with the [Hugging Face Model Hub](https://hf.co/models), which means you can not only
+load model from the Hub, but also push your models to the Hub with `push_to_hub()` method. Below is an example which downloads a vanilla Transformers model
 from the Hub and converts it to an optimum onnxruntime model and pushes it back into a new repository.
 
 <!-- TODO: Add Quantizer into example when UX improved -->
@@ -104,4 +104,8 @@ from the Hub and converts it to an optimum onnxruntime model and pushes it back 
 ## ORTModelForCausalLM
 
 [[autodoc]] onnxruntime.modeling_ort.ORTModelForCausalLM
+
+## ORTModelForImageClassification
+
+[[autodoc]] onnxruntime.modeling_ort.ORTModelForImageClassification
 
--- a/docs/source/pipelines.mdx
+++ b/docs/source/pipelines.mdx
@@ -12,7 +12,7 @@ specific language governing permissions and limitations under the License.
 
 # Optimum pipelines for inference
 
-The [`~pipelines.pipeline`] function makes it simple to use models from the [Model Hub](https://huggingface.co/models) for accelerated inference on a variety of tasks such as text classification.
+The [`~pipelines.pipeline`] function makes it simple to use models from the [Model Hub](https://huggingface.co/models) for accelerated inference on a variety of tasks such as question answering or image classification.
 Even if you don't have experience with a specific modality or understand the code powering the models, you can still use them with the [`~pipelines.pipeline`] function!
 
 <Tip>
@@ -31,11 +31,12 @@ Currenlty supported tasks are:
 * `question-answering`
 * `zero-shot-classification`
 * `text-generation`
+* `image-classification`
 
 ## Optimum pipeline usage
 
-While each task has an associated pipeline class, it is simpler to use the general [`~pipelines.pipeline`] function which wraps all the task-specific pipelines in one object. 
-The [`~pipelines.pipeline`] function automatically loads a default model and tokenizer capable of inference for your task. 
+While each task has an associated pipeline class, it is simpler to use the general [`~pipelines.pipeline`] function which wraps all the task-specific pipelines in one object.
+The [`~pipelines.pipeline`] function automatically loads a default model and tokenizer/feature-extractor capable of inference for your task.
 
 1. Start by creating a pipeline by specifying an inference task:
 
@@ -46,7 +47,7 @@ The [`~pipelines.pipeline`] function automatically loads a default model and tok
 
 ```
 
-2. Pass your input text to the [`~pipelines.pipeline`] function:
+2. Pass your input text/image to the [`~pipelines.pipeline`] function:
 
 ```python
 >>> classifier("I like you. I love you.")
@@ -57,9 +58,9 @@ _Note: The default models used in the [`~pipelines.pipeline`] function are not o
 
 ### Using vanilla Transformers model and converting to ONNX
 
-The [`~pipelines.pipeline`] function accepts any supported model from the [Model Hub](https://huggingface.co/models). 
-There are tags on the Model Hub that allow you to filter for a model you'd like to use for your task. 
-Once you've picked an appropriate model, load it with the `from_pretrained("{model_id}",from_transformers=True)` method associated with the `ORTModelFor*` 
+The [`~pipelines.pipeline`] function accepts any supported model from the [Model Hub](https://huggingface.co/models).
+There are tags on the Model Hub that allow you to filter for a model you'd like to use for your task.
+Once you've picked an appropriate model, load it with the `from_pretrained("{model_id}",from_transformers=True)` method associated with the `ORTModelFor*`
 `AutoTokenizer' class. For example, here's how you can load the [`~onnxruntime.ORTModelForQuestionAnswering`] class for question answering:
 
 ```python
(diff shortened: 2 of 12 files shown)


optimum/onnxruntime/quantization.py
```diff
@@ -279,7 +279,7 @@ def export(
         quantization_config: QuantizationConfig,
         calibration_tensors_range: Optional[Dict[NodeName, Tuple[float, float]]] = None,
         use_external_data_format: bool = False,
-        quantization_preprocessor: Optional[QuantizationPreprocessor] = None,
+        preprocessor: Optional[QuantizationPreprocessor] = None,
```
philschmid, line 282: wouldn't this break backward compatibility?

docs/source/onnxruntime/modeling_ort.mdx
```diff
@@ -12,13 +12,13 @@ specific language governing permissions and limitations under the License.
 
 # Optimum Inference with ONNX Runtime
 
-Optimum is a utility package for building and running inference with accelerated runtime like ONNX Runtime. 
-Optimum can be used to load optimized models from the [Hugging Face Hub](hf.co/models) and create pipelines 
+Optimum is a utility package for building and running inference with accelerated runtime like ONNX Runtime.
+Optimum can be used to load optimized models from the [Hugging Face Hub](hf.co/models) and create pipelines
```
philschmid, line 16: are you using some kind of lint/prettier? (Just asking)

docs/source/pipelines.mdx
```diff
@@ -198,16 +199,16 @@ Below you can find two examples on how you could [`~onnxruntime.ORTOptimizer`] a
...
+ `ORTModelFor*` class.
 
 ```diff
 from transformers import AutoTokenizer, pipeline
 -from transformers import AutoModelForQuestionAnswering
 +from optimum.onnxruntime import ORTModelForQuestionAnswering
 
 -model = AutoModelForQuestionAnswering.from_pretrained("deepset/roberta-base-squad2")
-+model = ORTModelForQuestionAnswering.from_transformers("optimum/roberta-base-squad2")
++model = ORTModelForQuestionAnswering.from_pretrained("optimum/roberta-base-squad2")
```
philschmid, line 211: Thanks for fix!

tests/onnxruntime/test_modeling_ort.py
```diff
@@ -286,15 +290,15 @@ def test_supported_transformers_architectures(self, *args, **kwargs):
 
     def test_load_vanilla_transformers_which_is_not_supported(self):
         with self.assertRaises(Exception) as context:
-            model = ORTModelForTokenClassification.from_pretrained("t5-small", from_transformers=Tru)
```
philschmid, line 289: the idea was here to use a "architecture" which we don't support with `from_transformers`

tests/onnxruntime/test_modeling_ort.py
```diff
@@ -357,15 +361,15 @@ def test_supported_transformers_architectures(self, *args, **kwargs):
 
     def test_load_vanilla_transformers_which_is_not_supported(self):
         with self.assertRaises(Exception) as context:
-            model = ORTModelForFeatureExtraction.from_pretrained("google/vit-base-patch16-224", from_transformers=Tru)
```
philschmid, line 360: the idea was here to use a "architecture" which we don't support with `from_transformers`

philschmid (approved):
LGTM!🔥✅ I left a few comments. Mostly regarding breaking backward compatibility. 
Other than this awesome PR! 🔥

optimum/onnxruntime/modeling_ort.py
```diff
@@ -730,3 +733,87 @@ def _prepare_attention_mask_for_generation(
         else:
             # Ensure attention mask is on the same device as the input IDs
             return torch.ones(inputs.shape[:2], dtype=torch.long, device=inputs.device)
+
+
+IMAGE_CLASSIFICATION_SAMPLE = r"""
```
michaelbenayoun, line 738: A bit nitpicking, but isn't it an "example" more than a "sample"?

michaelbenayoun (approved):
Great, thanks a lot for adding this @regisss !
Left a few comments, but those are minor stuff, LGTM!

docs/source/onnxruntime/modeling_ort.mdx
regisss: I switched on the setting in VSCode to trim extra whitespaces. But I can revert this if you think it should not be done in this PR.

optimum/onnxruntime/modeling_ort.py
regisss: I fully agree with you, I just copy-pasted this from other similar classes XD
I'm going to push a new commit to modify this :)

tests/onnxruntime/test_modeling_ort.py
regisss: Yep you're right, this is not the change I wanted to push actually XD
I wanted to add a "e" at `from_transformers=Tru`, I'm going to push this in a new commit.

optimum/onnxruntime/quantization.py
regisss: Yep it would, that's why I pushed a second commit to revert this. I think you did the review before I pushed this second commit, now this change should not appear anymore.

tests/onnxruntime/test_modeling_ort.py
regisss: Same here

docs/source/onnxruntime/modeling_ort.mdx
JingyaHuang: Hehe, I asked the same question before.

docs/source/pipelines.mdx
```diff
@@ -12,7 +12,7 @@ specific language governing permissions and limitations under the License.
 
 # Optimum pipelines for inference
 
-The [`~pipelines.pipeline`] function makes it simple to use models from the [Model Hub](https://huggingface.co/models) for accelerated inference on a variety of tasks such as text classification.
+The [`~pipelines.pipeline`] function makes it simple to use models from the [Model Hub](https://huggingface.co/models) for accelerated inference on a variety of tasks such as question answering or image classification.
```
JingyaHuang, line 15:
```suggestion
The [`~pipelines.pipeline`] function makes it simple to use models from the [Model Hub](https://huggingface.co/models) for accelerated inference on a variety of tasks such as text classification, question answering, and image classification.
```
I think that it makes sense to list them all, as they are typical and quite different examples.

docs/source/pipelines.mdx
```diff
@@ -198,16 +199,16 @@ Below you can find two examples on how you could [`~onnxruntime.ORTOptimizer`] a
 ## Transformers pipeline usage
 
 The [`~pipelines.pipeline`] function is just a light wrapper around the `transformers.pipeline` function to enable checks for supported tasks and additional features
-, like quantization and optimization. This being said you can use the `transformers.pipeline` and just replace your `AutoFor*` with the optimum 
- `ORTModelFor*` class. 
+, like quantization and optimization. This being said you can use the `transformers.pipeline` and just replace your `AutoFor*` with the optimum
```
JingyaHuang, line 202:
```suggestion
, like quantization and optimization. This being said you can use the `transformers.pipeline` and just replace your `AutoModelFor*` with the optimum
```

optimum/onnxruntime/optimization.py
```diff
@@ -54,22 +55,23 @@ def from_pretrained(
         Returns:
             An instance of `ORTOptimizer`.
         """
-        tokenizer = AutoTokenizer.from_pretrained(model_name_or_path)
+        preprocessor = get_preprocessor(model_name_or_path)
```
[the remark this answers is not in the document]
regisss: Actually it is from what I see [here](https://github.com/huggingface/transformers/blob/b6a65ae52a6df67180be8f1e162a3d1829a51bc0/src/transformers/onnx/utils.py#L71).

Regarding a unified type called `Preprocessor`, I'm in favour of it but on the other hand it is not clear in Transformers either and we also have a different kind of preprocessor for quantization. So I feel like it will always be quite ambiguous.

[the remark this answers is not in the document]
michaelbenayoun: I agree with you on the second point, maybe explicit is better for the end-user. For the first point, I think we need to figure it out!

optimum/onnxruntime/optimization.py
```diff
@@ -117,7 +119,7 @@ def export(
 
         # Export the model if it has not already been exported to ONNX IR
         if not onnx_model_path.exists():
-            export(self.tokenizer, self.model, self._onnx_config, self.opset, onnx_model_path)
+            export(self.preprocessor, self.model, self._onnx_config, self.opset, onnx_model_path)
```
JingyaHuang, line 122: I think that it would be better to add the keyword arguments on, although the term `preprocessor` doesn't exist for older version of transformers. WDYT?

JingyaHuang (approved):
LGTM(G means great here). :>

Just added some small nits.

tests/onnxruntime/test_modeling_ort.py
regisss: Thinking about it a bit more, I should actually remove `from_transformers=True` here right @philschmid ? Or setting it to `False`

philschmid: Yes!

optimum/onnxruntime/modeling_ort.py
philschmid: The naming is coming from transformers

tests/onnxruntime/test_modeling_ort.py
regisss: Okay I understand better. There was a bug in all the `test_load_vanilla_transformers_which_is_not_supported`. Since it was `from_transformers=Tru` and not `from_transformers=True`, an exception was raised as expected. However, then, `self.assertTrue("Unrecognized configuration class", context.exception)` will always pass because it assesses the truthness of the 1st argument (`"Unrecognized configuration class"` here). I guess what we want to do here is to check if `"Unrecognized configuration class"` is in `context.exception`. I'm going to fix this.

regisss: Also, for the feature extraction task, all models are supported (see [here](https://github.com/huggingface/transformers/blob/f47afefb210a120545a5e9745292aaf6c316d246/src/transformers/pipelines/feature_extraction.py#L15)) so I will just remove this test for this task.

docs/source/pipelines.mdx
regisss: Actually all the supported tasks are listed a few lines below. My point was to show that it is not only NLP and not only classification (which is why I replaced "text classification" by "question answering"). But your suggestion is fine, I'm going to integrate it :)

optimum/onnxruntime/optimization.py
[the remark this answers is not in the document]
regisss: Actually `AutoFeatureExtractor` and `AutoProcessor` will both return a `FeatureExtractorMixin` object.

optimum/onnxruntime/optimization.py
michaelbenayoun: The constraint would be transformers=4.7+ ?

regisss: Keeping it as it is for the moment to not break backwards compatibility.

davanstrien:
Thanks so much for doing this -- really happy to see this feature added 🤗

regisss:
> Thanks so much for doing this -- really happy to see this feature added hugs

Do not hesitate to show us how you use it :) @davanstrien

rhkenne:
Thank you very much for the great work. The description of this PR mentioned "perform quantization/optimization with computer vision models". Does this include CLIP model (CLIPProcessor, CLIPModel) ? I am looking for an approach to serialize CLIP models (to onnx) and perform quantization/optimization. Any idea how ?

regisss:
> Thank you very much for the great work. The description of this PR mentioned "perform quantization/optimization with computer vision models". Does this include CLIP model (CLIPProcessor, CLIPModel) ? I am looking for an approach to serialize CLIP models (to onnx) and perform quantization/optimization. Any idea how ? @NielsRogge @JingyaHuang

@rhkenne The first step would be to add ONNX support for CLIP in Transformers. Would you be interested in opening a PR to enable this?

Then, for optimization and quantization, additional work would be necessary because multimodal models are not supported yet.

regisss:
@rhkenne There is an ongoing PR [here](https://github.com/huggingface/transformers/pull/18515) to make CLIP exportable to ONNX.

PR #230, opened 2022-06-22
Modify the examples optimization level

echarlaix:
In this PR, we modify the onnx runtime optimization level in the examples in the quickstart and different READMEs for users to create hardware agnostic optimized graph.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,10 +154,9 @@ Then let's take a look at applying _graph optimizations_ techniques such as oper
 ```python
 from optimum.onnxruntime.configuration import OptimizationConfig
 
-# Here the optimization level is selected to be 1, enabling
-# basic optimizations such as redundant node eliminations
-# and constant folding. Higher optimization level will result 
-# in a hardware dependent optimized graph.
+# Here the optimization level is selected to be 1, enabling basic optimizations such as redundant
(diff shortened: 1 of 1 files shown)


tests/onnxruntime/test_onnxruntime.py
```diff
@@ -53,7 +53,7 @@ def test_optimize(self):
             "roberta-base",
             "google/electra-small-discriminator",
         }
-        optimization_config = OptimizationConfig(optimization_level=99, optimize_with_onnxruntime_only=False)
+        optimization_config = OptimizationConfig(optimization_level=2, optimize_with_onnxruntime_only=False)
```
regisss, line 56: Why 2 and not 1 here for `optimization_level`?

regisss (approved):
It looks good to me @echarlaix !!
I just have one question because I'm curious

tests/onnxruntime/test_onnxruntime.py
echarlaix: For the tests, I think it's better to verify that even the extended graph optimizations don't impact too much the models outputs. The optimization level of [99](https://onnxruntime.ai/docs/performance/graph-optimizations.html#layout-optimizations) doesn't seem useful in our use cases.

PR #231, opened 2022-06-22
Add unit tests for benchmark code

fxmarty:
* Add unit tests for text classification, token classification, question answering benchmark code. We check that the transformers models metrics match with those from `transformers/examples/pytorch/`, and that the quantized models metrics match with those from `optimum/examples/onnxruntime/quantization/`
* Fix text classification inference using pipeline according to https://github.com/huggingface/transformers/pull/17324

A next PR will be necessary to do some cleaning (add options to skip time tracking, select number of eval samples), but this is good for now.

## Before submitting


.github/workflows/test_benckmark.yml
```diff
@@ -0,0 +1,38 @@
...
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.8, 3.9]
+        os: [ubuntu-20.04, macos-10.15]
```
mfuntowicz, line 21: I would add Windows here

optimum/utils/runs.py
```diff
@@ -56,6 +57,15 @@ class QuantizationApproach(str, Enum):
     dynamic = "dynamic"
 
 
+def cpu_info_command():
+    if platform.system() == "Linux":
+        return "lscpu"
+    elif platform.system() == "Darwin":
+        return "sysctl -a | grep machdep.cpu"
+    else:
+        raise NotImplementedError("OS not supported.")
```
mfuntowicz, line 66: Maybe something like this for Windows?

```bash
$ wmic cpu get caption, deviceid, name, numberofcores, maxclockspeed, status
Caption                                DeviceID  MaxClockSpeed  Name                                   NumberOfCores  Status
Intel64 Family 6 Model 151 Stepping 2  CPU0      3600           12th Gen Intel(R) Core(TM) i7-12700KF  12             OK
```

mfuntowicz (approved):
LGTM 👍🏻 

What do you think about having `runs_base,py` inside the `benchmark` package in order to match with the name of the optional deps?

PR #232, opened 2022-06-22
Fx transformations

michaelbenayoun:
# What does this PR do?

- Provides an API to write graph transformations in optimum, a transformation is either:
  - A function that maps a graph module to a graph module
  - A `ReversibleTransformation` that implements two methods: `transform` and `reverse`, a transformation that can be reversed to get the original model back
 - An utility function, `compose`, allowing to create the composition of multiple transformations easily 
- A `ReversibleTransformation` that merges linear layers with the same input together
- A `ReversibleTransformation` that replaces the division by a static value by the multiplication by the inverse of the static value


optimum/fx/optimization/transformations.py
```diff
@@ -0,0 +1,308 @@
...
+        graph_module.graph.lint()
+        graph_module.recompile()
+        return graph_module
+
+
+def compose(*args: Transformation, inplace: bool = True) -> Callable[["GraphModule"], "GraphModule"]:
+    """
+    Composes a list of transformations together.
+    """
+    transformations = list(reversed(args))
```
regisss, line 283: Why do we need to reverse the list of transformations here?

regisss (approved):
Not a FX expert but LGTM!! Thanks @michaelbenayoun

optimum/fx/optimization/transformations.py
michaelbenayoun: That is because doing `composition = compose(f, g, h)` without reversing the list, the composition corresponds to `f o g o h`. So by reversing, the composition order happens in the same order as the one provided in the arguments.

PR #238, opened 2022-06-24
Add image classification quantization example (onnxruntime)

fxmarty:
# What does this PR do?

Add an example script for image classification, adapted from existing quantization scripts and https://github.com/huggingface/transformers/blob/09178705101b9803e7b9ea7f79a46b4c242dd4bf/examples/pytorch/image-classification/run_image_classification_no_trainer.py .

In a future PR, we could remove https://github.com/huggingface/optimum/blob/main/optimum/onnxruntime/model.py and use `ORTModelFor...` in examples accordingly. 

@regisss Let me know if you think this is good.


examples/onnxruntime/quantization/image-classification/run_image_classification.py
```diff
@@ -0,0 +1,434 @@
...
+        preds = p.predictions[0] if isinstance(p.predictions, tuple) else p.predictions
+        preds = np.argmax(preds, axis=1)
+
+        result = metric.compute(predictions=preds, references=p.label_ids)
+        return result
+
+    # Create the quantizer
+    quantizer = ORTQuantizer(
+        preprocessor=feature_extractor, model=model, feature="image-classification", opset=optim_args.opset
+    )
```
regisss, line 312: I think we should rather push the usage of `ORTQuantizer.from_pretrained` no?

examples/onnxruntime/quantization/image-classification/run_image_classification.py
```diff
@@ -0,0 +1,434 @@
...
+        calibration_dataset = quantizer.clean_calibration_dataset(calibration_dataset)
+
+        # Create the calibration configuration given the selected calibration method
+        if optim_args.calibration_method == "entropy":
+            calibration_config = AutoCalibrationConfig.entropy(calibration_dataset)
+        elif optim_args.calibration_method == "percentile":
+            calibration_config = AutoCalibrationConfig.percentiles(
+                calibration_dataset,
+                percentile=optim_args.calibration_histogram_percentile,
+            )
```
regisss, line 354: Depending on the operators we wish to quantize, entropy and percentile calibration may return an error on ORT's side. Just to be sure, do the 3 calibration methods work with the calibration config `qconfig` defined above?

regisss (commented):
No major comments, I just want to make sure the default quantization configuration runs well with any calibration methods. Doing this because I got issues when quantizing completely a Resnet-50 with entropy/percentile calibration.

examples/onnxruntime/quantization/image-classification/run_image_classification.py
fxmarty: There was the argument `ignore_mismatched_sizes` when loading the model in the transformers example, so that's why I loaded the model and quantizer manually. Following text-classification (where there is the same arg in transformers examples but not in Optimum), I removed it and used `ORTQuantizer.from_pretrained`.

examples/onnxruntime/quantization/image-classification/run_image_classification.py
fxmarty: Indeed, we do get an error for percentile (at least):

```
Collecting tensor data and making histogram ...
/home/fxmarty/anaconda3/envs/hf-inf/lib/python3.9/site-packages/onnxruntime/quantization/calibrate.py:539: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  data_arr = np.asarray(data_arr)
  File "/home/fxmarty/hf_internship/optimum/examples/onnxruntime/quantization/image-classification/run_image_classification.py", line 432, in <module>
    main()
  File "/home/fxmarty/hf_internship/optimum/examples/onnxruntime/quantization/image-classification/run_image_classification.py", line 369, in main
    quantizer.partial_fit(
  File "/home/fxmarty/hf_internship/optimum/optimum/onnxruntime/quantization.py", line 260, in partial_fit
    self._calibrator.collect_data(reader)
  File "/home/fxmarty/anaconda3/envs/hf-inf/lib/python3.9/site-packages/onnxruntime/quantization/calibrate.py", line 391, in collect_data
    self.collector.collect(clean_merged_dict)
  File "/home/fxmarty/anaconda3/envs/hf-inf/lib/python3.9/site-packages/onnxruntime/quantization/calibrate.py", line 502, in collect
    return self.collect_value(name_to_arr)
  File "/home/fxmarty/anaconda3/envs/hf-inf/lib/python3.9/site-packages/onnxruntime/quantization/calibrate.py", line 543, in collect_value
    min_value = np.min(data_arr)
  File "<__array_function__ internals>", line 5, in amin
  File "/home/fxmarty/anaconda3/envs/hf-inf/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 2879, in amin
    return _wrapreduction(a, np.minimum, 'min', axis, None, out,
  File "/home/fxmarty/anaconda3/envs/hf-inf/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 86, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

Most likely due to the warning when exporting to onnx ( https://github.com/huggingface/transformers/blob/401fcca6c561d61db6ce25d9b1cebb75325a034f/src/transformers/models/vit/modeling_vit.py#L171 )
```
/home/fxmarty/hf_internship/transformers/src/transformers/models/vit/modeling_vit.py:187: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if height != self.image_size[0] or width != self.image_size[1]:
```

mfuntowicz: Ok for me to add a check for `image-*` task and calibration methods to print out a message informing the user.

regisss (approved):
It looks great, thanks @fxmarty !!

PR #239, opened 2022-06-24
Fix doc versions

lewtun:
# What does this PR do?

This PR attempts to fix a problem with the way the docs are versioned with every push to `main` vs from a release branch.

I thought `doc-builder` would automatically build the `main` docs when the `--version` flag contained a `.dev0` suffix, but this turns out to be incorrect.

The current fix checks if `.dev0` exists in the version and then sets the appropriate flag in `doc-builder`

cc @fxmarty @mishig25 

## Before submitting


echarlaix (approved):
LGTM, thanks for taking care of this @lewtun !

PR #241, opened 2022-06-27
Enable possibility to not use past_key_values

echarlaix:
In this PR, we add the possibility to not use the precomputed `past_key_values` during ONNX Runtime inference of Seq2Seq models. This can be interesting for small models / short sequence length or if the user don't want to export the `decoder_with_past_model`


optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -290,7 +297,10 @@ def _from_pretrained(
                 the decoder model with a different name.
             kwargs (`Dict`, *optional*):
                 kwargs will be passed to the model during initialization.
+            use_past_key_values (`bool`, *optional*, defaults to `True`):
+                Whether or not use the pre-computed key/values hidden-states in order to speed up sequential decoding.
```
regisss, line 301:
```suggestion
                Whether or not to use the pre-computed key/values hidden-states in order to speed up sequential decoding.
```

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -363,13 +383,16 @@ def _from_transformers(
             cache_dir (`Union[str, Path]`, *optional*):
                 The path to a directory in which a downloaded pretrained model configuration should be cached if the
                 standard cache should not be used.
+            use_past_key_values (`bool`, *optional*, defaults to `True`):
+                Whether or not use the pre-computed key/values hidden-states in order to speed up sequential decoding.
```
regisss, line 387:
```suggestion
                Whether or not to use the pre-computed key/values hidden-states in order to speed up sequential decoding.
```

tests/onnxruntime/test_modeling.py
```diff
@@ -128,15 +135,26 @@ def test_save_model(self):
 
     def test_save_seq2seq_model(self):
         with tempfile.TemporaryDirectory() as tmpdirname:
-            model = ORTModelForSeq2SeqLM.from_pretrained(self.ONNX_SEQ2SEQ_MODEL_ID)
+            model = ORTModelForSeq2SeqLM.from_pretrained(self.ONNX_SEQ2SEQ_MODEL_ID, use_past_key_values=True)
             model.save_pretrained(tmpdirname)
             folder_contents = os.listdir(tmpdirname)
-            # Verify config and ONNX exported encoder, decoder and decoder present in folder
+            # Verify config and ONNX exported encoder, decoder and decoder with past present in folder
```
regisss, line 141:
```suggestion
            # Verify config and ONNX exported encoder, decoder and decoder with past are present in folder
```

regisss (approved):
It looks great @echarlaix!!
I just left a few nits regarding comments.

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -170,7 +170,12 @@ def __init__(
...
-        self.decoder_with_past = ORTDecoder(session=decoder_with_past_session, device=self._device)
+        self.use_past_key_values = decoder_with_past_session is not None
+        # If a decoder_with_past_path is provided, an inference session for the decoder with past key/values as inputs
+        # will be enabled
+        self.decoder_with_past = (
+            ORTDecoder(session=decoder_with_past_session, device=self._device) if self.use_past_key_values else None
+        )
         self.encoder_file_name = kwargs.get("last_encoder_model_name", ONNX_ENCODER_NAME)
         self.decoder_file_name = kwargs.get("last_decoder_model_name", ONNX_DECODER_NAME)
         self.decoder_file_with_past_name = kwargs.get("last_decoder_with_past_model_name", ONNX_DECODER_WITH_PAST_NAME)
```
philschmid, line 181: We should make this `None` when there is no past_key model file

echarlaix: `self.use_past_key_values` is set to `True` when a `decoder_with_past_session` is given, `False` otherwise (it seems equivalent to what you are saying)

PR #244, opened 2022-06-27
Fix YAML for GPU ORT tests

lewtun:
# What does this PR do?

This PR fixes another YAML formatting problem, this time related to #203 

Will be more systematic in future to catch these :)


regisss (approved):
Thanks again @lewtun !!

PR #248, opened 2022-06-30
Add CLM training example

carzh:
# What does this PR do?

* adds CLM training example + brief README related to it

Built off of work from the transformers/examples repo.


JingyaHuang:
Hi @carzh, thanks for adding the training example!

Can you reformat the changed files by `make style`, in the mean time, I will work on the review, thanks!

examples/onnxruntime/training/language-modeling/run_clm.py
```diff
@@ -0,0 +1,581 @@
...
+    trainer = trainer_class(
+        model=model,
+        args=training_args,
+        train_dataset=train_dataset if training_args.do_train else None,
+        eval_dataset=eval_dataset if training_args.do_eval else None,
+        tokenizer=tokenizer,
+        # Data collator will default to DataCollatorWithPadding, so we change it.
+        data_collator=default_data_collator,
+        compute_metrics=compute_metrics if training_args.do_eval and not is_torch_tpu_available() else None,
+    )
```
JingyaHuang, line 520:
```suggestion
    trainer = trainer_class(
        model=model,
        args=training_args,
        train_dataset=train_dataset if training_args.do_train else None,
        eval_dataset=eval_dataset if training_args.do_eval else None,
        tokenizer=tokenizer,
        # Data collator will default to DataCollatorWithPadding, so we change it.
        data_collator=default_data_collator,
        compute_metrics=compute_metrics if training_args.do_eval and not is_torch_tpu_available() else None,
        preprocess_logits_for_metrics=preprocess_logits_for_metrics
        if training_args.do_eval and not is_torch_tpu_available()
        else None,
    )
```

JingyaHuang (commented):
Hi @carzh, thanks for the example!

I left some small nits in the example. For the PyTorch baseline, are ort folks planning to use it for benchmarking purposes? Since we prefer to just use ORTTrainer as it is an ort training example. 

P.S. For the code quality check, can you do `pip install "optimum[quality]"` before `make style` to ensure that you are using the correct version of black and isort? This should help the scripts pass the quality check, thx!

examples/onnxruntime/training/language-modeling/run_clm.py
JingyaHuang: We will soon replace the bare inference sessions in ORTTrainer with [ORTModel](https://github.com/huggingface/optimum/blob/main/optimum/onnxruntime/modeling_ort.py#L674), so it could be good to keep the term although it is not useful for the moment.

carzh: Sounds good! Also, the preprocess_logits_for_metrics section was removed because ORTTrainer doesn't have that argument

JingyaHuang (approved):
Hi @carzh, thanks for iterating on this, LGTM!

For information, currently, the script will only work with the latest transformers version as demanded by ORTModels dependencies, thus gpt2 mixed-precision training will fail in this case. To solve the issue, we are working on:

* [transformers PR #18017](https://github.com/huggingface/transformers/pull/18017) - Fix gpt2 fp16 training which breaks after transformers>4.16.0 (with the fix users can do gpt2 fp16 training with the latest transformers)
* an upcoming PR #265  to decouple the import of different APIs (with the fix users can do gpt2 fp16 training with transformers 4.16.0)

carzh:
Hello, just wanted to confirm -- this PR is on standby until the GPT2 fp16 issue is resolved? Or can it be merged in sooner and is pending other checks? Thanks & sorry for any inconvenience!

JingyaHuang:
> Hello, just wanted to confirm -- this PR is on standby until the GPT2 fp16 issue is resolved? Or can it be merged in sooner and is pending other checks? Thanks & sorry for any inconvenience!

Hi @carzh, actually there are two possible modifications to the clm example:
* The fix of gpt2(the PR is waiting for the approval of transformers' core maintainers) -> need to change the minimum transformers version in the example.
* In the #295, new ORT training arguments were added which enable the use of ort fused adam optimizer -> need to replace training args by ort training args.

But I think that it would be better to add these modifications after these two PRs make their way to the main branch. I will merge this PR first!

PR #253, opened 2022-07-01
FuseBiasInLinear transformation

michaelbenayoun:
# What does this PR do?

- Adds a transformation that fuses the weight and the bias in linear modules

### DIFF
--- a/optimum/fx/optimization/transformations.py
+++ b/optimum/fx/optimization/transformations.py
@@ -22,6 +22,7 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, List
 
 import torch
+from transformers.utils.fx import _gen_constructor_wrapper
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


optimum/fx/optimization/transformations.py
```diff
@@ -293,15 +293,64 @@ def transform(self, graph_module: "GraphModule") -> "GraphModule":
...
+        return graph_module
+
+    def reverse(self, graph_module: "GraphModule") -> "GraphModule":
+        for node in graph_module.graph.nodes:
+            if getattr(node, "was_transformed", "") == "FuseBiasInLinear":
+                inserted_nodes = [node.args[0]]
+                node.args = node.original_node_args
+                while inserted_nodes:
+                    n = inserted_nodes.pop(0)
+                    if n is not node.original_node_args[0]:
```
regisss, line 343: When could this happen? When some linear layers have been merged?

regisss (approved):
LGTM @michaelbenayoun !

optimum/fx/optimization/transformations.py
michaelbenayoun: So this idea here is to delete all the inserted nodes by the `insert_concat` function in the `transform` method. 
This fuctions "starts" from `node.original_node_args[0]`, and the plan is to remove the nodes between this one and the output node. 

But while explaining it to you, I noticed that it is not what's happening here... thanks for catching this!

PR #269, opened 2022-07-07
Add lazy imports in optimum/onnxruntime

### LINKED ISSUE #265: Associated import lead to import fails
### System Info

Both transformers 4.16.0(currently demanded by gpt2 fp16 on ort training) and transformers 4.18.0 will fail when I tried to import `ORTTrainer`, as the import of ORTModel demands features newly integrated into transformers. I find this dependence not good, and a more robust import strategy shall be applied probably like what we have in transformers.

### Who can help?

@JingyaHuang @regisss @morgan

### Reproduction

```
>>> from optimum.onnxruntime import ORTTrainer
/home/ubuntu/anaconda3/envs/venv_ort_train/lib/python3.8/site-packages/cerberus/__init__.py:13: UserWarning: Module optimum was already imported from None, but /home/ubuntu/optimum is being added to sys.path
  from pkg_resources import get_distribution, DistributionNotFound
2022-07-07 09:59:03.368586: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /usr/local/cuda-10.2/lib64:
2022-07-07 09:59:03.368625: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/optimum/optimum/onnxruntime/__init__.py", line 53, in <module>
    from .modeling_ort import (
  File "/home/ubuntu/optimum/optimum/onnxruntime/modeling_ort.py", line 29, in <module>
    from transformers.onnx.utils import get_preprocessor
ImportError: cannot import name 'get_preprocessor' from 'transformers.onnx.utils' (/home/ubuntu/anaconda3/envs/venv_ort_train/lib/python3.8/site-packages/transformers/onnx/utils.py)
```

### Ideas

Importing `ORTTrainer` shouldn't rely on the dependencies of ORTModels. This could be done by importing with [`transformers.utils._LazyModule`](https://github.com/huggingface/transformers/blob/main/src/transformers/utils/import_utils.py#L884). WDYT?

regisss:
# What does this PR do?

Fixes #265. Modules in `optimum/onnxruntime` are now imported in a lazy manner as it is done in Transformers.

Also:
- absolute imports are replaced by relative imports in source files,
- `ORTQuantizableOperator` is moved to `optimum/onnxruntime/utils.py`,
- `AUTO_MINIMUM_SUPPORTED_ONNX_OPSET` is deleted because not used anywhere,
- `ORT_DEFAULT_CHANNEL_FOR_OPERATORS` and `ORT_FULLY_CONNECTED_OPERATORS` are moved to `optimum/onnxruntime/configuration.py` where they are only used.

## Before submitting


optimum/onnxruntime/quantizable_operators.py
```diff
@@ -0,0 +1,48 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from enum import Enum
+
+
+class ORTQuantizableOperator(Enum):
```
JingyaHuang, line 17: Why don't we put it in the onnxruntime/preprocessors/quantization.py ?

JingyaHuang (approved):
LGTM! Thx @regisss!

optimum/onnxruntime/__init__.py
```diff
@@ -11,56 +11,67 @@
...
+    "quantization": ["ORTQuantizer"],
+    "trainer": ["ORTTrainer"],
+    "trainer_seq2seq": ["ORTSeq2SeqTrainer"],
+    "utils": ["ONNX_DECODER_NAME", "ONNX_DECODER_WITH_PAST_NAME", "ONNX_ENCODER_NAME", "ONNX_WEIGHTS_NAME"],
+}
 
 
-AUTO_MINIMUM_SUPPORTED_ONNX_OPSET = None
+# Direct imports for type-checking
+if TYPE_CHECKING:
```
michaelbenayoun, line 46: Just out of curiosity: what does this direct import enable?

optimum/onnxruntime/quantizable_operators.py
michaelbenayoun, line 17: Why do we need this class? Isn't it something we can already have access to with ort?

michaelbenayoun (approved):
LGTM, thanks for this @regisss

optimum/onnxruntime/__init__.py
```diff
@@ -12,6 +12,9 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 from enum import Enum
+from typing import TYPE_CHECKING
+
+from transformers.utils import _LazyModule
```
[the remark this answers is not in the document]
regisss: Since it does not contain anything specific to Transformers, I think it's better to just import it from there and not maintain it in Optimum.

optimum/onnxruntime/__init__.py
regisss: It is only for type checking, otherwise modules are imported using `_LazyModule`. Not sure that answers your question though :thinking:

Issue #262, opened 2022-07-06
How can i set number of threads for Optimum exported model?


MiladMolazadeh:
### System Info

```shell
optimum==1.2.3
onnxruntime==1.11.1
onnx==1.12.0
transformers==4.20.1
python version 3.7.13
```

### Who can help?

@JingyaHuang @echarlaix 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Optimum doesn't have a config for assigning the number of threads
```
from onnxruntime import SessionOptions
SessionOptions().intra_op_num_threads = 1
```
### also limiting  on OS level doesn't work:

```bash
taskset -c 0-16 python inference_onnx.py
```
![1](https://user-images.githubusercontent.com/33181902/177484888-416316b9-aa00-4370-9937-437f2bb85c38.png)

```bash
taskset -c 0 python inference_onnx.py
```
![2](https://user-images.githubusercontent.com/33181902/177484909-b1f58bb9-019a-4099-abf7-1f1a2441f314.png)

### Expected behavior

```shell
Hi!

I can't specify the number of threads for inferencing Optimum ONNX models.
I didn't have such a problem with the default transformers model before.
Is there any Configuration in Optimum?
```

fxmarty:
Hello @MiladMolazadeh , by coincidence I run into the same issue today!

Would https://github.com/huggingface/optimum/pull/271 solve your issue?

I propose the following workflow:

```python
from functools import partial

from optimum.onnxruntime import ORTQuantizer
from optimum.onnxruntime.modeling_ort import ORTModelForSequenceClassification, ORTModel
from optimum.onnxruntime.configuration import AutoQuantizationConfig

import onnxruntime

import time
import torch
from tqdm import tqdm


model_name = "distilbert-base-uncased-finetuned-sst-2-english"
optimum_model_path = "/path/to/optimum_model.onnx"
optimum_quantized_model_path = "/path/to/optimum_quantized_model.onnx"

quantizer = ORTQuantizer.from_pretrained(model_name, feature="sequence-classification")

# Inference with Optimum
qconfig = AutoQuantizationConfig.avx512_vnni(is_static=False, per_channel=False)

quantizer.export(
    onnx_model_path=optimum_model_path,
    onnx_quantized_model_output_path=optimum_quantized_model_path,
[...]

PR #271, opened 2022-07-08
Add InferenceSession options and Provider to ORTModel

fxmarty:
`ORTModel.load_model` was not accepting custom session options, making it hard for example to use a given number of CPU cores with an `ORTModel` for inference. Indeed, onnxruntime obey neither `psutil.cpu_affinity` nor `taskset`.

This may close https://github.com/huggingface/optimum/issues/262

## Before submitting

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -116,7 +116,7 @@ class ORTModel(OptimizedModel):
         self._device = get_device_for_provider(self.providers[0])
 
         if self._device == None:
-            logger.warn(
+            logger.warning(
                 f"ORTModel outputs will be sent to CPU as the device could not be inferred from the execution provider {self.providers[0]}."
                 f" Use `ort_model.to()` to send the outputs to the wanted device."
             )
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -16,7 +16,7 @@ import logging
 import os
 import shutil
 from pathlib import Path
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 import torch
 from transformers import (
@@ -48,7 +48,7 @@ from transformers.onnx.utils import get_preprocessor
 import onnxruntime as ort
 from huggingface_hub import HfApi, hf_hub_download
 
-from ..modeling_base import OptimizedModel
+from ..modeling_base import FROM_PRETRAINED_START_DOCSTRING, OptimizedModel
 from .utils import ONNX_WEIGHTS_NAME, get_device_for_provider, get_provider_for_device
 
 
@@ -107,12 +107,19 @@ class ORTModel(OptimizedModel):
     base_model_prefix = "onnx_model"
     auto_model_class = AutoModel
 
-    def __init__(self, model=None, config=None, **kwargs):
+    def __init__(self, model: ort.InferenceSession, config, **kwargs):
         self.model = model
         self.config = config
         self.model_save_dir = kwargs.get("model_save_dir", None)
         self.latest_model_name = kwargs.get("latest_model_name", "model.onnx")
-        self._device = get_device_for_provider(self.model.get_providers()[0])
+        self.providers = model.get_providers()
+        self._device = get_device_for_provider(self.providers[0])
+
+        if self._device == None:
+            logger.warn(
+                f"ORTModel outputs will be sent to CPU as the device could not be inferred from the execution provider {self.providers[0]}."
+                f" Use `ort_model.to()` to send the outputs to the wanted device."
+            )
 
         # registers the ORTModelForXXX classes into the transformers AutoModel classes
         # to avoid warnings when create a pipeline https://github.com/huggingface/transformers/blob/cad61b68396a1a387287a8e2e2fef78a25b79383/src/transformers/pipelines/base.py#L863
@@ -128,10 +135,10 @@ class ORTModel(OptimizedModel):
         return self._device
 
     @device.setter
-    def device(self, value):
+    def device(self, value: torch.device):
         self._device = value
 
-    def to(self, device):
+    def to(self, device: torch.device):
         """
         Changes the ONNX Runtime provider according to the device.
         """
@@ -142,26 +149,45 @@ class ORTModel(OptimizedModel):
         self.device = device
         provider = get_provider_for_device(self.device)
         self.model.set_providers([provider])
+        self.providers = self.model.get_providers()
         return self
 
     def forward(self, *args, **kwargs):
         raise NotImplementedError
 
     @staticmethod
-    def load_model(path: Union[str, Path], provider=None):
+    def load_model(
+        path: Union[str, Path],
+        provider: Optional[Union[str, List[str]]] = "CPUExecutionProvider",
+        session_options: Optional[ort.SessionOptions] = None,
+        **kwargs
+    ):
         """
         Loads an ONNX Inference session with a given provider. Default provider is `CPUExecutionProvider` to match the default behaviour in PyTorch/TensorFlow/JAX.
 
         Arguments:
             path (`str` or `Path`):
                 Directory from which to load the model.
-            provider(`str`, *optional*):
-                ONNX Runtime provider to use for loading the model. Defaults to `CPUExecutionProvider`.
+            provider (`str` or `List[str]`, *optional*):
+                ONNX Runtime providers to use for loading the model. This can either be a single provider given as a string (for example
+                `CUDAExecutionProvider`) or a list of execution providers to use in priority order.
+                See https://onnxruntime.ai/docs/execution-providers/ for more details. Defaults to `CPUExecutionProvider`.
+            session_options (`onnxruntime.SessionOptions`, *optional*),:
+                ONNX Runtime session options to use for loading the model. Defaults to `None`.
         """
-        if provider is None:
-            provider = "CPUExecutionProvider"
+        if isinstance(provider, str):
+            providers = [provider]
+        elif isinstance(provider, list):
+            providers = provider
 
-        return ort.InferenceSession(path, providers=[provider])
+        available_providers = ort.get_available_providers()
+        for provider in providers:
+            if provider not in available_providers:
+                raise ValueError(
+                    f"Asked to use {provider} as an ONNX Runtime execution provider, but the available execution providers are {available_providers}."
+                )
+
+        return ort.InferenceSession(path, providers=providers, sess_options=session_options)
 
     def _save_pretrained(self, save_directory: Union[str, Path], file_name: Optional[str] = None, **kwargs):
         """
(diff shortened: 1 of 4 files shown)


philschmid (approved):
LGTM ✅ I left one comment on the session option if we should define a default value. 
@mfuntowicz what do you think?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -159,6 +185,43 @@ def _save_pretrained(self, save_directory: Union[str, Path], file_name: Optional
         dst_path = Path(save_directory).joinpath(model_file_name)
         shutil.copyfile(src_path, dst_path)
 
+    @classmethod
+    @add_start_docstrings(FROM_PRETRAINED_START_DOCSTRING)
```
philschmid, line 189: why did you add this method? For documentation reasons?
Nice idea with the `add_start_docstrings`!

optimum/onnxruntime/modeling_ort.py
```diff
@@ -213,8 +276,9 @@ def _from_pretrained(
             )
             kwargs["model_save_dir"] = Path(model_cache_path).parent
             kwargs["latest_model_name"] = Path(model_cache_path).name
-            model = ORTModel.load_model(model_cache_path)
+            model = ORTModel.load_model(model_cache_path, **kwargs)
```
philschmid, line 279: I think here we should define `self.provider` as well. That way a user can already access `model.provider` otherwise it wouldn't make much sense to add it in the `to` method.

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -201,12 +204,21 @@ def load_model(
             provider(`str`, *optional*):
                 The ONNX Runtime provider to use for loading the model. Defaults to `"CPUExecutionProvider"`.
         """
-        if provider is None:
-            provider = "CPUExecutionProvider"
+        if isinstance(provider, str):
+            providers = [provider]
+        elif isinstance(provider, list):
+            providers = provider
```
philschmid, line 210: same as for encoder why a list?

tests/onnxruntime/test_modeling.py
```diff
@@ -61,6 +61,31 @@ def test_load_model_from_hub(self):
         model = ORTModel.from_pretrained(self.ONNX_MODEL_ID)
         self.assertIsInstance(model.model, onnxruntime.capi.onnxruntime_inference_collection.InferenceSession)
         self.assertIsInstance(model.config, PretrainedConfig)
+        self.assertListEqual(model.providers, ["CPUExecutionProvider"])
+        self.assertEqual(model.device, torch.device("cpu"))
+
+    def test_load_model_cuda_provider_list(self):
```
philschmid, line 67: are those tests run on any device or noly when there is a GPU available?

tests/onnxruntime/test_modeling.py
```diff
@@ -61,6 +61,31 @@ def test_load_model_from_hub(self):
...
+        self.assertEqual(model.device, torch.device("cuda"))
+
+    def test_load_model_cuda_provider_str(self):
+        model = ORTModel.from_pretrained(self.ONNX_MODEL_ID, provider="CUDAExecutionProvider")
+        self.assertListEqual(model.providers, ["CUDAExecutionProvider", "CPUExecutionProvider"])
+        self.assertEqual(model.device, torch.device("cuda"))
+
+    def test_load_model_several_providers(self):
+        model = ORTModel.from_pretrained(
+            self.ONNX_MODEL_ID, provider=["CUDAExecutionProvider", "TensorrtExecutionProvider"]
```
philschmid, line 79: does it make sense? What happens if the model hast cuda and tensorrt?

tests/onnxruntime/test_modeling.py
```diff
@@ -61,6 +61,31 @@ def test_load_model_from_hub(self):
...
+        )
+        self.assertListEqual(
+            model.providers, ["CUDAExecutionProvider", "TensorrtExecutionProvider", "CPUExecutionProvider"]
```
philschmid, line 82: I don't understand this one? should `model.provider` return the value which provider is used for inference?

tests/onnxruntime/test_modeling.py
```diff
@@ -69,6 +94,29 @@ def test_load_seq2seq_model_from_hub(self):
...
+        )
+        self.assertEqual(model.device, torch.device("cuda"))
+
+    def test_load_seq2seq_model_unknown_provider(self):
+        with self.assertRaises(ValueError):
+            ORTModel.from_pretrained(self.ONNX_SEQ2SEQ_MODEL_ID, provider="FooExecutionProvider")
+
```
philschmid, line 119: same as above

philschmid (changes requested):
I added a few comments since some details are not clear yet to me.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -111,36 +118,55 @@ def device(self) -> torch.device:
...
 
-    def to(self, device):
+    def to(self, device: torch.device):
         """
         Changes the ONNX Runtime provider according to the device.
         """
         self.device = device
         provider = get_provider_for_device(self.device)
         self.model.set_providers([provider])
+        self.providers = self.model.get_providers()
```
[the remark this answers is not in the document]
fxmarty: Yes, it returns a list. Basically an InferenceSession can have several providers at the same time, by order of priority. So you could have `["TensorrtExecutionProvider", "CUDAExecutionProvider", "CPUExecutionProvider"]`, or `["CUDAExecutionProvider", "CPUExecutionProvider"]` for example. See https://onnxruntime.ai/docs/execution-providers/

optimum/onnxruntime/modeling_ort.py
```diff
@@ -111,36 +118,55 @@ def device(self) -> torch.device:
...
+                See https://onnxruntime.ai/docs/execution-providers/ for more details. Defaults to `CPUExecutionProvider`.
+            session_options (`onnxruntime.SessionOptions`, *optional*),:
+                ONNX Runtime session options to use for loading the model. Defaults to `None`.
         """
-        if provider is None:
-            provider = "CPUExecutionProvider"
+        if isinstance(provider, str):
+            providers = [provider]
+        elif isinstance(provider, list):
+            providers = provider
```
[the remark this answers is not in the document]
fxmarty: Because it is a feature from onnxruntime. But maybe not used so much, I am not sure, so if you prefer to support only a single execution provider I can change that, let me know!

optimum/onnxruntime/modeling_ort.py
fxmarty: I was thinking it makes sense to have `from_pretrained` in the `ORTModel` section of the documentation. If in the future we extend these pipeline wrappers to other backends (say, fx or OpenVino), they may have custom parameters that the documentation from `OptimizedModel` will not capture. So It makes sense to me to have the doc for `from_pretrained` in the subclasses as well.

What do you think?

optimum/onnxruntime/modeling_seq2seq.py
fxmarty: see above

optimum/onnxruntime/modeling_ort.py
fxmarty: I define it in the `__init__` from `ORTModel`: https://github.com/fxmarty/optimum/blob/21f4d919b202561968258fb6e8dbce67130e957b/optimum/onnxruntime/modeling_ort.py#L98 . Does it sound good?

tests/onnxruntime/test_modeling.py
fxmarty: An error will be raised here: https://github.com/fxmarty/optimum/blob/21f4d919b202561968258fb6e8dbce67130e957b/optimum/onnxruntime/modeling_ort.py#L163-L167

tests/onnxruntime/test_modeling.py
fxmarty: Yes it is confusing. ONNX Runtime adds the `CPUExecutionProvider` as a default if it is not provided, see: https://github.com/microsoft/onnxruntime/blob/0d6cbc6e57c2ecd0722cefaa87d300f301864e3b/onnxruntime/core/session/inference_session.cc#L1280-L1287

optimum/onnxruntime/modeling_ort.py
```diff
@@ -90,12 +90,19 @@ class ORTModel(OptimizedModel):
     base_model_prefix = "onnx_model"
     auto_model_class = AutoModel
 
-    def __init__(self, model=None, config=None, **kwargs):
+    def __init__(self, model: ort.InferenceSession, config, **kwargs):
```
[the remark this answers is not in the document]
fxmarty: Because to me `model` and `config` should be required arguments, it does not make sense to initialize an `ORTModel` without an InferenceSesson provided. What do you think?

tests/onnxruntime/test_modeling.py
fxmarty: Good catch, I use the decorator `@require_torch_gpu` now.

optimum/onnxruntime/modeling_ort.py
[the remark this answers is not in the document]
philschmid: Lets stick with what we are doing everywhere with named args

optimum/onnxruntime/modeling_ort.py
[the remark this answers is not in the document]
philschmid: cc @mfuntowicz on this

philschmid (approved):
LGTM, but let's get a review from @mfuntowicz and @echarlaix or @michaelbenayoun before merging. Especially, regarding the "providers" (list)

optimum/onnxruntime/modeling_ort.py
```diff
@@ -142,26 +149,44 @@ def to(self, device):
...
+                ONNX Runtime session options to use for loading the model. Defaults to `None`.
         """
-        if provider is None:
-            provider = "CPUExecutionProvider"
-
-        return ort.InferenceSession(path, providers=[provider])
+        if isinstance(provider, str):
+            providers = [provider]
+        elif isinstance(provider, list):
+            providers = provider
```
echarlaix, line 181: I would remove those lines

optimum/onnxruntime/modeling_ort.py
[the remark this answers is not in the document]
echarlaix: Currently we only support cpu and cuda so I don't think it makes a difference to give the possibility to give a list of execution providers with a priority order, so for clarity I would keep it as a string for the moment

echarlaix (approved):
LGTM ! 🔥

fxmarty:
Great! In any case, even if `CUDAExecutionProvider` is solely passed, the execution providers from onnxruntime is still a list, `["CUDAExecutionProvider", "CPUExecutionProvider"]`, so to me it makes still sense to have a list.

PR #273, opened 2022-07-08
Adding support for float16 conversion for GPUs

philschmid:
# What does this PR do?

This PR adds support for converting model weights from `fp32` to `fp16` by adding a new Optimization parameter. If the `fp16` arg is provided in the `OptimizationConfig` the weights are converted. I also added a test to make sure the model is not containing any fp32 weights

## Before submitting

### DIFF
--- a/tests/onnxruntime/test_optimization.py
+++ b/tests/onnxruntime/test_optimization.py
@@ -120,7 +120,9 @@ class ORTOptimizerTest(unittest.TestCase):
             for w in model.graph.initializer:
                 self.assertNotEqual(w.data_type, onnx.onnx_pb.TensorProto.FLOAT)
 
-            onnx_model = ORTModelForSequenceClassification.from_pretrained(tmp_dir, file_name="model-optimized.onnx")
+            onnx_model = ORTModelForSequenceClassification.from_pretrained(
+                output_dir.as_posix(), file_name="model-optimized.onnx"
+            )
             transformers_model = AutoModelForSequenceClassification.from_pretrained(model_name)
             tokenizer = AutoTokenizer.from_pretrained(model_name)
             tokens = tokenizer("This is a sample output", return_tensors="pt")
--- a/tests/onnxruntime/test_optimization.py
+++ b/tests/onnxruntime/test_optimization.py
@@ -20,6 +20,7 @@ from pathlib import Path
 
 import numpy as np
 import torch
+import onnx
 
 from onnx import load as onnx_load
 from onnxruntime import InferenceSession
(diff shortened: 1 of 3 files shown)


tests/onnxruntime/test_optimization.py
```diff
@@ -100,6 +101,23 @@ def test_optimization_details(self):
...
+            optimized_model_path = output_dir.joinpath("model-optimized.onnx")
+            optimizer = ORTOptimizer.from_pretrained(model_name, feature="sequence-classification")
+            optimizer.export(
+                onnx_model_path=model_path,
+                onnx_optimized_model_output_path=optimized_model_path,
+                optimization_config=optimization_config,
+            )
+            model = onnx.load(optimized_model_path.as_posix())
+            for w in model.graph.initializer:
+                self.assertNotEqual(w.data_type, onnx.onnx_pb.TensorProto.FLOAT)
```
regisss, line 119: Why not `self.assertEqual(w.data_type, onnx.onnx_pb.TensorProto.FLOAT16)`?

regisss (approved):
LGTM @philschmid !!
I just left two minor comments.

tests/onnxruntime/test_optimization.py
philschmid: Since there are constants and other int based weights and values

philschmid: for example the inputs are int64

regisss: Ah okay it makes sense, thanks :)

echarlaix (approved):
LGTM, thanks for this addition @philschmid

tests/onnxruntime/test_optimization.py
```diff
@@ -100,6 +101,23 @@ def test_optimization_details(self):
...
+            output_dir = Path(tmp_dir)
+            model_path = output_dir.joinpath("model.onnx")
+            optimized_model_path = output_dir.joinpath("model-optimized.onnx")
+            optimizer = ORTOptimizer.from_pretrained(model_name, feature="sequence-classification")
+            optimizer.export(
+                onnx_model_path=model_path,
+                onnx_optimized_model_output_path=optimized_model_path,
+                optimization_config=optimization_config,
+            )
+            model = onnx.load(optimized_model_path.as_posix())
```
mfuntowicz, line 117: Can you add a test about the relative difference in the output? I'm not a huge fan of changing dtype for all the layers like this.

philschmid: Not really sure why since we are not applying any custom logic. 
What if there is more difference? what do you expect from the test?

philschmid: Added the test but not sure what the benefit is

Issue #274, opened 2022-07-09
ORTModel isinstance check not working as expected


sam-h-bean:
### System Info

```shell
Databricks runtime running on GPU Spark cluster
Built optimum from source
Python 3.8
```

### Who can help?

@jingy

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

```python
from optimum.onnxruntime import ORTModel
from optimum.onnxruntime.configuration import OptimizationConfig
from optimum.onnxruntime import ORTOptimizer
from optimum.pipelines import pipeline as ort_pipeline
from transformers import AutoTokenizer, AutoModelForCausalLM


model = AutoModelForCausalLM.from_pretrained('Salesforce/codegen-2B-multi')
tokenizer = AutoTokenizer.from_pretrained('Salesforce/codegen-2B-multi')


optimization_config = OptimizationConfig(optimization_level=99, optimize_for_gpu=True)
optimizer = ORTOptimizer(
    model=model, preprocessor=tokenizer, feature="causal-lm", opset=13,
)

ort_model = ORTModel(
            "./model-optimized.onnx",
            optimizer._onnx_config,
            execution_provider="CUDAExecutionProvider",
)
ort_model.config = optimizer.model.config

generator = ort_pipeline(
    "text-generation",
    model=ort_model,
    tokenizer=tokenizer,
    max_length=128
)
```
<img width="1285" alt="Screen Shot 2022-07-08 at 11 52 22 PM" src="https://user-images.githubusercontent.com/43734688/178090476-e8be40ac-a8a7-408e-82b6-846c91e60c62.png">

```python
from optimum.onnxruntime.configuration import OptimizationConfig, ORTConfig
from optimum.onnxruntime import ORTModelForSequenceClassification, ORTOptimizer


model = AutoModelForCausalLM.from_pretrained('Salesforce/codegen-2B-multi')
tokenizer = AutoTokenizer.from_pretrained('Salesforce/codegen-2B-multi')

def get_optimized_pipeline(model, tokenizer):
    optimization_config = OptimizationConfig(optimization_level=99, optimize_for_gpu=True)
    model.eval()
    optimizer = ORTOptimizer(
        model=model, preprocessor=tokenizer, feature="causal-lm", opset=13,
    )

    optimizer.export(
        onnx_model_path="model.onnx",
[...]

PR #277, opened 2022-07-10
revert troublesome line

sam-h-bean:
# What does this PR do?
Does not default to CPU provider when multiple providers are available.

Fixes # (issue)
https://github.com/huggingface/optimum/issues/274

## Before submitting

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -32,8 +32,7 @@ import onnxruntime as ort
 from huggingface_hub import HfApi, hf_hub_download
 
 from ..modeling_base import OptimizedModel
-from .utils import ONNX_WEIGHTS_NAME, get_device_for_provider, get_provider_for_device
-
+from .utils import ONNX_WEIGHTS_NAME, get_device_for_provider, get_provider_for_device, _is_gpu_available
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


optimum/onnxruntime/modeling_ort.py
```diff
@@ -131,7 +130,7 @@ def load_model(path: Union[str, Path], provider=None):
                 ONNX Runtime provider to use for loading the model. Defaults to `CPUExecutionProvider`.
         """
         if provider is None:
-            provider = "CPUExecutionProvider"
+            provider = "CUDAExecutionProvider" if _is_gpu_available() else "CPUExecutionProvider"
```
sam-h-bean, line 133: I've been using TensorRT on another branch successfully. I'm curious how we imagine being able to pass that as an option to the InferenceSession

https://github.com/huggingface/optimum/compare/main...sam-h-bean:optimum:mrbean/tensorrt?expand=1

philschmid:
Please check my response in
*  #274 

You can easily move your model around with the `.to()` method.

Regarding your question about TensorRT. We haven't sketched out a plan on how to fully integrate it yet. Maybe we are not going to use onnxruntime for it and go with TensorRT directly. But when we add support for it we'll make sure the model can be correctly loaded.

sam-h-bean:
> Please check my response in
> 
>     * [Running into Issues Optimizing CausalLM Model for GPU #274](https://github.com/huggingface/optimum/issues/274)
> 
> 
> You can easily move your model around with the `.to()` method.
> 
> Regarding your question about TensorRT. We haven't sketched out a plan on how to fully integrate it yet. Maybe we are not going to use onnxruntime for it and go with TensorRT directly. But when we add support for it we'll make sure the model can be correctly loaded.

Responded in the issue. FYI it seems like if the TensorRT execution provider is available and you select it then it sort of just works. I have been able to get some really good performance

PR #278, opened 2022-07-11
Correct bug in post-processing benchmark/eval

fxmarty:
Correct a bug where for text-classification, stored label ids in config.json can be strings. This is temporary as we will use `TextClassificationEvaluator` from evaluate ultimately (when it supports regression and multi-column tasks).

### DIFF
--- a/optimum/utils/preprocessing/text_classification.py
+++ b/optimum/utils/preprocessing/text_classification.py
@@ -119,10 +119,9 @@ class TextClassificationProcessing(DatasetProcessing):
             preds = pipeline.postprocess(model_outputs, function_to_apply=ClassificationFunction.NONE)
 
             if not self.task_args["is_regression"]:
-                # the dataset label ids may be different from the label2id of predictions
                 if self.label_to_id is not None:
-                    preds = self.config.label2id[preds["label"]]
-                    preds = self.label_to_id[preds]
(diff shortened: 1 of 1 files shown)


optimum/utils/preprocessing/text_classification.py
```diff
@@ -119,10 +119,9 @@ def run_inference(self, eval_dataset: Dataset, pipeline: TextClassificationPipel
             preds = pipeline.postprocess(model_outputs, function_to_apply=ClassificationFunction.NONE)
 
             if not self.task_args["is_regression"]:
-                # the dataset label ids may be different from the label2id of predictions
                 if self.label_to_id is not None:
-                    preds = self.config.label2id[preds["label"]]
-                    preds = self.label_to_id[preds]
+                    preds = int(self.config.label2id[preds["label"]])
+                    preds = self.label_to_id[preds]  # dataset label ids than of the model
```
regisss, line 124: I think a word is missing in the comment.

regisss (approved):
LGTM @fxmarty!
Just a word missing in a comment I think.

optimum/utils/preprocessing/text_classification.py
fxmarty: Thanks a lot! I changed the comment.

PR #279, opened 2022-07-11
Fix the number of layer to the decoder ones when creating the past ke…

echarlaix:
This PR fixes the ONNX export of encoder-decoder models which do not have the same number of encoder and decoder layers.


regisss (approved):
LGTM, thanks @echarlaix!!

Issue #272, opened 2022-07-08
Missing documentation of `ORTModel.from_pretrained()`


fxmarty:
### Feature request

We miss the documentation of `ORTModel.from_pretrained()` in https://huggingface.co/docs/optimum/main/en/onnxruntime/modeling_ort . It's not much, just need to add the documentation to the method and modify the mdx file.

Also, compared to https://huggingface.co/docs/transformers/v4.20.1/en/main_classes/model#transformers.PreTrainedModel.from_pretrained , we don't have an argument `pretrained_model_name_or_path`. Currently to load a local `.onnx` file, we must do e.g. `ORTModelForSequenceClassification.from_pretrained("/path/to/folder/", file_name="model.onnx")`.

I wonder if, like for Tensorflow/Flax models in transformers, we could not do `ORTModelForSequenceClassification.from_pretrained("/path/to/folder/model.onnx"` and have the check `os.path.isfile()` like in the transformers method. Alternatively, have an argument `from_onnx=True`.

Here is how the check is done in transformers: https://github.com/huggingface/transformers/blob/d0acc9537829e7d067edbb791473bbceb2ecf056/src/transformers/modeling_utils.py#L1944

https://github.com/huggingface/optimum/blob/958b31757bd2453d1c0758c342aa873be4040646/optimum/modeling_base.py#L165-L174

https://github.com/huggingface/optimum/blob/958b31757bd2453d1c0758c342aa873be4040646/optimum/onnxruntime/modeling_ort.py#L155-L165

### Your contribution

I can work on this

philschmid:
Thank you @fxmarty for opening the issue. I compared the documentation with `transformers` as well. It seems that for `AutoModelForXXX` in transformers there is also no `from_pretrained` documentation: See [BERT](https://huggingface.co/docs/transformers/v4.20.1/en/model_doc/bert#transformers.BertForSequenceClassification). 

What we could do is add a `Main Class` section in the documentation and add the [OptimizedModel](https://github.com/huggingface/optimum/blob/42825ce3e73c381ed14c2fe906cd970e9090d325/optimum/modeling_base.py#L166) to it which implements the base `from_pretrained` method. For this we should add a similar nice docstring than `transformers`. 

Can you take a look on adding the documentation and the docstring? 

---

For the second part

> Currently to load a local .onnx file, we must do e.g. ORTModelForSequenceClassification.from_pretrained("/path/to/folder/", file_name="model.onnx").

[...]

PR #280, opened 2022-07-11
Add documentation of OptimizedModel methods

fxmarty:
# What does this PR do?

Add to the documentation `save_pretrained()` and `from_pretrained()` methods of `OptimizedModel`. I added it to a "Main classes" section, not sure if it is the best way to go, as e.g. the `ORTModelFor...` classes are in the ONNX Runtime --> Inference section. Was not sure how to go as well given that we want to merge the doc from subpackages (should we have a shared "Main classes" for all subpackages?).

If we prefer to have subclass-specific documentation (the one I wrote is too short I think, compare https://huggingface.co/docs/transformers/v4.20.1/en/main_classes/model#transformers.PreTrainedModel.from_pretrained), we could do in `ORTModel`:

```python
    @classmethod
    def from_pretrained(
        cls,
        model_id: Union[str, Path],
        from_transformers: bool = False,
        force_download: bool = True,
        use_auth_token: Optional[str] = None,
        cache_dir: Optional[str] = None,
        **model_kwargs,
    ):
    """subclass specific doc, with examples, etc"""
    super().from_pretrained(model_id, from_transformers, force_download, use_auth_token, cache_dir, **model_kwargs)
```

To me `ORTModel` class is a bit like a `PreTrainedModel`, `TFPreTrainedModel` or `FlaxPreTrainedModel`, and all of those have their own `from_pretrained()`, but it looks to be mostly duplicate code.

For example, following https://github.com/huggingface/optimum/pull/271, `sess_options` would be a keyword argument specific to `ORTModel`. We could use `@add_end_docstrings` to avoid repeating the doc, probably.

Partially fixes https://github.com/huggingface/optimum/issues/272

## Before submitting

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -23,6 +23,8 @@
     title: Optimization
   title: Torch FX
 - sections:
+  - local: modeling_base
+    title: Optimized models
   - local: benchmark
-    title: Benchmark Optimum models
-  title: API Reference
+    title: Benchmark
+  title: Main classes
--- a/docs/source/modeling_base.mdx
+++ b/docs/source/modeling_base.mdx
@@ -1,0 +1,17 @@
+<!--Copyright 2022 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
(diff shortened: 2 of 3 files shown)


philschmid (approved):
Thank you for adding the documentation

PR #281, opened 2022-07-11
Remove duplicated key in supported pipeline task

echarlaix:


regisss (approved):
Thanks for catching this @echarlaix!

PR #282, opened 2022-07-11
support passing in user specified provider to from_pretrained

sam-h-bean:
# What does this PR do?
Allow users to pass in their own provider to the `from_pretrained` method and spike it through to the load_model call in `ORTModel`. Currently the `load_model` method supports passing in a provider but it seems like this functionality is not supported via the `from_pretrained` code path.

Fixes # (issue)
https://github.com/huggingface/optimum/issues/274

## Before submitting

### DIFF
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -157,6 +157,7 @@ class OptimizedModel(ABC):
         revision: Optional[Union[str, None]] = None,
         force_download: bool = True,
         cache_dir: Optional[str] = None,
+        provider: Optional[str] = None,
         **kwargs,
     ):
         """Overwrite this method in subclass to define how to load your model from pretrained"""
@@ -170,6 +171,7 @@ class OptimizedModel(ABC):
         force_download: bool = True,
         use_auth_token: Optional[str] = None,
         cache_dir: Optional[str] = None,
+        provider: Optional[str] = None,
         **model_kwargs,
     ):
         revision = None
@@ -213,6 +215,7 @@ class OptimizedModel(ABC):
                 cache_dir=cache_dir,
                 force_download=force_download,
                 use_auth_token=use_auth_token,
+                provider=provider,
                 **model_kwargs,
             )
 
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -161,6 +161,7 @@ class ORTModel(OptimizedModel):
         force_download: bool = False,
         cache_dir: Optional[str] = None,
         file_name: Optional[str] = None,
+        provider: Optional[str] = None,
         **kwargs,
     ):
         """
@@ -206,7 +207,7 @@ class ORTModel(OptimizedModel):
             )
             kwargs["model_save_dir"] = Path(model_cache_path).parent
             kwargs["latest_model_name"] = Path(model_cache_path).name
-            model = ORTModel.load_model(model_cache_path)
+            model = ORTModel.load_model(model_cache_path, provider=provider)
             config = PretrainedConfig.from_dict(config_dict)
         return cls(model=model, config=config, **kwargs)
 


optimum/modeling_base.py
```diff
@@ -157,6 +157,7 @@ def _from_pretrained(
         revision: Optional[Union[str, None]] = None,
         force_download: bool = True,
         cache_dir: Optional[str] = None,
+        provider: Optional[str] = None,
```
philschmid, line 160: Lets not add `provider` to the default `model_base` since it is specific to ORT. Lets go with `kwargs`

optimum/onnxruntime/modeling_ort.py
```diff
@@ -206,7 +210,7 @@ def _from_pretrained(
             )
             kwargs["model_save_dir"] = Path(model_cache_path).parent
             kwargs["latest_model_name"] = Path(model_cache_path).name
-            model = ORTModel.load_model(model_cache_path)
+            model = ORTModel.load_model(model_cache_path, provider=provider)
```
philschmid, line 213: lets define in the `load_model` a `self.provider` attribute easily access the provider of the model.

philschmid (changes requested):
Thank you for working on this. I added a few comments. 
Also can you make sure that `make style` is run/ `make style_check` passes. 
And we should add test for it to make sure the correct provider is used when passed into the model

optimum/onnxruntime/modeling_ort.py
sam-h-bean: I don't know if we'd be able to access an attribute since this `load_model` call is static. Unsure what the request is here

optimum/modeling_base.py
sam-h-bean: done

optimum/onnxruntime/modeling_ort.py
```diff
@@ -138,6 +143,12 @@ def load_model(path: Union[str, Path], provider=None):
                 ONNX Runtime provider to use for loading the model. Defaults to `CPUExecutionProvider`.
         """
         if provider is None:
+            if check_if_multiple_available_providers():
+                raise ValueError(
+                    "No provider supplied and multiple providers available, "
+                    "can not deterministically set correct provider. "
+                    "Pass provider in explicitly."
+                )
```
sam-h-bean, line 151: @philschmid I added this check here so that in the future if you have multiple providers available and don't supply a provider like I did, `load_model` won't fail for nebulous ORT reasons

sam-h-bean:
@philschmid mind re-reviewing?

optimum/onnxruntime/modeling_ort.py
philschmid: Not sure about that. Since when you install `onnxruntime-gpu` you always have 2 providers. We should keep the current behavior that if you run `model.to` the model is moved to GPU (similar to `transformers`) and if you don'T define anything it is CPU.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -193,11 +204,12 @@ def _from_pretrained(
                 kwargs will be passed to the model during initialization
         """
         config_dict = kwargs.pop("config", {})
+        provider = kwargs.pop("provider", None)
```
philschmid, line 207: we should add named argument to the method preferably with a type and a default value

optimum/onnxruntime/modeling_ort.py
philschmid: we can extract the provider from the model and then set it/ provider it in the `cls` call. There should be an easy way for a user to know about the provider and device used for the model, e.g. `model.device`; `model.provider`.

tests/onnxruntime/test_modeling.py
```diff
@@ -86,6 +86,12 @@ def test_model_on_gpu(self):
         model.to(gpu)
         self.assertEqual(model.device, gpu)
 
+    @require_torch_gpu
+    def test_model_on_gpu_pass_cpu(self):
```
philschmid, line 90: can we also add a test for `tensorrt`?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -193,11 +204,12 @@ def _from_pretrained(
...
         model_file_name = file_name if file_name is not None else ONNX_WEIGHTS_NAME
         # load model from local directory
         if os.path.isdir(model_id):
             config = PretrainedConfig.from_dict(config_dict)
-            model = ORTModel.load_model(os.path.join(model_id, model_file_name))
+            model = ORTModel.load_model(os.path.join(model_id, model_file_name), provider=provider)
```
philschmid, line 212: We should also make sure that when a user provides a provider `CUDA` or `TensorRT` that the device will be `cuda`. Meaning that the model is "moved" automatically/the device is set correctly

philschmid (commented):
Added a few more comments. I think we are missing a few UX features before we can merge it.

fxmarty:
Closing this as https://github.com/huggingface/optimum/pull/271 was merged and include this option. Thanks @sam-h-bean for pointing it out, let us know if this solves your issue!

PR #286, opened 2022-07-12
fix activation dtype

sam-h-bean:
# What does this PR do?
Fix the activation type for quantizing for tensorrt

Fixes
```bash
ValueError: Invalid combination of use_static_quantization = False and activations_dtype = QuantType.QInt8. OnnxRuntime dynamic quantization requires activations_dtype = QuantType.QUInt8
```

## Before submitting


sam-h-bean:
@philschmid @JingyaHuang should we put this in? Feels like a bug

PR #287, opened 2022-07-12
Zhijxu/add/ort fused adam

zhijxu-MS:
integrate ort fused-adam implementation.

this implementation is mathematically equal to HF adam, and it's faster than HF adam.


zhijxu-MS:
@JingyaHuang  let's discuss here for ort-fused-adam.

with your new changes, i can run it locally, such as "question-answering/run_qa.py".

the [issue ](https://github.com/huggingface/optimum/pull/229#issuecomment-1181914030) you encountered looks like you are using an old ort, could you install a newer one. the install command is `pip uninstall -y onnxruntime-training
pip install --pre --no-deps --no-index onnxruntime-training -f https://download.onnxruntime.ai/onnxruntime_nightly_cu113.html
TORCH_CUDA_ARCH_LIST='7.0;8.0+PTX' python -m torch_ort.configure`

PR #289, opened 2022-07-13
Add the possibility to use an authentication token when loading the calibration dataset

regisss:
# What does this PR do?

This PR enables to use an authentication token to load the calibration dataset. This is necessary because some datasets (e.g. ImageNet) cannot be downloaded from the Hugging Face Hub without such token.

## Before submitting

### DIFF
--- a/optimum/onnxruntime/quantization.py
+++ b/optimum/onnxruntime/quantization.py
@@ -388,6 +388,7 @@ class ORTQuantizer(ABC):
         preprocess_function: Optional[Callable] = None,
         preprocess_batch: bool = True,
         seed: int = 2016,
+        use_auth_token: bool = False,
     ) -> Dataset:
         """
         Create the calibration `datasets.Dataset` to use for the post-training static quantization calibration step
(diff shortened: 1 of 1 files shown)


optimum/onnxruntime/quantization.py
```diff
@@ -404,10 +405,13 @@ def get_calibration_dataset(
...
             preprocess_function (`Callable`, *optional*):
                 Processing function to apply to each example after loading dataset.
-            preprocess_batch (`int`, defaults to `True`):
+            preprocess_batch (`bool`, defaults to `True`):
                 Whether the `preprocess_function` should be batched.
             seed (`int`, defaults to 2016):
                 The random seed to use when shuffling the calibration dataset.
+            use_auth_token (`bool`, defaults to `False`):
+                Whether to use the token generated when running `transformers-cli login` (necessary for some datasets "
+                "like ImageNet).
```
echarlaix, line 414:
```suggestion
            use_auth_token (`bool`, defaults to `False`):
                Whether to use the token generated when running `transformers-cli login` (necessary for some datasets
                like ImageNet).
```

echarlaix (approved):
Thanks for this addition @regisss

PR #296, opened 2022-07-14
Updated the operators_to_quantize doc string for the QuantizationConfig class

Shamik-07:
Added the set of default `operators_to_quantize` and the list of supported operators in the doc.

# What does this PR do?
It updates the `operators_to_quantize` doc string in QuantizationConfig class to list the default operators as well as include the other supported ones.

## Before submitting


optimum/onnxruntime/configuration.py
```diff
@@ -247,7 +247,8 @@ class QuantizationConfig:
         nodes_to_exclude (`list`):
             List of the nodes names to exclude when applying quantization.
         operators_to_quantize (`list`):
-            List of the operators types to quantize.
+            List of the operators types to quantize. Defaults to `["MatMul", "Add"]` but supports 
+            `['Conv','Attention', 'LSTM', 'Gather', 'Transpose', 'EmbedLayerNormalization'].`
```
regisss, line 251: Actually there are more supported operators than these in OnnxRuntime, take a look at [this piece of code](https://github.com/microsoft/onnxruntime/blob/bdd6b00c9a6a5cdd2925762a9058e5c46f4a787f/onnxruntime/python/tools/quantization/registry.py#L37).

So I would just keep the list of default operators and remove the supported ones, maybe with a reference to this link for the interested reader.

regisss (commented):
Thanks for documenting this @Shamik-07!!
I think we should just mention the default ops in the docstring because there are many more supported ops than the ones you mentioned.

regisss:
@Shamik-07 You probably need to rebase your branch on the latest commit so that all the tests pass.

optimum/onnxruntime/configuration.py
Shamik-07: @regisss 
Yes you are absolutely right. 
Thanks for pointing this out. 
Will make the necessary changes.

Shamik-07:
Made the necessary changes.

regisss (approved):
Thanks @Shamik-07!

PR #197, opened 2022-05-18
Fix wrong mapping between dataset label ids and model.config.label2id in examples

fxmarty:
# What does this PR do?

Fix a bug for certain text-classification models using a sentence pair input. 

In text classification, some models have different ids than the dataset label for the same label name. For example, in glue/mnli, the model `roberta-large-mnli` has

```
  "label2id": {
    "CONTRADICTION": 0,
    "ENTAILMENT": 2,
    "NEUTRAL": 1
  },
```

while the dataset labels has the order `"label": {"num_classes": 3, "names": ["entailment", "neutral", "contradiction"]`, effectively swapping "contradiction" and "entailment" labels.

In transformers example, this is fixed using an additional piece of code, see https://github.com/huggingface/transformers/blob/d6b8e9cec7301ba02f642588a6f12e78ec3b9798/examples/pytorch/text-classification/run_glue.py#L395-L413 . This piece of code is missing in the optimization and quantization examples (although not in the training example) and this PR add it back.

Note that, for example for glue/mnli, the `label_to_id` ends up being `{0: 2, 1: 1, 2: 0}`, correcting the swap between the labels 0 and 2.

## Before submitting
- [x] This PR fixes a bug

PR #298, opened 2022-07-15
Fix benchmark tests

fxmarty:
This PR improves the benchmark tests in several ways:
* Use temporary files/folders
* Do not clone optimum main, use the current version (useful in PRs)
* Use tiny models (<10 MB)
* Refactor in a single file
* Remove tests on MacOS taking time 
 
Note that the tests will NOT pass until https://github.com/huggingface/optimum/pull/197 is merged. This is because we use `hf-internal-testing/tiny-bert-for-token-classification` that has a different label mapping than conll2003.

cc @philschmid


philschmid:
@fxmarty what is the state here? When can we get those in?

fxmarty:
Hi @philschmid , feel free to merge https://github.com/huggingface/optimum/pull/197 (I have no merge rights), I can then rebase this PR and it can be merged. One week ago when I did it, I made sure that tests pass given https://github.com/huggingface/optimum/pull/197 is merged.

philschmid:
> Hi @philschmid , feel free to merge https://github.com/huggingface/optimum/pull/197 (I have no merge rights), I can then rebase this PR and it can be merged. One week ago when I did it, I made sure that tests pass given https://github.com/huggingface/optimum/pull/197 is merged.

Ah i didn't know that you cannot merge! 

What about the failing tests?

fxmarty:
One test fails because it uses a tiny model (`hf-internal-testing/tiny-bert-for-token-classification`) that has a different label mapping than conll2003, so the optimum example script gives very low performance results. That's why I was thinking merging https://github.com/huggingface/optimum/pull/197 first, otherwise I can do a PR on the model to fix the label mapping, or use an other (larger) model for the test.

Issue #297, opened 2022-07-15
Pipeline "zero-shot-classification" is not supported


Sudo-Swapnil:
### System Info

```shell
google colab:
   optimum==1.3.0, python==3.7.13

two ubuntu 20.04 systems:
   transformers==4.20.1
   optimum==1.2.3
   python==3.9.12

docker:
   transformers==4.20.1
   optimum==1.2.3
   python==3.9.12
```

### Who can help?

@philschmid 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

I am trying to load "zero-shot-classification" pipeline. I am getting the following error:

```
from optimum.pipelines import pipeline

classifier = pipeline(task="zero-shot-classification", accelerator="ort")
```

```
ValueError                                Traceback (most recent call last)
[<ipython-input-3-0f041b354f55>](https://localhost:8080/#) in <module>()
      1 from optimum.pipelines import pipeline
      2 
----> 3 classifier = pipeline(task="zero-shot-classification", accelerator="ort")

[/usr/local/lib/python3.7/dist-packages/optimum/pipelines.py](https://localhost:8080/#) in pipeline(task, model, tokenizer, feature_extractor, use_fast, use_auth_token, accelerator, **kwargs)
    106 
    107     if targeted_task not in list(SUPPORTED_TASKS.keys()):
--> 108         raise ValueError(f"Task {targeted_task} is not supported. Supported tasks are { list(SUPPORTED_TASKS.keys())}")
    109 
    110     if accelerator != "ort":

ValueError: Task zero-shot-classification is not supported. Supported tasks are []
```

### Expected behavior

Pipeline should download the default model.

regisss:
@Sudo-Swapnil  I don't manage to reproduce it. With Optimum 1.2.3, it works. With Optimum 1.3.0, I get another error:
```shell
Traceback (most recent call last):
  File "test_yolos.py", line 3, in <module>
    classifier = pipeline(task="zero-shot-classification", accelerator="ort")
  File "/home/regis/HuggingFace/dev/tmp_venv/lib/python3.8/site-packages/optimum/pipelines.py", line 113, in pipeline
    load_tokenizer = type(model.config) in TOKENIZER_MAPPING or model.config.tokenizer_class is not None
AttributeError: 'NoneType' object has no attribute 'config'
[...]

PR #301, opened 2022-07-17
Fix bug about accessing model.config when model is None in pipeline

regisss:
# What does this PR do?

This PR fixes the issue mentioned in #297: when instantiating a pipeline with `pipe = pipeline(task=my_task)`, the given `model` is `None` and we get the following error:
```shell
  File "test_yolos.py", line 3, in <module>
    classifier = pipeline(task="zero-shot-classification", accelerator="ort")
  File "/home/regis/HuggingFace/dev/tmp_venv/lib/python3.8/site-packages/optimum/pipelines.py", line 113, in pipeline
    load_tokenizer = type(model.config) in TOKENIZER_MAPPING or model.config.tokenizer_class is not None
AttributeError: 'NoneType' object has no attribute 'config'
```
`model.config` is not accessed anymore to set `load_tokenizer` and `load_feature_extractor`.

Also:

- the current version number is bumped,
- we were using the `pipeline` method from Transformers in `tests/onnxruntime/test_modeling.py`, this is changed for the `pipeline` method of Optimum.

## Before submitting

### DIFF
--- a/optimum/pipelines.py
+++ b/optimum/pipelines.py
@@ -19,6 +19,7 @@ from transformers.feature_extraction_utils import PreTrainedFeatureExtractor
 from transformers.models.auto.feature_extraction_auto import FEATURE_EXTRACTOR_MAPPING
 from transformers.models.auto.tokenization_auto import TOKENIZER_MAPPING
 from transformers.onnx.utils import get_preprocessor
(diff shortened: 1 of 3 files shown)


philschmid (approved):
LGTM

philschmid (approved):
LGTM. Nice that you reused the transformer's things of doing it. ✅

Issue #224, opened 2022-06-20
Different Result between SBERT and Optimum


jonhilgart22:
### System Info

```shell
optimum @ git+https://github.com/huggingface/optimum.git@65bb7fa4f635c6d2e6734485f1b8be58cb690fd0

Linux ubuntu 18.04
```

### Who can help?

@JingyaHuang @echarlaix 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction

Great package!

I'm trying to optimize an SBERT MNLI  model (https://www.sbert.net/) which I'm using to generate embeddings, and I see differences between the Optiumum config and the sbert output.

First, the Sbert output for a example input ['I love burriots'] is shape (1,384), while the output from Optimum is (1,8,384). I'm not sure where the 8 dimension comes from. 

Second, even if I average the output from the `ort_model`, it's not close to the output embeddings.

```python
test_encoding = ["I love burritos!"]
from sentence_transformers import SentenceTransformer
model=SentenceTransformer(all-MiniLM-L6-v2')
model.encode(test_encoding)
...
array([[-6.16856571e-03, -6.18794300e-02,  3.03266663e-02,
         4.37117592e-02, -1.06032923e-01, -5.10813743e-02,
         4.26754877e-02, -1.00264307e-02,  8.52041095e-02,
        -5.51405251e-02, -5.11156097e-02, -1.24764778e-01,
        -1.93562359e-02,  8.46508797e-03, -2.02773847e-02,
        -2.53673140e-02,  6.57850727e-02,  3.77209559e-02,
        -2.69741155e-02,  3.22519913e-02, -4.17455519e-03,
         8.19080994e-02,  2.70250998e-02,  9.85778719e-02,
        -1.37242666e-02,  6.78138584e-02,  3.51302326e-02,
         1.23526119e-02, -1.02652773e-01,  4.19400074e-03,
         3.17004174e-02, -2.85945758e-02,  2.98007298e-02,
        -4.73015457e-02,  2.54707262e-02, -6.96144551e-02,
         2.77759396e-02,  1.61638651e-02, -3.81675586e-02,
        -3.62055339e-02,  6.31522611e-02, -1.37887746e-02,
         8.40439573e-02,  2.10770611e-02, -2.60591563e-02,
[...]

PR #303, opened 2022-07-18
Add ORTModel support for custom tasks

JingyaHuang:
# What does this PR do?

`ORTModelForXXX.model` decided valid inputs and outputs of the model's forward method, thus the creation of inputs in the forward method can be abstract, and also the outputs. This would allow the ORTModels to be more flexible.

e.g. In `ORTTrainer`, the evaluation includes __labels__ as input and __loss__ as output. With the PR, it will enable us to replace bare inference sessions with ORTModels more easily.

## Before submitting


optimum/onnxruntime/modeling_ort.py
```diff
@@ -420,7 +420,8 @@ class ORTModelForQuestionAnswering(ORTModel):
 
     def __init__(self, model=None, config=None, **kwargs):
         super().__init__(model, config, **kwargs)
-        # create {name:idx} dict for model outputs
+        # create {name:idx} dict for model inputs and outputs
+        self.model_inputs = {output_key.name: idx for idx, output_key in enumerate(self.model.get_inputs())}
```
regisss, line 424:
```suggestion
        self.model_inputs = {input_key.name: idx for idx, input_key in enumerate(self.model.get_inputs())}
```
Same

optimum/onnxruntime/modeling_ort.py
```diff
@@ -520,9 +519,9 @@ class ORTModelForSequenceClassification(ORTModel):
 
     def __init__(self, model=None, config=None, **kwargs):
         super().__init__(model, config, **kwargs)
-        # create {name:idx} dict for model outputs
-        self.model_outputs = {output_key.name: idx for idx, output_key in enumerate(self.model.get_outputs())}
+        # create {name:idx} dict for model inputs and outputs
         self.model_inputs = {output_key.name: idx for idx, output_key in enumerate(self.model.get_inputs())}
```
regisss, line 523:
```suggestion
        self.model_inputs = {input_key.name: idx for idx, input_key in enumerate(self.model.get_inputs())}
```
Same

optimum/onnxruntime/modeling_ort.py
```diff
@@ -606,7 +603,8 @@ class ORTModelForTokenClassification(ORTModel):
 
     def __init__(self, model=None, config=None, **kwargs):
         super().__init__(model, config, **kwargs)
-        # create {name:idx} dict for model outputs
+        # create {name:idx} dict for model inputs and outputs
+        self.model_inputs = {output_key.name: idx for idx, output_key in enumerate(self.model.get_inputs())}
```
regisss, line 607:
```suggestion
        self.model_inputs = {input_key.name: idx for idx, input_key in enumerate(self.model.get_inputs())}
```
Same

optimum/onnxruntime/modeling_ort.py
```diff
@@ -689,8 +686,9 @@ class ORTModelForCausalLM(ORTModel, GenerationMixin):
 
     def __init__(self, model=None, config=None, **kwargs):
         super().__init__(model, config, **kwargs)
-        # create {name:idx} dict for model outputs
+        # create {name:idx} dict for model inputs and outputs
         self.main_input_name = "input_ids"
+        self.model_inputs = {output_key.name: idx for idx, output_key in enumerate(self.model.get_inputs())}
```
regisss, line 691:
```suggestion
        self.model_inputs = {input_key.name: idx for idx, input_key in enumerate(self.model.get_inputs())}
```
Same

optimum/onnxruntime/modeling_ort.py
```diff
@@ -807,7 +807,8 @@ class ORTModelForImageClassification(ORTModel):
 
     def __init__(self, model=None, config=None, **kwargs):
         super().__init__(model, config, **kwargs)
-        # create {name:idx} dict for model outputs
+        # create {name:idx} dict for model inputs and outputs
+        self.model_inputs = {output_key.name: idx for idx, output_key in enumerate(self.model.get_inputs())}
```
regisss, line 811:
```suggestion
        self.model_inputs = {input_key.name: idx for idx, input_key in enumerate(self.model.get_inputs())}
```

optimum/onnxruntime/modeling_ort.py
```diff
@@ -346,25 +347,24 @@ def __init__(self, model=None, config=None, **kwargs):
...
+        onnx_outputs = self.model.run(None, onnx_inputs)
+        outputs = dict(
+            map(
+                lambda output_name: (
+                    output_name,
+                    torch.from_numpy(onnx_outputs[self.model_outputs[output_name]]).to(self.device),
+                ),
+                self.model_outputs,
+            )
+        )
```
philschmid, line 365: Have you measured the performance difference for this? Also is there way to improve the readability of the code? having nested `dict`, `map` etc. is not so nice.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -346,25 +347,24 @@ def __init__(self, model=None, config=None, **kwargs):
...
         # converts pytorch inputs into numpy inputs for onnx
-        onnx_inputs = {
-            "input_ids": input_ids.cpu().detach().numpy(),
-            "attention_mask": attention_mask.cpu().detach().numpy(),
-        }
-        if token_type_ids is not None:
-            onnx_inputs["token_type_ids"] = token_type_ids.cpu().detach().numpy()
+        onnx_inputs = dict(
+            map(lambda input_name: (input_name, kwargs.pop(input_name).cpu().detach().numpy()), self.model_inputs)
+        )
```
philschmid, line 354: Similar for here. What are the performance changes by making this change. Also we plan to move to `IOBindinngs`.  
And i am also not a and of removing the `args` from the `forward` pass.

philschmid (changes requested):
I am not sure if we should make this change for enabling something, which the classes aren't designed for. 
You said 
> e.g. In ORTTrainer, the evaluation includes labels as input and loss as output. With the PR, it will enable us to replace bare inference sessions with ORTModels more easily.

This sounds more that we should use `evaluate` and `pipeline` or the ORTModel in the trainer with the post-processing outside of the model.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -335,7 +335,8 @@ class ORTModelForFeatureExtraction(ORTModel):
 
     def __init__(self, model=None, config=None, **kwargs):
         super().__init__(model, config, **kwargs)
-        # create {name:idx} dict for model outputs
+        # create {name:idx} dict for model inputs and outputs
+        self.model_inputs = {output_key.name: idx for idx, output_key in enumerate(self.model.get_inputs())}
```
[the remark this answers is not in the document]
JingyaHuang: Good point, will do!

optimum/onnxruntime/modeling_ort.py
JingyaHuang: IMO what users could probably have as inputs for models in transformers, they shall also be able to have them in the `forward` pass of ORTModels. E.g. for BERT model, we could have [various inputs](https://github.com/huggingface/transformers/blob/main/src/transformers/models/bert/modeling_bert.py#L916), but ORTModels only take `input_ids` and `attention_mask` as default and `token_type_ids` as optional. Although this corresponds to the vanilla definition in `FeatureManager`, we shall enable users to take all possible options in transformers as inputs, in case they have exported their model with customized OnnxConfig.
And this is what I did in the `ORTTrainer.evaluate`, I created [wrappers](https://github.com/huggingface/optimum/blob/main/optimum/onnx/configuration.py#L36) to include labels as input and loss as output. Thus the onnx model for evaluation will have labels and loss, and I don't think that could be possible to use pipeline or evaluate or ORTModel to calculate the extra output in such a situation with a a bare onnx model, as we lost information
```
>>> ort_model = ORTModelForSequenceClassification.from_pretrained(model_id="./assets", file_name="distilbert-base-uncased-wo-labels.onnx")
config.json NOT FOUND in HuggingFace Hub
>>> ort_model.config
PretrainedConfig {
  "transformers_version": "4.20.1"
}
```

JingyaHuang: This refactoring does pretty much the same thing as before just in a more abstract way, as the inputs and outputs are defined once we instantiate the ORTModels. 
I do agree that the mapping is not good as one could put no matter what in the `**kwargs` and iterating the invalid inputs could take computation. What about applying a filter for valid keys?

optimum/onnxruntime/modeling_ort.py
JingyaHuang: I can keep the args! That makes sense!

JingyaHuang: I didn't not consider the IO binding and just adapted our previous code. I open a PR for the replacement with IO binding for inputs and outputs on devices. You think the changes proposed by this PR would impact the replacement by IO binding?

optimum/onnxruntime/modeling_ort.py
philschmid: > IMO what users could probably have as inputs for models in transformers, they shall also be able to have them in the forward pass of ORTModels. 

I don't agree with that. The `ORTModelForXXX` are created for inference. To provide an easy abstraction from `transformers` to use onnxruntime for inference but keep the same API. 

But before we decide on any changes can you benchmark before and after. Since a `dict(map())` can cost quite a lot,  especially for generation models where the forward pass ist called multiple times.

JingyaHuang: For benchmarking, with the following snippet:
```python
tokens = tokenizer(["I love burritos and pizza and pasta!"]*10000, return_tensors="pt")
start_time = time.time()
pred = ort_model(**tokens)
runtime = time.time() - start_time
result = {"Inference_runtime": round(runtime, 4)}
print(result)
```
Previous ORTModel takes `{'Inference_runtime': 16.6773}` and the current one takes `{'Inference_runtime': 16.6948}`. 

I am agreed that `dict(map())` is not the best option in terms of computation speed. But I think making the ORTModels able to take various valid inputs is something that we should enable in terms of flexibility. What about putting back arguments and adding some `if` control flow as what is done with `token_type_ids`?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -283,6 +283,22 @@ def _from_transformers(
         # 3. load normal model
         return cls._from_pretrained(save_dir.as_posix(), **kwargs)
 
+    def prepare_onnx_inputs(self, **kwargs):
```
JingyaHuang, line 286: @philschmid What about something like that? This one took `{'Inference_runtime': 16.687}`, closer to what we had compared to my previous try.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -833,3 +781,39 @@ def forward(
         return ImageClassifierOutput(
             logits=torch.from_numpy(outputs[self.model_outputs["logits"]]),
         )
+
+
+class ORTModelForCustomTasks(ORTModel):
```
philschmid, line 786: i like it!

philschmid (approved):
can you add a test and then we should be good. Good idea! ✅ And if some use case emerges out of it we can add new task-specific model classes

JingyaHuang:
> I still do not understand the purpose of this change.
> 
> As mentioned before the `ORTModelForXX` were introduced for inference and not for training. The idea was to be able to have API compatible Model classes, which can be used with `pipelines` without the need to re-write the pre - & post-processing. Also, the idea is to add those inference model classes to other packages, e.g `optimum-intel`. Additionally, we are looking into removing the copying and adding `IOBindings` to reduce latency in the future.
> 
> The changes you suggest:
> 
> * slow down the inference
> * add a lot of complex dynamic code -> which we tried to exclude, that's why we have several `ORTModelForXX` classes rather than one
> * Add support for something `ORTTrainier` training specific.
> 
> The question I have is:
> 
> * Is this change needed for the ORTTrainer? how are we currently doing it?
> * What is the benefit for the customer?

Hi @philschmid, sorry for the late reply. I re-drafted the code, indeed it shouldn't be in other task-specific models as it will slow them down. The basic idea behind the PR is to leave some flexibility to users, it's like a fallback so that when they are using a more customized model they can still be able to benefit from the ORTModel foundation with a small sacrifice of speed.
`ORTTrainer.train()` is independent of the PR, but for the inference of `ORTTrainer` the `evaluate()` and `predict()`, I am using directly InferenceSession right now and I will replace it with ORTModel, for the predict it is pretty straight forward, but for the evaluate the model include loss thus I need something more customized, and things like `ORTModelForCustomTasks` shall be helpful.

optimum/onnxruntime/modeling_ort.py
JingyaHuang: Cool!! I will make it neater, it's a draft idea right now.

philschmid: Awesome! would you mind adding similar documentation to what we have for the others? Maybe you can be creative with a non-supported task/model, but which would work.

JingyaHuang: Example added!

optimum/onnxruntime/modeling_ort.py
```diff
@@ -833,3 +834,86 @@ def forward(
...
+    >>> from transformers import {processor_class}, pipeline
+    >>> from optimum.onnxruntime import {model_class}
+
+    >>> tokenizer = {processor_class}.from_pretrained("{checkpoint}")
+    >>> model = {model_class}.from_pretrained("{checkpoint}")
+    >>> onnx_extractor = pipeline("feature-extraction", model=model, tokenizer=tokenizer)
+
+    >>> text = "I love burritos!"
+    >>> pred = onnx_extractor(text)
+    ```
```
philschmid, line 869: Does that make sense? Since have the [ORTModelForFeautreExtraction](https://huggingface.co/docs/optimum/onnxruntime/modeling_ort#optimum.onnxruntime.ORTModelForFeatureExtraction)?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -833,3 +834,86 @@ def forward(
...
+    """
+    Onnx Model for any custom tasks. It can be used to leverage the inference acceleration with any custom exported ONNX model.
+    """,
+    ONNX_MODEL_START_DOCSTRING,
+)
+class ORTModelForCustomTasks(ORTModel):
+    """
+    Onnx Model for any custom tasks.
+    """
+
```
philschmid, line 883: Do we need to register it to some Transformers AutoClass similar to the others? that's needed to fix the warnings when pushing into a pipeline

optimum/onnxruntime/modeling_ort.py
```diff
@@ -833,3 +834,86 @@ def forward(
...
+        )
+    )
+    def forward(self, **kwargs):
+        # converts pytorch inputs into numpy inputs for onnx
+        onnx_inputs = self._prepare_onnx_inputs(**kwargs)
+        # run inference
+        onnx_outputs = self.model.run(None, onnx_inputs)
+        outputs = self._prepare_onnx_outputs(onnx_outputs)
+        # converts outputs to namedtuple for pipelines post-processing if applicable
+        return namedtuple("CustomTasksOutput", outputs.keys())(*outputs.values())
```
philschmid, line 901: Lets replace this with `ModelOutput` which is the underlying class for all outputs in transformers: Ref: https://github.com/huggingface/transformers/blob/ee67e7ad4fd7a766891b68f708cf03e30f609976/src/transformers/utils/generic.py#L147

philschmid (approved):
LGTM ✅ Left some minor comments

optimum/onnxruntime/modeling_ort.py
JingyaHuang: The example here is a [sentence BERT with `last_hidden_state` and `pooler_output`](https://huggingface.co/optimum/sbert-all-MiniLM-L6-with-pooler) as outputs. This can be traced back to [an issue](https://github.com/huggingface/optimum/issues/224) where the user wanted to leverage the `pooler_output` instead of the `last_hidden_state`. The [ORTModelForFeautreExtraction](https://huggingface.co/docs/optimum/onnxruntime/modeling_ort#optimum.onnxruntime.ORTModelForFeatureExtraction) only takes `last_hidden_state` as output so the example with pooler is somehow custom(but maybe not that custom...). WDYT? Otherwise, do you have a better example in mind?

philschmid: Nice!

optimum/onnxruntime/modeling_ort.py
JingyaHuang: Ok! I will go with 
```python
pipeline_task = "default"
auto_model_class = AutoModel
```

optimum/onnxruntime/modeling_ort.py
```diff
@@ -847,3 +848,88 @@ def forward(
...
         )
+
+
+CUSTOM_TASKS_EXAMPLE = r"""
+    Example of custom tasks(e.g. a sentence transformers taking `pooler_output` as output):
+
+    ```python
+    >>> from transformers import {processor_class}
+    >>> from optimum.onnxruntime import {model_class}
+    >>> import torch
```
echarlaix, line 859: torch doesn't seem to be used, could we remove it ?

echarlaix (approved):
LGTM, great job @JingyaHuang 🔥

PR #315, opened 2022-07-26
OnnxRuntime tests refactor to make them faster

michaelbenayoun:
# What does this PR do?

The goal of this PR is to make the Onnxruntime tests much faster, as they usually run for more than 1 hour. 
To do that, this PR:
- Fuses multiple tests together in `tests/test_modeling.py`:
  - `test_supported_transformers_architectures`, `test_model_call` are fused to `test_compare_to_transformers`
  - `test_default_pipeline_and_model_device` is fused to `test_pipeline_ort_model`
  - `test_generate_utils_with_input_ids` is fused to `test_generate_utils`
- Uses smaller model architectures to test, instead of big ones

### DIFF
--- a/tests/onnxruntime/test_modeling.py
+++ b/tests/onnxruntime/test_modeling.py
@@ -16,6 +16,7 @@ import gc
 import os
 import tempfile
 import unittest
+import pytest
 
 import torch
 from PIL import Image
@@ -282,6 +283,7 @@ class ORTModelForQuestionAnsweringIntegrationTest(unittest.TestCase):
 
         gc.collect()
 
+    @pytest.mark.run_in_series
     def test_pipeline_model_is_none(self):
         pipe = pipeline("question-answering")
         question = "Whats my name?"
@@ -371,6 +373,7 @@ class ORTModelForSequenceClassificationIntegrationTest(unittest.TestCase):
 
         gc.collect()
 
+    @pytest.mark.run_in_series
     def test_pipeline_model_is_none(self):
         pipe = pipeline("text-classification")
         text = "My Name is Philipp and i live in Germany."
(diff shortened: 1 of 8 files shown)


philschmid (approved):
LGTM

tests/onnxruntime/test_modeling.py
```diff
@@ -406,69 +419,64 @@ def test_pipeline_zero_shot_classification(self):
...
         onnx_model = ORTModelForTokenClassification.from_pretrained(model_id, from_transformers=True)
         tokenizer = get_preprocessor(model_id)
         pipe = pipeline("token-classification", model=onnx_model, tokenizer=tokenizer)
         text = "My Name is Philipp and i live in Germany."
         outputs = pipe(text)
 
-        # compare model output class
+        self.assertEqual(pipe.device, onnx_model.device)
+        # TODO: shouldn't it be all instead of any?
         self.assertTrue(any(item["score"] > 0.0 for item in outputs))
```
michaelbenayoun, line 475: @echarlaix @philschmid should it be any or all?
I noticed the use of any in multiple places, but I'm not sure if it should be all?

JingyaHuang (approved):
Excellent! Can't wait for the merge.

echarlaix (approved):
Thanks for taking care ot it @michaelbenayoun, it looks great !

tests/onnxruntime/test_modeling.py
echarlaix: Good catch !

philschmid: i think i added it to have some "sanity" checks that the output is correct, but yeah all sounds better.

PR #317, opened 2022-07-27
Fix seq2seq onnx config

echarlaix:
This PR allows more flexibility for ONNX Runtime inference of seq2seq models by removing assumption on the ONNX model input names

### DIFF
--- a/optimum/onnxruntime/modeling_seq2seq.py
+++ b/optimum/onnxruntime/modeling_seq2seq.py
@@ -506,6 +506,7 @@ class ORTDecoder:
         self._device = device
         self.input_names = {input_key.name: idx for idx, input_key in enumerate(self.session.get_inputs())}
         self.output_names = {output_key.name: idx for idx, output_key in enumerate(self.session.get_outputs())}
+        self.key_value_input_names = [key for key in self.input_names if "key_values" in key]
 
     @add_start_docstrings_to_model_forward(DECODER_INPUTS_DOCSTRING)
     def forward(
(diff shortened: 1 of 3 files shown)


JingyaHuang (approved):
LGTM! Just added a minor nit on the doc.

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -528,10 +529,9 @@ def forward(
         if past_key_values is not None:
             # Flatten the past_key_values
             past_key_values = [past_key_value for pkv_per_layer in past_key_values for past_key_value in pkv_per_layer]
-
             # Add the past_key_values to the decoder inputs
-            for i, past_key_value in enumerate(past_key_values):
-                onnx_inputs[f"past_key_values_{i}.1"] = past_key_value.cpu().detach().numpy()
+            for input_name, past_key_value in zip(self.key_value_input_names, past_key_values):
```
regisss, line 533:
```suggestion
            for input_name, key_value in zip(self.key_value_input_names, key_values):
```
Maybe the names should be updated since those are not necessarily past key values.

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -528,10 +529,9 @@ def forward(
...
+                onnx_inputs[input_name] = past_key_value.cpu().detach().numpy()
```
regisss, line 534:
```suggestion
                onnx_inputs[input_name] = key_value.cpu().detach().numpy()
```
Same

regisss (approved):
LGTM! :fire: 
I just left one nit.

optimum/onnxruntime/modeling_seq2seq.py
echarlaix: Thanks for your review ! Model inputs will always be "past" key/values (computed during the previous generation step). I have let some freedom in `self.key_value_input_names` to not be too restrictive on the ONNX model export, but I think keeping the variable name as `past_key_values` is good for clarity

PR #332, opened 2022-08-02
Improve the compatibility dealing with large ONNX proto in ORTOptimizer and ORTQuantizer

### LINKED ISSUE #222: [BUG] ONNX optimization fails when optimizing AlbertXXL despite the weights being under 2GB
### System Info

Optimum 1.2.3[onnxruntime-gpu],
PyTorch 1.12.0a0+bd13bc6, 
CUDA 11.6, Ubuntu 18.04, 
Transformers 4.19.0, 
Onnxruntime nightly build (ort-nightly-gpu 1.12.0.dev20220616003) because otherwise there's an error: `Traceback (most recent call last):
  File "run_qa.py", line 524, in <module>
    main()
  File "run_qa.py", line 311, in main
    optimizer.export(
  File "/opt/conda/lib/python3.8/site-packages/optimum/onnxruntime/optimization.py", line 131, in export
    optimizer = optimize_model(
  File "/opt/conda/lib/python3.8/site-packages/onnxruntime/transformers/optimizer.py", line 215, in optimize_model
    temp_model_path = optimize_by_onnxruntime(input,
  File "/opt/conda/lib/python3.8/site-packages/onnxruntime/transformers/optimizer.py", line 96, in optimize_by_onnxruntime
    session = onnxruntime.InferenceSession(onnx_model_path, sess_options, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 335, in __init__
    self._create_inference_session(providers, provider_options, disabled_optimizers)
  File "/opt/conda/lib/python3.8/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 363, in _create_inference_session
    raise ValueError("This ORT build has {} enabled. ".format(available_providers) +
ValueError: This ORT build has ['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider'] enabled. Since ORT 1.9, you are required to explicitly set the providers parameter when instantiating InferenceSession. For example, onnxruntime.InferenceSession(..., providers=['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider'], ...)`

The Onnxruntime needs a newer version than the most recently released tag because even the newest release (1.11.1) doesn't yet specify an explicit execution provider, hence we use the nightly build instead.

### Who can help?

@JingyaHuang @lewtun @mich

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

### Reproduction

The provided run_qa.py script in the `examples/onnxruntime/optimization/question-answering` doesn't work as expected with ahotrod/albert_xxlargev1_squad2_512.

To reproduce: 
`python run_qa.py \
    --model_name_or_path ahotrod/albert_xxlargev1_squad2_512 \
    --dataset_name squad_v2 \
    --optimization_level 99 \
    --do_eval \
    --output_dir /home/ubuntu/albert_xxlargev1_squad2_512_onnx_optimized \
    --execution_provider CUDAExecutionProvider \
    --optimize_for_gpu`

The resulting error:
`Traceback (most recent call last):
  File "run_qa.py", line 524, in <module>
    main()
  File "run_qa.py", line 311, in main
    optimizer.export(
  File "/opt/conda/lib/python3.8/site-packages/optimum/onnxruntime/optimization.py", line 142, in export
    optimizer.save_model_to_file(onnx_optimized_model_output_path, use_external_data_format)
  File "/opt/conda/lib/python3.8/site-packages/onnxruntime/transformers/models/gpt2/../../onnx_model.py", line 934, in save_model_to_file
    save_model(self.model, output_path)
  File "/opt/conda/lib/python3.8/site-packages/onnx/__init__.py", line 202, in save_model
    s = _serialize(proto)
  File "/opt/conda/lib/python3.8/site-packages/onnx/__init__.py", line 71, in _serialize
    result = proto.SerializeToString()
ValueError: Message onnx.ModelProto exceeds maximum protobuf size of 2GB: 3038449703`

This error triggers the ModelProto > 2gb error despite the model weights being less than 2gb (~870MB, rather).

### Expected behavior

```shell
The optimized ONNX model is successfully saved.
```

JingyaHuang:
# What does this PR do?

- [x] Add `all_tensors_to_one_file` option to `ORTOptimizer` when exporting the ONNX model.

P.S. only the ONNXModel of [optimization](https://github.com/microsoft/onnxruntime/blob/v1.12.1/onnxruntime/python/tools/transformers/onnx_model.py#L900) has added the option, not [quantization](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/onnx_model.py#L333).
- [x] Refactoring the compatibility of `ORTOptimizer` and `ORTQuantizer` in cases of large ONNX proto(export path / load ONNX)
- [x] Update the optimization and quantization examples with the new option.

Fixes #222 

Related to https://github.com/onnx/onnx/issues/4394

__P.S.__ 
* Currently the Optimizer exports the original ONNX to the same folder as the optimized ONNX model, `ORTOptimizer` would better cache the original model as the`ORTQuantizer`, this PR will not target the issue, it could be done either in the refactorization PR #294 or in a new PR.
* [onnx offers options](https://github.com/onnx/onnx/blob/main/onnx/__init__.py#L171): `save_as_external_data`, `all_tensors_to_one_file`, `size_threshold(1024)`, `convert_attribute(false)`, and the last two are not supported to be self-defined [in onnxruntime](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/python/tools/transformers/onnx_model.py#L926) yet. But in general, the first two are the most effective.
* Currently onnxruntime export external files and model proto after quantization to different directory and this might need to be fixed from their side. https://github.com/microsoft/onnxruntime/issues/12576


optimum/onnxruntime/quantization.py
```diff
@@ -315,7 +315,7 @@ def quantize(
             quantization_config.nodes_to_quantize = list(nodes_to_quantize)
             quantization_config.nodes_to_exclude = list(nodes_to_exclude)
 
-        onnx_model = onnx.load(self.onnx_model_path)
+        onnx_model = onnx.load(self.onnx_model_path.as_posix())
```
regisss, line 318: Was there not an issue with `as_posix()` lately? Not sure about this, I may be mixing things up.

regisss (commented):
Thanks for this PR @JingyaHuang!!
I just have one comment which may be unrelated to this PR.

optimum/onnxruntime/quantization.py
JingyaHuang: Hi Regis, after some tests, I found that onnx treats `str` better than `PosixPath` in case of large proto(with external files). It comes from the fact that [onnx._get_file_path](https://github.com/onnx/onnx/blob/e6547e8b41578c766dd2ccf298a01b2b1b6335cd/onnx/__init__.py#L50-L56) returns os.path.abspath(f.name) instead of os.path.abspath(f) which remove the folder from the path.

This means:
* When exporting onnx model with external files, the external tensors will be put outside of the folder where lies the model proto.
image

* When loading a model with external files, it will also search outside of the folder.

Theoretically, we can choose either str or PosixPath, but I think that it would be more logical to have everything in the `save_dir` defined by the users
.

Actually it would be more robust with `Path(self.onnx_model_path).as_posix()`, I will make the change!

regisss (approved):
LGTM, thanks Jingya!!

optimum/onnxruntime/optimization.py
```diff
@@ -118,7 +122,11 @@ def optimize(
         self.config.save_pretrained(save_dir)
 
         # Create and save the configuration summarizing all the parameters related to optimization
-        ort_config = ORTConfig(optimization=optimization_config)
+        ort_config = ORTConfig(
+            optimization=optimization_config,
+            use_external_data_format=use_external_data_format,
+            all_tensors_to_one_file=all_tensors_to_one_file,
```
JingyaHuang, line 128: Hey Ella, I added those two flags while saving the config, does that make sense to you?

examples/onnxruntime/optimization/multiple-choice/run_swag.py
```diff
@@ -164,17 +164,35 @@ class OptimizationArguments:
     )
 
 
+@dataclass
+class OnnxExportArguments:
+    """
+    Arguments to decide how the ModelProto will be saved.
+    """
+
+    use_external_data_format: bool = field(
```
echarlaix, line 173: I don't think `use_external_data_format` should be an argument here as it will not have any impact on the model onnx [export](https://github.com/huggingface/transformers/blob/main/src/transformers/onnx/convert.py#L173). Why not add an `ort_config` attribute to our `ORTModel` with the corresponding information (+  `opset`) and use it in when instantiating our `ORTOptimizer` / `ORTQuantizer` ?

optimum/onnxruntime/optimization.py
echarlaix: What is the advantage of setting `all_tensors_to_one_file` to `False` ? Also the name used by onnnxruntime is not super intuitive so I would be for renaming it to something clearer. It's good point to add the `use_external_data_format` to the ort config !

examples/onnxruntime/optimization/multiple-choice/run_swag.py
JingyaHuang: That's a good point! Indeed if the original export exceeds 2GB, there would be no need to consider the export after the optimization or the quantization. I will add the attribute in ORTModel as you suggest, thx!

JingyaHuang (commented):
Thanks for reviewing, I will put the arguments in ORTModel as suggested and make necessary changes in the examples.

optimum/onnxruntime/optimization.py
echarlaix: Could this parameter be inferred (depending on the model size) instead of being explicitly set ?

echarlaix: Thanks for the explanation, do you think you can add it in the docstrings ? Also the information that `all_tensors_to_one_file` is only relevant when `use_external_data_format` is set to `True`. And do you think we can rename `all_tensors_to_one_file` to something which explicitly link it to the external data format such as `one_external_file` or `not external_file_for_each_tensor` ?

examples/onnxruntime/optimization/multiple-choice/run_swag.py
echarlaix: My point is that it could be misleading for users as they could think that this will impact the model ONNX export (before optimization) so I think we should specify it. Also I think it could make sense to add an `use_external_data_format` attribute  to our `ORTModel` and save it in our `ORTConfig`  in order to use it in case `use_external_data_format` is not specified when instantiating an `ORTOptimizer` / `ORTQuantizer` as it could be a better default value than `False`, in any case this should be done on a separate PR

echarlaix (approved):
LGTM !

optimum/onnxruntime/optimization.py
JingyaHuang: @echarlaix `one_external_file` sounds good to me, will update the name and add description of its relationship with `use_external_data_format` in the docstring.

JingyaHuang:
Moved config(model and ORT) saving after saving the model, otherwise will get the following error as configs are saved in the folder:
```
RuntimeError: Output directory (results/albert_largev1_squad2) for external data is not empty.
```

While running the template example, I got the error message in the evaluation phase(no problem with the optimization, and models with external files are properly saved):
```
  File "run_qa.py", line 540, in <module>
    main()
  File "run_qa.py", line 492, in main
    metrics = compute_metrics(predictions)
  File "run_qa.py", line 460, in compute_metrics
    return metric.compute(predictions=p.predictions, references=p.label_ids)
  File "/usr/local/lib/python3.8/dist-packages/evaluate/module.py", line 444, in compute
    output = self._compute(**inputs, **compute_kwargs)
  File "/root/.cache/huggingface/modules/evaluate_modules/metrics/evaluate-metric--squad/b4e2dbca455821c7367faa26712f378254b69040ebaab90b64bdeb465e4a304d/squad.py", line 110, in _compute
    score = compute_score(dataset=dataset, predictions=pred_dict)
  File "/root/.cache/huggingface/modules/evaluate_modules/metrics/evaluate-metric--squad/b4e2dbca455821c7367faa26712f378254b69040ebaab90b64bdeb465e4a304d/compute_score.py", line 67, in compute_score
    exact_match += metric_max_over_ground_truths(exact_match_score, prediction, ground_truths)
  File "/root/.cache/huggingface/modules/evaluate_modules/metrics/evaluate-metric--squad/b4e2dbca455821c7367faa26712f378254b69040ebaab90b64bdeb465e4a304d/compute_score.py", line 52, in metric_max_over_ground_truths
    return max(scores_for_ground_truths)
ValueError: max() arg is an empty sequence
```
Seems like a bug with the metric, an update might need to be done with the examples. I can take it over if you don't have bandwidth @echarlaix.

PR #335, opened 2022-08-03
Fix bug in docs/combine_docs.py when there are subsections

regisss:
# What does this PR do?

Fixes a bug where building documentation fails because of subsections (for instance [here](https://github.com/huggingface/optimum-habana/blob/main/docs/source/_toctree.yml) where sections are not all at the same level). [This](https://github.com/huggingface/optimum/runs/7645249951?check_suite_focus=true) is an example where this bug occurred.

More precisely, in [`docs/combine_docs.py`](https://github.com/huggingface/optimum/blob/main/docs/combine_docs.py), local files are renamed to add a prefix corresponding to the associated subpackage:
```python
# Extend table of contents sections with subpackage name prefix
for item in subpackage_toc:
    for file in item["sections"]:
        file["local"] = f"{subpackage}_" + file["local"]
```
This will fail when `item["sections"]` contains another `"section"` key because the `"local"` key will then be inside the second section, leading to:
```
  File "docs/combine_docs.py", line 48, in <module>
    main()
  File "docs/combine_docs.py", line 40, in main
    file["local"] = f"{subpackage}_" + file["local"]
KeyError: 'local'
```

This PR proposes to check if `"local"` is in `item["sections"]` and to do a recursive call if this condition is not verified in order to reach the subsection level.

## Before submitting

### DIFF
--- a/docs/combine_docs.py
+++ b/docs/combine_docs.py
@@ -1,6 +1,7 @@
 import argparse
 import shutil
 from pathlib import Path
+from typing import Dict
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


regisss:
I added @lewtun as a reviewer so that he does not miss this PR when he is back.

philschmid (approved):
LGTM

PR #345, opened 2022-08-10
Fix CPU device parameter

mausch:
# What does this PR do?

Fixes a crash in Trainer using CPU.

This parameter is expected to be of type `torch.device` not `str`.

Otherwise this crashes with:
```
../../.mamba/envs/score/lib/python3.7/site-packages/optimum/onnxruntime/trainer.py:1315: in _export
    self.model.to("cpu")
../../.mamba/envs/score/lib/python3.7/site-packages/optimum/onnxruntime/modeling_ort.py:122: in to
    provider = get_provider_for_device(self.device)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

device = 'cpu'

    def get_provider_for_device(device: torch.device) -> str:
        """
        Gets the ONNX Runtime provider associated with the PyTorch device (CPU/CUDA).
        """
>       return "CUDAExecutionProvider" if device.type.lower() == "cuda" else "CPUExecutionProvider"
E       AttributeError: 'str' object has no attribute 'type'

../../.mamba/envs/score/lib/python3.7/site-packages/optimum/onnxruntime/utils.py:165: AttributeError
```

## Before submitting

### DIFF
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -1310,7 +1310,7 @@ class ORTTrainer(Trainer):
         else:
             if not (self.args.fp16 and self.args.deepspeed):
                 # Taking CPU to export the model
-                self.model.to("cpu")
+                self.model.to(torch.device("cpu"))
             model = unwrap_model(self.model)
 
         model_type, model_onnx_config = FeaturesManager.check_supported_model_or_raise(model, feature=self.feature)


JingyaHuang:
Hi @mausch,

The error message is a little bit ambiguous. Can you put the complete error message here or put a snippet for me to reproduce the error?

For the time being `ORTTrainer` takes only PyTorch model for instantiation. It seems to me that you have probably passed an ORTModel to it? Is that so?

PR #352, opened 2022-08-17
Free some memory in test

regisss:
# What does this PR do?

Fixes # (issue)

## Before submitting


regisss:
I ran the Windows ORT tests twice and they passed each time. It should be good now.

philschmid (approved):
Thank you so much for doing that🤗 
I thought we changed all models but seems we forgot one

regisss:
> Thank you so much for doing thathugs I thought we changed all models but seems we forgot one

Actually, the models defined in `MODEL_NAMES` in `test_modeling.py` were changed. But for this specific test that used to fail, we use `model=None` in the pipeline so we defer to the default pipeline models defined in `pipeline.py`.

PR #353, opened 2022-08-18
Add documentation dispatch

philschmid:
# What does this PR do?

Adds a way to trigger the action through a dispatch


regisss (approved):
Very clean and simple! Thanks @philschmid :fire: 

The doc build test didn't pass, I'm going to fix that quickly in optimum-habana to make sure we are all green here before merging.

lewtun (approved):
Nice solution @philschmid - LGTM with the fix from @regisss in the doc build

PR #356, opened 2022-08-23
Do not force download by default in ORTModel

fxmarty:
We passed by default `force_download=True`, although the caching works well when passing `force_download=False`. Especially when passing `from_transformers=False`, everything works as expected as the onnx model is loaded from the cache `hub/` subfolder.

When passing `from_transformers=True`, although we cache only the PyTorch model (before [the new transformers cache system ](https://github.com/huggingface/transformers/pull/18438) in `transformers/` cache subfolder, in `hub/` subfolder in the next transformers release), the `force_download` argument is anyway ignored in `_from_transformers` so this PR has no impact there:

https://github.com/huggingface/optimum/blob/e90332463f7d2c1ae7e6feb93cfb379f92ed6231/optimum/onnxruntime/modeling_ort.py#L239-L303

### DIFF
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -199,7 +199,6 @@ class OptimizedModel(ABC):
         Returns:
             `OptimizedModel`: The loaded optimized model.
         """
-
         revision = None
         if len(str(model_id).split("@")) == 2:
             model_id, revision = model_id.split("@")
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -193,10 +193,13 @@ class OptimizedModel(ABC):
             cache_dir (`str`, *optional*, defaults to `None`):
                 Path to a directory in which a downloaded pretrained model configuration should be cached if the
                 standard cache should not be used.
+            local_files_only(`bool`, *optional*, defaults to `False`):
+                Whether or not to only look at local files (i.e., do not try to download the model).
 
         Returns:
             `OptimizedModel`: The loaded optimized model.
         """
+
         revision = None
         if len(str(model_id).split("@")) == 2:
             model_id, revision = model_id.split("@")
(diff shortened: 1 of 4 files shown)


regisss (approved):
LGTM @fxmarty!

If the `force_download` argument is not used in `_from_transformers`, maybe we should remove it what do you think?

philschmid (changes requested):
That's awesome! But we have to add test to validate that it is working as expected. Especially since we are storing the "model.onnx" files at the top level with out any hash. So there is no way for the library to know if the cached "model.onnx" is the correct one or not.

We can add the `local_files_only` argument to the from_pretrained method and have then working and failing tests for "cached" and regular models. As well as for loading different models after each other to make sure we have the correct ones loaded. 
See documentation: https://huggingface.co/docs/huggingface_hub/package_reference/file_download#huggingface_hub.hf_hub_download.local_files_only

We should also check that this works for the seq2seq models. 

And we should check if the new cache system allows better distribution of onnx files. Currently, when a user loads and saves a onnx file it is saved at the top cache directory with `model.onnx`, which means if he loads a new model "from" cache it uses the old model. Thats why i went with "force_download" in the beginning.

fxmarty:
When loading an onnx model with `from_transformers=False`, here's where it is loaded from in the `hub/` subfolder:

image

For seq2seq models (with `from_transformers=False`), we have:

image


which does work as well. Thanks for the suggestion to add tests, I will try to add some!

When `from_transformers=True`, the argument `force_download` is ignored. 

In case we passed `from_transformers=True`, in the old caching system, we indeed get e.g. a `transformers/facebook/levit-256/model.onnx` file with no hash recorded. However, even if this file exists, it is NOT loaded into the `InferenceSession` and the cached pytorch model is converted to onnx and loaded into an InferenceSession.

Under the new caching system, if we pass `from_transformers=True`, we get e.g. a `hub/facebook/levit-256/model.onnx` with no hash recorded as well. Looking at the timestamps, this file is overwritten at every conversion.

In any case, I think the case `from_transformers=True` is independent from this PR.

philschmid (approved):
LGTM!

PR #363, opened 2022-08-26
fix(optimization): handle empty file suffix

### LINKED ISSUE #364: Empty file suffix when optimizing or quantizing is not handled
### System Info

```shell
latest master
```

### Who can help?

@echarlaix 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

### Reproduction

Optimize or quantize a model but user do not want a suffix so use : 
`file_suffix=None` => RuntimeError
`file_suffix=""` => file_name like `encoder_.onnx`

### Expected behavior

Have the same name as the original onnx file like `encoder.onnx`

#363 fixes that

ierezell:
# What does this PR do?
Fixes #364 

## Problem : 
Optimize or quantize a model but the user does not want a suffix so the user tries to use :

`file_suffix=None` => RuntimeError (for optimization)
`file_suffix=""` => saved file name is like `encoder_model_.onnx`

## Solution : 
Little logic to handle optional or empty file_suffix

## Doc : 
The doc was already saying the parameter could be optional (even if not with the current code)

### DIFF
--- a/optimum/onnxruntime/quantization.py
+++ b/optimum/onnxruntime/quantization.py
@@ -286,6 +286,8 @@ class ORTQuantizer(OptimumQuantizer):
             The path of the resulting quantized model.
         """
         use_qdq = quantization_config.is_static and quantization_config.format == QuantFormat.QDQ
+        save_dir = Path(save_dir)
+        save_dir.mkdir(parents=True, exist_ok=True)
 
         if not quantization_config.is_static:
             if quantization_config.mode != QuantizationMode.IntegerOps:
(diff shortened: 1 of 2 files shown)


ierezell:
Note that for quantization I moved the `os.makedirs` closer to the place it's used. 
It's just a personal preference so I can for sure revert it if you don't like that.

optimum/onnxruntime/quantization.py
```diff
@@ -343,6 +343,8 @@ def quantize(
         LOGGER.info("Quantizing model...")
         quantizer.quantize_model()
 
+        suffix = f"_{file_suffix}" if file_suffix else ""
+        quantized_model_path = Path(save_dir).joinpath(f"{self.onnx_model_path.stem}{suffix}").with_suffix(".onnx")
```
echarlaix, line 347:
```suggestion
        quantized_model_path = save_dir.joinpath(f"{self.onnx_model_path.stem}{suffix}").with_suffix(".onnx")
```

echarlaix (commented):
Thanks for the fix @ierezell !

optimum/onnxruntime/quantization.py
ierezell: Thanks ! It's changed.

echarlaix (approved):
LGTM ! 🔥

Issue #344, opened 2022-08-10
cannot import name 'cached_path' from 'transformers.file_utils'


tsaiHY:
### System Info

```shell
from optimum.onnxruntime import ORTQuantizer, ORTModelForTextClassification

then:

RuntimeError: Failed to import optimum.onnxruntime.quantization because of the following error (look up to see its traceback):
cannot import name 'cached_path' from 'transformers.file_utils'
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

from optimum.onnxruntime import ORTQuantizer, ORTModelForTextClassification

### Expected behavior

this would not be fail

tsaiHY:
transformers-4.21.1 won't fail

nbroad1881:
I believe this will need to be patched to work for two versions of transformers: before 4.22.0 and after 4.22.0

The file utils were modified in this pull request: https://github.com/huggingface/transformers/pull/18497 and a few later ones.  

Since many functions were removed and modified in 4.22.0, I looked into which functions would need to change. Below, the left side has the names that work for < 4.22.0, and the right side has the names of the new functions that can be used for >= 4.22.0. This was only from a brief skim of the old and new code:

cached_path -> cached_file

get_list_of_files -> (no clear replacement)

hf_bucket_url/is_remote_url -> get_file_from_repo

@echarlaix @michaelbenayoun

echarlaix:
Thanks for the detailed answer @nbroad1881 ! I opened https://github.com/huggingface/optimum/pull/368 to fix the issues you listed.

jermwatt:
This is still an issue for me using transformers 4.31.0 - downgrading to 4.21 solved as a workaround.

fxmarty:
Hi @jermwatt, this issue should not happen anymore: https://github.com/huggingface/optimum/blob/0b08a1fd19005b7334aa923433b3544bd2b11ff2/optimum/configuration_utils.py#L41 Which version of optimum are  you running?

PR #368, opened 2022-09-02
Adapt BaseConfig for next transformers release

echarlaix:
In this PR we fix https://github.com/huggingface/optimum/issues/344 by adapting `BaseConfig` in preview of transformers release v4.22.0

### DIFF
--- a/optimum/configuration_utils.py
+++ b/optimum/configuration_utils.py
@@ -22,7 +22,9 @@ from typing import Any, Dict, Optional, Tuple, Union
 from packaging import version
 from transformers import PretrainedConfig
 from transformers import __version__ as transformers_version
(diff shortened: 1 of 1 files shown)


philschmid (commented):
@echarlaix do we also need to make changes to the `ORTModel` classes for the `from_pretrained` method?

echarlaix:
@philschmid No as `ORTModel` does not use any function that will be impacted by the breaking changes introduced in transformers 4.22.0

JingyaHuang (approved):
LGTM! 
Maybe not for this PR, but in general for codes inspired by transformers, do you think we shall add a line of transformers version check?

echarlaix:
Good catch @JingyaHuang, I will add it before merging

PR #369, opened 2022-09-06
Fix documentation

philschmid:
# What does this PR do?

Fix wrong model class in documentation


echarlaix (approved):
Good catch @philschmid, it also seems that the same error is present in the optimization documentation

philschmid:
Made the changes for `optimization.mdx` as well

PR #373, opened 2022-09-08
Add default clone url for making doc

### LINKED ISSUE #372: `Combine subpackages documentation` step silently failing
image

[Push to repositories](https://github.com/huggingface/optimum/actions/runs/3006577791/workflow#L113) step is failing because [Combine subpackage documentation](https://github.com/huggingface/optimum/actions/runs/3006577791/workflow#L106) silently fails as you can see in the img above & [failed run/log](https://github.com/huggingface/optimum/runs/8225041105?check_suite_focus=true) here.

In the last step [Push to repositories](https://github.com/huggingface/optimum/actions/runs/3006577791/workflow#L113), we should have this folder structure:
```
build_dir/
├─ optimum/
│  ├─ main OR v.1.2.3/ # folder that contains built doc artifacts
│  ├─ _versions.yml
```

but we have this (without the doc built artifacts folder):
```
build_dir/
├─ optimum/
│  ├─ _versions.yml
```
because Combine subpackage documentation](https://github.com/huggingface/optimum/actions/runs/3006577791/workflow#L106) is failing

regisss:
# What does this PR do?

This PR fixes #372 where the generation of the main documentation fails. This happens because the clone URL is only taken from the PR which triggered the workflow, and since the main documentation build workflow is not triggered by a PR no clone URL is given. The fix consists in checking if the clone URL is empty or not, and give it a default value if it is.

## Before submitting

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,9 @@
 
 SHELL := /bin/bash
 CURRENT_DIR = $(shell pwd)
+DEFAULT_CLONE_URL := https://github.com/huggingface/optimum.git
+# If CLONE_URL is empty, revert to DEFAULT_CLONE_URL
+REAL_CLONE_URL = $(if $(CLONE_URL),$(CLONE_URL),$(DEFAULT_CLONE_URL))
 
 .PHONY:	style test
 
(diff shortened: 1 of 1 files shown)


mishig25 (approved):
lgtm

lewtun (approved):
Thanks for the fix, LGTM!

PR #375, opened 2022-09-08
Checkout branch name for new versions when building the doc

regisss:
# What does this PR do?

In the Docker container building the documentation, the default branch is `main` if no commit SHA/branch name/tag name is given as argument (`COMMIT_SHA_OPTIMUM=...`). This means that the documentation of the main branch is generated when a new version is pushed to a release branch, and not the documentation of the release branch. This PR fixes that.

## Before submitting


lewtun (approved):
Thanks for fixing this @regisss - you're a doc building pro now :)

LGTM!

PR #378, opened 2022-09-12
Remove extra "habana-deepspeed"

regisss:
# What does this PR do?

The extra "habana-deepspeed" has to be removed because `optimum-habana[deepspeed]` makes uploading `optimum-habana` on PyPI impossible since it does not accept direct dependencies specified as URLs like Habana Deepspeed. See https://github.com/huggingface/optimum-habana/pull/96.

## Before submitting


echarlaix (approved):
LGTM

PR #384, opened 2022-09-15
Add evaluate to setup

JingyaHuang:
# What does this PR do?

Benchmark suite CI failing due to the lack of evaluate.

P.S. Need to gradually replace `datasets.load_metric` in tests with `evaluate`.


JingyaHuang:
@fxmarty Can you have a check if that's good for benchmark test?

fxmarty:
Yes it's all good - in any case we stop using `datasets.load_metric` in https://github.com/huggingface/optimum/pull/318 . If you are up for reviewing it I can rebase and we merge it as well.

PR #386, opened 2022-09-16
BaseConfig update for transformers>=4.22.0

michaelbenayoun:
# What does this PR do?

This PR updates the `BaseConfig` class to synchronize with the updates performed on the `transformers.PretrainedConfig`, and fixes the issue induced by the `transformers` 4.22 release, breaking the current `BaseConfig.push_to_hub` method.

### DIFF
--- a/optimum/configuration_utils.py
+++ b/optimum/configuration_utils.py
@@ -1,3 +1,4 @@
+# coding=utf-8
 #  Copyright 2021 The HuggingFace Team. All rights reserved.
 #
 #  Licensed under the Apache License, Version 2.0 (the "License");
@@ -95,8 +96,10 @@ class BaseConfig(PretrainedConfig):
     def get_configuration_file(cls, configuration_files: List[str]) -> str:
         """
         Get the configuration file to use for this version of transformers.
+
         Args:
             configuration_files (`List[str]`): The list of available configuration files.
+
         Returns:
             `str`: The configuration file to use.
         """
@@ -109,6 +112,7 @@ class BaseConfig(PretrainedConfig):
                 configuration_files_map[v] = file_name
         available_versions = sorted(configuration_files_map.keys())
 
+        # Defaults to FULL_CONFIGURATION_FILE and then try to look at some newer versions.
         configuration_file = cls.CONFIG_NAME
         optimum_version = version.parse(__version__)
         for v in available_versions:
--- a/optimum/configuration_utils.py
+++ b/optimum/configuration_utils.py
@@ -24,7 +24,12 @@ from packaging import version
 from transformers import PretrainedConfig
 from transformers import __version__ as transformers_version
 from transformers.dynamic_module_utils import custom_object_save
-from transformers.utils import is_offline_mode
+from transformers.utils import (
+    cached_file,
+    download_url,
+    extract_commit_hash,
+    is_remote_url,
+)
 
 from huggingface_hub import hf_hub_download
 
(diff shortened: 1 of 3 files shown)


optimum/configuration_utils.py
```diff
@@ -117,148 +124,133 @@ def get_configuration_file(cls, configuration_files: List[str]) -> str:
...
-        """
-        original_kwargs = copy.deepcopy(kwargs)
-        # Get config dict associated with the base config file
-        config_dict, kwargs = cls._get_config_dict(pretrained_model_name_or_path, **kwargs)
-        if "_commit_hash" in config_dict:
-            original_kwargs["_commit_hash"] = config_dict["_commit_hash"]
-
-        # That config file may point us toward another config file to use.
-        if "configuration_files" in config_dict:
-            configuration_file = cls.get_configuration_file(config_dict["configuration_files"])
```
echarlaix, line 142: We should call `cls.get_configuration_file` here ([see](https://github.com/huggingface/transformers/blob/main/src/transformers/configuration_utils.py#L562)), so you shouldn't remove `get_config_dict()`

michaelbenayoun: You're right, thanks!

optimum/configuration_utils.py
```diff
@@ -151,114 +187,153 @@ def get_config_dict(
...
+                        cache_dir=cache_dir,
+                        force_download=force_download,
+                        resume_download=resume_download,
+                        proxies=proxies,
+                        use_auth_token=use_auth_token,
+                        local_files_only=local_files_only,
+                    )
+                else:
+                    # Load from local folder or from cache or download from model Hub and cache
+                    resolved_config_file = cached_file(
```
echarlaix, line 251: We should either use `cached_path` and `cached_file` (depending on which transformers version) or use `hf_hub_download` , I prefer the second option personnaly

michaelbenayoun: I changed, and use `cached_file` to match what was done for transformers <= 4.21.0
For the rest, I think it is better to match the `transformers` implementation since `BaseConfig` is just a customizable `transformers.PretrainedConfig`. We actually want to copy anything related to pushing to hub from this class, so we should use their implementation when it is possible, and here it is.

echarlaix: I don't see the gain in using `cached_path` and `cached_file` : our current implementation  with `hf_hub_download` is equivalent so I would be in favor of the option where we don't have additional statements to check which transformers version is used (option which will be more difficult to maintain)

michaelbenayoun: - `cached_path` is from the implementation from `transformers <= 4.21.0` and we are just matching the base class we are extending here
- `cached_file` uses `hf_hub_download` underneath, and does other stuff related to the offline mode and loading from the cache with a commit hash (which seems to be new). Once again, `BaseConfig` inherints from`transformers.PretrainedConfig`, so matching its behaviour will prevent issues: what if someday `cached_file` is further updated? Here we don't have a constraint on what to use, so we should use what's identical from what we are basing our code.

About the complexity, we will remove those if / else statements in a few release, once transformers 4.22 has been well adopted.

echarlaix: I would be in favor of removing the `PretrainedConfig` inheritence as some modifications in `PretrainedConfig` could potentially lead to some unexpected behavior and tracking every changes would be time consuming. I am ok with merging https://github.com/huggingface/optimum/pull/386 in the meantime

setup.py
```diff
@@ -15,7 +15,7 @@
 REQUIRED_PKGS = [
     "coloredlogs",
     "sympy",
-    "transformers[sentencepiece]>=4.20.1",
+    "transformers[sentencepiece]>=4.22.0",
```
echarlaix, line 18:
```suggestion
    "transformers[sentencepiece]>=4.20.1",
```

echarlaix (approved):
LGTM, thanks for taking care of this @michaelbenayoun

PR #398, opened 2022-09-26
Use more extra quantization options of ONNX Runtime

fxmarty:
Pass `AddQDQPairToWeight`, `DedicatedQDQPair`, `QDQOpTypePerChannelSupportToAxis` to `QDQQuantizer` or `ONNXQuantizer`, as those configuration attributes are available in `QuantizationConfig` but were not passed up to now.

This is useful if we later aim at using `TensorrtExecutionProvider` with a static quantized model.


philschmid (approved):

JingyaHuang (approved):
LGTM!

echarlaix (approved):
LGTM !

PR #401, opened 2022-09-26
Add possibility to pass `provider_options` to `ORTModel.from_pretrained()`

fxmarty:
`InferenceSession` init can take the `provider_options` argument, that may set some custom parameters depending on the execution provider: see https://onnxruntime.ai/docs/api/python/api_summary.html#inferencesession & https://onnxruntime.ai/docs/api/c/group___global.html .

This is useful for example when using `TensorrtExecutionProvider`, we need to pass `{"trt_int8_enable": True}` to enable TensorRT INT8 precision, otherwise getting the error `Internal Error (Int8 precision has been set for a layer or layer output, but int8 is not configured in the builder)`.

## Before submitting

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -160,6 +160,7 @@ class ORTModel(OptimizedModel):
         path: Union[str, Path],
         provider: Optional[str] = "CPUExecutionProvider",
         session_options: Optional[ort.SessionOptions] = None,
+        provider_options: Optional[Dict] = None,
         **kwargs
     ):
         """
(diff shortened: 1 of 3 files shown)


optimum/onnxruntime/modeling_ort.py
```diff
@@ -171,16 +172,25 @@ def load_model(
...
                 f"Asked to use {provider} as an ONNX Runtime execution provider, but the available execution providers are {available_providers}."
             )
 
-        return ort.InferenceSession(path, providers=[provider], sess_options=session_options)
+        # `providers` list must of be of the same length as `provider_options` list
+        return ort.InferenceSession(
+            path,
+            providers=[provider],
+            sess_options=session_options,
+            provider_options=None if provider_options is None else [provider_options],
```
philschmid, line 192: is `None` the default?

philschmid (approved):
LGTM

optimum/onnxruntime/modeling_ort.py
fxmarty: Of `InferenceSession` in ONNX Runtime, yes: https://github.com/microsoft/onnxruntime/blob/94e34ace15f4c7011590d387c5dcabaa3e446d7f/onnxruntime/python/onnxruntime_inference_collection.py#L295

Issue #288, opened 2022-07-13
Trying to Load Model Quantized for TensorRT Fails 


sam-h-bean:
### System Info

```shell
Building from source
Running inside the nvcr.io/nvidia/tensorrt:21.07-py3 docker container
```
(note there is a bug so you will have to build from #286)

### Who can help?

@philschmid @JingyaHuang 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

```python
from optimum.onnxruntime import ORTQuantizer
from optimum.onnxruntime.modeling_ort import ORTModelForCausalLM
from optimum.onnxruntime.configuration import AutoQuantizationConfig
from transformers import AutoTokenizer, AutoModelForCausalLM

model = AutoModelForCausalLM.from_pretrained('distilgpt2')
tokenizer = AutoTokenizer.from_pretrained('distilgpt2', padding=True, truncation=True)

qconfig = AutoQuantizationConfig.tensorrt(is_static=False, per_channel=False)
quantizer = ORTQuantizer(model=model, preprocessor=tokenizer, feature="causal-lm")
quantizer.export(
    onnx_model_path="model.onnx",
    onnx_quantized_model_output_path="model-quantized.onnx",
    quantization_config=qconfig,
)

tensorrt_model = ORTModelForCausalLM.load_model("./model-quantized.onnx", provider="TensorrtExecutionProvider")
```

Results in some error like
```bash
RuntimeError: /onnxruntime_src/onnxruntime/core/session/provider_bridge_ort.cc:1025 onnxruntime::Provider& onnxruntime::ProviderLibrary::Get() [ONNXRuntimeError] : 1 : FAIL : Failed to load library libonnxruntime_providers_tensorrt.so with error: /usr/local/lib/python3.8/dist-packages/onnxruntime/capi/libonnxruntime_providers_tensorrt.so: undefined symbol: getBuilderPluginRegistry
```

Using onnxruntime-gpu==1.11.1 I get this
```bash
[...]

PR #407, opened 2022-09-30
Add how-to guide to use Optimum-onnxruntime with GPU

fxmarty:
This adds an how-to guide on how to use Nvidia GPU with Optimum-onnxruntime.

To be done in next PR:
* Have a table with gains from `TensorrtExecutionProvider` vs `CUDAExecutionProvider`, both for quantized and non-quantized model (just to show why it can be worth using it) [probably best to wait for huggingface/optimum/pull/421 before]

Closes https://github.com/huggingface/optimum/issues/288 (dynamic quantization + tensorRT is a no-go)


fxmarty:
The added guide is https://moon-ci-docs.huggingface.co/docs/optimum/pr_407/en/onnxruntime/usage_guides/gpu

JingyaHuang:
FYI, models quantized dynamically with avx512_vnni config, have worse results compared with PyTorch both for w/ and w/o IO binding, whereas vanilla ONNX and ONNX with graph optimization perform better than PyTorch even for large sequence length when IO binding is applied.

fxmarty:
> FYI, models quantized dynamically with avx512_vnni config, have worse results compared with PyTorch both for w/ and w/o IO binding, whereas vanilla ONNX and ONNX with graph optimization perform better than PyTorch even for large sequence length when IO binding is applied.

In my experience, this is very dependent on:

* which, how many CPU cores are used
* batch size
* sequence length

On which CPU and setting did you try? Could be related https://github.com/microsoft/onnxruntime/issues/12925 

But yes, in my experience with large batch size and long sequence length ONNX Runtime with dynamic quantization on CPU is not super good.

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
+# Use Nvidia GPU for inference through Optimum-onnxruntime
+
+Although ONNX Runtime defaults to use CPU-only for inference, it is possible to place supported operations on an Nvidia GPU, while leaving unsupported ones (if remaining) on CPU. In most cases, all costly operations can be placed on GPU, allowing to leverage its compute power.
+
+ONNX Runtime provides an integration with two execution providers to make use of an Nvidia GPU, namely `CUDAExecutionProvider` and `TensorrtExecutionProvider`.
+
+In case you are using a vanilla model using floating-point arithmetic, both execution providers can be used. TensorRT generally has the best runtime performances, although it comes with its own caveats, notably limitations for inference with inputs having dynamic shape.
+
+When deploying a quantized model to a GPU, `CUDAExecutionProvider` can not be used, and only models using static quantization can be executed on `TensorrtExecutionProvider`.
```
regisss, line 9:
```suggestion
When deploying a quantized model to a GPU, `CUDAExecutionProvider` cannot be used, and only models using static quantization can be executed on `TensorrtExecutionProvider`.
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+_95), MatMul (MatMul_101), ...]
+```
+
+Here this result is expected: all the costly MatMul are placed on the CUDA execution provider.
+
+### Use CUDA execution provider with quantized models
+
+It is not possible to use quantized models with CUDAExecutionProvider.
+
+When using [🤗 Optimum dynamic quantization](quantization#dynamic-quantization-example), nodes as [`MatMulInteger`](https://github.com/onnx/onnx/blob/v1.12.0/docs/Operators.md#MatMulInteger), [`DynamicQuantizeLinear`](https://github.com/onnx/onnx/blob/v1.12.0/docs/Operators.md#DynamicQuantizeLinear) may be inserted in the computation graph, that can not be consumed by the CUDA execution provider. This is a limitation from ONNX Runtime.
```
regisss, line 92:
```suggestion
When using [🤗 Optimum dynamic quantization](quantization#dynamic-quantization-example), nodes as [`MatMulInteger`](https://github.com/onnx/onnx/blob/v1.12.0/docs/Operators.md#MatMulInteger), [`DynamicQuantizeLinear`](https://github.com/onnx/onnx/blob/v1.12.0/docs/Operators.md#DynamicQuantizeLinear) may be inserted in the computation graph, that cannot be consumed by the CUDA execution provider. This is a limitation from ONNX Runtime.
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+When using [static quantization](quantization#static-quantization-example), the onnx computation graph will contain matrix multiplications and convolutions in floating-point arithmetic, along with Quantize + Dequantize operations to simulate quantization. In this case, although the costly matrix multiplications and convolutions will be run on the GPU, they will use floating-point arithmetic as the `CUDAExecutionProvider` can not consume the Quantize + Dequantize nodes to replace them by the operations using integer arithmetic.
```
regisss, line 94:
```suggestion
When using [static quantization](quantization#static-quantization-example), the ONNX computation graph will contain matrix multiplications and convolutions in floating-point arithmetic, along with Quantize + Dequantize operations to simulate quantization. In this case, although the costly matrix multiplications and convolutions will be run on the GPU, they will use floating-point arithmetic as the `CUDAExecutionProvider` can not consume the Quantize + Dequantize nodes to replace them by the operations using integer arithmetic.
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+* CUDA toolkit: https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html
+* cuDNN: https://docs.nvidia.com/deeplearning/cudnn/install-guide/index.html
+* TensorRT: https://docs.nvidia.com/deeplearning/tensorrt/install-guide/index.html
+
+For TensorRT, we recommend the Tar File Installation method. Alternatively, TensorRT may be installable through the `nvidia-tensorrt` pip wheel available following [these instructions](https://github.com/microsoft/onnxruntime/issues/9986).
+
+The following environment variables need to be set for ONNX Runtime to detect TensorRT installation:
+```
+export CUDA_PATH=/usr/local/cuda
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda-11.7/lib64:/path/to/TensorRT-8.x.x/lib
```
regisss, line 116: Does it only work with CUDA 11.7?

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+from optimum.onnxruntime import ORTModelForSequenceClassification
+
+ort_model = ORTModelForSequenceClassification.from_pretrained(
+    "distilbert-base-uncased-finetuned-sst-2-english",
+    from_transformers=True,
+    provider="TensorrtExecutionProvider",
+)
+```
+
+[As previously for `CUDAExecutionProvider`](#use-cudaexecutionprovider-with-floatingpoint-model), by passing the session option `log_severity_level = 0` (verbose), we can check in the logs whether all nodes are indeed placed on the TensorRT execution provider or not:
```
regisss, line 159:
```suggestion
[As previously for `CUDAExecutionProvider`](#use-cuda-execution-provider-with-floatingpoint-models), by passing the session option `log_severity_level = 0` (verbose), we can check in the logs whether all nodes are indeed placed on the TensorRT execution provider or not:
```

regisss (commented):
Great addition @fxmarty :fire: 
One link should be fixed, otherwise just a few minor comments.

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
+# Use Nvidia GPU for inference through Optimum-onnxruntime
+
+Although ONNX Runtime defaults to use CPU-only for inference, it is possible to place supported operations on an Nvidia GPU, while leaving unsupported ones (if remaining) on CPU. In most cases, all costly operations can be placed on GPU, allowing to leverage its compute power.
+
+ONNX Runtime provides an integration with two execution providers to make use of an Nvidia GPU, namely `CUDAExecutionProvider` and `TensorrtExecutionProvider`.
```
lewtun, line 5:
```suggestion
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+In case you are using a vanilla model using floating-point arithmetic, both execution providers can be used. TensorRT generally has the best runtime performances, although it comes with its own caveats, notably limitations for inference with inputs having dynamic shape.
```
lewtun, line 7: Let's move the note about shapes to the TensorRT section
```suggestion
```

docs/source/onnxruntime/usage_guides/gpu.mdx
lewtun, line 9:
```suggestion
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+When deploying a quantized model to a GPU, `CUDAExecutionProvider` can not be used, and only models using static quantization can be executed on `TensorrtExecutionProvider`.
+
+## CUDAExecutionProvider
+
+### Installation
+
+Provided the CUDA and cuDNN [requirements](https://onnxruntime.ai/docs/execution-providers/CUDA-ExecutionProvider.html#requirements) are satisfied, the install is then straightforward with `pip install optimum[onnxruntime-gpu]`. Beforehand, to avoid conflicts between `onnxruntime` and `onnxruntime-gpu`, make sure the package `onnxruntime` is not installed by running `pip uninstall onnxruntime`.
```
lewtun, line 15:
```suggestion
Provided the CUDA and cuDNN [requirements](https://onnxruntime.ai/docs/execution-providers/CUDA-ExecutionProvider.html#requirements) are satisfied, install the additional dependencies by running:

 ```bash
 python -m pip install optimum[onnxruntime-gpu]
 ``` 

<Tip warning={true}>

To avoid conflicts between `onnxruntime` and `onnxruntime-gpu`, make sure the `onnxruntime` package is not installed by running `pip uninstall onnxruntime`.

</Tip>
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+ort_model = ORTModelForSequenceClassification.from_pretrained(
+    "philschmid/tiny-bert-sst2-distilled",
+    from_transformers=True,
+    provider="CUDAExecutionProvider",
+)
+print("providers:", ort_model.providers)
+
+tokenizer = AutoTokenizer.from_pretrained("philschmid/tiny-bert-sst2-distilled")
+inp = tokenizer("expectations were low, actual enjoyment was high", return_tensors="pt", padding=True)
```
lewtun, line 33:
```suggestion
inputs = tokenizer("expectations were low, actual enjoyment was high", return_tensors="pt", padding=True)
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+result = ort_model(**inp)
```
lewtun, line 35:
```suggestion
outputs = ort_model(**inputs)
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+```python
+from optimum.onnxruntime import ORTModelForSequenceClassification
+from transformers import AutoTokenizer
+
+ort_model = ORTModelForSequenceClassification.from_pretrained(
+    "philschmid/tiny-bert-sst2-distilled",
+    from_transformers=True,
+    provider="CUDAExecutionProvider",
+)
+print("providers:", ort_model.providers)
```
lewtun, line 30: I would move this to the end of the code block and consider using `assert ort_model.providers == ['CUDAExecutionProvider', 'CPUExecutionProvider']` to make it clear what the expected output is.

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+tokenizer = AutoTokenizer.from_pretrained("philschmid/tiny-bert-sst2-distilled")
+inp = tokenizer("expectations were low, actual enjoyment was high", return_tensors="pt", padding=True)
+
+result = ort_model(**inp)
+print(result)
```
lewtun, line 36:
```suggestion
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+```
+
+In case the printed result is `providers: ['CUDAExecutionProvider', 'CPUExecutionProvider']`, congratulations, the installation is successful!
```
lewtun, line 39: This could be deleted if you agree with my above suggestion to the code block

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+In case the error `ValueError: Asked to use CUDAExecutionProvider as an ONNX Runtime execution provider, but the available execution providers are ['CPUExecutionProvider'].` is raised, something is off in your installation.
```
lewtun, line 41:
```suggestion
If you encounter the following error:

```python
ValueError: Asked to use CUDAExecutionProvider as an ONNX Runtime execution provider, but the available execution providers are ['CPUExecutionProvider'].
```

then something is wrong with the ONNX Runtime installation.
```

File: docs/source/onnxruntime/usage_guides/gpu.mdx [hunk 15]
```diff
@@ -0,0 +1,225 @@
...
+
+### Use CUDA execution provider with floating-point models
+
+For non-quantized models, the use is straightforward, by simply using the `provider` argument in `ORTModel.from_pretrained()`. For example:
```
lewtun, line 45:
```suggestion
For non-quantized models, the use is straightforward. Simply specify the `provider` argument in the `ORTModel.from_pretrained()` method. Here's an example:
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+from optimum.onnxruntime import ORTModelForSequenceClassification
+
+ort_model = ORTModelForSequenceClassification.from_pretrained(
+    "distilbert-base-uncased-finetuned-sst-2-english",
+    from_transformers=True,
+    provider="CUDAExecutionProvider",
+)
+```
+
+Additionally, By passing the session option `log_severity_level = 0` (verbose), we can check whether all nodes are indeed placed on the CUDA execution provider or not:
```
lewtun, line 57:
```suggestion
Additionally, you can pass the session option `log_severity_level = 0` (verbose), to check whether all nodes are indeed placed on the CUDA execution provider or not:
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+```
```
lewtun, line 59:
```suggestion
```python
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+ort_model = ORTModelForSequenceClassification.from_pretrained(
+    "distilbert-base-uncased-finetuned-sst-2-english",
+    from_transformers=True,
+    provider="CUDAExecutionProvider",
+    session_options=session_options
+)
+```
+
+printing:
```
lewtun, line 73:
```suggestion
You should see the following logs:
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+```
```
lewtun, line 75:
```suggestion
```bash
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+    activations_symmetric=True,
+    weights_symmetric=True,
+    qdq_dedicated_pair=True,
+    qdq_add_pair_to_weight=True
+)
+```
+
+The parameter `qdq_dedicated_pair=True` is required by TensorRT, that expects a single node after each `QuantizeLinear` + `DequantizeLinear` (QDQ) pair.
+
+The parameter `qdq_add_pair_to_weight=True` is also required by TensorRT, that consumes a graph where the weights are stored in float32 with a QDQ pair. Normally, weights would be stored in fixed point 8-bits format and only a `DequantizeLinear` would be applied on the weights. As such, the storage savings from quantization can not be leveraged when we expect to later use the quantized `.onnx` model with TensorRT.
```
lewtun, line 182:
```suggestion
The parameter `qdq_add_pair_to_weight=True` is also required by TensorRT, that since it consumes a graph where the weights are stored in float32 with a QDQ pair. Normally, weights would be stored in fixed point 8-bits format and only a `DequantizeLinear` would be applied on the weights. As such, the storage savings from quantization can not be leveraged when we expect to later use the quantized ONNX model with TensorRT.
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+In the code sample below, after performing static quantization, the resulting model is loaded into the `ORTModel` class using TensorRT as the execution provider. ONNX Runtime graph optimization need to be disabled for the model to be consumed and optimized by TensorRT, and the fact that int8 operations are used needs to be specified to TensorRT.
```
lewtun, line 184:
```suggestion
In the code sample below, after performing static quantization, the resulting model is loaded into the `ORTModel` class using TensorRT as the execution provider. ONNX Runtime graph optimization need to be disabled for the model to be consumed and optimized by TensorRT, and the fact that INT8 operations are used needs to be specified to TensorRT.
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+The model can then be used with the common 🤗 Transformers API for inference and evaluation, such as pipelines:
+```python
+from optimum.pipelines import pipeline
+from transformers import AutoTokenizer
+
+tokenizer = AutoTokenizer.from_pretrained(model_name)
+
+pipe = pipeline(task="text-classification", model=ort_model, tokenizer=tokenizer)
+result = pipe("Both the music and visual were astounding, not to mention the actors performance.")
+print(result)
```
lewtun, line 209: Similar comment as above that you can include the actual outputs in the code block

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+```
+
+### TensorRT limitations for quantized models
+
+As highlighted in the previous section, TensorRT supports only a limited range of quantized model:
```
lewtun, line 214:
```suggestion
As highlighted in the previous section, TensorRT supports only a limited range of quantized models:
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -0,0 +1,225 @@
...
+* Static quantization only
+* Weights and activations quantization ranges are symmetric
+* Weights need to be stored in float32 in the `.onnx` model, thus there is no storage space saving from quantization.
```
lewtun, line 217:
```suggestion
* Weights need to be stored in float32 in the ONNX model, thus there is no storage space saving from quantization.
```

lewtun (commented):
Awesome guide @fxmarty - this is really neat 🔥 !

I've left a few nits and suggestions to tighten the text, but otherwise this looks great

docs/source/onnxruntime/usage_guides/gpu.mdx
fxmarty: Indeed no, good catch!

fxmarty:
Thanks a lot @regisss @lewtun for your thorough review! 🙏

I think it can be merged in its current state.

Improvements to this tutorial can be:
* Detail ONNX Runtime limitations on multi-GPU machines, that we can pin ORTModel to specific a GPU following https://github.com/huggingface/optimum/pull/427
* Precisely detail the limitations of TensorrtExecutionProvider vs native Tensor RT, see https://github.com/microsoft/onnxruntime/issues/12083#issuecomment-1285425725 . IMO there is no huge limitation, except that in order to do QAT you currently need to use an other framework than ONNX Runtime. Also, having dynamic input shape can be an issue as warmup seems to be required with fixed input shape (don't know if anybody familiar, maybe @mht-sharma knows more)

regisss (approved):
Thanks @fxmarty!!

fxmarty:
Note that there are as well other approaches to use TensorRT, notably fx2trt https://github.com/pytorch/TensorRT/blob/master/docsrc/tutorials/getting_started_with_fx_path.rst @michaelbenayoun . Not sure how it would compare to fx --> onnx --> ONNX Runtime with TensorrtExecutionProvider

Looks like torch.ao / fx and ONNX Runtime / onnx are a bit in competition, is it?

transformer-deploy uses the vanilla TensorRT Python bindings.

Issue #325, opened 2022-07-29
Add possibility to output attentions from the ONNX models


jegork:
### Feature request

Hello!

I am interested in the possibility of retrieving not only the logits from the ONNX exported model, but also the attentions (which can be set via output_attentions=True in transformers). 
I have tried setting output_attentions=True in the config.json file of the model repository, but an exception is raised when the onnxruntime InferenceSession is initialized.

```python
Fail: [ONNXRuntimeError] : 1 : FAIL : Load model from ort_model_optimized.onnx failed:/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-req-build-u762x_rp/onnxruntime/core/graph/graph.cc:1236 void onnxruntime::Graph::InitializeStateFromModelFileGraphProto() This is an invalid model. Graph output (attentions) does not exist in the graph
```

### Motivation

I am using attention scores for calculating the importance, but it is currently impossible to extract them in our production environment as we are using ONNX-exported transformers models.

### Your contribution

If you could guide me a bit on what to do, I could submit a PR.

JingyaHuang:
Hi @jegork, 

As the default __OnnxConfig__ takes [common outputs](https://github.com/huggingface/transformers/blob/main/src/transformers/onnx/config.py#L76) which do not include `output_attentions`, the output of exported ONNX model will not contain `output_attentions`.

We are working on making the onnx configs more flexible, for the moment a workaround would be creating a wrapper of the onnx config to include `output_attentions` as a part of the output. Here is an example of a custom sbert onnx config:
```python
class CustomSTOnnxConfig(OnnxConfig, ABC):
    def __init__(self, config: OnnxConfig):
        self.__dict__ = copy.deepcopy(config.__dict__)
        self._onnx_config = config

    @property
    def inputs(self):
        return self._onnx_config.inputs

    @property
    def outputs(self):
        return OrderedDict({"last_hidden_state": {0: "batch", 1: "sequence"}, "pooler_output": {0: "batch"}})
```
Then you can wrap your onnx config with the wrapper
```python
custom_onnx_config = CustomSTOnnxConfig(onnx_config)
```

And after the export, if you want to leverage the ORTModels in optimum for inference, here is a [PR in progress](https://github.com/huggingface/optimum/pull/303). With the class `ORTModelForCustomTasks`, you will be able to put any custom onnx model for inference.

jegork:
Thank you for the fast response @JingyaHuang

[...]

PR #412, opened 2022-10-04
Change field names in optimization configuration and expand the doc

regisss:
# What does this PR do?

This PR changes the field names for operator fusion in `OptimizationConfig`. The same name as ONNX Runtime used to be used but they were misleading: for example, `disable_gelu` means that Gelu fusion should is disabled, so it is now renamed as `disable_gelu_fusion`.

The documentation about optimization configuration is also expanded.

## Before submitting


optimum/onnxruntime/configuration.py
```diff
@@ -620,19 +620,19 @@ class OptimizationConfig:
...
-            Whether to only use ONNX Runtime to optimize the model and no graph fusion in Python.
-        disable_gelu (`bool`, defaults to `False`):
+        fuse_operators (`bool`, defaults to `True`):
+            Whether to fuse operators in addition to ONNX Runtime graph optimizations.
+        disable_gelu_fusion (`bool`, defaults to `False`):
             Whether to disable the Gelu fusion.
-        disable_layer_norm (`bool`, defaults to `False`):
+        disable_layer_norm_fusion (`bool`, defaults to `False`):
             Whether to disable Layer Normalization fusion.
-        disable_attention (`bool`, defaults to `False`):
```
JingyaHuang, line 629: Just kindly mention that changing the args name might be a problem for users who are already using these flags, like this https://github.com/huggingface/optimum/issues/325#issuecomment-1202467773

regisss: Yes I agree with you @JingyaHuang. Maybe I can add a warning in the `__post_init__` of the configuration? Or should we just keep the documentation update? What do you think? @echarlaix @michaelbenayoun

echarlaix: I think adding a warning in the `__post_init__` will not change much as an error will be thrown when users will try to instantiate an `OptimizationConfig` with the old argument names (before the warning). But maybe you meant allowing users to use the old argument names by modifying the `__init__` and add a warning in the `__post_init__` in order to smoothen this transition ?

regisss: Yes, you're right. I was thinking about something like that even though I don't like it that much because we would also need to make sure that the value of an argument does not contradict the value of its former version. But I'm not sure if there is a better solution. Or we just keep the current names and make sure it's clear for users that it only fuses operators.

michaelbenayoun: I think what can be done is this:

1. We allow both
2. If `difsable_attention` is set, it overwrites the other with a warning saying it overwrites the values, and that it will be deprecated in 3-4 releases
3. Remove the old one in 3-4 releases

What do you think?

regisss: Sounds good to me :+1:

regisss:
Closing this PR since it was integrated to https://github.com/huggingface/optimum/pull/457.

PR #419, opened 2022-10-09
Fix ORT training args log level bug

### LINKED ISSUE #418: "self.log_level" required to be string instead of int, while optimum convert it into int in function __init__
### System Info

```shell
optimum and transformers are all installed from source code.
```

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

the shell command to reproduce:
`python -m torch.distributed.launch --nproc_per_node=1 /optimum/examples/onnxruntime/training/text-classification/run_glue.py --model_name_or_path distilbert-base-uncased --task_name mnli --max_seq_length 128 --learning_rate 3e-6 --do_train --output_dir /dev/shm --overwrite_output_dir --max_steps 200 --logging_steps 20 --per_device_train_batch_size 1300 --fp16 --optim adamw_ort_fused`

the error log is 
image

the root cause is the transformer PR https://github.com/huggingface/transformers/pull/19239

### Expected behavior

the command should be runnable with default args.

JingyaHuang:
# What does this PR do?

Bugged after the recent change of log level in `transformers.TrainingArgs`: https://github.com/huggingface/transformers/pull/19239

* Fix by updating `ORTTrainingArguments`.
* CI will install transformers from source to track the possible breaking changes led by transformers' updates. 
* Add test for PyTorch DDP.
* Update training example minimal transformers version.

Fixes #418

### DIFF
--- a/tests/onnxruntime/nightly_test_trainer.py
+++ b/tests/onnxruntime/nightly_test_trainer.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 import gc
+import subprocess
+import sys
 import tempfile
 import unittest
 from typing import List, Optional
@@ -672,6 +674,37 @@ class ORTTrainerIntegrationDeepSpeedTest(unittest.TestCase):
             prediction = trainer.predict(test_dataset)
             # self.assertGreaterEqual(eval_metrics["eval_accuracy"], 0.75)
             gc.collect()
+
+
+# @slow
+# @require_torc
+class ORTTrainerIntegrationDDPTest(unittest.TestCase):
+    def test_trainer_ddp_glue(self):
+
+        subprocess.run(
+            f"cp examples/onnxruntime/training/text-classification/run_glue.py ./",
+            shell=True,
+        )
+
+        subprocess.run(
+            f"{sys.executable} -m torch.distributed.launch"
+            " --nproc_per_node=1"
+            " run_glue.py"
+            " --model_name_or_path distilbert-base-uncased"
+            " --task_name mnli"
+            " --max_seq_length 128"
+            " --learning_rate 3e-6"
+            " --do_train"
+            " --output_dir /tmp/distilbert"
+            " --overwrite_output_dir"
+            " --max_steps 200"
+            " --logging_steps 20"
+            " --per_device_train_batch_size 32"
+            " --fp16 --optim adamw_ort_fused"
+            " --max_train_samples 3000",
+            shell=True,
+            check=True,
+        )
 
 
 class ORTTrainerHyperParameterIntegrationTest(unittest.TestCase):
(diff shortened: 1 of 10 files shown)


tests/onnxruntime/nightly_test_trainer.py
```diff
@@ -674,6 +676,37 @@ def test_trainer_fp16_ds_stage2(self, test_name, model_name, feature, data_metri
             gc.collect()
 
 
+# @slow
+# @require_torc
```
regisss, line 680: Should we keep these two commented lines?

tests/onnxruntime/nightly_test_trainer.py
```diff
@@ -674,6 +676,37 @@ def test_trainer_fp16_ds_stage2(self, test_name, model_name, feature, data_metri
...
+    def test_trainer_ddp_glue(self):
+
+        subprocess.run(
+            f"cp examples/onnxruntime/training/text-classification/run_glue.py ./",
+            shell=True,
+        )
+
+        subprocess.run(
+            f"{sys.executable} -m torch.distributed.launch"
+            " --nproc_per_node=1"
```
regisss, line 691: Could we test it with more than one device?

regisss (commented):
Thanks for this fix @JingyaHuang!! I just have a couple of comments.

tests/onnxruntime/nightly_test_trainer.py
JingyaHuang: yes, good catch. I forgot to uncomment them, and an "h" disappears...

tests/onnxruntime/nightly_test_trainer.py
JingyaHuang: If not mistaken, the current ec2 instance in CI only has one GPU. Do you test with multiple for Habana?

echarlaix (approved):
LGTM @JingyaHuang !

tests/onnxruntime/nightly_test_trainer.py
JingyaHuang: @regisss, let's leave it to one process per node and increase the GPU number for CI in another PR if necessary.

tests/onnxruntime/nightly_test_trainer.py
JingyaHuang: Fixed in 2cc6b7dc5b78a8e2040291019f154656aacafdd5

tests/onnxruntime/nightly_test_trainer.py
regisss: Ah okay I thought the type of instance you use had more than one device! No worries it looks good to me :+1:

regisss (approved):

Issue #414, opened 2022-10-05
onnx speed is even slower


CatchDr:

Narsil:
Tagging @mfuntowicz @michaelbenayoun for visibility. Sorry if the transfer wasn't correct.

CatchDr:
from transformers import AutoTokenizer, pipeline
from optimum.onnxruntime import ORTModelForSeq2SeqLM
import warnings
text="Vehicle detection technology is of great significance for realizing automatic monitoring and AI-assisted driving systems. The state-of-the-art object detection method, namely, a class of YOLOv5, has often been used to detect vehicles."
warnings.filterwarnings("ignore")
import time
textlists=[text,text,text,text,text]
model_checkpoint = "Helsinki-NLP/opus-mt-en-zh"
model = ORTModelForSeq2SeqLM.from_pretrained(model_checkpoint, from_transformers=True)
tokenizer = AutoTokenizer.from_pretrained(model_checkpoint)
onnx_translation = pipeline("translation_en_to_zh", model=model, tokenizer=tokenizer)
t1=time.time()
result = onnx_translation(textlists)
print(result ,time.time()-t1)

from transformers import (
MarianTokenizer,
MarianMTModel,
)
modchoice = "Helsinki-NLP/opus-mt-en-zh"
tokenizer = MarianTokenizer.from_pretrained(modchoice)

model = MarianMTModel.from_pretrained(modchoice)

t1 = time.time()
encoded=tokenizer.prepare_seq2seq_batch(
textlists,
truncation=True,
padding="longest",
return_tensors="pt"
)
translated = model.generate(
**encoded
)

tgt_text = [tokenizer.decode(t, skip_special_tokens=True) for t in translated]
print(tgt_text,time.time() - t1)

Narsil:
@CatchDr this code is not useable, you NEED to use ``` to make it block code and show the proper indents.

CatchDr:
'''
from transformers import AutoTokenizer, pipeline
from optimum.onnxruntime import ORTModelForSeq2SeqLM
import warnings
text="Vehicle detection technology is of great significance for realizing automatic monitoring and AI-assisted driving systems. The state-of-the-art object detection method, namely, a class of YOLOv5, has often been used to detect vehicles."
warnings.filterwarnings("ignore")
import time
textlists=[text,text,text,text,text]
model_checkpoint = "Helsinki-NLP/opus-mt-en-zh"
[...]

Issue #404, opened 2022-09-28
Inference worse with onnxruntime-gpu than native pytorch for seq2seq model


Matthieu-Tinycoaching:
### System Info

```shell
Optimum: 1.4.1.dev0
torch: 1.12.1+cu116
onnx: 1.12.0
onnxruntime-gpu: 1.12.1
python: 3.8.13
CUDA: 11.6
cudnn: 8.4.1
RTX 3090
```

### Who can help?

@JingyaHuang @echarlaix 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

I compared inference on GPU of a native torch `Helsinki-NLP/opus-mt-fr-en` model with respect to the optimized onnx model thanks to Optimum library. So, I have defined a fastAPI microservice based on two classes below for GPU both torch and optimized ONNX, repsectively:

```
class Seq2SeqModel:
    tokenizer: Optional[MarianTokenizer]
    model: Optional[MarianMTModel]

    def load_model(self):
        """Loads the model"""
        # model_id="Helsinki-NLP/opus-mt-fr-en"
        model_path = Path("./app/artifacts/HF")
        tokenizer = AutoTokenizer.from_pretrained(model_path)
        model = AutoModelForSeq2SeqLM.from_pretrained(model_path).to("cuda")
        self.tokenizer = tokenizer
        self.model = model

    async def predict(self, input: PredictionInput) -> PredictionOutput:
        """Runs a prediction"""
        if not self.tokenizer or not self.model:
            raise RuntimeError("Model is not loaded")
        tokens = self.tokenizer(input.text, return_tensors="pt").to("cuda")
        translated = self.model.generate(**tokens, num_beams=beam_size)
        return PredictionOutput(translated_text=self.tokenizer.decode(translated[0], skip_special_tokens=True))

class OnnxOptimizedSeq2SeqModel:
    tokenizer: Optional[MarianTokenizer]
    model: Optional[ORTModelForSeq2SeqLM]

    def load_model(self):
        """Loads the model"""
        # model_id="Helsinki-NLP/opus-mt-fr-en"
        onnx_path = Path("./app/artifacts/OL_1")
        tokenizer = AutoTokenizer.from_pretrained(onnx_path)
        optimized_model = ORTModelForSeq2SeqLM.from_pretrained(
            onnx_path,
[...]

Issue #365, opened 2022-08-31
Inference performance drop 22X on GPU hardware with optimum[onnxruntime-gpu] (compared with transformer)


bugzyz:
### System Info

```shell
pip list:
onnx               1.12.0
onnxruntime-gpu    1.12.1
optimum            1.3.0
...
transformers       4.21.2

docker image: 
nvcr.io/nvidia/tensorrt:22.07-py3

gpu type:
Tesla V100-SXM2-16GB
```

### Who can help?

@JingyaHuang @echarlaix 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction
0. create docker container
```docker run --gpus all -it --name=tensorrtx -d --rm -v /home/ytseng/models:/models nvcr.io/nvidia/tensorrt:22.07-py3```
1. install required python package
```ln -s /usr/local/bin/pip /usr/bin/pip && pip install optimum[onnxruntime-gpu] transformers```
2. save huggingface model and export to onnx
cmd: ```python save_and_export_model.py```
```model_checkpoint = "facebook/m2m100_418M"
transformer_model = "/models/optimum_m2m100/m2m100_transformer"
onnx_model = "/models/optimum_m2m100/m2m100_onnx"


def download_transformer_model():
    from transformers import AutoTokenizer, AutoModelForSeq2SeqLM

    # Load tokenizer and PyTorch weights form the Hub
    tokenizer = AutoTokenizer.from_pretrained(model_checkpoint)
    pt_model = AutoModelForSeq2SeqLM.from_pretrained(model_checkpoint)

    # Save to disk
    tokenizer.save_pretrained(transformer_model)
    pt_model.save_pretrained(transformer_model)


def export_to_onnx_model():
    """
    -------- export to onnx
    """
    from optimum.onnxruntime import ORTModelForSeq2SeqLM

    # Load a model from transformers and export it through the ONNX format
    model = ORTModelForSeq2SeqLM.from_pretrained(model_checkpoint, from_transformers=True)

    # Save the onnx model
    model.save_pretrained(onnx_model)


if __name__ == '__main__':
    download_transformer_model()
    export_to_onnx_model()
```
3. run translation on transformer api and optimum api
testing cmd: ```python run_translation.py trans ; echo '---------------' ; python run_translation.py opt```
```
[...]

Issue #362, opened 2022-08-26
unexpect behavior GPU runtime with ORTModelForSeq2SeqLM 


tranmanhdat:
### System Info

```shell
OS: Ubuntu 20.04.4 LTS
CARD: RTX 3080

Libs:
python 3.10.4
onnx==1.12.0
onnxruntime-gpu==1.12.1
torch==1.12.1
transformers==4.21.2
```

### Who can help?

@lewtun @michaelbenayoun @JingyaHuang  @echarlaix 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction

Steps to reproceduce the behavior:
1. Convert a public translation from here: [vinai-translate-en2vi](https://huggingface.co/vinai/vinai-translate-en2vi)
```
from optimum.onnxruntime import ORTModelForSeq2SeqLM
save_directory = "models/en2vi_onnx"
# Load a model from transformers and export it through the ONNX format
model = ORTModelForSeq2SeqLM.from_pretrained('vinai/vinai-translate-en2vi', from_transformers=True)
# Save the onnx model and tokenizer
model.save_pretrained(save_directory)
```
2. Load model with modified from [example of origin creater model](https://github.com/VinAIResearch/VinAI_Translate#english-to-vietnamese-translation)
```
from transformers import AutoTokenizer, pipeline
from optimum.onnxruntime import ORTModelForSeq2SeqLM
import torch
import time
device = "cuda:0" if torch.cuda.is_available() else "cpu"
tokenizer_en2vi = AutoTokenizer.from_pretrained("vinai/vinai-translate-en2vi", src_lang="en_XX")
model_en2vi = ORTModelForSeq2SeqLM.from_pretrained("models/en2vi_onnx")
model_en2vi.to(device)

# onnx_en2vi = pipeline("translation_en_to_vi", model=model_en2vi, tokenizer=tokenizer_en2vi, device=0)
# en_text = '''It's very cold to go out.'''
# start = time.time()
# outpt = onnx_en2vi(en_text)
# end = time.time()
# print(outpt)
# print("time: ", end - start)

def translate_en2vi(en_text: str) -> str:
    start = time.time()
    input_ids = tokenizer_en2vi(en_text, return_tensors="pt").input_ids.to(device)
    end = time.time()
[...]

PR #421, opened 2022-10-13
Add IOBinding support to ONNX Runtime module 

JingyaHuang:
# Context

As reported by users, when using devices for acceleration, there is sometimes a significant performance drop. And the slow-down is especially significant for decoders(so also for seq2seq models). This is due to a large overhead while copying data across the host and device.

This PR will introduce IOBinding of ONNX Runtime to arrange inputs and pre-allocate outputs on device. 

# What does this PR do?

- [x] Create TypeHelper to prepare IO binding
- [x] Integrate IO binding into ORTModels
 

## Before submitting

__Associated issues__:
#362 #365 #404 #414


optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,159 @@
...
+            #     return IOBindingHelper.to_pytorch_via_cupy(ort_value)
+            # except Exception as e:
+            #     logging.error(traceback.format_exc())
+            #     logging.info("Unable to access output memory in CUDA, will offload to CPU")
+            #     return IOBindingHelper.to_pytorch_via_np(ort_value)
+
+    @staticmethod
+    def to_pytorch_via_np(ort_value: OrtValue) -> torch.Tensor:
+        ort_device = ort_value.device_name().lower()
+        return torch.tensor(ort_value.numpy()).to(ort_device)
```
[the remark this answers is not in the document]
JingyaHuang: When ort offloads the data to CPU with numpy, I need to turn numpy to torch tensor and send it back to GPU. Otherwise, during generation, the model is still on GPU while logit is in CPU, and it will fail.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -888,16 +889,23 @@ def forward(
...
-        }
-        # run inference
-        outputs = self.model.run(None, onnx_inputs)
-        logits = torch.from_numpy(outputs[self.model_outputs["logits"]]).to(self.device)
+
+        onnx_inputs = {"input_ids": input_ids, "attention_mask": attention_mask}
+        io_helper = IOBindingHelper(self.model, self.config, self.device)
+        io_binding = io_helper.prepare_io_binding(**onnx_inputs)
+
+        # run inference with binding
```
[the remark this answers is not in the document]
JingyaHuang: I am going to add an argument to enable users to turn it off, and when `onnxruntime-gpu` is not installed, the IO Binding shall be turned off automatically.

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,159 @@
...
+        return io_binding
+
+    @staticmethod
+    def to_pytorch(ort_value: OrtValue) -> torch.Tensor:
+        env = {pkg.key for pkg in pkg_resources.working_set}
+        if "onnxruntime-training" in env:
+            return IOBindingHelper.to_pytorch_via_dlpack(ort_value)
+        else:
+            return IOBindingHelper.to_pytorch_via_cupy(ort_value)
+            # try:
```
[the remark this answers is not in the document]
JingyaHuang: The design of `to_pytorch` is uncertain to me for the moment, as we have three approaches depending on different dependencies. I need the input of the team to see whether we leave some fallbacks, and how shall we decide which approach to use.

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+        name_to_np_type = TypeHelper.get_io_numpy_type_map(self.model)
+
+        # Bind inputs and outputs to onnxruntime session
+        io_binding = self.model.io_binding()
+
+        # Bind inputs
+        for input_name in self.model_input_names:
+            onnx_input = kwargs.pop(input_name)
+
+            assert onnx_input.is_contiguous()
```
lewtun, line 96: I suggest avoiding `assert` statements and raising a proper error with a helpful error message

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+        else:
+            try:
+                return IOBindingHelper.to_pytorch_via_cupy(ort_value)
+            except Exception as e:
+                logging.error(traceback.format_exc())
+                logging.info("Unable to access output memory in CUDA, will offload to CPU")
+                return IOBindingHelper.to_pytorch_via_np(ort_value)
+
+    @staticmethod
+    def to_pytorch_via_np(ort_value: OrtValue) -> torch.Tensor:
```
lewtun, line 127: I suggest being explicit where possible:
```suggestion
    def to_pytorch_via_numpy(ort_value: OrtValue) -> torch.Tensor:
```

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+        ort_device = ort_value.device_name().lower()
+        return torch.tensor(ort_value.numpy()).to(ort_device)
+
+    @staticmethod
+    def to_pytorch_via_cupy(ort_value: OrtValue) -> torch.Tensor:
+        ort_device = ort_value.device_name().lower()
+        assert ort_device == "cuda", f"Convert via CuPy only when device is CUDA, got: {ort_device}"
```
lewtun, line 134: Similar as above, I suggest raising an error here without using `assert`

optimum/onnxruntime/modeling_ort.py
```diff
@@ -66,6 +67,7 @@
             Initializing with a config file does not load the weights associated with the model, only the
             configuration. Check out the [`~onnxruntime.modeling_ort.ORTModel.from_pretrained`] method to load the model weights.
         model (`onnxruntime.InferenceSession`): [onnxruntime.InferenceSession](https://onnxruntime.ai/docs/api/python/api_summary.html#inferencesession) is the main class used to run a model. Check out the [`~onnxruntime.modeling_ort.ORTModel.load_model`] method for more information.
+        use_io_binding (`bool`, *optional*): Whether use IO Binding during inference to avoid memory copy between the host and devices. Will default to True if the device is CUDA, otherwise default to False.
```
lewtun, line 70: nit:
```suggestion
        use_io_binding (`bool`, *optional*): Whether use IOBinding during inference to avoid memory copy between the host and devices. Defaults to `True` if the device is CUDA, otherwise defaults to `False`.
```

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+import torch
+
+import onnxruntime as ort
+import pkg_resources
+from onnxruntime import InferenceSession
+from onnxruntime.capi.onnxruntime_inference_collection import OrtValue
+from onnxruntime.transformers.io_binding_helper import TypeHelper as ORTTypeHelper
+
+
+# adapted from https://github.com/microsoft/onnxruntime/blob/93e0a151177ad8222c2c95f814342bfa27f0a64d/onnxruntime/python/tools/transformers/io_binding_helper.py#L12
```
lewtun, line 29: turbo nit:

```suggestion
# Adapted from https://github.com/microsoft/onnxruntime/blob/93e0a151177ad8222c2c95f814342bfa27f0a64d/onnxruntime/python/tools/transformers/io_binding_helper.py#L12
```

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+class TypeHelper(ORTTypeHelper):
+    """
+    TypeHelper helps get data type information of the ONNX runtime inference session and offers the mapping from
+    OrtValues data type to the data type of other data structures(numpy, torch tensor...).
```
lewtun, line 33: nit:
```suggestion
    `OrtValue` data types to the data types of other frameworks (NumPy, PyTorch, etc).
```

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+import pkg_resources
+from onnxruntime import InferenceSession
+from onnxruntime.capi.onnxruntime_inference_collection import OrtValue
+from onnxruntime.transformers.io_binding_helper import TypeHelper as ORTTypeHelper
+
+
+# adapted from https://github.com/microsoft/onnxruntime/blob/93e0a151177ad8222c2c95f814342bfa27f0a64d/onnxruntime/python/tools/transformers/io_binding_helper.py#L12
+class TypeHelper(ORTTypeHelper):
+    """
+    TypeHelper helps get data type information of the ONNX runtime inference session and offers the mapping from
```
lewtun, line 32:
```suggestion
    Gets data type information of the ONNX Runtime inference session and provides the mapping from
```

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+            "tensor(float)": torch.float32,
+            "tensor(float16)": torch.float16,
+            "tensor(bool)": torch.bool,
+        }
+        if ort_type not in ort_type_to_torch_type_map:
+            raise ValueError(f"{ort_type} not found in map")
+        return ort_type_to_torch_type_map[ort_type]
+
+
+# adapted from https://github.com/microsoft/onnxruntime/blob/1ab11a111ce0717bfbfaca964d04a017cb9b1752/onnxruntime/python/tools/transformers/io_binding_helper.py#L97
```
lewtun, line 66:
```suggestion
# Adapted from https://github.com/microsoft/onnxruntime/blob/1ab11a111ce0717bfbfaca964d04a017cb9b1752/onnxruntime/python/tools/transformers/io_binding_helper.py#L97
```

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+class IOBindingHelper:
+    """
+    IOBindingHelper is a class that helps ORTModels to create buffers for inputs and outputs of an ONNX runtime
```
lewtun, line 69:
```suggestion
    A helper class to enable `ORTModel` instances create buffers for inputs and outputs of an ONNX Runtime
```

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+    inference session when using devices like GPU for acceleration. It helps reduce memory copy between the host
+    and device.
+    """
+
+    def __init__(self, model: ort.InferenceSession, config, device, **kwargs):
+        self.model = model
+        self.config = config
+        self.device = device
+        # create {name:idx} dict for model outputs
```
lewtun, line 78: dead code, so remove?

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+        self.model_inputs = {output_key.name: idx for idx, output_key in enumerate(model.get_inputs())}
+        self.model_outputs = {output_key.name: idx for idx, output_key in enumerate(model.get_outputs())}
+        self.model_input_names = list(self.model_inputs.keys())
+        self.model_output_names = list(self.model_outputs.keys())
+
+    def prepare_io_binding(self, **kwargs):
+        """Returnas IO binding object for a session."""
```
lewtun, line 85:
```suggestion
        """Returns IOBinding object for an inference session."""
```

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+        self.model = model
+        self.config = config
+        self.device = device
+        # create {name:idx} dict for model outputs
+        self.model_inputs = {output_key.name: idx for idx, output_key in enumerate(model.get_inputs())}
+        self.model_outputs = {output_key.name: idx for idx, output_key in enumerate(model.get_outputs())}
+        self.model_input_names = list(self.model_inputs.keys())
+        self.model_output_names = list(self.model_outputs.keys())
+
+    def prepare_io_binding(self, **kwargs):
```
lewtun, line 84: shoudl we add the return type?

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+        for name in self.model_output_names:
+            io_binding.bind_output(name, self.device.type, device_id=self.device.index)
+
+        return io_binding
+
+    @staticmethod
+    def to_pytorch(ort_value: OrtValue) -> torch.Tensor:
+        """Converts tensors held by OrtValues to torch tensor."""
+        env = {pkg.key for pkg in pkg_resources.working_set}
+        if "onnxruntime-training" in env:
```
lewtun, line 116: Would it be better to have a `bool` like `is_onnxruntime_training_available()` here?

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+        ort_device = ort_value.device_name().lower()
+        return torch.tensor(ort_value.numpy()).to(ort_device)
+
+    @staticmethod
+    def to_pytorch_via_cupy(ort_value: OrtValue) -> torch.Tensor:
+        ort_device = ort_value.device_name().lower()
+        assert ort_device == "cuda", f"Convert via CuPy only when device is CUDA, got: {ort_device}"
+
+        ort_type = ort_value.data_type()
+        np_type = TypeHelper.ort_type_to_numpy_type(ort_type)
```
lewtun, line 137: nit:
```suggestion
        numpy_type = TypeHelper.ort_type_to_numpy_type(ort_type)
```

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,170 @@
...
+        # If is boolean, the dtype will be uint8 and need to be convert back to bool.
+        if "bool" in ort_type:
+            torch_tensor = torch_tensor.to(torch.bool)
+
+        torch_tensor = torch_tensor.clone()
+
+        return torch_tensor
+
+    @staticmethod
+    # only `onnxruntime-training` supports dlpack for OrtValue
```
lewtun, line 156: Not sure I understand the implication of this comment. Could you expand it a bit?

lewtun (commented):
Thanks for working on this killer feature @JingyaHuang 🔥 !!

I've left a few nits, but the API design looks great to me. Would you mind sharing a small code example in the PR description once you have everything ready for a final review? 

I'm especially interested to know if quantization / optmization play nice with the current implementation :)

JingyaHuang (commented):
Hi @lewtun, thanks for the review!!! 

I will apply the helper on other ort models and update the PR description with a snippet once it is finished.

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,173 @@
...
+        ort_type_to_numpy_type_map = {
+            "tensor(int64)": np.int64,
+            "tensor(int32)": np.int32,
+            "tensor(int8)": np.int8,
+            "tensor(float)": np.float32,
+            "tensor(float16)": np.float16,
+            "tensor(bool)": np.uint8,
+        }
+        if ort_type not in ort_type_to_numpy_type_map:
+            raise ValueError(f"{ort_type} not found in map")
```
philschmid, line 48: We should improve the error message `map` might not be very helpful. 

Also, nit, but feel free to keep it.
I like it more if you check if the values are in the map and then return it rather than checking for an error.

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,173 @@
...
+        ort_type_to_torch_type_map = {
+            "tensor(int64)": torch.int64,
+            "tensor(int32)": torch.int32,
+            "tensor(int8)": torch.int8,
+            "tensor(float)": torch.float32,
+            "tensor(float16)": torch.float16,
+            "tensor(bool)": torch.bool,
+        }
+        if ort_type not in ort_type_to_torch_type_map:
+            raise ValueError(f"{ort_type} not found in map")
```
philschmid, line 62: We should improve the error message `map` might not be very helpful. 

Also, nit, but feel free to keep it.
I like it more if you check if the values are in the map and then return it rather than checking for an error.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -445,18 +450,40 @@ def forward(
...
+            self.model.run_with_iobinding(io_binding)
+            io_binding.synchronize_outputs()
+
+            # map outputs with names
+            outputs = {}
+            for name, output in zip(io_helper.model_output_names, io_binding._iobinding.get_outputs()):
+                outputs[name] = IOBindingHelper.to_pytorch(output)
+
+            # converts output to namedtuple for pipelines post-processing
+            return CausalLMOutputWithCrossAttentions(**outputs)
```
philschmid, line 471: why is the output class different between the io bindings and not?

philschmid (commented):
Awesome work @JingyaHuang! I added some first comments to the `io_binding_helper.py` and first model. 
Most of my comments are about performance. I am not sure if we need to keep things typer dynamic since we have dedicated classes for each "task" meaning, which allows us to have more "static/defined" code in the `forward` method to improve latency. 
It would be interested if you can take a look at my comments and try to evaluate the performance of replacing those "dynamic loops" with a more "static approach"

optimum/onnxruntime/modeling_ort.py
JingyaHuang: Good catch! I forgot to change the output type(I adapted other models by the code of causal...)

optimum/onnxruntime/modeling_ort.py
```diff
@@ -445,18 +450,40 @@ def forward(
...
+
+            # run inference with binding
+            io_binding.synchronize_inputs()
+            self.model.run_with_iobinding(io_binding)
+            io_binding.synchronize_outputs()
+
+            # map outputs with names
+            outputs = {}
+            for name, output in zip(io_helper.model_output_names, io_binding._iobinding.get_outputs()):
+                outputs[name] = IOBindingHelper.to_pytorch(output)
```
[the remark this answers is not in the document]
JingyaHuang: Indeed! I will remove zip(except for `ORTModelForCustomTasks`)

optimum/onnxruntime/modeling_ort.py
```diff
@@ -445,18 +450,40 @@ def forward(
...
-        outputs = self.model.run(None, onnx_inputs)
-        last_hidden_state = torch.from_numpy(outputs[self.model_outputs["last_hidden_state"]]).to(self.device)
-        # converts output to namedtuple for pipelines post-processing
-        return BaseModelOutput(last_hidden_state=last_hidden_state)
+        if self.device.type == "cuda" and self.use_io_binding:
+            onnx_inputs = {"input_ids": input_ids, "attention_mask": attention_mask}
+            if token_type_ids is not None:
+                onnx_inputs["token_type_ids"] = token_type_ids
+            io_helper = IOBindingHelper(self.model, self.config, self.device)
+            io_binding = io_helper.prepare_io_binding(**onnx_inputs)
```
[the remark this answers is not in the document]
JingyaHuang: Will check the penalty! Indeed for models with fixed inputs and outputs we can do that, then there will be much more lines of code also...

optimum/onnxruntime/modeling_ort.py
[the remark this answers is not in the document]
JingyaHuang: And `ORTDecoder`, as for `ORTDecoder` and `ORTModelForCustomTasks`, their outputs are dynamic and I need to map the outputs with their names.

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,173 @@
...
+        self.model = model
+        self.config = config
+        self.device = device
+        # Create {name:idx} dict for model inputs and outputs
+        self.model_inputs = {output_key.name: idx for idx, output_key in enumerate(model.get_inputs())}
+        self.model_outputs = {output_key.name: idx for idx, output_key in enumerate(model.get_outputs())}
+        self.model_input_names = list(self.model_inputs.keys())
+        self.model_output_names = list(self.model_outputs.keys())
+
+    def prepare_io_binding(self, **kwargs) -> ort.IOBinding:
```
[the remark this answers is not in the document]
JingyaHuang: Hi @philschmid, do you mean that I should change the name or should I not use keyword arguments here?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -445,18 +450,40 @@ def forward(
...
-        # run inference
-        outputs = self.model.run(None, onnx_inputs)
-        last_hidden_state = torch.from_numpy(outputs[self.model_outputs["last_hidden_state"]]).to(self.device)
-        # converts output to namedtuple for pipelines post-processing
-        return BaseModelOutput(last_hidden_state=last_hidden_state)
+        if self.device.type == "cuda" and self.use_io_binding:
+            onnx_inputs = {"input_ids": input_ids, "attention_mask": attention_mask}
+            if token_type_ids is not None:
+                onnx_inputs["token_type_ids"] = token_type_ids
+            io_helper = IOBindingHelper(self.model, self.config, self.device)
```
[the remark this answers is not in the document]
JingyaHuang: @philschmid actually I placed it here because it will throw an error if I place it in the `__init__`

```
  File "test_ort_infer_iobinding.py", line 104, in <module>
    output = v_onnx_model(**d)
  File "/workspace/optimum/modeling_base.py", line 57, in __call__
    return self.forward(*args, **kwargs)
  File "/workspace/optimum/onnxruntime/modeling_ort.py", line 1000, in forward
    io_binding = self.io_helper.prepare_io_binding(**onnx_inputs)
  File "/workspace/optimum/onnxruntime/io_binding/io_binding_helper.py", line 108, in prepare_io_binding
    io_binding.bind_input(
  File "/usr/local/lib/python3.8/dist-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 458, in bind_input
    C.OrtDevice(
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. onnxruntime.capi.onnxruntime_pybind11_state.OrtDevice(arg0: int, arg1: int, arg2: int)
```

optimum/onnxruntime/modeling_ort.py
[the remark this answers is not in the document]
JingyaHuang: Hi @philschmid, just ran a test (batch size:3000, iteration: 100)
* Bind with IO helper: {'Inference_runtime': 112.6225}
* Direct binding in the ort model: {'Inference_runtime': 86.8001}

Ok, there is a significant penalty, let's move the preparation of binding to each ort model. 
(And maybe also leave the dynamic method in helper to leave a door for dynamic cases like custom task ort model)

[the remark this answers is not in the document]
JingyaHuang: Thanks for the heads up!!  @philschmid

optimum/onnxruntime/io_binding/io_binding_helper.py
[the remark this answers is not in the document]
philschmid: We should define as much as possible named arguments rather than just having `**kwargs`

optimum/onnxruntime/modeling_ort.py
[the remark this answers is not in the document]
philschmid: > (And maybe also leave the dynamic method in helper to leave a door for dynamic cases like custom task ort model)

Do you think it is better to have it separate or in the CustomOrtModel. I think we can add it only the custom model since there are no other "users" of it right?

optimum/onnxruntime/modeling_ort.py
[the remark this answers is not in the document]
philschmid: Are you sure that's coming from placing it in the `__init__`? and might not be and order issue?

optimum/onnxruntime/modeling_ort.py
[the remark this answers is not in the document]
JingyaHuang: Sure, moved it to the ort model for custom tasks.

optimum/onnxruntime/io_binding/io_binding_helper.py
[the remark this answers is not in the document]
JingyaHuang: Solved as this function is customized and moved to ort models.

fxmarty:
Looks really neat, congratz! Have you already measured inference time with/without IOBindings? How does it compare?

optimum/onnxruntime/io_binding/io_binding_helper.py
fxmarty: So are the remaining `is_cupy_available` and `is_onnxruntime_training_available` methods useful here (maybe a future PR)?

JingyaHuang: @fxmarty, oh yeah, I will move them to another PR. Good catch!

optimum/onnxruntime/modeling_ort.py
```diff
@@ -110,6 +115,7 @@ class ORTModel(OptimizedModel):
     def __init__(self, model: ort.InferenceSession = None, config=None, **kwargs):
```
philschmid, line 115: we can make `use_io_binding` a named argument here and set it to `True`

optimum/onnxruntime/modeling_ort.py
```diff
@@ -440,6 +446,74 @@ def __init__(self, model=None, config=None, **kwargs):
...
+
+        return output_shape, output_buffer
+
+    def prepare_io_binding(
+        self,
+        input_ids: Optional[torch.Tensor] = None,
+        attention_mask: Optional[torch.Tensor] = None,
+        token_type_ids: Optional[torch.Tensor] = None,
+    ):
+        name_to_np_type = TypeHelper.get_io_numpy_type_map(self.model)
```
philschmid, line 465: is that needed to call on every `forward` ?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -884,6 +1323,60 @@ def prepare_inputs_for_generation(self, input_ids: torch.LongTensor, **kwargs) -
             inputs["attention_mask"] = kwargs["attention_mask"]
```
philschmid, line 1323: I couldn't comment above, but do we need to adjust the `prepare_inputs_for_generation`, to work with `.generate`?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -898,16 +1391,29 @@ def forward(
...
+
+            # run inference
+            outputs = self.model.run(None, onnx_inputs)
+            logits = torch.from_numpy(outputs[self.model_outputs["logits"]]).to(self.device)
+
+            # converts output to namedtuple for pipelines post-processing
+            return CausalLMOutputWithCrossAttentions(logits=logits)
 
     # Adapted from https://github.com/huggingface/transformers/blob/99289c08a1b16a805dd4ee46de029e9fd23cba3d/src/transformers/generation_utils.py#L490
     def _prepare_attention_mask_for_generation(
```
philschmid, line 1419: similar here

philschmid (commented):
I left some comments. I think we should add tests for all models to ensure the bindings are working and the outputs are similar to the ones with no bindings.

tests/onnxruntime/test_modeling.py
```diff
@@ -944,6 +1069,28 @@ def test_pipeline_on_gpu(self, model_arch):
 
         gc.collect()
 
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    @require_torch_gpu
+    def test_compare_to_io_binding(self, model_arch):
```
philschmid, line 1074: lets add a test for `.generate` as well

optimum/onnxruntime/modeling_ort.py
JingyaHuang: I don't think so, as the preparation will be done while calling forward during the generation.

JingyaHuang (commented):
@philschmid Thanks for reviewing! I have added some modifications according to the comments. And the tests are also added.

tests/onnxruntime/test_modeling.py
JingyaHuang: Added!

JingyaHuang:
Hi folks, 

Thanks for helping out and reviewing. I think the PR is ready for final review. Now the IO binding direct is applied to all ORTModels except for `ORTModelForCustomTasks`, for which I will open another PR.

Also since our last discussion, now the buffers' size for `ORTModelForSeq2SeqLM` has been reduced.

@lewtun Here is a snippet that I use for benchmarking, haven't tested quantization on TensorRT yet. 
```python
from pathlib import Path
import numpy as np
import pandas as pd
from time import perf_counter
import torch
from transformers import AutoConfig, AutoModelForSeq2SeqLM, AutoTokenizer
from optimum.onnxruntime.modeling_seq2seq import ORTModelForSeq2SeqLM
from optimum.onnxruntime import ORTOptimizer, ORTQuantizer
from optimum.onnxruntime.configuration import AutoQuantizationConfig, OptimizationConfig

model_id = "facebook/bart-base"
onnx_path = Path("results_seq2seq/")
seq_lengths = [8, 16, 32, 64, 128, 256, 512]

# Load vanilla onnx model
model = ORTModelForSeq2SeqLM.from_pretrained(model_id, from_transformers=True)
model.save_pretrained(onnx_path)
tokenizer.save_pretrained(onnx_path)

# Graph optimization
optimizer = ORTOptimizer.from_pretrained(model)
optimization_config = OptimizationConfig(optimization_level=2)  # enable all optimizations
optimizer.optimize(save_dir=onnx_path, optimization_config=optimization_config)

def benchmark(seq_len, model, tokenizer, device, iterations=200):
    # prepare date
    seq_len = "l " * (seq_len - 2)
    payload = tokenizer(seq_len, return_tensors="pt")
    payload = {key: val.to(device) for key, val in payload.items()}
    latencies = []
    # warm up
    for _ in range(10):
        _ = model.generate(**payload)
    # Timed run
    for _ in range(iterations):
        start_time = perf_counter()
        _ = model.generate(**payload)
        latency = perf_counter() - start_time
        latencies.append(latency)
    # Compute run statistics
    time_avg_ms = 1000 * np.mean(latencies)
    time_p95_ms = 1000 * np.percentile(latencies, 95)
    return {"seq_len": payload["input_ids"].shape[1], "time_avg_ms": time_avg_ms, "time_p95_ms": time_p95_ms}

device = torch.device("cuda:0")
# Baseline: PyTorch
config = AutoConfig.from_pretrained(model_id, use_cache=True)
pt_model = AutoModelForSeq2SeqLM.from_config(config)
pt_model.to(device)
# Case 1: Vanilla onnx with IO binding
v_onnx_model = ORTModelForSeq2SeqLM.from_pretrained(
    model_id, from_transformers=True, use_cache=True, use_io_binding=True
)
v_onnx_model.to(device)
# Case 2: graph optimized onnx with IOBinding
optim_onnx_model = ORTModelForSeq2SeqLM.from_pretrained(
    model_id="results_seq2seq",
    encoder_file_name="encoder_model_optimized.onnx",
    decoder_file_name="decoder_model_optimized.onnx",
    decoder_with_past_file_name="decoder_with_past_model_optimized.onnx",
)
optim_onnx_model.to(device)

# Benchmark
res = []
for seq_len in seq_lengths:
    print("seq_len: ", seq_len)
    pt = benchmark(seq_len, pt_model, tokenizer, device, iterations=500)
    res.append({**pt, "model": "pt"})

    v_onnx = benchmark(seq_len, v_onnx_model, tokenizer, device, iterations=500)
    res.append({**v_onnx, "model": "v_onnx"})

df = pd.DataFrame(res)
print(df)

chart_df = pd.merge(
    df[df.model == "pt"][["seq_len", "time_p95_ms"]],
    df[df.model == "v_onnx"][["seq_len", "time_p95_ms"]],
    on="seq_len",
)
chart_df = chart_df.rename(
    columns={
        "time_p95_ms_x": "pt_p95",
        "time_p95_ms_y": "v_onnx_p95",
    }
)
chart_df = pd.merge(
    chart_df,
    df[df.model == "optim_onnx"][["seq_len", "time_p95_ms"]],
    on="seq_len",
)
chart_df = chart_df.rename(
    columns={
        "time_p95_ms": "optim_onnx_p95",
    }
)

chart_df["io_improvement/pt"] = f"{round((chart_df['pt_p95'] - chart_df['v_onnx_p95']) / chart_df['pt_p95'] * 100,2)}%"
chart_df["io+optim/pt"] = f"{round((chart_df['pt_p95'] - chart_df['optim_onnx_p95']) / chart_df['pt_p95'] * 100,2)}%"

plt = chart_df.plot(x="seq_len", y=["pt_p95", "v_onnx_p95", "optim_onnx_p95"], kind="line")
plt.figure.savefig("gpu_res_iobinding_seq2seq.png", dpi=900)

print(chart_df.head(10))
chart_df.to_csv("gpu_res_iobinding_seq2seq.csv")
``` 

@echarlaix I add some seq2seq models to `ORTConfigManager` in order to pass the modeling tests. For they are tested with graph optimization. And whether a model can be well optimized is decided by `check_optimization_supported_model_or_raise` now(also updated in `ORTOptimizer`).

regisss (commented):
Huge PR @JingyaHuang :fire: :rocket: 
I just left a couple of minor comments.

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -0,0 +1,60 @@
...
+    @staticmethod
+    def ort_type_to_numpy_type(ort_type: str):
+        ort_type_to_numpy_type_map = {
+            "tensor(int64)": np.int64,
+            "tensor(int32)": np.int32,
+            "tensor(int8)": np.int8,
+            "tensor(float)": np.float32,
+            "tensor(float16)": np.float16,
+            "tensor(bool)": np.uint8,
+        }
```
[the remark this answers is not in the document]
JingyaHuang: Actually, it is overwritten from ORT's [TypeHelper](https://github.com/microsoft/onnxruntime/blob/93e0a151177ad8222c2c95f814342bfa27f0a64d/onnxruntime/python/tools/transformers/io_binding_helper.py#L29) to support more dtype, I would prefer to keep the original structure.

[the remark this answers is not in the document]
regisss: Ah okay, sounds good then :+1:

optimum/onnxruntime/modeling_ort.py
```diff
@@ -121,6 +131,12 @@ def __init__(self, model: ort.InferenceSession = None, config: PretrainedConfig
                 f" Use `ort_model.to()` to send the outputs to the wanted device."
             )
 
+        if "TensorrtExecutionProvider" in self.providers and self.use_io_binding:
+            logger.warning(
+                "There is no need to do IO binding for TensorrtExecutionProvider, `use_io_binding` will be set to False."
```
fxmarty, line 136: Maybe basic, but you know why it is? Is it hurtful to use IOBinding with TensorrtExecutionProvider?

tests/onnxruntime/test_modeling.py
```diff
@@ -944,6 +1079,47 @@ def test_pipeline_on_gpu(self, model_arch):
...
+        set_seed(SEED)
+        io_model = ORTModelForCausalLM.from_pretrained(model_id, from_transformers=True, use_io_binding=True)
+
+        tokenizer = get_preprocessor(model_id)
+        tokens = tokenizer("This is a sample output", return_tensors="pt")
+        onnx_outputs = onnx_model.generate(**tokens)
+        io_outputs = io_model.generate(**tokens)
+
+        # compare tensor outputs
+        self.assertTrue(torch.allclose(onnx_outputs, io_outputs, atol=1e-4))
```
philschmid, line 1119: Shouldn't those be equal and not only close? Since the `.generate` return the `input_ids`

tests/onnxruntime/test_modeling.py
```diff
@@ -1169,6 +1373,50 @@ def test_compare_with_and_without_past_key_values_model_outputs(self):
...
+        set_seed(SEED)
+        io_model = ORTModelForSeq2SeqLM.from_pretrained(model_id, from_transformers=True, use_io_binding=True)
+
+        tokenizer = get_preprocessor(model_id)
+        tokens = tokenizer("This is a sample output", return_tensors="pt")
+        onnx_outputs = onnx_model.generate(**tokens)
+        io_outputs = io_model.generate(**tokens)
+
+        # compare tensor outputs
+        self.assertTrue(torch.allclose(onnx_outputs, io_outputs, atol=1e-4))
```
philschmid, line 1416: Shouldn't those be equal and not only close? Since the `.generate` return the `input_ids`

philschmid (approved):
Awesome work @JingyaHuang 🚀✅ Looks good to me. Everything else can be a follow up PR. 
I left some two minor comments.

tests/onnxruntime/test_modeling.py
JingyaHuang: Yes, exactly. Will correct it!

optimum/onnxruntime/modeling_ort.py
JingyaHuang: Actually, there is no need to do IOBinding as TensorRT self-manages the memory. If you look into the source code of [TRT EP in ONNX Runtime](https://github.com/microsoft/onnxruntime/blob/2ecd1d662203d3fb5d77bf2aa97f6b8bcd7d1759/onnxruntime/core/providers/tensorrt/tensorrt_execution_provider.cc#L1778), you can see that they obtain the dimensions info of the inputs and outputs -> create buffers -> enqueue TRT context with created buffers -> run inference -> output put inside output buffer. There is no need to do the binding manually.

PR #427, opened 2022-10-20
Support specific device id for ORTModel

### LINKED ISSUE #426: Tried to load a model to both cores of a two-core GPUs, but only one core (device:0) is used.
I used Optimum to convert a finetuned `xlm-roberta-base` model to an ONNX model by following the tutorial in [this link](https://huggingface.co/docs/optimum/quicktour). I have a two core GPU. When I tried to load the original finetuned model, I was able to load it to both cores of my GPU. But for some reason, when I try it with the ONNX model, the model only gets loaded into one of the cores. And because of that, that core's memory gets maxed out. 

The code that deals with model loading is below, and any suggestion and advice would be much appreciated! 

```
def worker(i, sents, model_path):
    tokenizer = AutoTokenizer.from_pretrained(model_path)
    model = ORTModelForSequenceClassification.from_pretrained(model_path)
    if i % 2 == 0:        
        pipe = pipeline("text-classification", 
                        model=model
                        tokenizer=tokenizer, 
                        padding="max_length", 
                        max_length=96, 
                        truncation=True, 
                        device=0, 
                        batch_size=64)
    else:
        pipe = pipeline("text-classification", 
                        model=model, 
                        tokenizer=tokenizer, 
                        padding="max_length", 
                        max_length=96, 
                        truncation=True, 
                        device=1, 
                        batch_size=64)
    scores = pipe(sents)
    return scores
```

fxmarty:
# What does this PR do?

Support to pass a specific device id (e.g. as an int) for ORTModel, as transformers does for pipelines.

For example, `model.to(1)` will use the GPU with id 1.

Fixes #426 

## Before submitting
- [x] Did you write any new necessary tests? ==> yes but disabled by default

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -49,7 +49,7 @@ import onnxruntime as ort
 from huggingface_hub import HfApi, hf_hub_download
 
 from ..modeling_base import FROM_PRETRAINED_START_DOCSTRING, OptimizedModel
-from .utils import ONNX_WEIGHTS_NAME, get_device_for_provider, get_provider_for_device
+from .utils import ONNX_WEIGHTS_NAME, get_device_for_provider, get_provider_for_device, parse_device
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


optimum/onnxruntime/modeling_ort.py
```diff
@@ -138,18 +138,48 @@ def device(self) -> torch.device:
...
+        """Get the relevant torch.device from the passed device, and if relevant the provider options (e.g. to set the GPU id)."""
+        provider_options = {}
+        if isinstance(device, torch.device):
+            if device.type == "cuda" and device.index:
+                provider_options["device_id"] = device.index
+            device = device
+        elif isinstance(device, str):
+            if device.startswith("cuda") and device[-1].isdigit():
+                provider_options["device_id"] = int(device[-1])
+            device = torch.device(device)
```
JingyaHuang, line 151: Why not
```suggestion
        elif isinstance(device, str):
            device = torch.device(device)
            if device.startswith("cuda") and device[-1].isdigit():
                provider_options["device_id"] = device.index
```

optimum/onnxruntime/modeling_ort.py
```diff
@@ -138,18 +138,48 @@ def device(self) -> torch.device:
     def device(self, value: torch.device):
         self._device = value
 
-    def to(self, device: torch.device):
+    def _parse_device(self, device: Union[torch.device, str, int]):
```
JingyaHuang, line 141: If you just want to map a device with its device id, I think it should be in utils instead of ort modeling. Have you checked `torch._C._nn._parse_to` for the parsing?

JingyaHuang (commented):
Hi @fxmarty, thanks for working on it! Just left some questions.

optimum/onnxruntime/modeling_ort.py
fxmarty: I followed you advice below instead

optimum/onnxruntime/modeling_ort.py
fxmarty: Yes it's better indeed. I modified & retested according to your comment. Let me know if it is good.

JingyaHuang (approved):
LGTM, thanks for iterating this!

fxmarty:
I disabled the dummy image classification benchmark parity test where the expected accuracy was hardcoded, that was failing due to an issue in datasets https://github.com/huggingface/datasets/issues/5145

Their is also the parity test for token classification failing, due to an issue with downloading conll2003 from deepai.org , whose data.deepai.org SSL certificate expired today. It's unrelated so I will merge anyway.

PR #432, opened 2022-10-25
Fix Torch FX documentation

regisss:
# What does this PR do?

This PR renames the FX documentation repository from `fx` to `torch_fx` because `main/fx`is interpreted as a language version of the documentation (this would happen for any combination of 2 letters).

It also adds a Habana notebook in the corresponding section of the documentation.

## Before submitting


regisss:
@mishig25 The FX doc used to be displayed well in PR docs as far as I can remember. So we will know for sure if this works after merging.

mishig25 (approved):
lgtm!

PR #437, opened 2022-10-28
Add former subpackage_index.html

regisss:
# What does this PR do?

This PR generates a `{subpackage}_index.html` so that existing links to it still work. They currently don't work because of the doc overhaul (see #429).

## Before submitting


lewtun (approved):
Thanks for the fix @regisss ! I checked the tiles links on both the copy & source `index.html` file and they look fine to me.

PR #439, opened 2022-10-28
Add FXModel

fxmarty:
## Motivation

With `optimum.fx`, we can quantize and do graph transformations quite easily, but that's about it. What about using the models once you are done? Converting to ONNX?

## What does this PR do?

This PR adds a class wrapping a `torch.fx.GraphModule` into a `transformers.PreTrainedModel` for easy use of the transformers toolings (pipelines, evaluate, etc.).

We add as well a method `wrap_fx_model` allowing us to do this wrapping as a one-liner.

A couple of caveats here:
* Pipelines require `forward()` methods to have `return_dict=True` (as e.g. `model_outputs["logits"]` is used in postprocess functions), **but** to export quantized models with fx to ONNX, `torch.onnx.export` calls `torch.jit.trace` that does not accept dictionary as an output: see [here](https://github.com/pytorch/pytorch/blob/2fa8142cf9469fa6570507ec097d4dd5e0b13c92/torch/onnx/utils.py#L1028) and [here](https://github.com/pytorch/pytorch/pull/31860) (would need to pass `strict=False` which is not feasible with the current pytorch implementation).
* `transformers.onnx.export` checks that the model input keys (from `generate_dummy_inputs`) are a subset of the signature of `forward`. Hence by rewriting `def forward(self, *args, **kwargs)` we loose the signature information, and we need hence have a custom on-the-fly workaround.

What do you think @michaelbenayoun @younesbelkada anybody else working with fx? Would this be useful? The code is rather hacky, so if you find a better way feel free to suggest. 

## Example of usage

```python
from optimum.fx import wrap_fx_model

import transformers
from transformers.utils.fx import symbolic_trace
from transformers import AutoModelForSequenceClassification, AutoTokenizer, pipeline

from pathlib import Path

model_name = "philschmid/tiny-bert-sst2-distilled"

model = AutoModelForSequenceClassification.from_pretrained(model_name)
model.eval()

tokenizer = AutoTokenizer.from_pretrained(model_name)

fx_model = symbolic_trace(model, input_names=["input_ids", "token_type_ids", "attention_mask"])

# here do whatever modification you want on the graph, but preserving the input / output

# _w for wrapped
fx_model_w = wrap_fx_model(graph_module=fx_model, config=fx_model.config, task="sequence-classification")
```

```python
pipe = pipeline(task="text-classification", model=fx_model_w, tokenizer=tokenizer)

res = pipe("well let's hope the team will liked it!'")
print(res)
```

```python
from transformers.onnx import FeaturesManager

_, model_onnx_config = FeaturesManager.check_supported_model_or_raise(model, feature="sequence-classification")
onnx_config = model_onnx_config(model.config)

# fx_model_w.return_dict = False  # required for onnx export if we do some quantization with fx first

onnx_path = Path("model_fx.onnx")
res = transformers.onnx.export(tokenizer, fx_model_w, onnx_config, onnx_config.default_onnx_opset, onnx_path)
```

## Left to do
- [ ] Tests
- [ ] Documentation
- [ ] Tutorial on how to use with transforms / quantization

### DIFF
--- a/optimum/fx/__init__.py
+++ b/optimum/fx/__init__.py
@@ -13,3 +13,4 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 from . import optimization, quantization
+from .modeling_fx import FXModel, wrap_fx_model


younesbelkada:
Thanks a lot @fxmarty  ! 
This could be super useful in the context of #423 ! We were wondering with @michaelbenayoun how to use fx without "forgetting" the class functions after transformations 🔥

michaelbenayoun:
Nice @fxmarty !
There is also the approach we talked about offline. Let's discuss what is more adapted for us.

PR #443, opened 2022-11-02
Add Model loading from subfolders similar to transformers

### LINKED ISSUE #308: Add Model loading from subfolders similar to transformers
### Feature request

Transformers is adding support for loading models from subfolders: https://github.com/huggingface/transformers/pull/18184

```python
from transformers import BertModel
BertModel.from_pretrained("fusing/latent-diffusion-text2im-large", revision="d5eab56", subfolder="bert")
```

### Motivation

Makes it easier to manage optimized models in existing repositories

### Your contribution

.

fxmarty:
# What does this PR do?

Allow to pass the argument `subfolder` to `ORTModel.from_pretrained()` in a similar fashion to transformers `PreTrainedModel`. It allows to host several models (e.g. optimized) in the same model repository.

The preprocessor is expected to be loaded from the parent directory.

Fixes #308

## Before submitting

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -175,7 +175,7 @@ class ORTModel(OptimizedModel):
 
         Arguments:
             path (`str` or `Path`):
-                Directory from which to load the model.
+                Path of the ONNX model.
             provider (`str`, *optional*):
                 ONNX Runtime provider to use for loading the model. See https://onnxruntime.ai/docs/execution-providers/
                 for possible providers. Defaults to `CPUExecutionProvider`.
@@ -230,6 +230,7 @@ class ORTModel(OptimizedModel):
         force_download: bool = False,
         use_auth_token: Optional[str] = None,
         cache_dir: Optional[str] = None,
+        subfolder: Optional[str] = "",
         provider: Optional[str] = "CPUExecutionProvider",
         session_options: Optional[ort.SessionOptions] = None,
         provider_options: Optional[Dict] = None,
@@ -255,6 +256,7 @@ class ORTModel(OptimizedModel):
             force_download,
             use_auth_token,
             cache_dir,
+            subfolder,
             provider=provider,
             session_options=session_options,
             provider_options=provider_options,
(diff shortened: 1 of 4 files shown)


optimum/onnxruntime/modeling_ort.py
```diff
@@ -356,23 +361,27 @@ def _from_transformers(
...
+        save_dir = Path(save_dir).joinpath(model_id, subfolder)
         save_dir.mkdir(parents=True, exist_ok=True)
         kwargs["model_save_dir"] = save_dir
 
         # reads pipeline task from ORTModelForXXX class if available else tries to extract from hub
         if cls.export_feature is not None:
             task = cls.export_feature
         else:
-            task = HfApi().model_info(model_id, revision=revision).pipeline_tag
+            task = HfApi().model_info(model_id, revision=revision).pipeline_tag  # FIXME: load from subfolder?
```
regisss, line 372: Out of curiosity, the FIXME comment means that we are not sure if this can be done with `HfApi`?

regisss (approved):
LGTM @fxmarty!

fxmarty:
Actually this could be useful if we want to save ONNX models in subfolders on the Hub? cc @JingyaHuang

optimum/onnxruntime/modeling_ort.py
fxmarty: One of the issue is that the ONNX export in transformers relies on preprocessors to generate dummy inputs, hence we use `preprocessor = get_preprocessor(model_id)` that has no option to specify a subfolder. We would need to do a PR in transformers for that. Therefore, the tokenizer must be at the top level of the repo, as well as the `config.json` that may be required in `AutoTokenizer` to find the right class.

I changed a bit the code so that we don't need to expect a `config.json` in the subfolder, we fall back to using the one at the top level if there is not in the subfolder. Does it sound good to you?

regisss: Yes sounds good! In that case, maybe we could log a warning saying that the config was not found in the given subfolder and that we are looking for it in the parent folder.

PR #445, opened 2022-11-03
Exporters automatic task detection

michaelbenayoun:
# What does this PR do?

This PR enables automatic task detection for `optimum.exporters`.

- [ ] Task detection for local directories
- [x] Task detection for model repos on the Hub

### DIFF
--- a/optimum/exporters/onnx/__main__.py
+++ b/optimum/exporters/onnx/__main__.py
@@ -76,7 +76,7 @@ def main():
     # Infer the task
     task = args.task
     if task == "auto":
-        task = TasksManager.infer_task_from_model_info(args.model)
+        task = TasksManager.infer_task_from_model(args.model)
 
     # Allocate the model
     model = TasksManager.get_model_from_task(task, args.model, framework=args.framework, cache_dir=args.cache_dir)
@@ -127,7 +127,11 @@ def main():
         if isinstance(args.atol, dict):
             args.atol = args.atol[task.replace("-with-past", "")]
 
-    validate_model_outputs(onnx_config, model, args.output, onnx_outputs, args.atol)
+    try:
+        validate_model_outputs(onnx_config, model, args.output, onnx_outputs, args.atol)
+    except ValueError:
+        logger.error(f"An error occured, but the model was saved at: {args.output.as_posix()}")
+        return
     logger.info(f"All good, model saved at: {args.output.as_posix()}")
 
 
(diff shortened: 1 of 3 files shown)


optimum/exporters/tasks.py
```diff
@@ -659,6 +661,37 @@ def determine_framework(model: str, framework: str = None) -> str:
...
+        else:
+            tasks_to_automodels = TasksManager._TASKS_TO_TF_AUTOMODELS
+            class_name_prefix = "TF"
+
+        inferred_task_name = None
+        is_local = os.path.isdir(model_name_or_path)
+
+        if is_local:
+            # TODO: implement this.
+            raise NotImplementedError("Cannot infer the task from a local directory yet.")
```
julien-c, line 679: not sure we even want to support this, in a first version at least

optimum/exporters/tasks.py
```diff
@@ -659,6 +661,37 @@ def determine_framework(model: str, framework: str = None) -> str:
...
+        else:
+            model_info = huggingface_hub.model_info(model_name_or_path)
+            transformers_info = model_info.transformersInfo
+            if transformers_info is None or transformers_info.get("auto_model") is None:
+                raise RuntimeError(f"Could not infer the task from the model repo {model_name_or_path}")
+            auto_model_class_name = f"{class_name_prefix}{transformers_info['auto_model']}"
```
julien-c, line 685: i think `auto_model` can start with `TF` already in some cases so you just only prepend it if it's not already there

optimum/exporters/onnx/__main__.py
```diff
@@ -36,7 +36,7 @@ def main():
     )
     parser.add_argument(
         "--task",
-        default="default",
+        default="auto",
         help="The type of tasks to export the model with.",
```
julien-c, line 40: outside the scope of this PR but

`The type of task` (singular, not plural), no?

julien-c (commented):
looks food from quick glance!

julien-c:
BTW context for anyone reading this PR: the Hub determines AutoModel-type from models' config.json using:
- this file auto-generated from transformers: https://huggingface.co/datasets/huggingface/transformers-metadata/blob/main/pipeline_tags.json
- some legacy heuristics for older models

optimum/exporters/tasks.py
michaelbenayoun: Alright, changed the message and the exception to something that does not open the door for future support (even though it might come someday).

fxmarty:
LGTM, hopefully we put out a space tonight or tomorrow!

fxmarty (approved):

PR #453, opened 2022-11-09
Fix cuda device index

fxmarty:
cc @JingyaHuang 

Fix the second issue in https://github.com/huggingface/optimum/issues/365


JingyaHuang (approved):
LGTM, thanks for fixing!

PR #454, opened 2022-11-09
Fix IOBinding bugs

fxmarty:
`test_pipeline_on_gpu` test was broken for question answering and feature extraction. Fixing the bug.

Additionally, adding a missing `device=0` for the test for pipeline on gpu for several tasks.

There are still bugs to be solved, will open other PRs shortly.

cc @JingyaHuang


optimum/onnxruntime/modeling_ort.py
```diff
@@ -524,6 +524,7 @@ def prepare_io_binding(
             batch_size=input_ids.size(0),
```
JingyaHuang, line 524: Good catch, can you also update the comment from
`# bind logits` to `# bindlast_hidden_state` ?

JingyaHuang (approved):
LGTM, thanks for fixing!! Might better wait for the fix of GPU CI to merge.

optimum/onnxruntime/modeling_ort.py
fxmarty: Thanks!

PR #455, opened 2022-11-09
Fix ORTModelForSeq2SeqLM test

fxmarty:
# What does this PR do?

Fix a test that was not passing.

### DIFF
--- a/tests/onnxruntime/test_modeling.py
+++ b/tests/onnxruntime/test_modeling.py
@@ -1382,14 +1382,14 @@ class ORTModelForSeq2SeqLMIntegrationTest(unittest.TestCase):
         tokenizer = get_preprocessor(model_id)
         pipe = pipeline("translation_en_to_de", model=onnx_model, tokenizer=tokenizer, device=0)
         text = "My Name is Philipp and i live"
-        outputs = pipe(text)
+        outputs = pipe(text, min_length=len(text) + 1)
         # check model device
         self.assertEqual(pipe.model.device.type.lower(), "cuda")
         # compare model output class
         self.assertTrue(isinstance(outputs[0]["translation_text"], str))
 
         # does not pass with hf-internal-testing/tiny-random-t5 , but OK with t5-base
-        # self.assertTrue(len(outputs[0]["translation_text"]) > len(text))
+        self.assertTrue(len(outputs[0]["translation_text"]) > len(text))
 
     def test_compare_with_and_without_past_key_values_model_outputs(self):
         model_id = MODEL_NAMES["t5"]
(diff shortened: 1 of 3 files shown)


tests/onnxruntime/test_modeling.py
```diff
@@ -1386,8 +1386,10 @@ def test_pipeline_on_gpu(self, model_arch):
         # check model device
         self.assertEqual(pipe.model.device.type.lower(), "cuda")
         # compare model output class
-        self.assertTrue(isinstance(outputs[0]["generated_text"], str))
-        self.assertTrue(len(outputs[0]["generated_text"]) > len(text))
+        self.assertTrue(isinstance(outputs[0]["translation_text"], str))
+
+        # does not pass with hf-internal-testing/tiny-random-t5 , but OK with t5-base
+        # self.assertTrue(len(outputs[0]["translation_text"]) > len(text))
```
philschmid, line 1392: why is not passing?

tests/onnxruntime/test_modeling.py
```diff
@@ -1386,8 +1386,10 @@ def test_pipeline_on_gpu(self, model_arch):
         # check model device
```
philschmid, line 1386: should be added 1 line above: 
We can also add `generation_args` to have a `min_lenght` for the test. to pass?

tests/onnxruntime/test_modeling.py
fxmarty: no idea, it's not passing either with `hf-internal-testing/tiny-random-t5` + transformer model

fxmarty: The test was not run up to know due to the require gpu decorator

philschmid: can you check why? like whats the issue?

fxmarty: The output `translation_text` is `""`. I just think it's the model being bad.

philschmid: even when setting explicitly `min_length`?

fxmarty:
Yup, could be a bug in transformers.

```python
from transformers import pipeline as tf_pipeline
from transformers import AutoModelForSeq2SeqLM, AutoTokenizer

model_id = "hf-internal-testing/tiny-random-t5"

model = AutoModelForSeq2SeqLM.from_pretrained(model_id)

pipe = tf_pipeline("translation_en_to_de", model=model, tokenizer=tokenizer, min_length=len(text) + 1, max_length=2 * len(text) + 1)
outputs = pipe(text)

print(outputs)
print(len(outputs[0]["translation_text"]) > len(text))

## prints
[{'translation_text': ''}]
False
```

I'll look more into it ASAP.

tests/onnxruntime/test_modeling.py
fxmarty: When the pipeline parameter  `return_tensors` is `False`, the padding tokens are removed from the output. Hence the issue. Fixed now.

michaelbenayoun (approved):
Great, thanks for fixing this!

PR #457, opened 2022-11-09
ORTOptimizer refactor

michaelbenayoun:
# What does this PR do?

- Renames the attributes for the `OptimizationConfig`: 
    - The previous naming was not very clear, because it was matching the naming from what was done in `onnxruntime` for instance `disable_attention` for disabling attention fusion.
    - Make things clearer in terms of optimization tool by explicitly specifying which attributes tune the general purpose ONNX Runtime optimizer (compared to the other which are tuning the transformers specific optimizer).
- Adds an `AutoOptimizationConfig` class, with several optimization levels that cover the most common use cases, to make it easy to create an `OptimizationConfig` without having to dive in the ONNX Runtime ecosystem. 
- Uses `NormalizedConfig` when possible.

To be done:

- [ ] Write tests for configuration classes

### DIFF
--- a/docs/source/onnxruntime/usage_guides/optimization.mdx
+++ b/docs/source/onnxruntime/usage_guides/optimization.mdx
@@ -108,3 +108,27 @@ Below you will find an easy end-to-end example on how to optimize a Seq2Seq mode
 >>> outputs = optimized_model.generate(**tokens)
 ```
 
+
+## Optimization Configuration
+
+# TODO: edit this before merging, once the names have been chosen.
+
+In the optimization configuration, there are 4 possible optimization levels:
+- `general_optimization_level=0` to disable all optimizations,
+- `general_optimization_level=1` to enable basic optimizations such as constant folding or redundant node eliminations,
+- `general_optimization_level=2` to enable extended graph optimizations such as node fusions,
+- `general_optimization_level=99` to enable data layout optimizations.
+
+Choosing a level enables the optimizations of that level, as well as the optimizations of all preceding levels.
+More information [here](https://onnxruntime.ai/docs/performance/graph-optimizations.html).
+
+`fuse_operators=True` means that graph fusion is performed in addition to ONNX Runtime optimizations described above.
+Here is the list of all the operator fusions you can enable:
+- Gelu fusion with `disable_gelu_fusion=False`,
+- Layer Normalization fusion with `disable_layer_norm_fusion=False`,
+- Attention fusion with `disable_attention_fusion=False`,
+- SkipLayerNormalization fusion with `disable_skip_layer_norm_fusion=False`,
+- Add Bias and SkipLayerNormalization fusion with `disable_bias_skip_layer_norm_fusion=False`,
+- Add Bias and Gelu / FastGelu fusion with `disable_bias_gelu_fusion=False`.
+
+All other optimization configuration options are detailed in [`~onnxruntime.configuration.OptimizationConfig`].
--- a/docs/source/onnxruntime/usage_guides/optimization.mdx
+++ b/docs/source/onnxruntime/usage_guides/optimization.mdx
@@ -114,10 +114,10 @@ Below you will find an easy end-to-end example on how to optimize a Seq2Seq mode
 # TODO: edit this before merging, once the names have been chosen.
 
 In the optimization configuration, there are 4 possible optimization levels:
-- `general_optimization_level=0` to disable all optimizations,
-- `general_optimization_level=1` to enable basic optimizations such as constant folding or redundant node eliminations,
-- `general_optimization_level=2` to enable extended graph optimizations such as node fusions,
-- `general_optimization_level=99` to enable data layout optimizations.
+- `optimization_level=0` to disable all optimizations,
+- `optimization_level=1` to enable basic optimizations such as constant folding or redundant node eliminations,
+- `optimization_level=2` to enable extended graph optimizations such as node fusions,
+- `optimization_level=99` to enable data layout optimizations.
 
 Choosing a level enables the optimizations of that level, as well as the optimizations of all preceding levels.
 More information [here](https://onnxruntime.ai/docs/performance/graph-optimizations.html).
--- a/docs/source/onnxruntime/usage_guides/optimization.mdx
+++ b/docs/source/onnxruntime/usage_guides/optimization.mdx
@@ -64,9 +64,10 @@ Here is a list of the possible optimizations you can enable:
 - Gelu approximation with `enable_gelu_approximation=True`.
 
 While [`~onnxruntime.configuration.OptimizationConfig`] gives you full control on how to do optimization, it can be hard to know what to enable / disable. Instead, you can use [`~onnxruntime.configuration.AutoOptimizationConfig`] which provides 3 common optimizations levels:
-- O1: basic general optimizations
-- O2: basic and extended general optimizations, `transformers`-specific fusions, and Gelu approximation
-- O3: same as O2, but adds mixed precision.
+- O1: basic general optimizations.
+- O2: basic and extended general optimizations, `transformers`-specific fusions.
+- O3: same as O2 with Gelu approximation.
+- O4: same as O3 with mixed precision.
 
 Example: Loading a O2 [`~onnxruntime.configuration.OptimizationConfig`]
 
--- a/docs/source/onnxruntime/usage_guides/optimization.mdx
+++ b/docs/source/onnxruntime/usage_guides/optimization.mdx
@@ -17,7 +17,7 @@ specific language governing permissions and limitations under the License.
 
 ## Creating an `ORTOptimizer`
 
-The `ORTOptimizer` class is used to optimize your ONNX model. The class can be initialized using the `from_pretrained()` method, which supports different checkpoint formats.
+The [`~onnxruntime.ORTOptimizer`] class is used to optimize your ONNX model. The class can be initialized using the `from_pretrained()` method, which supports different checkpoint formats.
 
 1. Using an already initialized `ORTModelForXXX` class.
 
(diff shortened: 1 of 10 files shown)


optimum/onnxruntime/configuration.py
```diff
@@ -604,73 +608,222 @@ def tensorrt(
...
     OptimizationConfig is the configuration class handling all the ONNX Runtime optimization parameters.
+    There are two stacks of optimizations:
+        1. The ONNX Runtime general-purpose optimization tool: it can work on any ONNX model.
+        2. The ONNX Runtime transformers optimization tool: it can only work on a subset of transformers models.
 
-    Args:
-        optimization_level (`int`, defaults to 1):
-            ONNX opset version to export the model with.
+    Attributes:
+        general_optimization_level (`int`, defaults to 1):
```
michaelbenayoun, line 616: Not a big fan of this name, but basically trying to convey that this `optimization_level` is about the general ONNX Runtime tool, and not the `transformers`-specific optimizations.

Do you have any better suggestion @echarlaix @regisss ?

optimum/onnxruntime/configuration.py
```diff
@@ -604,73 +608,222 @@ def tensorrt(
...
+                - 99: will enable all available optimizations including layout optimizations
         optimize_for_gpu (`bool`, defaults to `False`):
             Whether to optimize the model for GPU inference.
             The optimized graph might contain operators for GPU or CPU only when `optimization_level` > 1.
         fp16 (`bool`, defaults to `False`):
             Whether all weights and nodes should be converted from float32 to float16.
-        optimize_with_onnxruntime_only (`bool`, defaults to `False`):
-            Whether to only use ONNX Runtime to optimize the model and no graph fusion in Python.
-        disable_gelu (`bool`, defaults to `False`):
+        enable_transformers_specific_optimizations (`bool`, defaults to `True`):
```
michaelbenayoun, line 629: I went for `enable_transformers_specific_optimizations`, which I think is nice because it conveys that we will use a `transformers` specific tool.
 @echarlaix @regisss (did not use `fuse_operators` since you also have approximations such as fast gelu)

docs/source/onnxruntime/usage_guides/optimization.mdx
```diff
@@ -108,3 +108,27 @@ Below you will find an easy end-to-end example on how to optimize a Seq2Seq mode
 >>> outputs = optimized_model.generate(**tokens)
 ```
 
+
+## Optimization Configuration
+
+# TODO: edit this before merging, once the names have been chosen.
```
michaelbenayoun, line 114: I will edit that once we settled with the names.

optimum/onnxruntime/configuration.py
regisss: I think we should keep `optimization_level` here because I have the feeling more accurate names will confuse users. This is not great but I don't think there is a perfect solution.
That means that we need to make it very clear in the documentation.

optimum/onnxruntime/configuration.py
regisss: Sounds good to me :+1:

optimum/onnxruntime/configuration.py
```diff
@@ -604,73 +608,222 @@ def tensorrt(
...
+                    f"{old_name} will be deprecated soon, use {new_name} instead.",
+                    FutureWarning,
+                )
+                if mapping_func is None:
+
+                    def identity(x):
+                        return x
+
+                    mapping_func = identity
+                setattr(self, new_name, mapping_func(getattr(self, old_name)))
```
regisss, line 707: Maybe we could put the deprecation warning after setting the value of the new attribute and add to the warning message something like: `f"{new_name} is set to {mapping_func(getattr(self, old_name))}"`.

optimum/onnxruntime/configuration.py
```diff
@@ -730,14 +732,87 @@ def create_fusion_options(self, model_type: str) -> FusionOptions:
...
             if hasattr(args, attr):
                 continue
             setattr(args, attr, value)
 
         return FusionOptions.parse(args)
 
 
+class AutoOptimizationConfig:
+    _LEVELS = [
+        {
```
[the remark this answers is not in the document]
regisss: I feel like 6 levels is a lot. Here is a proposal:
- 0: `general_optimization_level=1` and that's it (basic),
- 1: `general_optimization_level=2`, `enable_transformers_specific_optimizations=True` and `enable_gelu_approximation=True` (medium),
- 2: `general_optimization_level=99`, `enable_transformers_specific_optimizations=True`, `enable_gelu_approximation=True` and `fp16=True` (advanced).

I would also maybe name them O0, O1 and O2 to avoid any confusion with ORT optimization levels. WDYT?

optimum/onnxruntime/optimization.py
```diff
@@ -75,29 +85,28 @@ def from_pretrained(cls, model_or_path: Union[str, os.PathLike, ORTModel], file_
...
         else:
             raise ValueError(f"Unable to load the model from {model_or_path}.")
+        return cls(onnx_model_path, config=config)
 
     def optimize(
         self,
         optimization_config: OptimizationConfig,
         save_dir: Union[str, os.PathLike],
-        file_suffix: Optional[str] = "optimized",
+        file_suffix: str = "optimized",
```
regisss, line 105: The `Optional` hint should still be there

regisss (commented):
It looks good @michaelbenayoun! I left a few additional comments but since it's still a draft feel free to ignore them.

optimum/onnxruntime/configuration.py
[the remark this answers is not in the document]
michaelbenayoun: I agree!
What is the expected drop in accuracy with `enable_gelu_approximation=True`? If it's significant, I would suggest adding one level with only `general_optimization_level=2, enable_transformers_specific_optimizations=True `

docs/source/onnxruntime/usage_guides/optimization.mdx
```diff
@@ -108,3 +108,27 @@ Below you will find an easy end-to-end example on how to optimize a Seq2Seq mode
...
+In the optimization configuration, there are 4 possible optimization levels:
+- `general_optimization_level=0` to disable all optimizations,
+- `general_optimization_level=1` to enable basic optimizations such as constant folding or redundant node eliminations,
+- `general_optimization_level=2` to enable extended graph optimizations such as node fusions,
+- `general_optimization_level=99` to enable data layout optimizations.
+
+Choosing a level enables the optimizations of that level, as well as the optimizations of all preceding levels.
+More information [here](https://onnxruntime.ai/docs/performance/graph-optimizations.html).
+
+`fuse_operators=True` means that graph fusion is performed in addition to ONNX Runtime optimizations described above.
```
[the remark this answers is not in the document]
michaelbenayoun: Also, I will edit that now that we have settled the naming.

optimum/exporters/onnx/__main__.py
```diff
@@ -64,11 +64,11 @@ def main():
         ),
     )
     parser.add_argument("--cache_dir", type=str, default=None, help="Path indicating where to store cache.")
-    parser.add_argument("output", type=Path, help="Path indicating where to store generated ONNX model.")
+    parser.add_argument("output", type=Path, help="Path indicating the directory where to store generated ONNX model.")
 
     # Retrieve CLI arguments
     args = parser.parse_args()
-    args.output = args.output if args.output.is_file() else args.output.joinpath("model.onnx")
+    args.output = args.output.joinpath("model.onnx")
```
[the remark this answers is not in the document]
michaelbenayoun: No, but if the file did not exist it was failing, and I feel like it should be consistent => producing a directory with the file inside.
You think we should keep things as they were before?

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -711,18 +712,18 @@ def prepare_output_buffer(
...
             output_buffer = torch.empty(1, dtype=torch_type, device=self._device).contiguous()
         elif output_name == "logits":
-            output_shape = (batch_size, sequence_length, self.config.vocab_size)
+            output_shape = (batch_size, sequence_length, self.normalized_config.vocab_size)
             output_buffer = torch.empty(np.prod(output_shape), dtype=torch_type, device=self._device).contiguous()
         elif "key_values" in output_name:
-            num_heads = getattr(self.config, ORTConfigManager.get_num_heads_name(self.config.model_type))
-            hidden_size = getattr(self.config, ORTConfigManager.get_hidden_size_name(self.config.model_type))
-            embed_size_per_head = hidden_size // num_heads
+            num_attention_heads = self.normalized_confi.num_attention_heads
```
echarlaix, line 718:
```suggestion
            num_attention_heads = self.normalized_config.num_attention_heads
```

optimum/onnxruntime/configuration.py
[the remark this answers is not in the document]
echarlaix: An `AutoOptimizationConfig` could be interesting in the case where we want to optimize a model while keeping it compatible with quantization or if already have a quantized model and want to automatically define the optimizations that can be applied on it. I think we should wait before adding it as the different options could vary as there is no clear options which currently makes more sense than the others to me

echarlaix (commented):
It looks great ! 🔥

optimum/onnxruntime/configuration.py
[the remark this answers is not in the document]
michaelbenayoun: The current `OptimizationConfig` is not intuitive, hence the `AutoOptimizationConfig` hiding those details to the user. 
Also, we can add those features later, for now the `AutoOptimizationConfig` provides "good" common optimizations, independent from quantization, exactly as what `AutoQuantizationConfig` does for quantization.

optimum/exporters/onnx/__main__.py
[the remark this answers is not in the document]
regisss: No it's fine like this. I was just curious, it sounds good to me :+1:

optimum/onnxruntime/configuration.py
```diff
@@ -604,73 +608,248 @@ def tensorrt(
...
+                Whether the model to optimize will run on GPU, some optimizations depends on the hardware the model
+                will run on. Only needed for optimization_level > 1.
+            kwargs (`Dict[str, Any]`, *optional*):
+                Arguments to provide to the [`~OptimizationConfig`] constructor.
+
+        Returns:
+            `OptimizationConfig`: The `OptimizationConfig` corresponding to the requested optimization level.
+        """
+        if optimization_level not in cls._LEVELS:
+            raise ValueError(f"optimization_level must be in {', '.join(cls._LEVELS.keys())}")
```
regisss, line 792: Nit: I would add at the end of the error message: "got {optimization_level}"
```suggestion
            raise ValueError(f"optimization_level must be in {', '.join(cls._LEVELS.keys())}, got {optimization_level}")
```

optimum/onnxruntime/configuration.py
```diff
@@ -604,73 +608,248 @@ def tensorrt(
...
+
+    _LEVELS = {
+        "O1": {
+            "optimization_level": 1,
+            "enable_transformers_specific_optimizations": False,
+        },
+        "O2": {
+            "optimization_level": 2,
+            "enable_transformers_specific_optimizations": True,
+            "enable_gelu_approximation": True,
```
regisss, line 761: Yeah I don't know if `enable_gelu_approximation` makes sense in a medium config :thinking: 
I don't have a strong opinion on this since I don't really know to which extent it impacts the model behaviour.

regisss (approved):
Great PR @michaelbenayoun :rocket: 
I just left a nit.

optimum/onnxruntime/configuration.py
michaelbenayoun: I just added another optimization level.

PR #460, opened 2022-11-10
Fix python call for gpu EC2 test

fxmarty:
Second try.


fxmarty:
Sorry for the bad commit message, forgot to edit

PR #463, opened 2022-11-10
Fix onnxruntime gpu test yml

fxmarty:
As discussed on slack, one issue with the current EC2 AMI used is that CUDA/cuDNN are not installed, thus using `CUDAExecutionProvider` is not possible. So currently all tests from `test_onnxruntime_gpu.yml`.

As hinted by @JingyaHuang , it may be better to use docker for this.

Something I have noticed though in the AWS sandbox account is that the AMI ami-0dc1c26161f869ed1 used for test_onnxruntime_train.yml and test_onnxruntime_gpu.yml is >300 GB. Should we worry about this?

### DIFF
--- a/.github/workflows/test_onnxruntime_gpu.yml
+++ b/.github/workflows/test_onnxruntime_gpu.yml
@@ -51,14 +51,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Install dependencies
+      - name: Build image
         run: |
-          sudo apt -y update && sudo pip install --upgrade pip
-          pip install .[onnxruntime-gpu,tests]
-      - name: Test with unittest
-        working-directory: tests
+          docker build -f tests/onnxruntime/Dockerfile_onnxruntime_gpu -t onnxruntime-gpu .
+      - name: Test with unittest within docker container
         run: |
-          python3 -m unittest discover -s onnxruntime -p 'test_*.py'
+          docker run --gpus all --workdir=/workspace/optimum/tests onnxruntime-gpu:latest
 
   stop-runner:
     name: Stop self-hosted EC2 runner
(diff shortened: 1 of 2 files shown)


regisss:
> Something I have noticed though in the AWS sandbox account is that the AMI ami-0dc1c26161f869ed1 used for test_onnxruntime_train.yml and test_onnxruntime_gpu.yml is >300 GB. Should we worry about this?

I guess that the AMI was created from an image which had a few models and datasets cached. I don't think it's a big deal, but on the other hand it wouldn't take much to create a new "cleaner" AMI.

.github/workflows/test_onnxruntime_gpu.yml
```diff
@@ -51,14 +51,12 @@ jobs:
     steps:
       - name: Checkout
```
JingyaHuang, line 52: What about triggering the workflow both when a PR is merged to main and nightly(and workflow_dispatch of course), something like this
```
on:
  workflow_dispatch:
  pull_request:
    branches: 
      - main
    types: [closed]
  schedule:
    - cron: 0 7 * * * #every day at 7 am
```

.github/workflows/test_onnxruntime_gpu.yml
```diff
@@ -51,14 +51,12 @@ jobs:
...
         run: |
-          sudo apt -y update && sudo pip install --upgrade pip
-          pip install .[onnxruntime-gpu,tests]
-      - name: Test with unittest
-        working-directory: tests
+          docker build -f tests/onnxruntime/Dockerfile_onnxruntime_gpu -t onnxruntime-gpu .
+      - name: Test with unittest within docker container
         run: |
-          python3 -m unittest discover -s onnxruntime -p 'test_*.py'
+          docker run --gpus all --workdir=/workspace/optimum/tests onnxruntime-gpu:latest
```
JingyaHuang, line 59: Do you need to do that? I mounted and run `run.sh` in the training image because I couldn't configure `torch_ort` while building the image.

JingyaHuang: Never mind, it's better to do it your way.

JingyaHuang (approved):

.github/workflows/test_onnxruntime_gpu.yml
fxmarty: added!

PR #436, opened 2022-10-27
Use PretrainedConfig child classes for ORTModel config attribute

fxmarty:
# What does this PR do?

Currently, ORTModel `config` attribute was loading the config key/items from a vanilla dict using `PretrainedConfig.from_dict()`. With this PR, we rather load the config with `AutoConfig.from_pretrained()` that allows to instanciate child classes from `PretrainedConfig`, e.g. `DistilBertConfig`, `XLMRobertaConfig`, etc.

The current way may have been intended, is it the case @philschmid ? Do you think this change is OK? With the current code, the key `model_type` was not saved in the `config.json` when using `save_pretrained`.

Fixes #402

PR #466, opened 2022-11-14
Explicit error message when model not found for ORTModel

fxmarty:
https://github.com/huggingface/optimum/pull/436 broke the error message when the specified model is neither on the Hub or a local path. We would only get `UnboundLocalError: local variable 'config' referenced before assignment`

This PR fixes the issue.


regisss (approved):
LGTM!

PR #467, opened 2022-11-14
Fix docstring example for ORTModelForCausalLM

fxmarty:
There was no `optimum/gpt2`. An alternative could be to add `optimum/gpt2`, here the advantage is that if `gpt2` is already cached no need to redownload it.

Note that the modified docstring `TEXT_GENERATION_EXAMPLE` is used only once.


regisss (approved):
Thanks for fixing this @fxmarty!!

PR #480, opened 2022-11-17
Add doc for IOBinding

JingyaHuang:
# What does this PR do?

Add documentation for IOBinding on the GPU guide.

### DIFF
--- a/docs/source/onnxruntime/usage_guides/gpu.mdx
+++ b/docs/source/onnxruntime/usage_guides/gpu.mdx
@@ -123,7 +123,7 @@ Due to current limitations in ONNX Runtime, it is not possible to use quantized 
 
 ### Reduce memory footprint with IOBinding
 
-[IOBinding](https://onnxruntime.ai/docs/api/python/api_summary.html#iobinding) is an efficient way to avoid expensive data copying when using GPUs. By default, ONNX Runtime will copy the input from the CPU(even if the tensors are already copied to the targeted device), and assume that outputs also need to be copied back to the CPU from GPUs after the run. These data copying overheads between the host and devices are expensive, and __can lead to worse inference latency than vanilla PyTorch__ especially for decoding.
+[IOBinding](https://onnxruntime.ai/docs/api/python/api_summary.html#iobinding) is an efficient way to avoid expensive data copying when using GPUs. By default, ONNX Runtime will copy the input from the CPU (even if the tensors are already copied to the targeted device), and assume that outputs also need to be copied back to the CPU from GPUs after the run. These data copying overheads between the host and devices are expensive, and __can lead to worse inference latency than vanilla PyTorch__ especially for decoding.
 
 To avoid the slowdown, 🤗 Optimum adopts the IOBinding to copy inputs onto GPUs and pre-allocate memory for outputs prior the inference.  When instanciating the `ORTModel`, set the value of the argument `use_io_binding` to choose whether to turn on the IOBinding during the inference. `use_io_binding`is set to `True` by default, if you choose CUDA as execution provider.
 
--- a/docs/source/onnxruntime/usage_guides/gpu.mdx
+++ b/docs/source/onnxruntime/usage_guides/gpu.mdx
@@ -123,7 +123,7 @@ Due to current limitations in ONNX Runtime, it is not possible to use quantized 
 
 ### Reduce memory footprint with IOBinding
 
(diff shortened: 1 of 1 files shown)


docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -121,6 +121,27 @@ Due to current limitations in ONNX Runtime, it is not possible to use quantized
 
 * When using [static quantization](quantization#static-quantization-example), the ONNX computation graph will contain matrix multiplications and convolutions in floating-point arithmetic, along with Quantize + Dequantize operations to simulate quantization. In this case, although the costly matrix multiplications and convolutions will be run on the GPU, they will use floating-point arithmetic as the `CUDAExecutionProvider` can not consume the Quantize + Dequantize nodes to replace them by the operations using integer arithmetic.
 
+### Reduce memory footprint with IOBinding
+
+[IOBinding](https://onnxruntime.ai/docs/api/python/api_summary.html#iobinding) is an efficient way to avoid expensive data copying when using GPUs. By default, ONNX Runtime will copy the input from the CPU(even if the tensors are already copied to the targeted device), and assume that outputs also need to be copied back to the CPU from GPUs after the run. These data copying overheads between the host and devices are expensive, and __can lead to worse inference latency than vanilla PyTorch__ especially for decoding.
```
younesbelkada, line 126:
```suggestion
[IOBinding](https://onnxruntime.ai/docs/api/python/api_summary.html#iobinding) is an efficient way to avoid expensive data copying when using GPUs. By default, ONNX Runtime will copy the input from the CPU (even if the tensors are already copied to the targeted device), and assume that outputs also need to be copied back to the CPU from GPUs after the run. These data copying overheads between the host and devices are expensive, and __can lead to worse inference latency than vanilla PyTorch__ especially for decoding.
```

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -121,6 +121,27 @@ Due to current limitations in ONNX Runtime, it is not possible to use quantized
...
+
+To avoid the slowdown, Optimum adopts the IOBinding to copy inputs onto GPUs and pre-allocate memory for outputs prior the inference.  When instanciating the `ORTModel`, set the value of the argument `use_io_binding` to choose whether to turn on the IOBinding during the inference. `use_io_binding`is set to `True` by default, if you choose CUDA as execution provider.
```
younesbelkada, line 128:
```suggestion
To avoid the slowdown, 🤗 Optimum adopts the IOBinding to copy inputs onto GPUs and pre-allocate memory for outputs prior the inference.  When instanciating the `ORTModel`, set the value of the argument `use_io_binding` to choose whether to turn on the IOBinding during the inference. `use_io_binding`is set to `True` by default, if you choose CUDA as execution provider.
```

younesbelkada (approved):
Thanks a lot for working on this 💪 💯

fxmarty (approved):
Awesome!

docs/source/onnxruntime/usage_guides/gpu.mdx
```diff
@@ -123,7 +123,7 @@ Due to current limitations in ONNX Runtime, it is not possible to use quantized
 
 ### Reduce memory footprint with IOBinding
 
-[IOBinding](https://onnxruntime.ai/docs/api/python/api_summary.html#iobinding) is an efficient way to avoid expensive data copying when using GPUs. By default, ONNX Runtime will copy the input from the CPU(even if the tensors are already copied to the targeted device), and assume that outputs also need to be copied back to the CPU from GPUs after the run. These data copying overheads between the host and devices are expensive, and __can lead to worse inference latency than vanilla PyTorch__ especially for decoding.
+[IOBinding](https://onnxruntime.ai/docs/api/python/api_summary.html#iobinding) is an efficient way to avoid expensive data copying when using GPUs. By default, ONNX Runtime will copy the input from the CPU (even if the tensors are already copied to the targeted device), and assume that outputs also need to be copied back to the CPU from GPUs after the run. These data copying overheads between the host and devices are expensive, and __can lead to worse inference latency than vanilla PyTorch__ especially for decoding.
```
michaelbenayoun, line 126:
```suggestion
[IOBinding](https://onnxruntime.ai/docs/api/python/api_summary.html#iobinding) is an efficient way to avoid expensive data copying when using GPUs. By default, ONNX Runtime will copy the input from the CPU (even if the tensors are already copied to the targeted device), and assume that outputs also need to be copied back to the CPU from GPUs after the run. These data copying overheads between the host and devices are expensive, and __can lead to worse inference latency than vanilla PyTorch__ especially for decoding tasls.
```

JingyaHuang: I prefer to go with `decoding process`, to avoid confusion with ORTModel tasks.

michaelbenayoun: Decoding process is perfect!

PR #489, opened 2022-11-18
Fix tensorflow CI

fxmarty:
See https://github.com/huggingface/transformers/commit/37e016331f6fee7a6b8c430057c08129da48c97f and the corresponding slack thread


fxmarty:
Since it fixes the bug with exporters tests, I'll merge.

PR #497, opened 2022-11-21
Add ort export in exporters for encoder-decoder models

### LINKED ISSUE #496: Export `encoder` and `decoder` in the `encoder-decoder` models separately using exporters
### Feature request

Currently, the `encoder-decoder` models are exported as single `model.onnx` file using `exporters.onnx`. Add support to export the `encoder` and `decoder` separately using `exporters` cli based on`cmdline` arguments

### Motivation

The exported models can then be reused with `ORTModelXXX` class.

### Your contribution

I would implement this.

mht-sharma:
# What does this PR do?

Adds support to export the `encoder` and `decoder` separately for `encoder-decoder` models using `exporters` cli based on cmd line arguments.

Fixes #496 

Example usage:

```python
python -m optimum.exporters.onnx --model="openai/whisper-tiny.en" --task=speech2seq-lm-with-past output_whisper  --for-ort
```

## Before submitting


optimum/exporters/onnx/base.py
```diff
@@ -303,6 +303,56 @@ def flatten_output_collection_property(cls, name: str, field: Iterable[Any]) ->
         """
         return {f"{name}.{idx}": item for idx, item in enumerate(itertools.chain.from_iterable(field))}
 
+    def generate_dummy_inputs_onnxruntime(self, reference_model_inputs: Mapping[str, Any]) -> Mapping[str, Any]:
```
mht-sharma, line 306: Discuss with lewtun regarding the use of the function. https://github.com/huggingface/transformers/pull/19525#discussion_r1007728512

mht-sharma:
Currently only `Whisper` is exported for ORT in this PR.  Should I add the support for all the models or atleast a few more (example T5) in this PR. Or they could have their own PRs? @michaelbenayoun  WDYT.

optimum/exporters/onnx/__main__.py
```diff
@@ -64,14 +64,21 @@ def main():
         ),
     )
     parser.add_argument("--cache_dir", type=str, default=None, help="Path indicating where to store cache.")
+    parser.add_argument(
+        "--for-ort",
+        action="store_true",
+        help=(
+            "This generates ONNX models to run inference with ONNX Runtime ORTModelXXX for encoder-decoder models."
```
michaelbenayoun, line 71:
```suggestion
            "This exports models ready to be run with optimum.onnxruntime ORTModelXXX. Useful for encoder-decoder models."
```

optimum/exporters/onnx/__main__.py
```diff
@@ -115,41 +122,65 @@ def main():
...
+            (model, decoder_onnx_config, "decoder_model.onnx"),
+        ]
+
+        if "-with-past" in task:
+            decoder_onnx_config_with_past = onnx_config.get_decoder_onnx_config(
+                decoder_model.config, task.replace("-with-past", ""), use_past=True
+            )
+            models_for_export.append((model, decoder_onnx_config_with_past, "decoder_with_past_model.onnx"))
+    else:
+        models_for_export = [(model, onnx_config, "model.onnx")]
```
michaelbenayoun, line 150: Would it be possible to add this logic inside a function, and to use this function in `export`, or in both `export_pytorch` and `export_tensorflow`?

optimum/exporters/onnx/__main__.py
```diff
@@ -115,41 +122,65 @@ def main():
...
+    if args.atol is None:
+        args.atol = onnx_config.ATOL_FOR_VALIDATION
+        if isinstance(args.atol, dict):
+            args.atol = args.atol[task.replace("-with-past", "")]
+
+    if model.config.is_encoder_decoder and args.for_ort:
+        encoder_model = model.get_encoder()
+        encoder_onnx_config = onnx_config.get_encoder_onnx_config(encoder_model.config)
+
+        decoder_model = model.get_decoder()
```
michaelbenayoun, line 134: Does it handle heads after the decoder?

optimum/exporters/onnx/base.py
michaelbenayoun: It is needed to generate the inputs for the separate encoder and decoder models?
Is it only used for validation?

optimum/exporters/onnx/base.py
```diff
@@ -303,6 +303,56 @@ def flatten_output_collection_property(cls, name: str, field: Iterable[Any]) ->
...
+                Whether to export the model with past_key_values.
+
+        Returns:
+            `OnnxConfig`: An instance of the ONNX configuration object.
+        """
+        raise NotImplementedError(
+            f"{config.model_type} decoder export is not supported yet. ",
+            f"If you want to support {config.model_type} please propose a PR or open up an issue.",
+        )
+
```
michaelbenayoun, line 355: I think this can be added to `OnnxSeq2SeqConfigWithPast` instead of the base class, since it is specifically for those guys that we need to handle this case?

tests/exporters/test_onnx_export.py
```diff
@@ -266,6 +270,69 @@ def _onnx_export(self, test_name, name, model_name, task, onnx_config_class_cons
...
+                    f" {onnx_config.MIN_TORCH_VERSION}, got: {TORCH_VERSION}"
+                )
+
+        encoder_model = model.get_encoder()
+        encoder_onnx_config = onnx_config.get_encoder_onnx_config(encoder_model.config)
+
+        decoder_model = model.get_decoder()
+        decoder_onnx_config = onnx_config.get_decoder_onnx_config(
+            decoder_model.config, task.replace("-with-past", ""), use_past=False
+        )
```
michaelbenayoun, line 297: I think this is not neat, we need to either make sure that when `task` does not contain `"with-past"` we have `use_past=False` (isn't it the case already?), or that setting `use_past` to `False`, would ignore the `"with-past"` in `task`, wdyt?

michaelbenayoun:
To answer your question: you can also add support for other models if you feel like it!

optimum/exporters/onnx/base.py
mht-sharma: It is needed only for validation using onnxruntime.  Since the onnx model and torch model will have different input signatures when using `encoder_outputs` for exporting the model.

optimum/exporters/onnx/__main__.py
mht-sharma: Could be possible to move to `export` itself. But along with this the `validate_model_outputs` function would also need to be updated support multiple model validation.

optimum/exporters/onnx/__main__.py
mht-sharma: This line is just used to get the decoder config from the model. And the entire model is used to export the `decoder+heads` in lines below.

The export uses `encoder_outputs` as input which would ensure that all modules after the `encoder` is present in the exported `decoder` model.

tests/exporters/test_onnx_export.py
mht-sharma: Thought so. This function is used to return the  `decoder_onnx_config` object. `OnnxConfig` initalizer do not accept `task` with `-with-past` suffix or it would throw an error. Same is being applied here: [tasks.py:101](https://github.com/huggingface/optimum/blob/main/optimum/exporters/tasks.py#L101).

```python
return SpeechSeq2SeqDecoderOnnxConfig(config, task, use_past=use_past)
```

But in this specific case, we would want to export models with `use_past=True` and `use_past=False` when `-with-past` is present in the task.  Hence, we anyways would want to remove the `with-past` from task explicitly.

optimum/exporters/onnx/__main__.py
michaelbenayoun: You mean moving the validation inside the export function?

optimum/exporters/onnx/__main__.py
michaelbenayoun: Nice!

optimum/exporters/onnx/__main__.py
mht-sharma: Not exactly. If we the above logic for ORT inside the `export` function we would return 3 `onnx_outputs`. The same logic to create 3 models should also be part of the `validate_model_outputs`. Hence both the functions would change.

optimum/exporters/onnx/__main__.py
mht-sharma: Updated

optimum/exporters/onnx/__main__.py
mht-sharma: Created separate functions to export and validate  encoder-decoder models for conditional generation. Found this to be more cleaner than adding new args to the existing export function.

optimum/exporters/onnx/base.py
mht-sharma: Done!

mht-sharma:
> To answer your question: you can also add support for other models if you feel like it!

Added support for exporting the Seq2Seq-lm models also. AFAIK I covered the models with existing support let me know if I missed something.

optimum/exporters/onnx/__main__.py
```diff
@@ -115,12 +128,20 @@ def main():
             f"At least  {onnx_config.DEFAULT_ONNX_OPSET} is required."
         )
 
-    onnx_inputs, onnx_outputs = export(
-        model,
-        onnx_config,
-        args.opset,
-        args.output,
-    )
+    use_past = True if "-with-past" in task else False
```
michaelbenayoun, line 131: I think you do not need that since it is already in the onnx config no?

optimum/exporters/onnx/base.py
michaelbenayoun: What about calling it `generate_dummy_inputs_for_validation`?
I think it can be misleading otherwise (more so now that we use a `for-ort` argument that does not mean the same thing)

optimum/exporters/onnx/base.py
```diff
@@ -454,3 +466,43 @@ def flatten_past_key_values(self, flattened_output, name, idx, t):
...
+            config (`PretrainedConfig`):
+                The encoder model's configuration to use when exporting to ONNX.
+
+        Returns:
+            `OnnxConfig`: An instance of the ONNX configuration object.
+        """
+        raise NotImplementedError(
+            f"{config.model_type} encoder export is not supported yet. ",
+            f"If you want to support {config.model_type} please propose a PR or open up an issue.",
+        )
```
michaelbenayoun, line 485: Is it needed here?
Otherwise I would add it to `OnnxSeq2SeqConfigWithPast`

optimum/exporters/onnx/base.py
```diff
@@ -454,3 +466,43 @@ def flatten_past_key_values(self, flattened_output, name, idx, t):
...
+
```
michaelbenayoun, line 486:
```suggestion
@abstractmethod
```

optimum/exporters/onnx/base.py
```diff
@@ -454,3 +466,43 @@ def flatten_past_key_values(self, flattened_output, name, idx, t):
         flattened_output[f"{name}.{idx}.decoder.value"] = t[1]
         flattened_output[f"{name}.{idx}.encoder.key"] = t[2]
         flattened_output[f"{name}.{idx}.encoder.value"] = t[3]
+
```
michaelbenayoun, line 469:
```suggestion
@abstractmethod
```

optimum/exporters/onnx/base.py
michaelbenayoun, line 485: Maybe you did not set it as an abstract method to handle this case?

optimum/exporters/onnx/convert.py
```diff
@@ -61,6 +61,66 @@ def check_dummy_inputs_are_allowed(
...
+def validate_encoder_decoder_model_outputs(
+    config: OnnxConfig,
+    reference_model: Union["PreTrainedModel", "TFPreTrainedModel"],
+    onnx_named_outputs: List[str],
+    atol: float,
+    task: str,
+    use_past: bool,
+    encoder_onnx_model: Path,
+    decoder_onnx_model: Path,
+    decoder_with_past_onnx_model: Path = None,
```
michaelbenayoun, line 73:
```suggestion
    decoder_with_past_onnx_model: Optional[Path] = None,
```

optimum/exporters/onnx/convert.py
```diff
@@ -61,6 +61,66 @@ def check_dummy_inputs_are_allowed(
...
+            The absolute tolerance in terms of outputs difference between the reference and the exported model.
+        task (`str`)
+            The type of task to export the model with.
+        use_past (`bool`, *optional*, defaults to `None`):
+            Whether to export the model with past_key_values.
+        encoder_onnx_model (`Path`):
+            The path to the exported encoder ONNX model.
+        decoder_onnx_model (`Path`):
+            The path to the exported decoder ONNX model.
+        decoder_with_past_onnx_model (`Path`, *optional*, defaults to `None`):
```
michaelbenayoun, line 96:
```suggestion
        decoder_with_past_onnx_model (`Optional[Path]`, defaults to `None`):
```

optimum/exporters/onnx/convert.py
```diff
@@ -61,6 +61,66 @@ def check_dummy_inputs_are_allowed(
...
+            The configuration used to export the model.
+        reference_model ([`~PreTrainedModel`] or [`~TFPreTrainedModel`]):
+            The model used for the export.
+        onnx_named_outputs (`List[str]`):
+            The names of the outputs to check.
+        atol (`float`):
+            The absolute tolerance in terms of outputs difference between the reference and the exported model.
+        task (`str`)
+            The type of task to export the model with.
+        use_past (`bool`, *optional*, defaults to `None`):
```
michaelbenayoun, line 90:
```suggestion
        use_past (`bool`):
```

michaelbenayoun: Is it needed since we both have the `task` and the `config` arguments?

optimum/exporters/onnx/convert.py
```diff
@@ -223,7 +286,9 @@ def export_pytorch(
         device = torch.device(device)
         if device.type == "cuda" and torch.cuda.is_available():
             model.to(device)
-            dummy_inputs = tree_map(lambda value: value.to(device), dummy_inputs)
+            dummy_inputs = tree_map(
+                lambda value: value.to(device) if isinstance(value, torch.Tensor) else None, dummy_inputs
```
michaelbenayoun, line 290: Not sure about this, but just checking:
```suggestion
                lambda value: value.to(device) if isinstance(value, torch.Tensor) else value, dummy_inputs
```

optimum/exporters/onnx/convert.py
```diff
@@ -321,6 +386,68 @@ def export_tensorflow(
...
 
+def export_encoder_decoder_model(
+    model: Union["PreTrainedModel", "TFPreTrainedModel"],
+    config: OnnxConfig,
+    opset: int,
+    task: str,
+    use_past: bool,
+    encoder_output: Path,
+    decoder_output: Path,
+    decoder_with_past_output: Path = None,
```
michaelbenayoun, line 397:
```suggestion
    decoder_with_past_output: Optional[Path] = None,
```

optimum/exporters/onnx/convert.py
```diff
@@ -321,6 +386,68 @@ def export_tensorflow(
...
+            The version of the ONNX operator set to use.
+        task (`str`)
+            The type of task to export the model with.
+        use_past (`bool`):
+            Whether to export the model with past_key_values.
+        encoder_output (`Path`):
+            Directory to store the exported encoder ONNX model.
+        decoder_output (`Path`):
+            Directory to store the exported decoder ONNX model.
+        decoder_with_past_output (`Path`, *optional*, defaults to `None`):
```
michaelbenayoun, line 420:
```suggestion
        decoder_with_past_output (`Optional[Path]`, defaults to `None`):
```

optimum/exporters/onnx/model_configs.py
```diff
@@ -49,6 +49,70 @@
...
+        return {
+            "decoder_input_ids": "input_ids",
+            "encoder_outputs": "encoder_hidden_states",
+            "attention_mask": "encoder_attention_mask",
+        }
+
+    @property
+    def outputs(self) -> Mapping[str, Mapping[int, str]]:
+        common_outputs = super().outputs
+        self.add_past_key_values(common_outputs, direction="outputs")
```
michaelbenayoun, line 96: Regardless of `use_past`?

optimum/exporters/onnx/utils.py
```diff
@@ -69,3 +79,40 @@ def check_onnxruntime_requirements(minimum_version: packaging.version.Version):
             f"but we require the version to be >= {minimum_version} to enable all the conversions options.\n"
             "Please update ONNX Runtime by running `pip install --upgrade onnxruntime`"
         )
+
+
+def get_encoder_decoder_models_for_export(
+    model: Union["PreTrainedModel", "TFPreTrainedModel"], config: "OnnxConfig", task: str, use_past: bool
+) -> Dict[str, Tuple[Union["PreTrainedModel", "TFPreTrainedModel"], "OnnxConfig"]]:
+    """
+    Exports a Pytorch or TensorFlow model to an ONNX Intermediate Representation.
```
michaelbenayoun, line 88: Not sure about this docstring

optimum/utils/input_generators.py
```diff
@@ -296,6 +296,7 @@ class DummySeq2SeqDecoderTextInputGenerator(DummyDecoderTextInputGenerator):
         "decoder_input_ids",
         "decoder_attention_mask",
         "encoder_outputs",
+        "attention_mask",
```
michaelbenayoun, line 299: Can't we use the regular `DummyEncoderTextInputGenerator` in places where we need that instead?

michaelbenayoun (commented):
That's really great!
Left a few comments and questions, but huge work @mht-sharma !

optimum/exporters/onnx/base.py
mht-sharma: Yes, creating an abstract method will require the implementation of the methods. But it can be deferred for later.

mht-sharma: It's under the `OnnxSeq2SeqConfigWithPast` class. The GitHub view of the files might cause the confusion.

optimum/exporters/onnx/model_configs.py
mht-sharma: Yes, the `ORTModelForConditionalGeneration` requires the decoder to always output the hidden states, hence this needs to enabled.

Note: When `use_past` is False then the hidden states are present in the output. But the input will not contain the past_key_values.

optimum/exporters/onnx/__main__.py
mht-sharma: Yes, was probably thinking of how the `ORTModelForConditionalGeneration` is currently implemented which create onnx config always without past. But that can be modified later. 

Currently, updated to use `use_past` from the onnx config.

optimum/exporters/onnx/convert.py
mht-sharma: Removed.

optimum/exporters/onnx/utils.py
mht-sharma: Updated

optimum/utils/input_generators.py
mht-sharma: Yes, not sure why I removed the encoder text generator. I have updated the input generator.

optimum/exporters/onnx/model_configs.py
michaelbenayoun: Alright, I think you explained that to me once, sorry for asking again haha

michaelbenayoun (approved):
LGTM!

JingyaHuang (approved):
LGTM, love the consistency of new ONNX configs.

PR #500, opened 2022-11-22
fix NormalizedConfig init error message

PaulQbFeng:
# What does this PR do?

The error message displayed was incomplete. With this addition, the user is reminded that by setting allow_new=True, a new attribute can be added.

## Before submitting


JingyaHuang:
Thanks for completing the error message! @PaulQbFeng

PR #506, opened 2022-11-23
Add mobilebertlayer

raghavanone:
# What does this PR do?

Adds mobilebertlayer


optimum/bettertransformer/models/__init__.py
```diff
@@ -18,6 +18,7 @@
     BartEncoderLayerBetterTransformer,
     BertLayerBetterTransformer,
     DistilBertLayerBetterTransformer,
+    MobileBertLayer,
```
younesbelkada, line 21:
```suggestion
    MobileBertLayerBetterTransformer,
```

optimum/bettertransformer/models/__init__.py
```diff
@@ -31,6 +32,7 @@
     "Data2VecTextLayer": BertLayerBetterTransformer,
     "CamembertLayer": BertLayerBetterTransformer,
     "MarkupLMLayer": BertLayerBetterTransformer,
+    "MobileBertLayer": MobileBertLayer,
```
younesbelkada, line 35:
```suggestion
    "MobileBertLayer": MobileBertLayerBetterTransformer,
```

optimum/bettertransformer/models/encoder_models.py
```diff
@@ -751,3 +751,111 @@ def forward(self, hidden_states, attention_mask, **__):
         if hidden_states.is_nested and self.is_last_layer:
             hidden_states = hidden_states.to_padded_tensor(0.0)
         return (hidden_states,)
+
+
+class MobileBertLayer(BetterTransformerBaseLayer):
```
younesbelkada, line 756:
```suggestion
class MobileBertLayerBetterTransformer(BetterTransformerBaseLayer):
```

optimum/bettertransformer/models/__init__.py
```diff
@@ -18,6 +18,7 @@
     BartEncoderLayerBetterTransformer,
```
younesbelkada, line 18: I think that you forgot to import `MobileBertLayerBetterTransformer` above!

younesbelkada (commented):
Thanks a lot for your great work @raghavanone and for start adding `BetterTransformer` support for `MobileBert`!
I left few comments, several points needs to be addressed before merging. Starting from the naming of the new class. We prefer to call the new class as `xxxLayerBetterTransformer` to be able to correctly trace whether the conversion has been successfull or not. 
Also could you make sure this implementation is tested by adding `hf-internal-testing/tiny-random-MobileBertModel` [here](https://github.com/huggingface/optimum/blob/ed95b9fa8019af29ce1904ac3cfef8729eb4f4be/tests/bettertransformer/test_bettertransformer_encoder.py#L47)
Let us know if you face into any issue here!

younesbelkada:
hey @raghavanone !
do you need any help converting the model? It seems that you did forget to add `hf-internal-testing/tiny-random-MobileBertModel` in the testing suite?

raghavanone:
> hey @raghavanone ! do you need any help converting the model? It seems that you did forget to add `hf-internal-testing/tiny-random-MobileBertModel` in the testing suite?

Done

optimum/bettertransformer/models/encoder_models.py
```diff
@@ -751,3 +751,111 @@ def forward(self, hidden_states, attention_mask, **__):
...
+        # Model hyper parameters
+        self.num_heads = mobilebert_layer.attention.self.num_attention_heads
+        self.embed_dim = mobilebert_layer.attention.self.all_head_size
+
+        # Last step: set the last layer to `False` -> this will be set to `True` when converting the model
+        self.is_last_layer = False
+
+        self.validate_bettertransformer()
+
+    def forward(self, hidden_states, attention_mask, *_):
```
michaelbenayoun, line 817: What is the purpose of `*_` in the end of the arguments?

raghavanone: I was not sure, I saw *_ in every other forward call definition.

michaelbenayoun: Alright, maybe we should also add `**__` or something to handle potential keyword arguments?
@younesbelkada

raghavanone: I saw both some forward def having both *_ and **_ . We should possibly add more information about it in the guide published.

michaelbenayoun: Yes basically it is to "handle" potential arguments that were passed to the original forward function, by allowing this forward function to get them, but ignoring them right after.

younesbelkada:
The test is failing for few reasons, firstly MobileBERT uses a special mechanism called `bottleneck_attention` and `botteneck`, is these are enabled it seems that the conversion criteras are not met. Therefore I made a special model here: `ybelkada/tiny-mobilebertmodel` that you can replace on the testing suite and hopefully the test will pass!

optimum/bettertransformer/models/__init__.py
```diff
@@ -18,7 +18,7 @@
...
     BertLayerBetterTransformer,
     DistilBertLayerBetterTransformer,
-    ViltLayerBetterTransformer,
```
younesbelkada, line 21: This has been removed by mistake, could you please revert that change? 🙏

raghavanone: Done

michaelbenayoun:
There are conflicts, you need to fetch on the main branch and merge / rebase on this branch.

PR #509, opened 2022-11-23
[BT] put decorator on the correct place

younesbelkada:
# What does this PR do?

This PR puts the decorator `check_if_pytorch_greater_112` in the correct place. In fact, putting the decorator above the definition of the class does not somehow activates the check

cc @michaelbenayoun @fxmarty

### DIFF
--- a/optimum/bettertransformer/transformation.py
+++ b/optimum/bettertransformer/transformation.py
@@ -113,7 +113,6 @@ def set_last_layer(model):
         return False
 
 
-@check_if_pytorch_greater_112()
 class BetterTransformer(object):
     r"""
     A conversion wrapper that takes as an input the `transformers` model to be converted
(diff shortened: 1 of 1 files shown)


fxmarty:
Do we have tests of BetterTransformer for both 1.12 and 1.13?

younesbelkada:
Actually we don't have any tests for PT 1.12 and we can, the thing is I was just afraid it will add more boilerplate on the current testing suite (we already have ~40 tests and 2 CI runners), wdyt?

fxmarty:
I'm not sure, when trying with torchserve + pytorch 1.12. I recall having errors with 1.12. I'll come up with a reproducible example ASAP.

Let's merge this anywas, LGTM

fxmarty (approved):

PR #510, opened 2022-11-23
[BT] clearer error message for `norm_first`

younesbelkada:
# What does this PR do?

This PR fixes a small nit, when users run `BetterTransformer` under some models that uses `norm_first` the error message is not clear. This PR adds a sanity check and a clear explanation of what a user should do in case they're using a BT model under PyTorch 1.12 that uses `norm_first` 

cc @fxmarty @michaelbenayoun

### DIFF
--- a/optimum/bettertransformer/models/base.py
+++ b/optimum/bettertransformer/models/base.py
@@ -14,6 +14,8 @@
 import torch
 import torch.nn as nn
 
+from optimum.utils import is_pytorch_greater_version
+
 
 KNOWN_ACTIVATION_ATTRIBUTES = ["hidden_act", "activation", "act_fn", "activation_function"]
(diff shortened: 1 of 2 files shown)


optimum/bettertransformer/models/base.py
```diff
@@ -91,6 +93,14 @@ def validate_bettertransformer(self):
                 f" Number of heads must be even."
             )
 
+        # Check if norm_first is activated for pytorch==1.12
+        if self.norm_first and not is_pytorch_greater_version("1.13.0"):
```
fxmarty, line 97:
```suggestion
        if self.norm_first == True and not is_pytorch_greater_version("1.13.0"):
```

just personal taste, feel free to ignore

fxmarty (approved):
LGTM 🚀

Issue #511, opened 2022-11-23
Improve error messages if CUDAExecutionProvider/TensorrtExecutionProvider are tentively used without the required dependencies


fxmarty:
### Feature request

For example if `onnxruntime` is installed instead of `onnxruntime-gpu`. The error messages are not really helpful currently.

### Motivation

/

### Your contribution

/

fxmarty:
Closed in https://github.com/huggingface/optimum/pull/514

Issue #416, opened 2022-10-05
 Failed to create CUDAExecutionProvider.


hungnmai:
### System Info

```shell
Linux
Python version: 3.8.13
Cuda version: 11.6
GPU Tesla T4 16Gb
Python Libraries:

- transformers==4.22.1
- optimum==1.3.0
- onnx==1.12.0
- onnxruntime-gpu==1.12.1

```

### Who can help?

Hello @JingyaHuang and @echarlaix
I have an issue when running with onnx-gpu. I get an error  **Failed to create CUDAExecutionProvider.** when load model:
`model = ORTModelForQuestionAnswering.from_pretrained(
model_name,
                                                     file_name="model-optimized.onnx",
                                                     use_auth_token=access_token,
                                                     cache_dir=onnx_path)`

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

`model = ORTModelForQuestionAnswering.from_pretrained(model_name,file_name="model-optimized.onnx",
                                                     use_auth_token=access_token,
                                                     cache_dir=onnx_path)`

### Expected behavior

ONNX model can run with GPU

michaelbenayoun:
Hi @hungnmai, 
Can you share the full error message please, I am able to run this on my end:
```python
from transformers import AutoTokenizer
from optimum.onnxruntime import ORTModelForQuestionAnswering

tok = AutoTokenizer.from_pretrained("bert-base-uncased")
model = ORTModelForQuestionAnswering.from_pretrained(
    "bert-base-uncased",
    from_transformers=True, 
    provider="CUDAExecutionProvider"
)

model(**tok("this is a test", return_tensors="pt"))
```

It might be an installation issue, but I cannot confirm without the full error message.

BNDSllx:
Maybe your ORT version and CUDA version don't match, try checking the [version requirement](https://onnxruntime.ai/docs/reference/execution-providers/CUDA-ExecutionProvider.html)

fxmarty:
closed in https://github.com/huggingface/optimum/pull/514

PR #514, opened 2022-11-24
Clearer error messages when initilizing the requested ONNX Runtime execution provider fails

fxmarty:
# What does this PR do?

This adds more precise error messages with guidance to the user when the ONNX Runtime configuration is badly installed for some execution provider (notably CUDA, Tensorrt).

Note the fuzzy stuff with tokenizers env variable to avoid the warnings https://github.com/huggingface/transformers/issues/5486 , not sure how to do it better.

I did not add tests as this would require to install / uninstall pip packages to run (or have several docker images, on for each case), not sure if I should? What do you think?

Note that there remains an issue: in case `onnxruntime-gpu` is indeed installed but the dependencies are not met, ONNX Runtime will show `CUDAExecutionProvider` and `TensorrtExecutionProvider` in `get_available_providers()` but initializing the inference session with them will fail anyway, **only with a warning message in C++**. I'm wondering if we should not redirect the warning in Python (tried https://stackoverflow.com/questions/24277488/in-python-how-to-capture-the-stdout-from-a-c-shared-library-to-a-variable , is there better?) to raise an error with a better message.

Related issues https://github.com/huggingface/optimum/issues/511 https://github.com/huggingface/optimum/issues/416

### DIFF
--- a/optimum/utils/__init__.py
+++ b/optimum/utils/__init__.py
@@ -13,13 +13,9 @@
 #  limitations under the License.
 import importlib.util
 import inspect
-import os
-import subprocess
 from contextlib import contextmanager
 
 from packaging import version
-
-import onnxruntime
 
 
 CONFIG_NAME = "config.json"
@@ -30,7 +26,6 @@ _accelerate_available = importlib.util.find_spec("accelerate") is not None
 
 
 def is_onnxruntime_available():
-    """Not strict (will pass if either `onnxruntime`, `onnxruntime-gpu` or `onnxruntime-training` is installed)."""
     try:
         # Try to import the source file of onnxruntime - if you run the tests from `tests` the function gets
         # confused since there a folder named `onnxruntime` in `tests`. Therefore, `_onnxruntime_available`
--- a/optimum/utils/__init__.py
+++ b/optimum/utils/__init__.py
@@ -13,6 +13,8 @@
 #  limitations under the License.
 import importlib.util
 import inspect
+import os
+import subprocess
 from contextlib import contextmanager
 
 from packaging import version
@@ -26,6 +28,7 @@ _accelerate_available = importlib.util.find_spec("accelerate") is not None
 
 
 def is_onnxruntime_available():
+    """Not strict (will pass if either `onnxruntime` or `onnxruntime-gpu` is installed)."""
     try:
         # Try to import the source file of onnxruntime - if you run the tests from `tests` the function gets
         # confused since there a folder named `onnxruntime` in `tests`. Therefore, `_onnxruntime_available`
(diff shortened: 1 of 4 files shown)


optimum/utils/__init__.py
```diff
@@ -37,6 +40,22 @@ def is_onnxruntime_available():
     return _onnxruntime_available
 
 
+def is_onnxruntime_strict_available():
```
michaelbenayoun, line 43: Why "strict"?

optimum/utils/__init__.py
```diff
@@ -37,6 +40,22 @@ def is_onnxruntime_available():
...
+    os.environ["TOKENIZERS_PARALLELISM"] = tokenizers_parallelism
+    return result
+
+
+def is_onnxruntime_gpu_available():
+    tokenizers_parallelism = os.environ.get("TOKENIZERS_PARALLELISM", "")
+    os.environ["TOKENIZERS_PARALLELISM"] = "false"
+    result = subprocess.check_output(["pip list | grep onnxruntime-gpu || true"], shell=True).decode("utf-8") != ""
+    os.environ["TOKENIZERS_PARALLELISM"] = tokenizers_parallelism
+    return result
```
michaelbenayoun, line 56: I think those functions are quite heavy for just checking that.

fxmarty: Agreed, so far I have not find a way directly in Python to know whether `onnxruntime` is installed through `onnxruntime` or `onnxruntime-gpu` package. Will look more into it.

optimum/utils/__init__.py
fxmarty: It is true if and only if the pypi package `onnxruntime is installed` (not `onnxruntime-gpu`)

optimum/utils/__init__.py
JingyaHuang: Maybe with something shallower? Like:
```python
>>> import pkg_resources
>>> env = {pkg.key for pkg in pkg_resources.working_set}
>>> "onnxruntime" in env
```
What about also adding the check of `onnxruntime-training` ? I think they are in the same category of problem.

fxmarty: Thanks! Unfortunately `"onnxruntime" in env` returns True even if there is only onnxruntime-gpu installed. Let me have a look for onnxruntime-training!

optimum/utils/__init__.py
regisss: I agree with Michael, "strict" sounds a bit weird but I don't have much better ideas either. Would `is_onnxruntime_cpu_available` make sense? Or can we still run things on the GPU with `onnxruntime` but without `onnxruntime-gpu`?

fxmarty:
I changed much of the code, it should be much cleaner now, no call to `subprocess`, no issue with tokenizer. We just check the `capi` from onnxruntime and check the required libs can indeed be loaded or not.

regisss (approved):
LGTM!

PR #515, opened 2022-11-24
Optimum ONNX Runtime API improvement

michaelbenayoun:
# What does this PR do?

- `ORTModel`, `ORTModelDecoder` and `ORTModelForConditionalGeneration` can load any ONNX model files regardless of their names, allowing to load optimized and quantized models without having to specify the `file_name` argument.
- `ORTModel._from_transformers` now downloads and loads the model in a temporary directory instead of the cache, which was not a right place to store it.  
- `ORTQuantizer` saves both the model configuration and the tokenizer / processor / feature extractor, making the exported directory usable end-to-end.
- `ORTOptimizer` was already saving the model configuration but not the tokenize / processor / feature extractor, it does now.
- `OnnxConfigWithPast` now has two class atributes (that will be used when no value is specified in the `__init__` to avoid having to overwrite the `__init__` just for that) : 
   - `USE_PAST_FOR_INPUTS`: if True, past key values will be in the inputs of the model
   -  `USE_PRESENT_FOR_OUTPUTS`: if True, present key values will be in the outputs of the model
   - **This is useful for decoder and encoder-decoder models (cc @mht-sharma)**

Those features combined allow the following:

```python
from optimum.onnxruntime import (
    ORTQuantizer,
    ORTOptimizer,
    ORTConfig,
    AutoQuantizationConfig,
    AutoOptimizationConfig,
    ORTModelForSequenceClassification
)
model_name = "distilbert-base-uncased-finetuned-sst-2-english" 

model = ORTModelForSequenceClassification.from_pretrained(model_name, from_transformers=True)
optimizer = ORTOptimizer.from_pretrained(model)

optimizer.optimize(
AutoOptimizationConfig.with_optimization_level("O2"), "optimized_model"
)

quantizer = ORTQuantizer.from_pretrained(model)
quantizer.quantize(
     AutoQuantizationConfig.avx512_vnni(is_static=False), "quantized_model"
)

optimized_model = ORTModelForSequenceClassification.from_pretrained("optimized_model")
quantized_model = ORTModelForSequenceClassification.from_pretrained("quantized_model")

tokenizer = AutoTokenizer.from_pretrained("optimized_model")
inputs = tokenizer("This is easier now!", return_tensors="pt")

print(model(**inputs))
print(optimized_model(**inputs))
print(quantized_model(**inputs))
```
This was not possible before.

### DIFF
--- a/optimum/exporters/onnx/__main__.py
+++ b/optimum/exporters/onnx/__main__.py
@@ -14,12 +14,13 @@
 # limitations under the License.
 """Entry point to the optimum.exporters.onnx command line."""
 
+import logging
 from argparse import ArgumentParser
 from pathlib import Path
 
-from transformers import AutoFeatureExtractor, AutoTokenizer
+from transformers import AutoTokenizer
 
-from ...utils import logging
+from ...utils.save_utils import maybe_save_tokenizer_or_processor_or_feature_extractor
 from ..tasks import TasksManager
 from .base import OnnxConfigWithPast
 from .convert import (
@@ -30,7 +31,7 @@ from .convert import (
 )
 
 
-logger = logging.get_logger()  # pylint: disable=invalid-name
+logger = logging.get_logger(__name__)
 logger.setLevel(logging.INFO)
 
 
@@ -143,18 +144,7 @@ def main():
     # Saving the model config as this is needed sometimes.
     model.config.save_pretrained(args.output.parent)
 
-    # Saving the tokenizer / feature extractor as well.
-    try:
-        tokenizer = AutoTokenizer.from_pretrained(args.model)
-        tokenizer.save_pretrained(args.output.parent)
-    except Exception:
-        pass
-
-    try:
-        feature_extractor = AutoFeatureExtractor.from_pretrained(args.model)
-        feature_extractor.save_pretrained(args.output.parent)
-    except Exception:
-        pass
+    maybe_save_tokenizer_or_processor_or_feature_extractor(args.model, args.output.parent)
 
     if args.atol is None:
         args.atol = onnx_config.ATOL_FOR_VALIDATION
--- a/optimum/exporters/onnx/__main__.py
+++ b/optimum/exporters/onnx/__main__.py
@@ -14,12 +14,12 @@
 # limitations under the License.
 """Entry point to the optimum.exporters.onnx command line."""
 
-import logging
 from argparse import ArgumentParser
 from pathlib import Path
 
 from transformers import AutoTokenizer
 
+from ...utils import logging
 from ...utils.save_utils import maybe_save_tokenizer_or_processor_or_feature_extractor
 from ..tasks import TasksManager
 from .base import OnnxConfigWithPast
@@ -31,7 +31,7 @@ from .convert import (
 )
 
 
-logger = logging.get_logger(__name__)
+logger = logging.get_logger()
 logger.setLevel(logging.INFO)
 
 
--- a/optimum/exporters/onnx/base.py
+++ b/optimum/exporters/onnx/base.py
@@ -354,10 +354,14 @@ class OnnxConfigWithPast(OnnxConfig, ABC):
         task: str = "default",
         patching_specs: List[PatchingSpec] = None,
         use_past: bool = False,
-        use_past_in_inputs: Optional[bool] = USE_PAST_IN_INPUTS,
-        use_present_in_outputs: Optional[bool] = USE_PRESENT_IN_OUTPUTS,
+        use_past_in_inputs: Optional[bool] = None,
+        use_present_in_outputs: Optional[bool] = None,
     ):
         self.use_past = use_past
+        if use_past_in_inputs is None:
+            use_past_in_inputs = self.USE_PAST_IN_INPUTS
+        if use_present_in_outputs is None:
+            use_present_in_outputs = self.USE_PRESENT_IN_OUTPUTS
         self.use_past_in_inputs = use_past if use_past_in_inputs is None else use_past_in_inputs
         self.use_present_in_outputs = use_past if use_present_in_outputs is None else use_present_in_outputs
         super().__init__(config, task=task, patching_specs=patching_specs)
--- a/optimum/exporters/onnx/base.py
+++ b/optimum/exporters/onnx/base.py
@@ -344,7 +344,9 @@ class OnnxConfigWithPast(OnnxConfig, ABC):
     Inherits from [`~exporters.onnx.OnnxConfig`]. A base class to handle the ONNX configuration of decoder-only models.
     """
 
-    PAD_ATTENTION_MASK_TO_MATCH_TOTAL_SEQUENCE_LENGTH = True
+    PAD_ATTENTION_MASK_TO_MATCH_TOTAL_SEQUENCE_LENGTH: bool = True
+    USE_PAST_IN_INPUTS: Optional[bool] = None
+    USE_PRESENT_IN_OUTPUTS: Optional[bool] = None
 
     def __init__(
         self,
@@ -352,8 +354,12 @@ class OnnxConfigWithPast(OnnxConfig, ABC):
         task: str = "default",
         patching_specs: List[PatchingSpec] = None,
         use_past: bool = False,
+        use_past_in_inputs: Optional[bool] = USE_PAST_IN_INPUTS,
+        use_present_in_outputs: Optional[bool] = USE_PRESENT_IN_OUTPUTS,
     ):
         self.use_past = use_past
+        self.use_past_in_inputs = use_past if use_past_in_inputs is None else use_past_in_inputs
+        self.use_present_in_outputs = use_past if use_present_in_outputs is None else use_present_in_outputs
         super().__init__(config, task=task, patching_specs=patching_specs)
 
     @classmethod
@@ -375,15 +381,14 @@ class OnnxConfigWithPast(OnnxConfig, ABC):
     @property
     def outputs(self) -> Mapping[str, Mapping[int, str]]:
         common_outputs = super().outputs
-        if self.use_past:
+        if self.use_present_in_outputs:
             self.add_past_key_values(common_outputs, direction="outputs")
-
         return common_outputs
 
     @property
     def values_override(self) -> Optional[Mapping[str, Any]]:
         if hasattr(self._config, "use_cache"):
-            return {"use_cache": self.use_past}
+            return {"use_cache": self.use_past_in_inputs or self.use_present_in_outputs}
 
     @add_dynamic_docstring(text=GENERATE_DUMMY_DOCSTRING, dynamic_elements=DEFAULT_DUMMY_SHAPES)
     def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
@@ -407,7 +412,7 @@ class OnnxConfigWithPast(OnnxConfig, ABC):
 
         if (
             self.PAD_ATTENTION_MASK_TO_MATCH_TOTAL_SEQUENCE_LENGTH
-            and self.use_past
+            and self.use_past_in_inputs
             and "attention_mask" in dummy_inputs
         ):
             past_length = dummy_inputs["past_key_values"][0][0].shape[2]
@@ -473,7 +478,7 @@ class OnnxSeq2SeqConfigWithPast(OnnxConfigWithPast):
                 # We reset the value as the order in common_outputs (OrderedDict) is lost otherwise
                 else:
                     axes_names[axis_idx] = name
-        if self.use_past:
+        if self.use_present_in_outputs:
             self.add_past_key_values(common_outputs, direction="outputs")
 
         return common_outputs
--- a/optimum/exporters/onnx/base.py
+++ b/optimum/exporters/onnx/base.py
@@ -364,6 +364,16 @@ class OnnxConfigWithPast(OnnxConfig, ABC):
             use_present_in_outputs = self.USE_PRESENT_IN_OUTPUTS
         self.use_past_in_inputs = use_past if use_past_in_inputs is None else use_past_in_inputs
         self.use_present_in_outputs = use_past if use_present_in_outputs is None else use_present_in_outputs
+        if use_past != self.use_past_in_inputs:
+            logger.warning(
+                f"use_past = {use_past} is different than use_past_in_inputs = {use_past_in_inputs}, the value of "
+                "use_past_in_inputs will used for the inputs."
+            )
+        if use_past != self.use_present_in_outputs:
+            logger.warning(
+                f"use_past = {use_past} is different than use_present_in_outputs = {use_present_in_outputs}, the value "
+                "of use_present_in_outputs value will used for the outputs."
+            )
         super().__init__(config, task=task, patching_specs=patching_specs)
 
     @classmethod
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -132,10 +132,10 @@ class OptimizedModel(ABC):
 
     def push_to_hub(
         self,
-        save_directory: str = None,
-        repository_id: Optional[str] = None,
+        save_directory: str,
+        repository_id: str,
         private: Optional[bool] = None,
-        use_auth_token: Optional[Union[bool, str]] = None,
+        use_auth_token: Union[bool, str] = True,
     ) -> str:
         if isinstance(use_auth_token, str):
             huggingface_token = use_auth_token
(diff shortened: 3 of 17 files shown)


optimum/exporters/onnx/base.py
```diff
@@ -344,16 +344,26 @@ class OnnxConfigWithPast(OnnxConfig, ABC):
...
+        use_past_in_inputs: Optional[bool] = None,
+        use_present_in_outputs: Optional[bool] = None,
     ):
         self.use_past = use_past
+        if use_past_in_inputs is None:
+            use_past_in_inputs = self.USE_PAST_IN_INPUTS
+        if use_present_in_outputs is None:
+            use_present_in_outputs = self.USE_PRESENT_IN_OUTPUTS
+        self.use_past_in_inputs = use_past if use_past_in_inputs is None else use_past_in_inputs
+        self.use_present_in_outputs = use_past if use_present_in_outputs is None else use_present_in_outputs
```
regisss, line 366: Would it be worth it to log a warning when `use_past` is equal to the complementary of `use_past_in_inputs`/`use_present_in_outputs`?

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -198,474 +201,160 @@
...
-            The name of the ONNX file containing the decoder part of the model.
-        decoder_file_with_past_name (`str`, defaults to `optimum.onnxruntime.utils.ONNX_DECODER_WITH_PAST_NAME`):
-            The name of the ONNX file containing the decoder with past key/values part of the model.
-        model_save_dir (`str`, defaults to `""`):
-            The directory under which the model exported to ONNX was saved.
+    Encoder model for ONNX Runtime inference.
 
+    Args:
+        session (`ort.InferenceSession`):
+            The ONNX Runtime inference session associated to the encoder.
```
regisss, line 210: Why only `session`? The `__init__` seems to have more arguments.
I also think the `Args: ...` part should be moved to the `__init__`'s docstring, even though I'm not sure there is any consistency regarding this in HF libraries.

optimum/exporters/onnx/__main__.py
```diff
@@ -143,18 +144,7 @@ def main():
...
-        tokenizer.save_pretrained(args.output.parent)
-    except Exception:
-        pass
-
-    try:
-        feature_extractor = AutoFeatureExtractor.from_pretrained(args.model)
-        feature_extractor.save_pretrained(args.output.parent)
-    except Exception:
-        pass
+    maybe_save_tokenizer_or_processor_or_feature_extractor(args.model, args.output.parent)
```
mht-sharma, line 147: I remember discussing to move this inside export, is that still the case?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -465,121 +484,214 @@ def _from_pretrained(
...
+            force_download=force_download,
+        )
+
+        model_type = model.config.model_type.replace("_", "-")
+        model_name = getattr(model, "name", None)
+
+        onnx_config_constructor = TasksManager.get_exporter_config_constructor(
+            model_type, "onnx", task=task, model_name=model_name
+        )
+        onnx_config = onnx_config_constructor(model.config, use_present_in_outputs=True)
```
mht-sharma, line 664: I see this being repeated in multiple files. Would it make sense to have a helper function for these lines?

optimum/modeling_base.py
```diff
@@ -117,6 +118,8 @@ def save_pretrained(
         os.makedirs(save_directory, exist_ok=True)
 
         self.config.save_pretrained(save_directory)
+        for preprocessor in self._preprocessors:
```
fxmarty, line 121: `PreTrainedModel` does not save preprocessors with `save_pretrained`, why would we introduce this asymmetry in Optimum? Personnally I don't understand the motivation for this `self._preprocessors`. What's the issue with fully dissociating preprocessing and modeling, like it's done in transformers?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -14,25 +14,26 @@
...
 import shutil
 from pathlib import Path
-from typing import Any, Dict, Optional, Tuple, Union
+from tempfile import TemporaryDirectory
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
 import transformers
 from transformers import AutoModelForCausalLM, PretrainedConfig
```
fxmarty, line 26: if `PretrainedConfig` is imported only when doing type checking, we can add

```
from typing import TYPE_CHECKING

if TYPECHECKING:
    from transformers import PretrainedConfig
```

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -321,55 +325,71 @@ class ORTModelDecoder(ORTModel):
 
     def __init__(
         self,
-        config: transformers.PretrainedConfig,
         decoder_session: onnxruntime.InferenceSession,
+        config: transformers.PretrainedConfig,
```
fxmarty, line 329:
```suggestion
        config: "PretrainedConfig",
```

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -321,55 +325,71 @@ class ORTModelDecoder(ORTModel):
...
         """
+        # TODO: remove at version 2.0
+        def show_deprecated_argument(arg_name):
+            if kwargs.pop(arg_name, None) is not None:
+                logger.warning(
+                    f"The {arg_name} argument to create an {self.__class__.__name__} is deprecated, and not used "
+                    "anymore."
+                )
+
+        show_deprecated_argument("last_decoder_model_name:")
```
fxmarty, line 361:
```suggestion
        show_deprecated_argument("last_decoder_model_name")
```

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -465,121 +484,214 @@ def _from_pretrained(
...
+        model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
     ):
-        file_names = {}
-        # Load model from a local directory
-        if os.path.isdir(os.path.join(model_id, subfolder)):
-            decoder_with_past_path = (
-                os.path.join(model_id, subfolder, decoder_with_past_file_name) if use_cache else None
+        model_path = Path(model_id)
+
+        def validate_filename(filename):
```
fxmarty, line 502: Would this be better out of `_from_pretrained`. Otherwise this is redefined every time `_from_pretrained` is called, am I correct?

fxmarty, line 502: Can we add a short comment as to what this function does

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -465,121 +484,214 @@ def _from_pretrained(
...
+            if model_path.is_dir():
+                return (model_path / subfolder / filename).is_file()
+            succeeded = True
+            try:
+                get_hf_file_metadata(hf_hub_url(model_id, filename, subfolder=subfolder, revision=revision))
+            except Exception:
+                succeeded = False
+            return succeeded
+
+        def infer_filename(pattern: str, argument_name: str, fail_if_not_found: bool = True) -> str:
```
fxmarty, line 512: Same as above (Can we add a short comment as to what this function does + wouldn't it be better out)

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -465,121 +484,214 @@ def _from_pretrained(
...
+        provider: str = "CPUExecutionProvider",
+        session_options: Optional[onnxruntime.SessionOptions] = None,
+        provider_options: Optional[Dict[str, Any]] = None,
+        use_io_binding: bool = True,
+        task: Optional[str] = None,
+    ) -> "ORTModelDecoder":
+        if task is None:
+            task = cls._AUTOMODELS_TO_TASKS[cls.auto_model_class]
+
+        save_dir = TemporaryDirectory()
```
fxmarty, line 643: Why not save to the cache? Here this means that every time a model is loaded with `from_transformers=True`, we will need to redownload the model? Why is this better?

optimum/exporters/onnx/__main__.py
michaelbenayoun: I actually do not know if we should put this inside the export itself, good question.

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -198,474 +201,160 @@
...
-        self.providers = encoder_session.get_providers()
-        self._device = get_device_for_provider(encoder_session.get_providers()[0])
+    def prepare_io_binding(
+        self,
+        input_ids: Optional[torch.Tensor] = None,
+        attention_mask: Optional[torch.Tensor] = None,
+    ):
+        io_binding = self.session.io_binding()
 
-        if "TensorrtExecutionProvider" in self.providers and self.use_io_binding:
```
fxmarty, line 288: Is this removed? I could not find it elsewhere. If removed, why?

optimum/onnxruntime/quantization.py
```diff
@@ -195,32 +209,32 @@ def partial_fit(
...
             calibration_config (`CalibrationConfig`):
                 The configuration containing the parameters related to the calibration step.
-            onnx_augmented_model_name (`Union[str, os.PathLike]`):
+            onnx_augmented_model_name (`Union[str, Path]`, *optional*, defaults to `"augmented_model.onnx"`):
                 The path used to save the augmented model used to collect the quantization ranges.
-            operators_to_quantize (`list`, *optional*):
+            operators_to_quantize (`Optional[List[NodeType]]`, *optional*):
                 List of the operators types to quantize.
-            batch_size (`int`, defaults to 1):
+            batch_size (`int`, *optional*, defaults to 1):
```
fxmarty, line 231: I'm thought we aligned on not using the `*optional*`, is that not? https://huggingface.slack.com/archives/C02P0559X9S/p1669628754896019

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -130,8 +131,11 @@ def __init__(
         self.session_outputs = {output_key.name: idx for idx, output_key in enumerate(self.session.get_outputs())}
         self.session_input_names = list(self.session_inputs.keys())
         self.session_output_names = list(self.session_outputs.keys())
-        self.key_value_input_names = [key for key in self.session_input_names if "key_values" in key]
-        self.key_value_output_names = [key for key in self.session_output_names if "key_values" in key]
+        # TODO: make this less hacky.
+        self.key_value_input_names = [key for key in self.session_input_names if (".key" in key) or (".value" in key)]
```
fxmarty, line 135: did not understand it all - why is there this change?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -465,121 +484,214 @@ def _from_pretrained(
...
+                        f"Too many ONNX model files were found in {path}, specify which one to load by using the "
+                        f"{argument_name} argument."
+                    )
+            return onnx_files[0].name
+
+        if not validate_filename(decoder_file_name):
+            decoder_file_name = infer_filename(r"(.*)?decoder((?!with_past).)*?\.onnx", "decoder_file_name")
+        if not validate_filename(decoder_with_past_file_name):
+            decoder_with_past_file_name = infer_filename(
+                r"(.*)?decoder(.*)?with_past(.*)?\.onnx", "decoder_with_past_file_name", fail_if_not_found=False
```
fxmarty, line 543: Can we document somewhere in a docstring how the lookup is done.

optimum/onnxruntime/modeling_decoder.py
michaelbenayoun: I tried implementing it, but end up with circular imports!

optimum/exporters/onnx/base.py
michaelbenayoun: Done!

optimum/onnxruntime/modeling_seq2seq.py
michaelbenayoun: You are right!

optimum/modeling_base.py
michaelbenayoun: If it is not there, we do not do anything, otherwise we save it. The reason for this is because let's say I export a model with the `ORTModel` classes, I want to be able to use the save directory as a model repo end-to-end. Same thing for graph optimization / quantization. Otherwise we would need to provide a different name for the `ORTModel.from_pretrained` and the `AutoTokenizer.from_pretrained`.

optimum/onnxruntime/modeling_decoder.py
michaelbenayoun: Done!

optimum/onnxruntime/modeling_ort.py
```diff
@@ -42,11 +42,12 @@
 )
 
 import onnxruntime as ort
-from huggingface_hub import hf_hub_download
+from huggingface_hub import HfApi, HfFolder, hf_hub_download
 
 from ..exporters import TasksManager
 from ..exporters.onnx import export
 from ..modeling_base import FROM_PRETRAINED_START_DOCSTRING, OptimizedModel
+from ..utils.save_utils import maybe_load_preprocessors, maybe_save_tokenizer_or_processor_or_feature_extractor
```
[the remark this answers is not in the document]
michaelbenayoun: Because it might load / save or not. If nothing is found, it won't do anything.

optimum/onnxruntime/modeling_seq2seq.py
michaelbenayoun: It is in `ORTModel`, now `ORTModelForConditionalGeneration` uses the base class initialiazer, so we do not need that anymore here.

optimum/onnxruntime/quantization.py
michaelbenayoun: Yes, I will make a PR for this later (to handle everything)

optimum/utils/save_utils.py
```diff
@@ -0,0 +1,66 @@
...
+        pass
+
+    try:
+        preprocessors.append(AutoFeatureExtractor.from_pretrained(src_name_or_path, subfolder=subfolder))
+    except Exception:
+        pass
+    return preprocessors
+
+
+def maybe_save_tokenizer_or_processor_or_feature_extractor(
```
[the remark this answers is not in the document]
michaelbenayoun: Done (partially)

optimum/onnxruntime/modeling_decoder.py
michaelbenayoun: Because the test was not working for the past key values input / output names

optimum/onnxruntime/quantization.py
fxmarty: all good

optimum/onnxruntime/modeling_decoder.py
fxmarty: why have it as a function in the first time then?

michaelbenayoun: We use it multiple time inside this function, but never outside. We could make that a method, but that is not needed specifically. While it is true that there is redundancy between this and `modeling_seq2seq.py`, I wanted to keep the function simple (making it a method woulrd require more arguments and so on).

fxmarty:
https://github.com/huggingface/optimum/pull/515#discussion_r1039625535 => I still don't understand why we introduce `self._preprocessors` that does not exist in transformers.

https://github.com/huggingface/optimum/pull/515#discussion_r1039638141 => could not find a description

https://github.com/huggingface/optimum/pull/515#discussion_r1039638417 => same, I disagree it is not needed, it would help readability

https://github.com/huggingface/optimum/pull/515#discussion_r1039659032 => could not find a documentation

michaelbenayoun:
- The idea behind `self._preprocessors` is to be able to save all the model's preprocessors when possible, this will make using the saved model much easier.

**Example**:  You load `bert-base-uncased` and quantize it. Now you have a `bert_quantized` directory. 

With `self._preprocessors` we are able to save the preprocessors during quantization, meaning that you ca do:
```python
model_name = "bert_quantized"
tok = AutoTokenizer.from_pretrained(model_name)
quantized_model = ORTModel.from_pretrained(model_name)
```
instead of:

```python
tok = AutoTokenizer.from_pretrained("bert-base-uncased")
quantized_model = ORTModel.from_pretrained("bert_quantized")
```

This makes things a lot easier in terms of usage, especially if we think about production.

- I am not really in favor of documenting inner functions, especially this one, it does what the name implies: it infers the filename of the onnx file. But I will document it if you think this is needed here. 
- Same comment, `validate_file_exists` is quite explicit it terms of naming, it checks that the file called "filename" exists. I will put a comment if you think this is needed too.
- You mean that you want to say in the docstring that we use a regex to find potential files?

fxmarty:
@michaelbenayoun What is the difference between the two code samples?

PR #528, opened 2022-11-30
Add Swin support in exporters.onnx

fxmarty:
# What does this PR do?

As per title. It's a model where a robust test during the export is needed.

In a next PR we should make the tests for exporters faster on PRs and main.

### DIFF
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -610,6 +610,10 @@ class YolosOnnxConfig(ViTOnnxConfig):
     DEFAULT_ONNX_OPSET = 12
 
 
+class SwinOnnxConfig(ViTOnnxConfig):
+    DEFAULT_ONNX_OPSET = 12
+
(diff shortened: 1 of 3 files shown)


optimum/exporters/onnx/model_configs.py
```diff
@@ -610,6 +610,10 @@ class YolosOnnxConfig(ViTOnnxConfig):
     DEFAULT_ONNX_OPSET = 12
 
 
+class SwinOnnxConfig(ViTOnnxConfig):
+    DEFAULT_ONNX_OPSET = 12
```
michaelbenayoun, line 614: Is it needed for sure? Just checking

michaelbenayoun (approved):
LGTM!

optimum/exporters/onnx/model_configs.py
fxmarty: Right, goes well with opset = 11 as well!

PR #534, opened 2022-12-02
Add CLIP BetterTransformer

fxmarty:
The support is only for the vision model, as the text model uses causal attention that is likely not supported by PyTorch Better Transformer.

### DIFF
--- a/optimum/bettertransformer/models/encoder_models.py
+++ b/optimum/bettertransformer/models/encoder_models.py
@@ -1108,6 +1108,8 @@ class CLIPLayerBetterTransformer(BetterTransformerBaseLayer):
         r"""
         A simple conversion of the CLIPEncoderLayer to its `BetterTransformer` implementation.
 
+        **The implementation is valid only for the vision model, that does not use `causal_attention_mask`.**
+
         Args:
             layer (`torch.nn.Module`):
                 The original `CLIPEncoderLayer` where the weights needs to be retrieved.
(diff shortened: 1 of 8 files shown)


optimum/bettertransformer/transformation.py
```diff
@@ -152,7 +152,7 @@ def transform(
         else:
             load_accelerate = False
 
-        hf_config = model.config
+        hf_config = cfg
```
younesbelkada, line 155: Why you need this arg? 🤔

fxmarty: it's draft

optimum/bettertransformer/models/encoder_models.py
```diff
@@ -1095,3 +1101,108 @@ def forward(self, hidden_states, attention_mask, position_bias=None, *_, **__):
...
+        self.validate_bettertransformer()
+
+    def forward(self, hidden_states, attention_mask, *_, **__):
+        r"""
+        This is just a wrapper around the forward function proposed in:
+        https://github.com/huggingface/transformers/pull/19553
+        """
+        super().forward_checker()
+
+        assert attention_mask is None  # we expect attention_mask to be None in the vision model
```
younesbelkada, line 1177:
```suggestion
        # we expect attention_mask to be None in the vision model 
        if attention_mask is not None: 
            raise ValueError("Please do not use attention masks when using `BetterTransformer` converted models") 
```

fxmarty:
Tests pass locally on GPU, so I'll merge. I don't think there is any workflow for GPU tests?

PR #535, opened 2022-12-02
Fold sections by default in the documentation's side-bar

regisss:
# What does this PR do?

In the side-bar of the documentation of Optimum, all sections are currently unfolded (see [here](https://huggingface.co/docs/optimum/index)). This looks overwhelming and it would be clearer if main sections are folded.

With this PR, only the "Overview" section is unfolded so that the "Installation" and "Quick tour" subsections are visible and easily accessible.

This will need to be done for the doc of subpackages too.

## Before submitting

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -11,12 +11,14 @@
     - local: concept_guides/quantization
       title: Quantization
     title: Conceptual guides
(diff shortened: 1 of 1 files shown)


echarlaix (approved):
It looks great and makes the documentation much clearer, thanks @regisss !

PR #544, opened 2022-12-04
enable ORTModelForSpeechSeq2Seq optimization

flozi00:
# What does this PR do?

Fixes # (issue)

## Before submitting


flozi00:
@michaelbenayoun @echarlaix

optimum/onnxruntime/optimization.py
```diff
@@ -74,7 +74,7 @@ def from_pretrained(
         onnx_model_path = []
         config = None
         if isinstance(model_or_path, ORTModel):
-            if isinstance(model_or_path, ORTModelForSeq2SeqLM):
+            if isinstance(model_or_path, ORTModelForSeq2SeqLM) or isinstance(model_or_path, ORTModelForSpeechSeq2Seq):
```
michaelbenayoun, line 77:
```suggestion
            if isinstance(model_or_path, ORTModelForConditionalGeneration):
```

michaelbenayoun: I think this would be the most general case here, @echarlaix wdyt?

michaelbenayoun:
There is a bug because you did not import `ORTModelForConditionalGeneration` in `optimization.py`.
You need to add this and also please remove unused imports at the same time! (if there is any)

PR #546, opened 2022-12-05
Use exporters in ORTTrainer

mht-sharma:
# What does this PR do?

ORTTrainer uses optimum.exporters.onnx

Changes in the PR
1. Adds `OnnxConfigWithLoss` class in exporters to integrate with `ORTTrainer`
2. Onnx export with `Tensorflow` now creates dynamic shaped inputs in onnx model
3. Removed deprecated code

Fixes # (issue)

## Before submitting


optimum/exporters/onnx/base.py
```diff
@@ -558,3 +558,104 @@ def get_decoder_onnx_config(
             f"{config.model_type} decoder export is not supported yet. ",
             f"If you want to support {config.model_type} please propose a PR or open up an issue.",
         )
+
+
+class OnnxConfigWithLoss(OnnxConfig, ABC):
```
mht-sharma, line 563: Created only a single wrapper class to for training. Earlier there were 3 configs for `OnnxConfigWithLoss`, `OnnxConfigWithPastAndLoss`, `OnnxSeq2SeqConfigWithPastAndLoss`. The single config should be able to handle all the above use-cases. 

@JingyaHuang  to comment if we need them separately.

JingyaHuang: Yeah one wrapper in great!

JingyaHuang: Actually, with_loss configs were temporary, it is necessary for the `evaluate` in the trainer, and by the end of the day we should have a more generic wrapper to allow users to export all possible outputs as in transformers. This is something that we can tackle in upcoming PRs.

JingyaHuang: @michaelbenayoun @fxmarty

optimum/exporters/onnx/base.py
```diff
@@ -558,3 +558,93 @@ def get_decoder_onnx_config(
...
+    @property
+    def outputs(self) -> Mapping[str, Mapping[int, str]]:
+        common_outputs = self._onnx_config.outputs
+        extra_outputs = self._tasks_to_extra_outputs["default"]
+        common_outputs.update(extra_outputs)
+        for key in reversed(extra_outputs.keys()):
+            common_outputs.move_to_end(key, last=False)
+        return copy.deepcopy(common_outputs)
+
+    def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
```
michaelbenayoun, line 614: Why is it needed?

optimum/exporters/onnx/base.py
```diff
@@ -558,3 +558,93 @@ def get_decoder_onnx_config(
...
+    }
+    _tasks_to_extra_outputs = {
+        "default": OrderedDict({"loss": {}}),
+    }
+
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyTrainingLabelsInputGenerator,)
+
+    def __init__(self, config: OnnxConfig):
+        self.__dict__ = copy.deepcopy(config.__dict__)
+        self._onnx_config = config
```
michaelbenayoun, line 593: I do not think this is nice for the `OnnxConfigWithLoss` to be both an `OnnxConfig`, with the same state as `config` and to have it as an attribute that is being used in methods. If it has the same state, why should we need to store `config`?

For instance the two properties `torch_to_onnx_input_map` and `values_override` are not needed no?
Same thing for `generate_dummy_inputs_for_validation`?

Maybe I am missing something but wanted to raise the question!

optimum/exporters/onnx/convert.py
```diff
@@ -370,7 +370,9 @@ def export_tensorflow(
     output_names = list(config.outputs.keys())
 
     config.patch_ops()
-    input_signature = [tf.TensorSpec.from_tensor(tensor, name=key) for key, tensor in dummy_inputs.items()]
+    input_signature = [
+        tf.TensorSpec([None] * tensor.ndim, dtype=tensor.dtype, name=key) for key, tensor in dummy_inputs.items()
```
michaelbenayoun, line 374: It means that we do not specify the shape for any dimension?

optimum/exporters/onnx/base.py
mht-sharma: Yes you are right.  The old `OnnxConfigWithLoss` also has both the state and config copied. But I guess we don't need both. I could remove the copying of the state, this will also remove the new condition in `normalised_config.py` for deep copy.

Answering why `torch_to_onnx_input_map` and `values_override`  are needed - The `OnnxConfigWithLoss` would be a wrapper for every model config, be it Bert, bart, t5 etc. These methods could have different implementations for each model in their config, plus there could be additional methods as well. So it would be better to return the `_onnx_config` method output when they are called.

Same case is for `generate_dummy_inputs_for_validation`. This method also uses the `_onnx_config` to return the old inputs and then appends the extra inputs.

michaelbenayoun: Ok I understand now! Then yes, I think `OnnxConfigWithLoss` should not be an `OnnxConfig` and just a wrapper around it.

optimum/exporters/onnx/convert.py
mht-sharma: Yes, but not sure if it's the best way. I added this because of a failing test. Transformers [transformers:convert.py#LL279](https://github.com/huggingface/transformers/blob/main/src/transformers/onnx/convert.py#LL279) also uses the same method to export.

Could try to augment it with the `config.inputs` dynamic_axes.

optimum/onnxruntime/trainer.py
```diff
@@ -1455,24 +1455,23 @@ def _export(
                 self.model.to("cpu")
             model = unwrap_model(self.model)
 
-        _, model_onnx_config = FeaturesManager.check_supported_model_or_raise(model, feature=self.feature)
-        onnx_config = model_onnx_config(model.config)
-        opset = onnx_config.default_onnx_opset if opset is None else opset
+        model_type = model.config.model_type.replace("_", "-")
+        model_name = getattr(model, "name", None)
```
JingyaHuang, line 1459: I saw this also in ORTModels' export, why didn't we make it something inside ` TasksManager.get_exporter_config_constructor`?

optimum/onnxruntime/trainer.py
```diff
@@ -1455,24 +1455,23 @@ def _export(
...
+        onnx_config = onnx_config_constructor(model.config)
+
+        opset = onnx_config.DEFAULT_ONNX_OPSET if opset is None else opset
 
         if with_loss:
             onnx_config = wrap_onnx_config_for_loss(onnx_config)
             opset = max(opset, 12)  # Operators like `nll_loss`are added for opset>=12
 
         # transformers >= 4.21.0 is required to export with specified device
         check_min_version("4.21.0")
```
JingyaHuang, line 1473: We are not using the export in transformers anymore, you can remove the minimal version check.

optimum/onnxruntime/trainer_seq2seq.py
```diff
@@ -795,30 +791,37 @@ def _export(
...
             onnx_config_decoder_with_past = wrap_onnx_config_for_loss(onnx_config_decoder_with_past)
             opset = max(opset, 12)  # Operators like `nll_loss`are added for opset>=12
 
-        # Extract the encoder for ONNX export
-        encoder = model.get_encoder()
-        # Concatenate the decoder with the language model head for ONNX export
-        decoder_with_lm_head = _DecoderWithLMhead(model)
-
         # transformers >= 4.21.0 is required to export with specified device
         check_min_version("4.21.0")
```
JingyaHuang, line 821: Same here, no need to check anymore.

optimum/utils/input_generators.py
```diff
@@ -575,3 +571,21 @@ def generate(self, input_name: str, framework: str = "pt"):
...
+        max_value = 1 if self.task != "seq2seq-lm" else self.vocab_size
+        shape = [self.batch_size, self.sequence_length]
+        if self.task in [
+            "default",
+            "sequence-classification",
+            "multiple-choice",
+            "question-answering",
+            "image-classification",
+        ]:
+            shape = [self.batch_size]
```
JingyaHuang, line 589: I prefer to check the label shape info in the mapping `_tasks_to_extra_inputs`, otherwise we need to update both in case of new tasks.

But as it is currently in the onnx config, there would be some circular import issue. I think that we should put it somewhere else, and for future shape info of other custom inputs, we can put all that here as well.

optimum/exporters/onnx/base.py
```diff
@@ -558,3 +558,95 @@ def get_decoder_onnx_config(
...
+        common_outputs = self._onnx_config.outputs
+        extra_outputs = self._tasks_to_extra_outputs["default"]
+        common_outputs.update(extra_outputs)
+        for key in reversed(extra_outputs.keys()):
+            common_outputs.move_to_end(key, last=False)
+        return copy.deepcopy(common_outputs)
+
+    def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
+        dummy_inputs = self._onnx_config.generate_dummy_inputs(framework=framework, **kwargs)
+        input_name, _ = next(iter(self._onnx_config.inputs.items()))
```
JingyaHuang, line 618: The way of getting the batch_size here seems a little bit hacky for me, but we might target it later I guess.

optimum/exporters/onnx/base.py
```diff
@@ -558,3 +558,95 @@ def get_decoder_onnx_config(
...
+            input_was_inserted = False
+            for dummy_input_gen in dummy_inputs_generators:
+                if dummy_input_gen.supports_input(input_name):
+                    dummy_inputs[input_name] = dummy_input_gen.generate(input_name, framework=framework)
+                    input_was_inserted = True
+                    break
+            if not input_was_inserted:
+                raise RuntimeError(
+                    f'Could not generate dummy input for "{input_name}". Try adding a proper dummy input generator to the model ONNX config.'
+                )
```
JingyaHuang, line 639: These conditional statements are prepared for the case that we want to insert inputs other than labels?

optimum/exporters/onnx/base.py
```diff
@@ -558,3 +558,95 @@ def get_decoder_onnx_config(
...
+
+        return dummy_inputs
+
+    def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[str, Any]) -> Mapping[str, Any]:
```
JingyaHuang, line 643: Why do we need this function?

JingyaHuang (commented):
Thanks @mht-sharma, it looks good!

Just left a few nits and some questions about the design.

optimum/exporters/onnx/convert.py
mht-sharma: Updated, only the axes having dynamic shapes will be updated with None.

PS. Probably a separate issue, the tf export do not work when `past_key_values` are in the input.

optimum/onnxruntime/trainer.py
mht-sharma: Yes, I think we can do that. Updated with the change.  

cc: @michaelbenayoun if there is needs to be reverted

optimum/onnxruntime/trainer.py
mht-sharma: Done

optimum/exporters/onnx/base.py
mht-sharma: Yes they should be, this is the same snippet used in the `OnnxConfig` get_dymmy_inputs.

optimum/utils/input_generators.py
mht-sharma: I am not too sure about using the dynamic axes in the input generator. Dynamic axes values are quite specific to ONNX while `DummyInputGenerator` could be used for other exporters. Plus there could be other inputs with ndim > dynamic axes. WDYT?

optimum/onnxruntime/trainer.py
michaelbenayoun: I guess it is good for now (I do not have any potential issue in mind)

optimum/exporters/onnx/base.py
mht-sharma: When exporting the decoder models for onnxruntime we use `encoder_outputs` of shape `Tuple(encoder_hidden_states, None, None) `.  After export to onnx the inputs to the onnx model are just the  `encoder_hidden_states`.

Thus the inputs for the export and validation change. In these cases, we use the helper function to modify the inputs.

mht-sharma:
@michaelbenayoun @JingyaHuang rebased the PR. Let me know if it looks good and we can merge.

JingyaHuang (approved):
Looks great! Thanks for iterating, let's merge it!

optimum/exporters/onnx/base.py
michaelbenayoun: Maybe a small nit, but should we put it in another file?
This file is like for base classes, and this one seems to be ready to go, it would make more sense to put it in `model_configs.py` for instance, but even there, I think it would be out of place... So maybe a new file?

WDYT @JingyaHuang @mht-sharma ?

optimum/exporters/tasks.py
```diff
@@ -863,24 +863,39 @@ def get_model_from_task(
...
             `ExportConfigConstructor`: The `ExportConfig` constructor for the requested backend.
         """
+        model_type = getattr(model.config, "model_type", model_type)
+
+        if model_type is None:
+            raise ValueError("Model type cannot be inferred. Please provide the model_type for the model!")
+
+        model_type = model_type.replace("_", "-")
+        if model_name is None:
+            model_name = getattr(model, "name", None)
```
michaelbenayoun, line 897: To follow the same "pattern" as `model_type`:
```suggestion
        model_name = getattr(model, "name", model_name)
```

michaelbenayoun (approved):
Left small nits.
LGTM, thanks a lot @mht-sharma !

optimum/exporters/onnx/base.py
mht-sharma: Do not have a strong opinion on this.  Could move this config in a new file `training_config.py`?

PR #548, opened 2022-12-06
Adds `ASTLayer` support for `BetterTransformer`

ravenouse:
# What does this PR do?

Adds `ASTLayer` support for `BetterTransformer`

Fixes [Community contribution - BetterTransformer integration for more models! #20372
](https://github.com/huggingface/transformers/issues/20372)

## Questions:
1. I used `"MIT/ast-finetuned-audioset-10-10-0.4593` as the test model to run `pytest` but some tests failed. I think one test model, like `"hf-internal-testing/tiny-random-MBartModel"`, is needed to test the `ASTLayerBetterTransformer`.

截屏2022-12-05 下午9 59 59

2. I have one question about the `BetterTransformerBaseLayer` class. I notice that we set `self. use_gelu` as false for the default setting but lots of the supported transformer models actually use the gelu activation function, like `bert`.  Could you provide more information about it? 

Thank you so much for your effort!!

Please let me know what else I need to do!

### DIFF
--- a/optimum/bettertransformer/models/__init__.py
+++ b/optimum/bettertransformer/models/__init__.py
@@ -31,7 +31,7 @@ from .encoder_models import (
 
 class BetterTransformerManager:
     MODEL_MAPPING = {
-        "ast":("ASTLayer": ASTLayerBetterTransformer),
+        "audio-spectrogram-transformer": ("ASTLayer", ASTLayerBetterTransformer),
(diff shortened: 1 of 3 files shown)


younesbelkada:
Hi @ravenouse !
This is great! Thanks for adding this, could you please rebase with main branch so that we can merge this PR? 🙏 
Thanks!

PR #560, opened 2022-12-07
prevent optimum from re-wrapping ORTModule

prathikr:
# What does this PR do?

In the case of running optimum + accelerate's auto_batch_size, ORTModule is recursively wrapped which causes export errors the second time batch_size is halved as ORT will try to export an already exported model. This fix adds a quick instance check to prevent ORTModule being double wrapped when optimum + accelerate's auto_batch_size is used.

### DIFF
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -421,7 +421,7 @@ class ORTTrainer(Trainer):
 
         # Wrap the model with `ORTModule`
         logger.info("Wrap ORTModule for ONNX Runtime training.")
-        model = ORTModule(self.model)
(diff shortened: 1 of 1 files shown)


JingyaHuang:
Hi @prathikr,

Thanks for the PR, the double wrapping problem is also handled in #561 by unwrapping `self.model` when all wrapping is finished. Can you take a look at this PR to see if it can solve the issue here?

Thank you!

JingyaHuang:
Hi @prathikr, the rewrapping issue shall have been fixed with `unwarp_model`:
https://github.com/huggingface/optimum/blob/dbb90b4ce286b22615c767a4ac4339dfac78d7c7/optimum/onnxruntime/trainer.py#L450
If things go well from your side, can I close the pull request?

Issue #549, opened 2022-12-06
Improve quantization documentation


michaelbenayoun:
### Feature request

As per title

### Motivation

The current documentation is quite empty

### Your contribution

/

PR #565, opened 2022-12-08
Documentation on quantization

michaelbenayoun:
# What does this PR do?

- Small fix in the docstring related to quantization in ONNX Runtime
- Conceptual guide on what quantization is all about

Fixes: #549

### DIFF
--- a/docs/source/concept_guides/quantization.mdx
+++ b/docs/source/concept_guides/quantization.mdx
@@ -31,8 +31,163 @@ C = A + B
 Here the result is much bigger than the biggest representable value in `int8`, which is `127`. Hence the need for a larger
 precision data type to avoid a huge precision loss that would make the whole quantization process useless.
 
+## Quantization
 
-### How do machines represent numbers?
+The two most common quantization cases are `float32 -> float16` and `float32 -> int8`.
+
+### Quantization to `float16`
+
+Performing quantization to go from `float32` to `float16` is quite straightforward since both data types follow the same
+representation scheme. The questions to ask yourself when quantizing an operation to `float16` are:
+
+- Does my operation have a `float16` implementation?
+- Does my hardware suport `float16`? For instance, Intel CPUs [have been supporting `float16` as a storage type, but
+computation is done after converting to `float32`](https://scicomp.stackexchange.com/a/35193). Full support will come
+in Cooper Lake and Sapphire Rapids.
+- Is my operation sensitive to lower precision?
+For instance the value of epsilon in `LayerNorm` is usually very small (~ `1e-12`), but the smallest representable value in
+`float16` is ~ `6e-5`, this can cause `NaN` issues.  The same applies for big values.
+
+### Quantization to `int8`
+
+Performing quantization to go from `float32` to `int8` is more tricky. Only 256 values can be represented in `int8`,
+while `float32` can represent a very wide range of values. The idea is to find the best way to project our range `[a, b]`
+of `float32` values  to the `int8` space.
+
+Let's consider a float `x` in `[a, b]`, then we can write the following quantization scheme, also called the *affine
+quantization scheme*:
+
+```
+x = S * (x_q - Z)
+```
+
+where:
+
+- `x_q` is the quantized `int8` value associated to `x`
+- `S` and `Z` are the quantization parameters
+  - `S` is the scale, and is a positive `float32`
+  - `Z` is called the zero-point, it is the `int8` value corresponding to the value `0` in the `float32` realm. This is
+  important to be able to represent exactly the value `0` because it is used everywhere throughout machine learning
+  models.
+
+
+The quantized value `x_q` of `x` in `[a, b]` can be computed as follows:
+
+```
+x_q = round(x/S + Z)
+```
+
+And `float32` values outside of the `[a, b]` range are clipped to the closest representable value, so for any
+floating-point number `x`:
+
+```
+x_q = clip(round(x/S + Z), round(a/S + Z), round(b/S + Z))
+
+```
+
+<Tip>
+
+Usually `round(a/S + Z)` corresponds to the smallest representable value in the considered data type, and `round(b/S + Z)`
+to the biggest one. But this can vary, for instance when using a *symmetric quantization scheme* as you will see in the next
+paragraph.
+
+</Tip>
+
+### Symmetric and affine quantization schemes
+
+The equation above is called the *affine quantization sheme* because the mapping from `[a, b]` to `int8` is an affine one.
+
+A common special case of this scheme is the *symmetric quantization scheme*, where we consider a symmetric range of float values `[-a, a]`.
+In this case the integer space is usally `[-127, 127]`, meaning that the `-128` is opted out of the regular `[-128, 127]` signed `int8` range.
+The reason being that having both ranges symmetric allows to have `Z = 0`. While one value out of the 256 representable
+values is lost, it can provide a speedup since a lot of addition operations can be skipped.
+
+**Note**: To learn how the quantization parameters `S` and `Z` are computed, you can read the
+[Quantization and Training of Neural Networks for Efficient Integer-Arithmetic-Only Inference](https://arxiv.org/abs/1712.05877)
+paper, or [Lei Mao's blog post](https://leimao.github.io/article/Neural-Networks-Quantization) on the subject.
+
+
+### Per-tensor and per-channel quantization
+
+Depending on the accuracy / latency trade-off you are targetting you can play with the granularity of the quantization parameters:
+
+- Quantization parameters can be computed on a *per-tensor* basis, meaning that one pair of `(S, Z)` will be used per
+tensor.
+- Quantization parameters can be computed on a *per-channel* basis, meaning that it is possible to store a pair of
+`(S, Z)` per element along one of the dimensions of a tensor. For example for a tensor of shape `[N, C, H, W]`, having
+*per-channel* quantization parameters for the second dimension would result in having `C` pairs of `(S, Z)`. While this
+can give a better accuracy, it requires more memory.
+
+### Calibration
+
+The section above described how quantization from `float32` to `int8` works, but one question
+remains: how is the `[a, b]` range of `float32` values determined? That is where calibration comes in to play.
+
+Calibration is the step during quantization where the `float32` ranges are computed. For weights it is quite easy since
+the actual range is know at *quantization-time*. But it is less clear for activations, and different approaches exist:
+
+1. Post training **dynamic quantization**: the range for each activation is computed on the fly at *runtime*. While this
+gives great results without too much work, it can be a bit slower than static quantization because of the overhead
+introduced by computing the range each time.
+It is also not an option on certain hardware.
+2. Post training **static quantization**: the range for each activation is computed at in advance at *quantization-time*,
+typically by passing representative data through the model and recording the activation values. In practice, the steps are:
+  1. Observers are put on activations to record their values.
+  2. A certain number of forward passes on a calibration dataset is done (around `200` examples is enough).
+  3. The ranges for each computation are computed according to some *calibration technique*.
+3. **Quantization aware training**: the range for each activation is computed at *training-time*, following the same idea
+than post training static quantization. But "fake quantize" operators are used instead of observers: they record
+values just as observers do, but they also simulate the error induced by quantization to let the model adapt to it.
+
+
+For both post training static quantization and quantization aware training, it is necessary to define calibration
+techniques, the most common are:
+
+- Min-max: the computed range is `[min observed value, max observed value]`, this works well with weights.
+- Moving average min-max: the computed range is `[moving average min observed value, moving average max observed value]`,
+thiw works well with activations.
+- Histogram: records a histogram of values along with min and max values, then chooses according to some criterion.
+  - Entropy: the range is computed as the one minimizing the error between the full-precision and the quantized data.
+  - Mean Square Error: the range is computed as the one minimizing the mean square error between the full-precision and
+  the quantized data.
+  - Percentile: the range is computed using a given percentile value `p` on the observed values. The idea is to try to have
+  `p%` of the observed values in the computed range. While this is possible when doing affine quantization, it is not always
+  possible to exactly match that when doing symmetric quantization. You can check [how it is done in ONNX
+  Runtime](https://github.com/microsoft/onnxruntime/blob/2cb12caf9317f1ded37f6db125cb03ba99320c40/onnxruntime/python/tools/quantization/calibrate.py#L698)
+  for more details.
+
+
+### Pratical steps to follow to quantize a model to `int8`
+
+To effectively quantize a model to `int8`, the steps to follow are:
+
+1. Choose which operators to quantize. Good operators to quantize are the one dominating it terms of computation time,
+for instance linear projections and matrix multiplications.
+2. Try post-training dynamic quantization, if it is fast enough stop here, otherwise continue to step 3.
+3. Try post-training static quantization which can be faster than dynamic quantization but often with a drop in terms of
+accuracy. Apply observers to your models in places where you want to quantize.
+implies defining which quantization scheme to use.
+4. Perform calibration.
+5. Convert the model to its quantized form: the observers are removed and the `float32` operators are converted to their `int8`
+coutnerparts.
+6. Evaluate the quantized model: is the accuracy good enough? If yes, stop here, otherwise start again at step 3 but
+with quantization aware training this time.
+
+## Supported tools to perform quantization in 🤗 Optimum
+
+🤗 Optimum provides APIs to perform quantization using different tools for different targets:
+
+- The `optimum.onnxruntime` package allows to
+[quantize and run ONNX models](https://huggingface.co/docs/optimum/onnxruntime/usage_guides/quantization) using the
+ONNX Runtime tool.
+- The `optimum.intel` package enables to [quantize](https://huggingface.co/docs/optimum/intel/optimization_inc) 🤗 Transformers
+models while respecting accuracy and latency constraints.
+- The `optimum.fx` package provides wrappers around the
+[PyTorch quantization functions](https://pytorch.org/docs/stable/quantization-support.html#torch-quantization-quantize-fx)
+to allow graph-mode quantization of 🤗 Transformers models in PyTorch. This is a lower-level API compared to the two
+mentioned above, giving more flexibility, but requiring more work on your end.
+
+## Going further: How do machines represent numbers?
 
 <Tip>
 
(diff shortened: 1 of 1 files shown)


docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,251 @@
...
+For both post training static quantization and quantization aware training, it is necessary to define calibration
+techniques, the most common are:
+
+- Min-max: the computed range is `[min observed value, max observed value]`, this works well with weights.
+- Moving average min-max: the computed range is `[moving average min observed value, moving average max observed value]`,
+thiw works well with activations.
+- Histogram: records a histogram of values along with min and max values, then chooses according to some criterion.
+  - Entropy: the range is computed as the one minimizing the error between the full-precision and the quantized data.
+  - Mean Square Error: the range is computed as the one minimizing the mean square error between the full-precision and
+  the quantized data.
```
regisss, line 204: Add percentile maybe?

regisss (commented):
Great initiative @michaelbenayoun!!

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,251 @@
...
+data type in question. For examples, let's consider:
+
+```
+C = A x B
+```
+
+where `A` is a `m x n` matrix of `int8` values, and `B` is a `n x p` matrix of `int8` values.
+
+Then for `i` in `[0, m - 1]` and `j` in `[0, p - 1]`, `C[i, j]` is an `int32` because it the sum of products of
+`int8` values.
```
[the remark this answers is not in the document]
michaelbenayoun: Done, thanks for the great suggestion!

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,251 @@
...
+
+### How do machines represent numbers?
```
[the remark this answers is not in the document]
michaelbenayoun: I assumed people would know those data types, but might not know how they work underneath, but I get your point, where should we put it in your opinion?

docs/source/concept_guides/quantization.mdx
michaelbenayoun: Done, can you validate that it is correct please?

regisss: Actually, according to what I see [here](https://github.com/microsoft/onnxruntime/blob/2cb12caf9317f1ded37f6db125cb03ba99320c40/onnxruntime/python/tools/quantization/calibrate.py#L706) In ORT's source code, it seems that it's rather `[p/2% of the observed values are bigger than this one, p/2% of the observed value are smaller than this one]`.

michaelbenayoun: It depends on whether the scheme is symmetric or not, from what I could understand:

- When symmetric, only one value can be chosen since the range needs to be `[-a, a]` in the end. So `a` is the value so that p% of the people are smaller.
- When asymmetric, the idea is to have a range when p% of the people fall into, so we cut both ends by the p/2% of the people that do not fall.

regisss: Yeah the symmetric case is a bit different but I didn't want to make things more complicated :grin: 
Yep I understand it the same way as you do. Your explanation sounds good to me, no need to make the documentation too complicated here :+1:

docs/source/onnxruntime/usage_guides/quantization.mdx
```diff
@@ -126,9 +147,13 @@ The `ORTQuantizer` class can be used to statically quantize your ONNX model. Bel
 
 ## Quantize Seq2Seq models
 
-The `ORTQuantizer` currently doesn't support multi-file models, like `ORTModelForSeq2SeqLM`. If you want to quantize a Seq2Seq model, you have to quantize each model's component individually using the `ORTQuantizer` class. Currently, only dynamic quantization is supported for Seq2Seq model.
+The [`~optimum.onnxruntime.ORTQuantizer`] class currently doesn't support multi-file models, like
+[`~optimum.onnxruntime.ORTModelForSeq2SeqLM`]. If you want to quantize a Seq2Seq model, you have to quantize each
+model's component individually.
+
+**Note**: Currently, only dynamic quantization is supported for Seq2Seq model.
```
regisss, line 154:
```suggestion
<Tip warning={true}>

Currently, only dynamic quantization is supported for Seq2Seq models.

</Tip>
```

regisss (approved):
Awesome work @michaelbenayoun :rocket: :rocket: :rocket: 
I left one more nit.

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+
+```
+x = sign x mantissa x (2^exponent)
+```
+
+### Quantization
+
+The two most common quantization cases are `float32 -> float16` and `float32 -> int8`.
+
+#### Quantization to `float16`
```
fxmarty, line 102: As float8 becomes more common (e.g. support in Trainium, Inferentia), I add it here, feel free to ignore if you feel it's irrelevant.

```suggestion
#### Quantization to `float16`, `float8`
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+
+Performing quantization to go from `float32` to `float16` is quite straightforward since both data types follow the same
```
fxmarty, line 104:
```suggestion
Performing quantization to go from `float32` to lower precision floating point formats (as `float16` and `float8`) is quite straightforward since both data types follow the same
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+
+- Does my operation have a `float16` implementation?
+- Does my hardware suport `float16`? For instance, Intel CPUs [have been supporting `float16` as a storage type, but
+computation is done after converting to `float32`](https://scicomp.stackexchange.com/a/35193). Full support will come
+in Cooper Lake and Sapphire Rapids.
+- Is my operation sensitive to lower precision?
+For instance the value of epsilon in `LayerNorm` is usually very small (~ `1e-12`), but the smallest representable value in
+`float16` is ~ `6e-5`, this can cause `NaN` issues.  The same applies for big values.
+
+#### Quantization to `int8`
```
fxmarty, line 115:
```suggestion
#### Quantization to integer format (typically `int8`)
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+
+```
+x_q = round(x/S + Z)
+```
+
+And `float32` values outside of the `[a, b]` range are clipped to the closest representable value, so for any
+floating-point number `x`:
+
+```
+x_q = clip(round(x/S + Z), round(a/S + Z), round(b/S + Z))
```
fxmarty, line 148:
```suggestion
x_q = clamp(x_q, min_representable_value, max_representable_value)
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+
```
fxmarty, line 149:
```suggestion
For example, in the case of int8 quantization, `min_representable_value = -127`, `max_representable_value = 128.

```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+```
+
+### Quantization
+
+The two most common quantization cases are `float32 -> float16` and `float32 -> int8`.
+
+#### Quantization to `float16`
+
+Performing quantization to go from `float32` to `float16` is quite straightforward since both data types follow the same
+representation scheme. The questions to ask yourself when quantizing an operation to `float16` are:
```
fxmarty, line 105:
```suggestion
representation scheme. The questions to ask yourself when quantizing an operation to a lower precision float format:
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+floating-point number `x`:
+
+```
+x_q = clip(round(x/S + Z), round(a/S + Z), round(b/S + Z))
+
+```
+
+This is called the *affine quantization sheme* because the mapping from `[a, b]` to `int8` is an affine one.
+
+A common special case of this scheme is the *scale quantization scheme*, where we consider a symmetric range of float values `[-a, a]`.
```
fxmarty, line 154: I never read this term. To me what this paragraph describes is symmetric vs asymmetric no?

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+In this case the integer space is usally `[-127, 127]`, meaning that the `-128` is opted out of the regular `[-128, 127]` signed `int8` range.
+The reason being that having both ranges symmetric allows to have `Z = 0`. While one value out of the 256 representable
+values is lost, it can provide a speedup since a lot of addition operations can be skipped.
+
+**Note**: To learn how the quantization parameters `S` and `Z` are computed, you can read the
+[Quantization and Training of Neural Networks for Efficient Integer-Arithmetic-Only Inference](https://arxiv.org/abs/1712.05877)
+paper, or [Lei Mao's blog post](https://leimao.github.io/article/Neural-Networks-Quantization) on the subject.
+
+
+#### Number of quantization parameters
```
fxmarty, line 164:
```suggestion
### Per-tensor, per-channel, and group-wise quantization
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+Depending on the accuracy / latency trade-off you are targetting you can play with the granularity of the quantization parameters:
+
+- Quantization parameters can be computed on a *per-tensor* basis, meaning that one pair of `(S, Z)` will be used per
+tensor.
+- Quantization parameters can be computed on a *per-channel* basis, meaning that it is possible to store a pair of
+`(S, Z)` per element along one of the dimensions of a tensor. For example for a tensor of shape `[N, C, H, W]`, having
+*per-channel* quantization parameters for the second dimension would result in having `C` pairs of `(S, Z)`. While this
+can give a better accuracy, it requires more memory.
+
+#### Calibration
```
fxmarty, line 175:
```suggestion
#### Dynamic and static quantization
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+
+The section above described how quantization from `float32` to `int8` works, but one question
+remains: how is the `[a, b]` range of `float32` values determined? That is where calibration comes in to play.
+
+Calibration is the step during quantization where the `float32` ranges are computed. For weights it is quite easy since
+the actual range is know at *quantization-time*. But it is less clear for activations, and different approaches exist:
+
+1. Post training dynamic quantization: the range for each activation is computed at *runtime*. While this gives great
```
fxmarty, line 183:
```suggestion
1. Dynamic quantization: the range for each activation is computed at *runtime*. While this gives great
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+*per-channel* quantization parameters for the second dimension would result in having `C` pairs of `(S, Z)`. While this
+can give a better accuracy, it requires more memory.
+
+#### Calibration
+
+The section above described how quantization from `float32` to `int8` works, but one question
+remains: how is the `[a, b]` range of `float32` values determined? That is where calibration comes in to play.
+
+Calibration is the step during quantization where the `float32` ranges are computed. For weights it is quite easy since
+the actual range is know at *quantization-time*. But it is less clear for activations, and different approaches exist:
```
fxmarty, line 181:
```suggestion
the actual range can be known and computed in advance before runtime. But it is less clear for activations, and different approaches exist:
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+
+1. Post training dynamic quantization: the range for each activation is computed at *runtime*. While this gives great
+results without too much work, it can be a bit slower than static quantization because the range is computed each time.
+It is also not an option on certain hardware.
+2. Post training static quantization: the range for each activation is computed at *quantization-time*.
```
fxmarty, line 186:
```suggestion
2. Static quantization: the range for each activation is computed at *quantization-time*.
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+#### Calibration
+
+The section above described how quantization from `float32` to `int8` works, but one question
+remains: how is the `[a, b]` range of `float32` values determined? That is where calibration comes in to play.
+
+Calibration is the step during quantization where the `float32` ranges are computed. For weights it is quite easy since
+the actual range is know at *quantization-time*. But it is less clear for activations, and different approaches exist:
+
+1. Post training dynamic quantization: the range for each activation is computed at *runtime*. While this gives great
+results without too much work, it can be a bit slower than static quantization because the range is computed each time.
```
fxmarty, line 184:
```suggestion
results without too much work, it can be a bit slower than static quantization because over the overhead introduced of computing the range each time.
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+  - 5 bits in `float16`
+  - 8 bits in `bfloat16`
+  - 8 bits in `float32`
+  - 11 bits in `float64`
+2. The mantissa
+  - 11 bits in `float16` (10 explictly stored)
+  - 8 bits in `bfloat16` (7 explicitly stored)
+  - 24 bits in `float32` (23 explictly stored)
+  - 53 bits in `float64` (52 explictly stored)
+
```
JingyaHuang, line 88: I feel that this part could be clearer if we use directly an illustration.

JingyaHuang: Not sure if we shall add fp8 as there are two formats could be too complicated here.

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+Calibration is the step during quantization where the `float32` ranges are computed. For weights it is quite easy since
+the actual range is know at *quantization-time*. But it is less clear for activations, and different approaches exist:
+
+1. Post training dynamic quantization: the range for each activation is computed at *runtime*. While this gives great
+results without too much work, it can be a bit slower than static quantization because the range is computed each time.
+It is also not an option on certain hardware.
+2. Post training static quantization: the range for each activation is computed at *quantization-time*.
+  - Observers are put on activations to record their values.
+  - A certain number of forward passes on a calibration dataset is done (around `200` examples is enough).
+  - The ranges for each computation are computed according to some *calibration technique*.
```
fxmarty, line 189:
```suggestion
    * Observers are put on activations to record their values.
    * A certain number of forward passes on a calibration dataset is done (around `200` examples is enough).
    * The ranges for each computation are computed according to some *calibration technique*.
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+- Histogram: records a histogram of values along with min and max values, then chooses according to some criterion.
+  - Entropy: the range is computed as the one minimizing the error between the full-precision and the quantized data.
+  - Mean Square Error: the range is computed as the one minimizing the mean square error between the full-precision and
+  the quantized data.
+  - Percentile: the range is computed using a given percentile value `p` on the observed values. The idea is to try to have
+  `p%` of the observed values in the computed range. While this is possible when doing affine quantization, it is not always
+  possible to exactly match that when doing symmetric quantization. You can check [how it is done in ONNX
+  Runtime](https://github.com/microsoft/onnxruntime/blob/2cb12caf9317f1ded37f6db125cb03ba99320c40/onnxruntime/python/tools/quantization/calibrate.py#L698)
+  for more details.
+
```
fxmarty, line 210:
```suggestion

#### Quantization-Aware Training (QAT) and Post-Training Quantization (PTQ)

There are two main strategies when it comes to *when to perform quantization*:

* **Post-training quantization:** No retraining is involved. The original trained full precision model is used, and quantization is applied on top. In both dynamic and static quantization, the quantization parameters for the weights are pre-computed before inference, and weights can be stored as `int8`. Activations are quantized on the fly during inference in the case of dynamic quantization, while static quantization precomputes the quantization ranges using a calibration dataset.
* **Quantization-aware training:**  The range for weights and activations are computed at *training-time*, following the same idea than post training static quantization. But "fake quantize" operators are used instead of observers: they record alues just as observers do, but they also **simulate the error induced by quantization during training** to let the model adapt to it.
```

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+### Quantization
+
+The two most common quantization cases are `float32 -> float16` and `float32 -> int8`.
+
+#### Quantization to `float16`
+
+Performing quantization to go from `float32` to `float16` is quite straightforward since both data types follow the same
+representation scheme. The questions to ask yourself when quantizing an operation to `float16` are:
+
+- Does my operation have a `float16` implementation?
```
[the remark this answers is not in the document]
JingyaHuang: I feel like float8 is not quite relevant for Optimum, it is not supported by ONNX Runtime, does Gaudi support it @regisss ?
(Maybe add fp87 as it is a general guide and point out that it is really limited for device like Hopper, Trainium...)

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+  - Percentile: the range is computed using a given percentile value `p` on the observed values. The idea is to try to have
+  `p%` of the observed values in the computed range. While this is possible when doing affine quantization, it is not always
+  possible to exactly match that when doing symmetric quantization. You can check [how it is done in ONNX
+  Runtime](https://github.com/microsoft/onnxruntime/blob/2cb12caf9317f1ded37f6db125cb03ba99320c40/onnxruntime/python/tools/quantization/calibrate.py#L698)
+  for more details.
+
+
+#### Pratical steps to follow to quantize a model to `int8`
+
+To effectively quantize a model to `int8`, the steps to follow are:
```
fxmarty, line 214: probably the tradeoff speed / prediction quality should be highlighted here.

image

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+
+1. Choose which operators to quantize. Good operators to quantize are the one dominating it terms of computation time,
+for instance linear projections and matrix multiplications.
+2. Try dynamic quantization, if it is fast enough stop here, otherwise continue to step 3.
+3. Apply observers to your models in places where you want to quantize. It implies defining which quantization scheme to use.
```
fxmarty, line 219:
```suggestion
3. In 3., 4., 5., try post-training static quantization that can be faster than dynamic quantization, but is harder to maintain prediction quality. Apply observers to your models in places where you want to quantize. It implies defining which quantization scheme to use.
```

fxmarty (approved):
Awesome! I left a few comments, feel free to address if you want (especially how to order and name sections, the rest of my comments are minor)

JingyaHuang (approved):
LGTM, thanks for adding this guide!!! Just left a few minor comments.

docs/source/concept_guides/quantization.mdx
```diff
@@ -1,5 +1,256 @@
...
+  - Observers are put on activations to record their values.
+  - A certain number of forward passes on a calibration dataset is done (around `200` examples is enough).
+  - The ranges for each computation are computed according to some *calibration technique*.
+3. Quantization aware training: the range for each activation is computed at *training-time*, following the same idea
+than post training static quantization. But "fake quantize" operators are used instead of observers: they record
+values just as observers do, but they also simulate the error induced by quantization to let the model adapt to it.
+
+
+For both post training static quantization and quantization aware training, it is necessary to define calibration
+techniques, the most common are:
```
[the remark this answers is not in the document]
michaelbenayoun: Why not keep this?

PR #570, opened 2022-12-09
Add Stable Diffusion ONNX export

echarlaix:
In this PR we add the ONNX export of Stable Diffusion models, by exporting the U-Net, text encoder and VAE components.


optimum/exporters/onnx/__main__.py
```diff
@@ -87,6 +96,33 @@ def main():
...
+    # TODO : infer stable-diffusion when auto
+    elif task == "stable-diffusion":
+        pipeline = StableDiffusionPipeline.from_pretrained(args.model)
+        output_names = ["text_encoder/model.onnx", "unet/model.onnx", "vae_decoder/model.onnx"]
+        onnx_inputs, onnx_outputs = export_models(
+            model=pipeline,
+            onnx_config=None,
+            opset=args.opset or 14,
+            output_dir=args.output.parent,
+            fn_get_models_from_config=get_stable_diffusion_models_for_export,
```
echarlaix, line 108: I

optimum/utils/input_generators.py
```diff
@@ -575,3 +575,72 @@ def generate(self, input_name: str, framework: str = "pt"):
                 max_value=1,
                 framework=framework,
             )
+
+
+class DummyStableDiffusionInputGenerator(DummyInputGenerator):
```
michaelbenayoun, line 580: I think this should be broken down in parts for the proper modalities.

- `input_ids` are handled already somewhere else.
- `encoder_hidden_states` are handled in `DummySeq2SeqDecoderTextInputGenerator` as `encoder_outputs`
- `sample` and `latent_sample` could come from a `DummyNoiseInputGenerator` or something of the likes
- I do not know where `timestep` should be handled because I do not know its purpose, but we could do something similar.

optimum/utils/normalized_config.py
```diff
@@ -31,8 +33,8 @@ class NormalizedConfig:
             The config to normalize.
     """
 
-    def __init__(self, config: "PretrainedConfig", allow_new: bool = False, **kwargs):
-        self.config = config
+    def __init__(self, config: Union["PretrainedConfig", Dict], allow_new: bool = False, **kwargs):
```
michaelbenayoun, line 36:
```suggestion
    def __init__(self, config: Union[PretrainedConfig, Dict], allow_new: bool = False, **kwargs):
```

optimum/utils/normalized_config.py
```diff
@@ -15,7 +15,9 @@
 """Normalization configuration classes."""
 
 import functools
-from typing import TYPE_CHECKING, Callable, Type
+from typing import TYPE_CHECKING, Callable, Dict, Type, Union
+
+from transformers import PretrainedConfig
 
 
 if TYPE_CHECKING:
```
michaelbenayoun, line 23: This should be removed if we import `PretrainedConfig` above

optimum/exporters/onnx/convert.py
```diff
@@ -22,14 +22,12 @@
 import numpy as np
 from transformers.utils import is_tf_available, is_torch_available
 
+from optimum.utils import is_diffusers_available
```
fxmarty, line 25:
```suggestion
from ...utils import is_diffusers_available
```

optimum/exporters/onnx/convert.py
```diff
@@ -22,14 +22,12 @@
...
+
 from ...utils import logging
```
fxmarty, line 27:
```suggestion
from ...utils import is_diffusers_available, logging
```

optimum/exporters/onnx/convert.py
```diff
@@ -67,55 +68,44 @@ def check_dummy_inputs_are_allowed(
 
 
 def validate_models_outputs(
-    onnx_config: OnnxConfig,
-    reference_model: Union["PreTrainedModel", "TFPreTrainedModel"],
+    models: Dict[str, Tuple[Union["PreTrainedModel", "ModelMixin"], "OnnxConfig"]],
```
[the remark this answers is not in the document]
fxmarty: We always do `models_for_validation = fn_get_models_from_config(reference_model, onnx_config)` so it could make sense to remove the arg

optimum/exporters/onnx/base.py
```diff
@@ -338,6 +338,17 @@ def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[s
         """
         return reference_model_inputs
 
+    def output_names_for_validation(self, reference_output_names: List[str]) -> List[str]:
```
fxmarty, line 341: I think the name is not explicit, is it? Maybe it should be rather something like `filter_output_names`?

optimum/exporters/onnx/convert.py
```diff
@@ -199,6 +190,9 @@ def validate_model_outputs(
     # Compute outputs from the ONNX model
     onnx_outputs = session.run(onnx_named_outputs, onnx_inputs)
 
+    # Create onnxruntime output names for comparison with the reference model inputs
```
fxmarty, line 193: the comment somewhat contradicts the docstring I think. I would rather say "filter the output names" or something, and maybe add a `logger.info`

optimum/exporters/tasks.py
```diff
@@ -239,6 +239,10 @@ class TasksManager:
             "default",
             onnx="CLIPOnnxConfig",
         ),
+        "cliptext": supported_tasks_mapping(
```
fxmarty, line 242: how different is it from CLIP?

optimum/exporters/tasks.py
```diff
@@ -512,6 +516,14 @@ class TasksManager:
             "seq2seq-lm-with-past",
             onnx="T5OnnxConfig",
         ),
+        "unet": supported_tasks_mapping(
```
fxmarty, line 519: Just to understand: what is the motive for having different parts of the model in different configs? People are interested in exporting different parts standalone?

setup.py
```diff
@@ -51,6 +51,7 @@
     "neural-compressor": "optimum-intel[neural-compressor]",
     "graphcore": "optimum-graphcore",
     "habana": "optimum-habana",
+    "diffusers": "diffusers",
```
fxmarty, line 54: why create a new subpackage? For people to be able to opt-out? Otherwise I would have put it in `exporters`

optimum/exporters/tasks.py
echarlaix: Currently we export each model's component separately (U-NET, text encoder and VAE) in order for users to be able to different schedulers as well as to use the resulting exported model for different tasks. Also this integration will make it possible to run inference with `diffusers.OnnxStableDiffusionPipeline`.  It would be great to extend this work in future PRs

optimum/exporters/onnx/convert.py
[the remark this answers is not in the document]
echarlaix: For me it makes more sense to create `models_for_validation` outside the `export_models` and `validate_models_outputs`  functions: we only need to compute it once, currently the same dictionnary is generated two times. Also the arguments of `fn_get_models_from_config` could vary depending on the model type (for stable diffusion model, we only need the reference model). I agree that the argument name is misleading though, we can change it for `models_to_config` wdyt ?

optimum/exporters/onnx/base.py
```diff
@@ -338,6 +338,17 @@ def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[s
...
+        """
+        Generate the output names for the comparison between the ONNX and the reference model.
```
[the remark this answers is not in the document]
echarlaix: It's useful whenwhen the output names between the ONNX model and the reference model differ. Just updated the docstring to make it clearer.

optimum/exporters/onnx/base.py
echarlaix: It maps the output names of the ONNX model to the output names of the reference model and returns the output names to use for validation (useful when the output names between the ONNX model and the reference model differ).

optimum/exporters/onnx/convert.py
```diff
@@ -496,7 +479,10 @@ def export(
             "Please install torch or tensorflow first."
         )
 
-    if is_torch_available() and issubclass(type(model), PreTrainedModel):
+    output.parent.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
echarlaix: Used to create the directory where the ONNX model will be saved if the latter doesn't exist, allowing to pass the corresponding `output_names` to the `export_models` function (for stable diffusion, each component is saved in a different directory)

optimum/exporters/onnx/base.py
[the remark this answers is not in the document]
michaelbenayoun: Just to understand: why would they differ?

optimum/exporters/onnx/convert.py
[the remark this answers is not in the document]
michaelbenayoun: I would rather go with `models_and_onnx_configs` or something similar, the point being that it's a tuple not a dict.

optimum/exporters/tasks.py
echarlaix: This was added to export a `CLIPTextModel`  (only composed of the text encoder)

optimum/exporters/onnx/convert.py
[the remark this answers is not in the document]
echarlaix: `models` is a dictionnary not a tuple, what about `models_to_onnx_configs` ?

optimum/exporters/onnx/base.py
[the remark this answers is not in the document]
echarlaix: In our case the reference model input and output names are both `"sample"`, which cannot be the case for the exported ONNX model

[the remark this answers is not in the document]
michaelbenayoun: Thanks!

optimum/exporters/onnx/convert.py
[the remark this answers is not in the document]
michaelbenayoun: I mean you match the model "name" like `"encoder"` to a tuple looking like this `(model, onnx config)` right? Or the type annotation is wrong.

optimum/exporters/onnx/model_configs.py
```diff
@@ -645,6 +646,87 @@ def outputs(self) -> Mapping[str, Mapping[int, str]]:
         }
 
 
+class CLIPTextOnnxConfig(TextEncoderOnnxConfig):
+    NORMALIZED_CONFIG_CLASS = NormalizedConfig.with_args(
+        vocab_size="vocab_size",
+        sequence_length="max_position_embeddings",
+        allow_new=True,
+    )
```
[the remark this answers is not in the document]
michaelbenayoun: Yes, I am just mentioning that a `CLIPNormalizedConfig` is defined later in this file, which *maybe* could be used here (not sure, hence my question)

optimum/exporters/onnx/convert.py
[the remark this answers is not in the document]
echarlaix: Yes ! In your previous comment you were proposing to replace the name with `models_and_onnx_configs` to emphasize on the fact that it was a tuple no ?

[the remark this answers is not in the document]
michaelbenayoun: The values are tuples. The tuple both contains the model, and the onnx config. So `model_TO_onnx_configs` seems misleading to me since it sounds like values are dictionaries.

optimum/exporters/tasks.py
fxmarty: I see thanks

optimum/exporters/onnx/convert.py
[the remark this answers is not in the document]
echarlaix: I don't have a strong opinion on this so happy to change it with `models_and_onnx_configs` to close this conversation

optimum/exporters/onnx/base.py
fxmarty: Then I don't understand https://github.com/huggingface/optimum/blob/df602513d4452959b03f79fc5b2866201f7f3db5/optimum/exporters/onnx/model_configs.py#L700

I actually don't understand what `Maps the output names of the ONNX model to the output names of the reference model.` means. In one case, the method is identity, in the second one it just returns ["sample"]

optimum/exporters/tasks.py
fxmarty: So this means that with this PR I would be able to export the unet part alone, if I want to? How does it work with exporters since different parts of the model are in subfolders?

For example:

```
python -m optimum.exporters.onnx -m https://huggingface.co/runwayml/stable-diffusion-v1-5 out_onnx/
```

will export all the subparts.

But given that `unet` and such are added in the TasksManager, I would expect that this means I am able to use the exporters CLI to export only part of the model. I don't think this would be possible currently since parts of the models are stored in subfolders, e.g. https://huggingface.co/runwayml/stable-diffusion-v1-5/tree/main

So I'm not sure it is good to add those here.

optimum/exporters/onnx/model_configs.py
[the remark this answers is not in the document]
echarlaix: No `CLIPNormalizedConfig` inherits from `NormalizedTextAndVisionConfig` while `CLIPTextOnnxConfig` should inherit from `TextEncoderOnnxConfig` (a `CLIPTextModel` is only composed of the text encoder)

optimum/exporters/onnx/base.py
echarlaix: For a U-Net model, the output names of the original model is `"sample"` while for the ONNX model the output name is `"out_sample"`. `output_names_for_validation` allows to match the output names so we can compare both outputs.

optimum/exporters/tasks.py
echarlaix: To export one specific component or a subset of them, you should have your adapted `fn_get_models_from_config` but I feel this is a very specific case that we can tackle in a future PR in case it's requested by users

optimum/exporters/onnx/base.py
fxmarty: Could we have it more explicit? Like the `torch_to_onnx_input_map`, where we would explicitely see the map

```
{"out_sample": "sample"}
```

or something like this, otherwise I don't understand where the `maps` description comes in play

optimum/exporters/tasks.py
michaelbenayoun: I think this is not a big deal for now, what we can do is add a `TODO` comment in the code for now. Maybe also create an issue to remember to work on it later.

optimum/exporters/onnx/__main__.py
```diff
@@ -90,84 +97,93 @@ def main():
...
+        # Saving the model config and preprocessor as this is needed sometimes.
+        model.config.save_pretrained(args.output.parent)
+        maybe_save_preprocessors(args.model, args.output.parent)
+
+    if task == "stable-diffusion" or (
+        args.for_ort and (model.config.is_encoder_decoder or task.startswith("causal-lm"))
+    ):
+        if task == "stable-diffusion":
+            args.opset = args.opset or 14
+            args.atol = args.atol or 1e-3
```
michaelbenayoun, line 150: I don't think we should hardcode values here, and we actually set those args to default values above.

optimum/exporters/onnx/convert.py
```diff
@@ -22,20 +22,17 @@
...
-    get_encoder_decoder_models_for_export,
-    is_torch_onnx_support_available,
-)
 
 
 if is_torch_available():
     from transformers.modeling_utils import PreTrainedModel
     from transformers.pytorch_utils import is_torch_less_than_1_11
 
+    from diffusers import ModelMixin
```
michaelbenayoun, line 34: We do not need a `is_diffusers_available`? 
I actually do not know since we add `diffusers` as extras for `exporters`...
Feel free to ignore!

optimum/exporters/tasks.py
fxmarty: What would be the place to list all supported model types and modalities supported in the exporters ONNX CLI then?

I understand this is not a big deal, I just think it is good to have a way to know what is supported and what is not. This change is breaking this.

optimum/exporters/onnx/model_configs.py
```diff
@@ -645,6 +646,102 @@ def outputs(self) -> Mapping[str, Mapping[int, str]]:
...
+            "last_hidden_state": {0: "batch_size", 1: "sequence_length", 2: "feature_dim"},
+            "pooler_output": {0: "batch_size", 1: "feature_dim"},
+        }
+
+    def generate_dummy_inputs(self, framework: str = "pt"):
+        dummy_inputs = super().generate_dummy_inputs(framework=framework)
+        if framework == "pt":
+            import torch
+
+            dummy_inputs["input_ids"] = dummy_inputs["input_ids"].to(dtype=torch.int32)
```
michaelbenayoun, line 674: Why? Dummy inputs are not already in `int32`?

optimum/utils/input_generators.py
```diff
@@ -338,7 +339,7 @@ def __init__(
         self.hidden_size = normalized_config.hidden_size
 
     def generate(self, input_name: str, framework: str = "pt"):
-        if input_name == "encoder_outputs":
+        if input_name == "encoder_outputs" or input_name == "encoder_hidden_states":
```
michaelbenayoun, line 342: nit:
```suggestion
        if input_name in ["encoder_outputs", "encoder_hidden_states"]:
```

optimum/utils/input_generators.py
```diff
@@ -575,3 +578,31 @@ def generate(self, input_name: str, framework: str = "pt"):
...
+
+    def __init__(
+        self,
+        task: str,
+        normalized_config: NormalizedConfig,
+        batch_size: int = DEFAULT_DUMMY_SHAPES["batch_size"],
+        random_batch_size_range: Optional[Tuple[int, int]] = None,
+    ):
+        self.task = task
+        self.vocab_size = getattr(normalized_config, "vocab_size", 49408)
```
michaelbenayoun, line 598: I suggested a possible solution for this above.

michaelbenayoun (commented):
Left a few comments, most of them are nits and questions, except the one for the vocab size. LGTM once this is solved!

optimum/exporters/tasks.py
echarlaix: Like `"seq2seq-lm"` will load the model using the class `AutoModelForSeq2SeqLM`, `"stable-diffusion"` will use `StableDiffusionPipeline` to load the model's different component. I don't think `"text-to-image"` is adapted here because this task include other types of models which should not be loaded with `StableDiffusionPipeline` and also exclude other tasks such as `image-to-image`.

optimum/exporters/onnx/__main__.py
echarlaix: This means that the default ospet and atol of the onnx config will not be used when not specified by the user. I would rather the make the opset argument optional in `export `, and use the onnx config default value when not specified

optimum/exporters/onnx/convert.py
echarlaix: We don't since `diffusers` was added as an extra for `exporters`

optimum/exporters/onnx/model_configs.py
echarlaix: `torch.randint` returns a tensor of type `torch.int64` by default

optimum/utils/input_generators.py
echarlaix: Good catch ! We don't need to set a default value there, I will remove it

optimum/exporters/tasks.py
fxmarty: The difference is that arguably `seq2seq-lm` is a task while `stable-diffusion` is a model architecture. Stable diffusion is not a task.

fxmarty: agreed!

michaelbenayoun: I think @fxmarty makes a good point, but to not make this PR last to long, I think it would be perfect to create issues for that so we can work on them later.

optimum/exporters/onnx/__main__.py
michaelbenayoun: But already we check if `args.opset` is `None` and set it to the onnx config value if that's the case. Same thing for the `atol`.

fxmarty:
@echarlaix Feel free to ask re-review once it's all good, I'll have a last look. The fix for the CLI tests is merged https://github.com/huggingface/optimum/pull/584 , so hopefully it will be green now!

optimum/exporters/onnx/__main__.py
echarlaix: Yes, this is done if the model has an `OnnxConfig`, but in our case each components has it's own (so we don't have a default `opset` or `atol` to give to `export_models` as every components should have his own default value). I will make the `opset` / `atol` optional in both `export` and `validate_model_outputs`

michaelbenayoun (approved):
LGTM (once all the tests pass), thanks @echarlaix !

optimum/exporters/tasks.py
michaelbenayoun: Created an issue => #596

optimum/utils/input_generators.py
```diff
@@ -50,8 +50,8 @@ def wrapper(*args, **kwargs):
     "sequence_length": 16,
     "num_choices": 4,
     # image
-    "width": 224,
-    "height": 224,
+    "width": 64,
```
fxmarty, line 53: 224 was for imagenet, not sure how the export for vision models will work there

optimum/exporters/onnx/convert.py
```diff
@@ -149,7 +141,7 @@ def validate_model_outputs(
             The path to the exported model.
         onnx_named_outputs (`List[str]`):
             The names of the outputs to check.
-        atol (`float`):
+        atol (`float`, defaults to `None`):
```
fxmarty, line 144:
```suggestion
        atol (`Optional[float]`, defaults to `None`):
```

Aligning with https://huggingface.slack.com/archives/C02P0559X9S/p1669628754896019

optimum/exporters/onnx/convert.py
```diff
@@ -67,75 +68,66 @@ def check_dummy_inputs_are_allowed(
...
         onnx_named_outputs (`List[str]`):
             The names of the outputs to check.
-        atol (`float`):
-            The absolute tolerance in terms of outputs difference between the reference and the exported model.
         output_dir (`Path`):
             Output directory where the exported ONNX models are stored.
-        fn_get_models_from_config (`Callable[[Union[PreTrainedModel, TFPreTrainedModel], OnnxConfig], Dict[str, Tuple[Union[PreTrainedModel, TFPreTrainedModel], OnnxConfig]]]`):
-            Function outputing a dictionary of submodels and downstream onnx configurations, for example to export the `model` in several pieces,
-            as it is the case with encoder-decoder models.
+        atol (`float`, defaults to `None`):
```
fxmarty, line 90:
```suggestion
        atol (`Optional[float]`, defaults to `None`):
```

optimum/exporters/onnx/convert.py
```diff
@@ -412,48 +410,42 @@ def export_models(
...
-        opset (`int`):
-            The version of the ONNX operator set to use.
+        models_and_onnx_configs (`Dict[str, Tuple[Union[`PreTrainedModel`, `TFPreTrainedModel`], `OnnxConfig`]]):
+             A dictionnary containing the models to export and their corresponding onnx configs.
         output_dir (`Path`):
             Output directory to store the exported ONNX models.
-        fn_get_models_from_config (`Callable[[Union[PreTrainedModel, TFPreTrainedModel], OnnxConfig], Dict[str, Tuple[Union[PreTrainedModel, TFPreTrainedModel], OnnxConfig]]]`):
-            Function outputing a dictionary of submodels and downstream onnx configurations, for example to export the `model` in several pieces,
-            as it is the case with encoder-decoder models.
+        opset (`int`, defaults to `None`):
```
fxmarty, line 417:
```suggestion
        opset (`Optional[int]`, defaults to `None`):
```

optimum/exporters/onnx/convert.py
```diff
@@ -477,10 +469,10 @@ def export(
             The model to export.
         config ([`~exporters.onnx.config.OnnxConfig`]):
             The ONNX configuration associated with the exported model.
-        opset (`int`):
-            The version of the ONNX operator set to use.
         output (`Path`):
             Directory to store the exported ONNX model.
+        opset (`int`, defaults to `None`):
```
fxmarty, line 474:
```suggestion
        opset (`Optional[int]`, defaults to `None`):
```

fxmarty (approved):
LGTM, awesome! It would be great to update the wording in https://huggingface.co/docs/optimum/main/en/exporters/onnx/usage_guides/export_a_model to indicate that the ONNX export is not only for Transformers.

And maybe update https://huggingface.co/docs/optimum/main/en/exporters/onnx/package_reference/configuration with relevant configs if any

optimum/utils/input_generators.py
echarlaix: I modified the default dummy inputs width and height as the validation step comparing the reference and the ONNX model is quite memory intensive for stable diffusion model (leading to OOM error). I fixed the `LayoutLMv3OnnxConfig` to get the `image_size` from the config as it should be fixed. For the other architectures, I'm expecting that their export should work for any weight / height default values, if not we should fix their corresponding `OnnxConfig` no ?

fxmarty: Yes discussing with Michael this should be fine. I would recommend using tiny models for the tests.

echarlaix: Yes it's already the case for the tests ! It's more for common users using the CLI with classic stable diffusion models.

anton-l (approved):
LGTM, thank you so much @echarlaix!

Feature requests for next PRs:
* copying over the scheduler & tokenizer configs to have a complete pipeline as a result
* bundling the 2GB+ UNet weights as an [external file](https://huggingface.co/runwayml/stable-diffusion-v1-5/tree/onnx/unet). Currently the ops and weights are exploding into multiple files which will severely affect down/upload speeds to the Hub, as I've noticed with the first versions of the diffusers exporter

PR #539, opened 2022-12-02
Add the ORTModelForSemanticSegmentation class

TheoMrc:
# What does this PR do?

This PR aims to implement the ORTModelForImageSegmentation class to provide support for image segmentation .onnx models, and full integration of such models through transformers pipelines for CPU or GPU onnxruntime inference (see  Issue #382)

# Implementation details

The ORTModelForImageSegmentation was based on the already implemented ORTModelForImageClassification in `optimum/onnxruntime/modeling_ort.py` with several modifications: 

1) For CPU and GPU inference :
- class was added to `optimum/onnxruntime/__init__.py`
- `self.forward` method returns a `SemanticSegmenterOutput` instead of `ImageClassifierOutput`
- correct `auto_model_class `and `export_feature` referenced
- Copied all tests from the `ORTModelForImageClassificationIntegrationTest` in `tests/onnxruntime/test_modeling.py`

2) For GPU inference
- logits_shape was changed `ORTModelForImageSegmentation.prepare_logits_buffer` to return a 4 dimensional tensor shape 2D of shape `(input_batch_size, self.config.num_labels, output_height, output_width)`. 
**The issue is that I did not find a way to get model output size, which is different from input size from config.json, or any other attribute of ORTModelForImageSegmentation or ORTModelForImageSegmentation.model.**

CPU inference works as following: 
```
from optimum.onnxruntime.modeling_ort import ORTModelForImageSegmentation
session = ORTModelForImageSegmentation.load_model(onnx_path)
onnx_model = ORTModelForImageSegmentation(session)
inputs = feature_extractor(pil_image, return_tensors="pt")
outputs = onnx_model(**inputs)
```

I could not test GPU inference because I could not manage to make onnxruntime-gpu work:
```
onnx_model.to('cuda:0')
>>>  File "C:\Users\theol\Documents\GitHub\Repositories\optimum\optimum\onnxruntime\modeling_ort.py", line 202, in to
    validate_provider_availability(provider)  # raise error if the provider is not available
>>>  File "C:\Users\theol\Documents\GitHub\Repositories\optimum\optimum\onnxruntime\utils.py", line 227, in validate_provider_availability
    raise ImportError(
>>>ImportError: Asked to use CUDAExecutionProvider, but `onnxruntime-gpu` package was not found. Make sure to install `onnxruntime-gpu` package instead of `onnxruntime`.
```
Might be because of local venv setup issues on my side.
My CUDA installation is working for transformers with torch models.
[...]

PR #571, opened 2022-12-09
Move custom IOBinding to IOBindingHelper

JingyaHuang:
# What does this PR do?

Move the `prepare_io_binding` function from `ORTModelForCustomTasks` to `IOBindingHelper` as this could be helpful and reusable for other ORT models, namely the ongoing `ORTModelForSemanticSegmentation`(#539).

### DIFF
--- a/optimum/onnxruntime/io_binding/io_binding_helper.py
+++ b/optimum/onnxruntime/io_binding/io_binding_helper.py
@@ -155,7 +155,7 @@ class IOBindingHelper:
     @staticmethod
     def prepare_io_binding(ort_model: "ORTModel", **inputs) -> ort.IOBinding:
         """
-        Returns IOBinding object for an inference session. This method is created for general purpose, if the inputs and outputs
+        Returns an IOBinding object for an inference session. This method is general purpose, if the inputs and outputs
         are determined, you can prepare data buffers directly to avoid tensor transfers across frameworks.
         """
 
--- a/optimum/onnxruntime/io_binding/io_binding_helper.py
+++ b/optimum/onnxruntime/io_binding/io_binding_helper.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 import logging
 import traceback
+from typing import TYPE_CHECKING
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


optimum/onnxruntime/modeling_ort.py
```diff
@@ -1576,7 +1576,7 @@ def forward(
 
 @add_start_docstrings(
     """
-    Onnx Model for any custom tasks. It can be used to leverage the inference acceleration with any custom exported ONNX model.
+    Onnx Model for any custom tasks. It can be used to leverage the inference acceleration with custom encoder or decoder-only ONNX model.
```
fxmarty, line 1579:
```suggestion
    ONNX Model for any custom tasks. It can be used to leverage the inference acceleration for any single-file ONNX model.
```

fxmarty: or something like this

fxmarty (approved):
Thanks!

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -147,3 +151,35 @@ def get_device_index(device):
...
+        """
+
+        name_to_np_type = TypeHelper.get_io_numpy_type_map(ort_model.model)
+
+        # Bind inputs and outputs to onnxruntime session
+        io_binding = ort_model.model.io_binding()
+
+        # Bind inputs
+        for input_name in ort_model.model_input_names:
+            onnx_input = inputs.pop(input_name)
```
michaelbenayoun, line 169: I think we should check whether it is there or not, otherwise the code will fail later.

optimum/onnxruntime/io_binding/io_binding_helper.py
```diff
@@ -147,3 +151,35 @@ def get_device_index(device):
         elif isinstance(device, int):
             return device
         return 0 if device.index is None else device.index
+
+    @staticmethod
+    def prepare_io_binding(ort_model: "ORTModel", **inputs) -> ort.IOBinding:
+        """
+        Returns IOBinding object for an inference session. This method is created for general purpose, if the inputs and outputs
```
michaelbenayoun, line 158:
```suggestion
        Returns an IOBinding object for an inference session. This method is general purpose, if the inputs and outputs
```

michaelbenayoun (approved):
LGTM!

PR #573, opened 2022-12-12
Add test for exporters.onnx CLI

fxmarty:
As per title.

Simply check that the CPI does not raise errors.

Next PR: testing on various shapes, slow and fast tests.

## Before submitting


tests/exporters/exporters_utils.py
```diff
@@ -0,0 +1,58 @@
+PYTORCH_EXPORT_MODELS_TINY = {
```
michaelbenayoun, line 1: I think this could be the occasion to get rid of the `PYTORCH_EXPORT_MODELS` in `test_onnx_export.py` and use this everywhere, wdty?

michaelbenayoun (approved):
LGTM!

regisss (approved):
LGTM! Thanks @fxmarty :fire:

tests/exporters/exporters_utils.py
fxmarty: Yes, I want to do it in a next PR where we'll have nightly tests for the ONNX export with various shapes, potentially a bit larger models, while the normal (on PR and main) exporters tests would use this `TINY` dict!

fxmarty:
Will merge and fix the failing tests in an other PR.

PR #586, opened 2022-12-13
Export ONNX models to multiple subfolders

NouamaneTazi:


NouamaneTazi:
Saving is correctly done as we discussed @fxmarty, but loading deserves some more discussion.
I'm trying to load the different submodels from the different subfolders (for example [here](https://github.com/nouamanetazi/optimum/blob/ca7a38b510006462b5ff01daccf72abf470b1a41/optimum/onnxruntime/modeling_seq2seq.py#L914-L933)). I think an easy (but bloated) solution would be to have multiple subfolder arguments, like we have for file_name
```python

        encoder_file_name: str = ONNX_ENCODER_NAME,
        decoder_file_name: str = ONNX_DECODER_NAME,
        decoder_with_past_file_name: str = ONNX_DECODER_WITH_PAST_NAME,
        subfolder: str = "",

```
Wdyt?

optimum/exporters/onnx/convert.py
```diff
@@ -444,11 +444,17 @@ def export_models(
...
         submodel, sub_onnx_config = models_for_export[model_name]
-        output_path = (
-            output_dir.joinpath(output_names[i])
-            if output_names is not None
-            else output_dir.joinpath(model_name + ".onnx")
-        )
+        output_name = output_names[i] if output_names is not None else Path(model_name + ".onnx")
+
+        # If we export several models, we need to create a subfolder for each of them
+        if len(models_for_export) > 1:
```
fxmarty, line 450: This should always be the case in the function

fxmarty:
Some recommended logic in `from_pretrained`:

```python
        for file_name in [encoder_file_name, decoder_file_name, decoder_with_past_file_name]:
            if (
                validate_file_exists(model_id, file_name, subfolder=subfolder, revision=revision)
                and validate_file_exists(model_id, file_name, subfolder=Path(subfolder, Path(file_name).stem), revision=revision)
            ):
                logger.warning(
                    f"The ONNX file {file_name} was found both in ./{subfolder} and ./{Path(subfolder, Path(file_name).stem)}. The file"
                    f" ./{Path(subfolder, Path(file_name).stem)} will be used by default."
                )

        file_names = {
            "encoder_file_name": {"name": encoder_file_name, "pattern": ENCODER_ONNX_FILE_PATTERN},
            "decoder_file_name": {"name": decoder_file_name, "pattern": DECODER_ONNX_FILE_PATTERN},
            "decoder_with_past_file_name": {"name": decoder_with_past_file_name, "pattern": DECODER_WITH_PAST_ONNX_FILE_PATTERN}
        }

        for name_key in file_names:
            if not validate_file_exists(model_id, file_names[name_key]["name"], subfolder=subfolder, revision=revision):
                file_names[name_key]["name"] = ORTModelForConditionalGeneration.infer_onnx_filename(
                    model_id,
                    file_names[name_key]["pattern"],
                    name_key,
                    subfolder=subfolder,
                    use_auth_token=use_auth_token,
                    revision=revision,
                    fail_if_not_found=True if name_key != "decoder_with_past_file_name" else use_cache
                )

        for name_key, standard_name in zip(file_names.keys(), [ONNX_ENCODER_NAME, ONNX_DECODER_NAME, ONNX_DECODER_WITH_PAST_NAME]):
            regular_onnx_filenames = ORTModelForConditionalGeneration._generate_regular_names_for_filename(standard_name)
        
            if file_names[name_key]["name"] not in regular_onnx_filenames:
                logger.warning(
                    f"The ONNX file {file_names[name_key]['name']} is not a regular name used in optimum.onnxruntime, the "
                    "ORTModelForConditionalGeneration might not behave as expected."
                )
```

fxmarty:
So we can maybe align on:

* use ONNX from subfolders with hard-coded names if the subfolders are found
* otherwise use top-level (backward compatible)

fxmarty:
Great! I tried it, and in this case, I get several warnings:

```
The ONNX file encoder_model/encoder_model.onnx is not a regular name used in optimum.onnxruntime, the ORTModelForConditionalGeneration might not behave as expected.
The ONNX file decoder_model/decoder_model.onnx is not a regular name used in optimum.onnxruntime, the ORTModelForConditionalGeneration might not behave as expected.
The ONNX file decoder_with_past_model/decoder_with_past_model.onnx is not a regular name used in optimum.onnxruntime, the ORTModelForConditionalGeneration might not behave as expected.
```

image

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -1152,6 +1157,42 @@ def to(self, device: Union[torch.device, str, int]):
         return self
 
 
+    @staticmethod
+    def infer_onnx_filename(
```
fxmarty, line 1161: it seems the only difference with the method in `modeling_ort` is the `return onnx_files[0].parent.name + "/" + onnx_files[0].name`, is the duplication really necessary?

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -900,14 +901,18 @@ def _save_pretrained(
...
         """
-        src_file_names = [self.encoder_model_path, self.decoder_model_path]
-        dst_file_names = [encoder_file_name, decoder_file_name]
+        src_paths = [self.encoder_model_path, self.decoder_model_path]
+        dst_file_names = ["encoder_model/" + encoder_file_name, "decoder_model/" + decoder_file_name]
         if self.use_cache:
-            src_file_names.append(self.decoder_with_past_model_path)
-            dst_file_names.append(decoder_with_past_file_name)
+            src_paths.append(self.decoder_with_past_model_path)
+            dst_file_names.append("decoder_with_past_model/" + decoder_with_past_file_name)
```
michaelbenayoun, line 908: Why not use path join instead?

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -1152,6 +1157,42 @@ def to(self, device: Union[torch.device, str, int]):
...
+            if fail_if_not_found:
+                raise FileNotFoundError(f"Could not find any ONNX model file in {path}")
+            return None
+        elif len(onnx_files) > 1:
+            if argument_name is not None:
+                raise RuntimeError(
+                    f"Too many ONNX model files were found in {path}, specify which one to load by using the "
+                    f"{argument_name} argument."
+                )
+        return onnx_files[0].parent.name + "/" + onnx_files[0].name
```
michaelbenayoun, line 1193: Why not use path join instead? Also it seems that you want to produce the "parent/filename" string, I guess you can do using Path directly.

optimum/onnxruntime/modeling_seq2seq.py
NouamaneTazi: Because `dst_file_names` were strings initially. It's true that it would be more proper to have them as Paths, but then I would have to make sure everything else still works

optimum/onnxruntime/modeling_seq2seq.py
NouamaneTazi: The point is to have `infer_onnx_filename()` return `encoder/encoder_model.onnx` instead of just `encoder_model.onnx`
So we can either separate the function's logic for each `modeling_ort` `modeling_decoder` and `modeling_seq2seq`  or we can add an argument to `infer_onnx_filename()` which lets us control that

fxmarty: yes, I'd rather do a

`for name, folder name in zip([...], [...]):` with the three files than duplicate the method.

optimum/onnxruntime/modeling_seq2seq.py
fxmarty: You can create a path doing `Path(x, y, z)` if that's what you want, with `x, y, z` either strings or `Path`

mht-sharma:
Could we convert all the external data files into a single file. ONNX provides the following api to do so:
```python
onnx_model = ... # Your model in memory as ModelProto
onnx.save_model(onnx_model, "path/to/save/the/model.onnx", save_as_external_data=True, all_tensors_to_one_file=True, location="filename", size_threshold=1024, convert_attribute=False)
```
After conversion the files could be like this:
```bash
- encoder_model.onnx
- encoder_model.data
- decoder_model.onnx
- decoder_model.data
.....
```

Just a thought, I do not think it brings any kind of optimisation, but would make the handling of files easier? @NouamaneTazi @fxmarty @michaelbenayoun

fxmarty:
oh interesting, did not know this was possible. So basically all our discussion from yesterday is fruitless if we can do this @NouamaneTazi :D

michaelbenayoun:
That seems interesting!
Can it be handled by ONNX Runtime? If so, then it would be much cleaner IMO.

mht-sharma:
Unfortunately no it cannot be handled by ONNXRuntime. We need to add the snippet just after the ONNX export in `convert.py` to make it feasible.

JingyaHuang:
Not sure how relevant it is, but for `ORTOptimizer`, I have added the possibility to export all external tensors in one file #332. Better be aligned with the naming if you finally decide to also enable it in the exporter.

fxmarty:
@mht-sharma What do you mean by `Unfortunately no it cannot be handled by ONNXRuntime`?

Does it mean that ONNX Runtime InferenceSession can not load

```
encoder_model.onnx
encoder_model.data
```

?

mht-sharma:
> @mht-sharma What do you mean by `Unfortunately no it cannot be handled by ONNXRuntime`?
> 
> Does it mean that ONNX Runtime InferenceSession can not load
> 
> ```
> encoder_model.onnx
> encoder_model.data
> ```
> 
> ?

@fxmarty  I guess I misunderstood the question. I thought ichael was asking the conversion can be done by ort.

But yeh the above format can be loaded with onnxruntime and we do not need to make any changes there. The single file format will just add offsets to the onnx model from where one wants to load the data and the file name will just point to the single file.

fxmarty:
@mht-sharma @NouamaneTazi So from what I understand, the proposal from Mohit is very elegant, but the only issue it has is that the `torch.onnx.export` will automatically export many external files. So to convert them to a single external `.data`, we would need to reload in memory the external files, and dump them to a single `.data` using the onnx python API. From there, `encoder_model.onnx + encoder_model.data` can be loaded into an InferenceSession.

So there is some downside to it as well?

optimum/onnxruntime/modeling_seq2seq.py
NouamaneTazi: I was hesitant of changing `infer_onnx_filename`'s behaviour, but I just noticed that it's only used in `modeling_decoder.py` and `modeling_seq2seq.py` so yes I think that's a very good suggestion, thanks!

optimum/onnxruntime/utils.py
```diff
@@ -270,3 +272,22 @@ class ORTQuantizableOperator(Enum):
     Resize = "Resize"
     AveragePool = "AveragePool"
     Concat = "Concat"
+
+
+def _get_external_data_paths(src_paths: List[Path], dst_file_names: List[str]) -> Tuple[List[Path], List[str]]:
```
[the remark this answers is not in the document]
NouamaneTazi: It's still useful for when we don't save external data in a subfolder (which is the default case when we don't have multiple submodels like encoder/decoder etc). Unless we want to always force onnx export to a subfolder even when external data is not created

mht-sharma:
@NouamaneTazi Currently there is no argument in the `torch.onnx.export` to enforce creation of a single file. 

To avoid creation of separate folders, we need to add the conversion statements just after the `torch.onnx.export` in `convert.py`, so that we have a clean export after calling the `from_pretrained`.

The downside is there that we need to reload and dump. But that is just a small latency during the loading stage and does not affect the inference.

However, we need to delete the redundant weight files after the data is dumped in single files to avoid additional disk usage.

The function could be something like this:

```python
def convert_external_data_to_single_file(path_to_onnx_model):
     external_data_file_paths = get_external_weight_files()
     if external_data:
         # load the model
         # convert the model
         # delete the external data    
```

fxmarty:
Is it really a small latency? We could be handling 10s of GB here

NouamaneTazi:
After exporting is it still cleaner to keep each model in a subfolder
```
encoder_model/
  encoder_model.onnx
  encoder_model.data
decoder_model/
  decoder_model.onnx
  decoder_model.data
```

Or do you prefer something like this
```
encoder_model.onnx
encoder_model.data
decoder_model.onnx
decoder_model.data
```
cc @fxmarty @michaelbenayoun @mht-sharma

mht-sharma:
> Is it really a small latency? We could be handling 10s of GB here

True in absolute terms it can be high and can depend on many factor like HDD in use. The conversion should take nearly 2x the export time or maybe less. Hence, could be a small fraction for some models and can be high for others.

Once done some of the things like `save_pretrained` or loading / downloading model should be relatively faster? (Copying single large file Vs 1000 small files)

Hence, we need to check if the latency would be worth it.

NouamaneTazi:
Saving to a single file seems like more of a hassle than it should be. So I believe we can do that in another PR.
For now we're saving all external data files to separate folders, and are able to load them. Which is good enough I think

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -934,25 +939,29 @@ def _from_pretrained(
         model_path = Path(model_id)
 
         if not validate_file_exists(model_id, encoder_file_name, subfolder=subfolder, revision=revision):
```
NouamaneTazi, line 941: If we keep the subfolders structure, I don't think this condition is ever realized, shall we remove it?

NouamaneTazi:
Saving a seq2seq ORT model now gives
```
.
 |-onnx
 | |-decoder_with_past_model
 | | |-decoder_with_past_model.onnx
 | | |-decoder_with_past_model.onnx_data
 | |-decoder_model
 | | |-decoder_model.onnx_data
 | | |-decoder_model.onnx
 | |-encoder_model
 | | |-encoder_model.onnx
 | |-config.json
 |-pytorch_model.bin
 |-config.json
```

NouamaneTazi:
Since the when uploading files to the hub, it saves everything to root, I think it's cleaner for now to keep everything in root like @mht-sharma suggested
```
.
 |-onnx
 | |-decoder_model.onnx_data
 | |-decoder_with_past_model.onnx
 | |-decoder_model.onnx
 | |-encoder_model.onnx
 | |-encoder_model.onnx_data
 | |-decoder_with_past_model.onnx_data
 | |-config.json
```

fxmarty:
@NouamaneTazi can you make sure the tests pass? You can run `make style`, and also the exporters test should not fail (it's green on main)

PoodleWang:
I tried your method. Now the problem is loading the onnx model and transfer it to become the torchrt engine cost a lot of time. It is more then 1 hour. 
Below is my code:
    "Salesforce/codegen-6B-mono/",
    from_transformers=True,
    provider="TensorrtExecutionProvider"
)
model.save_pretrained("./codegen-6B-mono-onnx")

above code generate the below structure:
--codegen-6B-mono-onnx
----config.json
----decoder_model.onnx
----decoder_model.onnx_data
----decoder_with_past_model.onnx
----decoder_with_past_model.onnx_data

Now, I reload the model 
    "./codegen-6B-mono-onnx,
    provider="TensorrtExecutionProvider",
)

it need more then one hour to load.
Then I run the code:
local_rank = int(os.getenv('LOCAL_RANK', '0'))
world_size = int(os.getenv('WORLD_SIZE', '1'))
generator = pipeline('text-generation', model=model, device=local_rank, tokenizer=tokenizer)

sample = generator("expectations were low, actual enjoyment was high", pad_token_id=tokenizer.eos_token_id,max_new_tokens=128, max_time=20.0, do_sample=True,
                                        temperature=0.8, top_p=0.95, use_cache=True, num_return_sequences=1)
print (sample)

The error is there is no pad token in the tokenizer, I will attach the log later.

PoodleWang:
Another feature request, or maybe I should open a new issue. Can you save the tensorrt engine as well? because building the engine cost a lot of time.

JingyaHuang:
> I tried your method. Now the problem is loading the onnx model and transfer it to become the torchrt engine cost a lot of time. It is more then 1 hour. Below is my code: model = ORTModelForCausalLM.from_pretrained( "Salesforce/codegen-6B-mono/", from_transformers=True, provider="TensorrtExecutionProvider" ) model.save_pretrained("./codegen-6B-mono-onnx")
> 
> above code generate the below structure: --codegen-6B-mono-onnx ----config.json ----decoder_model.onnx ----decoder_model.onnx_data ----decoder_with_past_model.onnx ----decoder_with_past_model.onnx_data
> 
> Now, I reload the model model = ORTModelForCausalLM.from_pretrained( "./codegen-6B-mono-onnx, provider="TensorrtExecutionProvider", )
> 
> it need more then one hour to load. Below is the most of logs that I encounter: 2022-12-17 20:42:33.230697938 [W:onnxruntime:Default, tensorrt_execution_provider.h:60 log] [2022-12-17 12:42:33 WARNING] external/onnx-tensorrt/onnx2trt_utils.cpp:395: One or more weights outside the range of INT32 was clamped 2022-12-17 20:42:33.230871907 [W:onnxruntime:Default, tensorrt_execution_provider.h:60 log] [2022-12-17 12:42:33 WARNING] external/onnx-tensorrt/onnx2trt_utils.cpp:395: One or more weights outside the range of INT32 was clamped
> 
> Then I run the code: local_rank = int(os.get

Hi @PoodleWang, yes please open a new issue for it with an error log.

And for saving the pre-compiled trt engine, can you set it up [in `provider_options`](https://huggingface.co/docs/optimum/onnxruntime/package_reference/modeling_ort#optimum.onnxruntime.ORTModel.from_pretrained.provider_options)

https://github.com/huggingface/optimum/blob/a36bf71163ea7769be8ea205f1772111ca7f6aca/optimum/onnxruntime/modeling_ort.py#L516

With the following:
```python
provider_options = {
    "trt_engine_cache_enable": 1,
    "trt_engine_cache_path": "/path/to/cache",
}
```

PoodleWang:
> > I tried your method. Now the problem is loading the onnx model and transfer it to become the torchrt engine cost a lot of time. It is more then 1 hour. Below is my code: model = ORTModelForCausalLM.from_pretrained( "Salesforce/codegen-6B-mono/", from_transformers=True, provider="TensorrtExecutionProvider" ) model.save_pretrained("./codegen-6B-mono-onnx")
> > above code generate the below structure: --codegen-6B-mono-onnx ----config.json ----decoder_model.onnx ----decoder_model.onnx_data ----decoder_with_past_model.onnx ----decoder_with_past_model.onnx_data
> > Now, I reload the model model = ORTModelForCausalLM.from_pretrained( "./codegen-6B-mono-onnx, provider="TensorrtExecutionProvider", )
> > it need more then one hour to load. Below is the most of logs that I encounter: 2022-12-17 20:42:33.230697938 [W:onnxruntime:Default, tensorrt_execution_provider.h:60 log] [2022-12-17 12:42:33 WARNING] external/onnx-tensorrt/onnx2trt_utils.cpp:395: One or more weights outside the range of INT32 was clamped 2022-12-17 20:42:33.230871907 [W:onnxruntime:Default, tensorrt_execution_provider.h:60 log] [2022-12-17 12:42:33 WARNING] external/onnx-tensorrt/onnx2trt_utils.cpp:395: One or more weights outside the range of INT32 was clamped
> > Then I run the code: local_rank = int(os.get
> 
> Hi @PoodleWang, yes please open a new issue with an error log if you have any further questions.
> 
> And for saving the pre-compiled trt engine, can you try setting it up [in `provider_options`](https://huggingface.co/docs/optimum/onnxruntime/package_reference/modeling_ort#optimum.onnxruntime.ORTModel.from_pretrained.provider_options)?
> 
> https://github.com/huggingface/optimum/blob/a36bf71163ea7769be8ea205f1772111ca7f6aca/optimum/onnxruntime/modeling_ort.py#L516
> 
> With the following:
> 
> ```python
> provider_options = {
>     "trt_engine_cache_enable": 1,
>     "trt_engine_cache_path": "/path/to/cache",
> }
> ```

I tried your method, but it is failed:


> > I tried your method. Now the problem is loading the onnx model and transfer it to become the torchrt engine cost a lot of time. It is more then 1 hour. Below is my code: model = ORTModelForCausalLM.from_pretrained( "Salesforce/codegen-6B-mono/", from_transformers=True, provider="TensorrtExecutionProvider" ) model.save_pretrained("./codegen-6B-mono-onnx")
> > above code generate the below structure: --codegen-6B-mono-onnx ----config.json ----decoder_model.onnx ----decoder_model.onnx_data ----decoder_with_past_model.onnx ----decoder_with_past_model.onnx_data
> > Now, I reload the model model = ORTModelForCausalLM.from_pretrained( "./codegen-6B-mono-onnx, provider="TensorrtExecutionProvider", )
> > it need more then one hour to load. Below is the most of logs that I encounter: 2022-12-17 20:42:33.230697938 [W:onnxruntime:Default, tensorrt_execution_provider.h:60 log] [2022-12-17 12:42:33 WARNING] external/onnx-tensorrt/onnx2trt_utils.cpp:395: One or more weights outside the range of INT32 was clamped 2022-12-17 20:42:33.230871907 [W:onnxruntime:Default, tensorrt_execution_provider.h:60 log] [2022-12-17 12:42:33 WARNING] external/onnx-tensorrt/onnx2trt_utils.cpp:395: One or more weights outside the range of INT32 was clamped
> > Then I run the code: local_rank = int(os.get
> 
> Hi @PoodleWang, yes please open a new issue with an error log if you have any further questions.
> 
> And for saving the pre-compiled trt engine, can you try setting it up [in `provider_options`](https://huggingface.co/docs/optimum/onnxruntime/package_reference/modeling_ort#optimum.onnxruntime.ORTModel.from_pretrained.provider_options)?
> 
> https://github.com/huggingface/optimum/blob/a36bf71163ea7769be8ea205f1772111ca7f6aca/optimum/onnxruntime/modeling_ort.py#L516
> 
> With the following:
> 
> ```python
> provider_options = {
>     "trt_engine_cache_enable": 1,
>     "trt_engine_cache_path": "/path/to/cache",
> }
> ```

Unfortunately, there is still a bug. 

Code:
provider_options = {
    "trt_engine_cache_enable": 1,
    "trt_engine_cache_path": "/opt/tiger/genius/checkpoints/codegen-6B-mono/trt",
}
    "/opt/tiger/genius/checkpoints/codegen-6B-mono/onnx",
    provider="TensorrtExecutionProvider",
    provider_options = provider_options,
)

LOG:
2022-12-18 13:25:25.887143: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 AVX512F AVX512_VNNI FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2022-12-18 13:25:26.000533: I tensorflow/core/util/port.cc:104] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
2022-12-18 13:25:26.510609: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libnvinfer.so.7'; dlerror: libnvinfer.so.7: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/tiger/yarn_deploy/jdk/jre/lib/amd64/server:/opt/tiger/yarn_deploy/hadoop_current/lib/native:/opt/tiger/yarn_deploy/hadoop_current/lib/native/ufs:/opt/tiger/yarn_deploy/hadoop/lib/native:/opt/tiger/yarn_deploy/hadoop_current/lib/native:/opt/tiger/yarn_deploy/hadoop/lib/native:/opt/tiger/jdk/jdk1.8/jre/lib/amd64/server:/usr/local/cuda/lib64::/usr/local/cuda/compat
2022-12-18 13:25:26.510701: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libnvinfer_plugin.so.7'; dlerror: libnvinfer_plugin.so.7: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/tiger/yarn_deploy/jdk/jre/lib/amd64/server:/opt/tiger/yarn_deploy/hadoop_current/lib/native:/opt/tiger/yarn_deploy/hadoop_current/lib/native/ufs:/opt/tiger/yarn_deploy/hadoop/lib/native:/opt/tiger/yarn_deploy/hadoop_current/lib/native:/opt/tiger/yarn_deploy/hadoop/lib/native:/opt/tiger/jdk/jdk1.8/jre/lib/amd64/server:/usr/local/cuda/lib64::/usr/local/cuda/compat
2022-12-18 13:25:26.510711: W tensorflow/compiler/tf2tensorrt/utils/py_utils.cc:38] TF-TRT Warning: Cannot dlopen some TensorRT libraries. If you would like to use Nvidia GPU with TensorRT, please make sure the missing libraries mentioned above are installed properly.
EP Error using ['TensorrtExecutionProvider', 'CUDAExecutionProvider']
Falling back to ['CUDAExecutionProvider', 'CPUExecutionProvider'] and retrying.
EP Error using ['TensorrtExecutionProvider', 'CUDAExecutionProvider']
Falling back to ['CUDAExecutionProvider', 'CPUExecutionProvider'] and retrying.

fxmarty:
@PoodleWang can you please open an issue?

PoodleWang:
> note: issue open here #605 @PoodleWang

let me create a new issue for it. They are different~

PoodleWang:
> > note: issue open here #605 @PoodleWang
> 
> let me create a new issue for it. They are different~

New issue here: https://github.com/huggingface/optimum/issues/606

NouamaneTazi:
Added some tests for saving/loading from local folder and from hub
The tests which sae to hub such as `test_push_ort_model_with_external_data_to_hub` seem to fail when tested locally because it saves to my personal repo on the hub, then it tries to load it from `hf-internal-testing`. It's because of [this part in push_to_hub](https://github.com/nouamanetazi/optimum/blob/2511eaab71b3d124d96b090bc0dd6d18374228be/optimum/modeling_base.py#L167-L172). I'm wondering if it works fine on the CI and if we should do anything about it? 🤔

Otherwise the PR should be good to merge

tests/onnxruntime/test_modeling.py
```diff
@@ -455,6 +460,63 @@ def test_save_model_with_different_name(self):
...
+            # save transformers model to be able to load it with `ORTModel...`
+            model.save_pretrained(tmpdirname)
+
+            model = ORTModelForSeq2SeqLM.from_pretrained(tmpdirname, from_transformers=True)
+            model.save_pretrained(tmpdirname + "/onnx")
+
+            # Verify config and ONNX exported encoder, decoder and decoder with past are present each in their own folder
+            folder_contents = os.listdir(tmpdirname + "/onnx")
+            self.assertTrue(CONFIG_NAME in folder_contents)
+
```
NouamaneTazi, line 511: This test may slow down the CI a little. We could consider adding `@slow` decorators for such tests

cc @michaelbenayoun @fxmarty @mht-sharma

fxmarty: yes!

NouamaneTazi: This is the only test failing it seems. It probably fails because it tries to load the whole model and save convert it to external data. We're getting the error
```
worker 'gw0' crashed while running 'tests/onnxruntime/test_modeling.py::ORTModelIntegrationTest::test_save_seq2seq_model_with_external_data'
```
I'm wondering if we should just use a smaller model instead, and use `FORCE_ONNX_EXTERNAL_DATA` ?

optimum/exporters/onnx/convert.py
```diff
@@ -323,6 +328,34 @@ def export_pytorch(
...
+                onnx_model = onnx.load(
+                    str(output), load_external_data=True
+                )  # TODO: this will probably be too memory heavy, shall we free `model` memory?
+                onnx.save(
+                    onnx_model,
+                    str(output),
+                    save_as_external_data=True,
+                    all_tensors_to_one_file=True,
+                    location=output.name + "_data",
+                    size_threshold=1024 if not FORCE_ONNX_EXTERNAL_DATA else 0,
```
mht-sharma, line 351: Any use_case why we want the threshold to be 0 or `FORCE_ONNX_EXTERNAL_DATA` in general.

mht-sharma, line 351: Should we create a function for this? Probably would be cleaner. 

Not sure how `tf2onnx` handles files >2GB.  Could this be used in the `export_tensorflow`?

NouamaneTazi: I was testing mainly with pytorch. It'll be better if somebody else made another PR to apply the same modifications to `export_tensorflow`

tests/onnxruntime/test_modeling.py
michaelbenayoun: I think it would be good yes, that way we both test the feature and get a faster (and less prone to OOM) CI.

fxmarty:
@NouamaneTazi Do you think it is likely to have it merged today/tomorrow? I think it should be in the release

NouamaneTazi:
Should be ready to merge once all tests pass @fxmarty  🙌

fxmarty (approved):
LGTM, just waiting for the tests

PR #588, opened 2022-12-14
Add vision encoder decoder model in exporters

mht-sharma:
# What does this PR do?

Supports vision-encoder-decoder export to onnx in optimum

Changes
* Adds generic class for exporting Encoder-Decoder type models [doc](https://huggingface.co/docs/transformers/model_doc/encoder-decoder)
* Adds export for `vision-encoder-decoder` models using the above class

Limitation
* Past key value export not supported with TROcr
* Donut model is not supported

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 """Common ONNX configuration classes that handle most of the features for building model specific configurations."""
 
-from typing import TYPE_CHECKING, Any, List, Mapping
+from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Mapping, Optional
 
 from ...utils import (
     DummyAudioInputGenerator,
@@ -33,6 +33,7 @@ if TYPE_CHECKING:
     from transformers import PretrainedConfig
 
     from ...utils import DummyInputGenerator
+    from .base import PatchingSpec
 
 logger = logging.get_logger(__name__)
 
@@ -214,3 +215,64 @@ class AudioToTextOnnxConfig(OnnxSeq2SeqConfigWithPast):
             reference_model_inputs["encoder_hidden_states"] = reference_model_inputs.pop("encoder_outputs")[0]
 
         return reference_model_inputs
+
+
+class EncoderDecoderOnnxConfig(OnnxSeq2SeqConfigWithPast):
+    DUMMY_INPUT_GENERATOR_CLASSES = (
+        DummyTextInputGenerator,
+        DummySeq2SeqDecoderTextInputGenerator,
+    )
+
+    def __init__(
+        self,
+        config: "PretrainedConfig",
+        task: str = "default",
+        patching_specs: Optional[List["PatchingSpec"]] = None,
+        use_past: bool = False,
+        use_past_in_inputs: Optional[bool] = None,
+        use_present_in_outputs: Optional[bool] = None,
+        behavior: ConfigBehavior = ConfigBehavior.MONOLITH,
+    ):
+        super().__init__(
+            config,
+            task=task,
+            patching_specs=patching_specs,
+            use_past=use_past,
+            use_past_in_inputs=use_past_in_inputs,
+            use_present_in_outputs=use_present_in_outputs,
+            behavior=behavior,
+        )
+
+        from ..tasks import TasksManager
+
+        from pdb import set_trace; set_trace()
+        # retrieve the encoder config
+        encoder_onnx_config_constructor = TasksManager.get_exporter_config_constructor(
+            exporter="onnx", task="default", model_type=config.encoder.model_type
+        )
+        self._encoder_onnx_config = encoder_onnx_config_constructor(config.encoder)
+
+        # retrieve the decoder config
+        decoder_onnx_config_constructor = TasksManager.get_exporter_config_constructor(
+            exporter="onnx", task="default", model_type=config.decoder.model_type
+        )
+        self._decoder_onnx_config = decoder_onnx_config_constructor(config.decoder, use_past=use_past)
+
+        self._normalized_config.ENCODER_NORMALIZED_CONFIG_CLASS = self._encoder_onnx_config._normalized_config
+        self._normalized_config.DECODER_NORMALIZED_CONFIG_CLASS = self._decoder_onnx_config._normalized_config
+
+        if isinstance(self.decoder_onnx_config, OnnxSeq2SeqConfigWithPast):
+            self._past_key_values_generator = (DummySeq2SeqPastKeyValuesGenerator,)
+        else:
+            self._past_key_values_generator = (DummyPastKeyValuesGenerator,)
+
+        self.DUMMY_INPUT_GENERATOR_CLASSES += self._past_key_values_generator
+
+    def add_past_key_values(self, inputs_or_outputs: Mapping[str, Mapping[int, str]], direction: str):
+        return self._decoder_onnx_config.add_past_key_values(inputs_or_outputs, direction)
+
+    def flatten_past_key_values(self, flattened_output, name, idx, t):
+        return self._decoder_onnx_config.flatten_past_key_values(flattened_output, name, idx, t)
+
+    def flatten_output_collection_property(self, name: str, field: Iterable[Any]) -> Dict[str, Any]:
+        return self._decoder_onnx_config.flatten_output_collection_property(name, field)
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -245,27 +245,45 @@ class EncoderDecoderOnnxConfig(OnnxSeq2SeqConfigWithPast):
 
         from ..tasks import TasksManager
 
-        # retrieve the encoder config
-        encoder_onnx_config_constructor = TasksManager.get_exporter_config_constructor(
-            exporter="onnx", task="default", model_type=config.encoder.model_type
-        )
-        self._encoder_onnx_config = encoder_onnx_config_constructor(config.encoder)
+        if self._behavior is not ConfigBehavior.DECODER:
+            # retrieve the encoder config
+            encoder_onnx_config_constructor = TasksManager.get_exporter_config_constructor(
+                exporter="onnx", task="default", model_type=config.encoder.model_type
+            )
+            self._encoder_onnx_config = encoder_onnx_config_constructor(config.encoder)
+            self._normalized_config.ENCODER_NORMALIZED_CONFIG_CLASS = self._encoder_onnx_config._normalized_config
 
-        # retrieve the decoder config
-        decoder_onnx_config_constructor = TasksManager.get_exporter_config_constructor(
-            exporter="onnx", task="default", model_type=config.decoder.model_type
-        )
-        self._decoder_onnx_config = decoder_onnx_config_constructor(config.decoder, use_past=use_past)
+        if self._behavior is not ConfigBehavior.ENCODER:
+            # retrieve the decoder config
+            decoder_onnx_config_constructor = TasksManager.get_exporter_config_constructor(
+                exporter="onnx", task="default", model_type=config.decoder.model_type
+            )
+            self._decoder_onnx_config = decoder_onnx_config_constructor(config.decoder, use_past=use_past)
 
-        self._normalized_config.ENCODER_NORMALIZED_CONFIG_CLASS = self._encoder_onnx_config._normalized_config
-        self._normalized_config.DECODER_NORMALIZED_CONFIG_CLASS = self._decoder_onnx_config._normalized_config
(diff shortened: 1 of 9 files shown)


tests/exporters/exporters_utils.py
```diff
@@ -110,6 +110,7 @@
     "speech-to-text": "hf-internal-testing/tiny-random-Speech2TextModel",
     "xlm": "hf-internal-testing/tiny-random-XLMModel",
     "xlm-roberta": "hf-internal-testing/tiny-xlm-roberta",
+    "vision-encoder-decoder": "hf-internal-testing/tiny-random-VisionEncoderDecoderModel-vit-gpt2",
```
fxmarty, line 113: Could you add a test for `trocr` and `donut-swin` as well?

optimum/exporters/onnx/config.py
```diff
@@ -211,3 +212,88 @@ def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[s
...
+            config,
+            task=task,
+            patching_specs=patching_specs,
+            use_past=use_past,
+            use_past_in_inputs=use_past_in_inputs,
+            use_present_in_outputs=use_present_in_outputs,
+            behavior=behavior,
+        )
+
+        from ..tasks import TasksManager
```
fxmarty, line 243: would we have it at the top level?

optimum/utils/normalized_config.py
```diff
@@ -94,6 +94,30 @@ def __getattr__(self, attr_name):
...
+        ):
+            return self.ENCODER_NORMALIZED_CONFIG_CLASS.__getattr__(attr_name)
+        if (
+            self.DECODER_CONFIG is not None
+            and self.DECODER_NORMALIZED_CONFIG_CLASS is not None
+            and attr_name.upper() in dir(self.DECODER_NORMALIZED_CONFIG_CLASS)
+        ):
+            return self.DECODER_NORMALIZED_CONFIG_CLASS.__getattr__(attr_name)
+
+        return super().__getattr__(attr_name)
```
fxmarty, line 118: why do we need this?

optimum/exporters/onnx/model_configs.py
```diff
@@ -903,3 +909,67 @@ class Speech2TextOnnxConfig(AudioToTextOnnxConfig):
...
+
+
+# TODO: Replace the TextSeq2SeqOnnxConfig inheritance with VisionToTextOnnxConfig when added.
+# The change below however does not affect the export for the model
+class TrOCROnnxConfig(TextSeq2SeqOnnxConfig):
+    NORMALIZED_CONFIG_CLASS = NormalizedSeq2SeqConfig.with_args(
+        decoder_num_layers="decoder_layers",
+        num_layers="decoder_layers",
+        decoder_num_attention_heads="decoder_attention_heads",
+        eos_token_id="eos_token_id",
```
fxmarty, line 921:
```suggestion
```

tests/exporters/exporters_utils.py
mht-sharma: added for trocr, donut-skin has threshold issue while exporting would look into it separately

optimum/exporters/onnx/config.py
mht-sharma: It causes circular export on top

optimum/utils/normalized_config.py
mht-sharma: These type of models can have any decoder and encoder head and it is difficult to manage every normalised class. So created this helper class to work with these models

tests/exporters/exporters_utils.py
fxmarty: Then could we not add it in tasks.py? And handle adding donut-swin in a separate PR altogether?

optimum/exporters/onnx/config.py
fxmarty: This could mean that some refactor is needed, I guess.

tests/exporters/exporters_utils.py
mht-sharma: The test model has high atol requirement (4e-2). The export for the model in doc is slow but works with 1e-3 threshold, so not putting that here.

optimum/exporters/onnx/config.py
mht-sharma: Deferring import would still be ok IMO. But if there is a better way would look into it.

optimum/exporters/onnx/model_configs.py
```diff
@@ -903,3 +909,66 @@ class Speech2TextOnnxConfig(AudioToTextOnnxConfig):
...
+        task: str = "default",
+        patching_specs: Optional[List["PatchingSpec"]] = None,
+        use_past: bool = False,
+        use_past_in_inputs: Optional[bool] = None,
+        use_present_in_outputs: Optional[bool] = None,
+        behavior: ConfigBehavior = ConfigBehavior.MONOLITH,
+    ):
+        super().__init__(config, task, patching_specs, use_past, use_past_in_inputs, use_present_in_outputs, behavior)
+
+        # TODO: Check modelling code to fix the issue with use_cache for trocr
```
michaelbenayoun, line 948:
```suggestion
        # TODO: Check modeling code to fix the issue with use_cache for trocr
```

optimum/utils/normalized_config.py
```diff
@@ -94,6 +94,30 @@ def __getattr__(self, attr_name):
         return super().__getattr__(attr_name)
 
 
+class NormalizedEncoderDecoderConfig(NormalizedSeq2SeqConfig, NormalizedVisionConfig):
```
michaelbenayoun, line 97: Can this be refactored with `NormalizedTextAndVisionConfig`?

Also, what are `ENCODER_CONFIG` and `DECODER_CONFIG`?

optimum/exporters/onnx/config.py
```diff
@@ -211,3 +212,88 @@ def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[s
...
+                self._past_key_values_generator = (
+                    DummySeq2SeqDecoderTextInputGenerator,
+                    DummySeq2SeqPastKeyValuesGenerator,
+                )
+            else:
+                self._past_key_values_generator = (
+                    DummySeq2SeqDecoderTextInputGenerator,
+                    DummyPastKeyValuesGenerator,
+                )
+
```
michaelbenayoun, line 272: I would add the nromalized config creation hear instead of setting attributes in the if branches. I would also add a type check on it.

optimum/exporters/onnx/config.py
```diff
@@ -211,3 +212,88 @@ def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[s
...
+    def flatten_past_key_values(self, flattened_output, name, idx, t):
+        return self._decoder_onnx_config.flatten_past_key_values(flattened_output, name, idx, t)
+
+    def flatten_output_collection_property(self, name: str, field: Iterable[Any]) -> Dict[str, Any]:
+        return self._decoder_onnx_config.flatten_output_collection_property(name, field)
+
+    def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[str, Any]) -> Mapping[str, Any]:
+        if self._behavior is ConfigBehavior.DECODER:
+            reference_model_inputs["input_ids"] = reference_model_inputs.pop("decoder_input_ids")
+            reference_model_inputs["encoder_hidden_states"] = reference_model_inputs.pop("encoder_outputs")[0]
```
michaelbenayoun, line 297: Why not use `self._decoder_onnx_config.generate_dummy_inputs_for_validation`?

optimum/exporters/onnx/config.py
```diff
@@ -211,3 +212,88 @@ def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[s
...
+            use_past=use_past,
+            use_past_in_inputs=use_past_in_inputs,
+            use_present_in_outputs=use_present_in_outputs,
+            behavior=behavior,
+        )
+
+        from ..tasks import TasksManager
+
+        if self._behavior is not ConfigBehavior.DECODER:
+            # retrieve the encoder config
```
michaelbenayoun, line 246: Unnecessary imo
```suggestion
```

optimum/exporters/onnx/config.py
```diff
@@ -211,3 +212,88 @@ def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[s
...
+            encoder_onnx_config_constructor = TasksManager.get_exporter_config_constructor(
+                exporter="onnx", task="default", model_type=config.encoder.model_type
+            )
+            self._encoder_onnx_config = encoder_onnx_config_constructor(config.encoder)
+            self._normalized_config.ENCODER_NORMALIZED_CONFIG_CLASS = self._encoder_onnx_config._normalized_config
+
+        if self._behavior is not ConfigBehavior.ENCODER:
+            # retrieve the decoder config
```
michaelbenayoun, line 254: Unnecessary imo
```suggestion
```

optimum/utils/normalized_config.py
mht-sharma: Probably not as the use of both the configs is different. Though built in similar way, the new normalised config is to be used with any type of `EncoderDecoder` type models text, vision, speech etc. The idea is to not force any normalised config beforehand and it takes the configs from the encoder and decoder part of the model while initialising the onnx config. Removed the incorrect inheritance to avoid confusion.

I have removed the `ENCODER_CONFIG` and `DECODER_CONFIG` since they are not used anymore.

optimum/exporters/onnx/config.py
mht-sharma: Is there a reason to set the values in constructor than setting attribute? Plus why would type check be necessary?

tests/exporters/exporters_utils.py
fxmarty: What do you mean by `the model in doc`?

fxmarty: I'd just like that what is said to be supported in `tasks.py` is actually supported, and we can make sure of it by adding an unit test.

mht-sharma: Commented Donut. Will handle this separately

optimum/exporters/onnx/config.py
mht-sharma: There can be decoders which inherited from `OnnxConfigWithPast` and would have the `dummy_inputs_for_validation` empty. We have to rewrite to handle those cases.

optimum/utils/normalized_config.py
```diff
@@ -101,6 +101,23 @@ def __getattr__(self, attr_name):
...
+        if self.ENCODER_NORMALIZED_CONFIG_CLASS is not None and attr_name.upper() in dir(
+            self.ENCODER_NORMALIZED_CONFIG_CLASS
+        ):
+            return self.ENCODER_NORMALIZED_CONFIG_CLASS.__getattr__(attr_name)
+        if self.DECODER_NORMALIZED_CONFIG_CLASS is not None and attr_name.upper() in dir(
+            self.DECODER_NORMALIZED_CONFIG_CLASS
+        ):
+            return self.DECODER_NORMALIZED_CONFIG_CLASS.__getattr__(attr_name)
+
+        return super().__getattr__(attr_name)
```
fxmarty, line 118: Couldn't the `attr_name` be in both the encoder and decoder config? e.g. `bos_token_id` in https://huggingface.co/microsoft/trocr-small-handwritten/blob/main/config.json . Could this result in unexpected behavior? We should rather be able to access attributes recursively no, like `normalized_config.encoder.bos_token_id` or `normalized_config.decoder.bos_token_id`, if this makes sense!

fxmarty (approved):
LGTM, thanks!

michaelbenayoun (approved):

optimum/utils/normalized_config.py
mht-sharma: Yes, it could be possible. This seems to be a common problem for the other hybrid configs also. Would make a note  of it and handle it separately. 

Currently, this would not affect the ORT export (where we export the encoder and decoder separately) as one of the configs would always be None. So considering it low priority.

Mir-Umar:
Hi @mht-sharma can you kindly provide some insights about how to remove the following limitation from the above issue:
1. Past key value export not supported with TROcr
I am actually working with a custom trained TROCR model and want to convert it to ONNX for faster inference.

fxmarty:
@Mir-Umar Feel free to open an issue as well.

mht-sharma:
@Mir-Umar There is an existing issue for this https://github.com/huggingface/optimum/issues/744. I did not got the time to work on this yet. 

The  `use_cache` argument for the TrOCR model does not work in Transformers so the export was failing. Probably the TROcr code in Transformers needs to be looked at to debug the issue further. Let me know if you would like to work on the above issue.

PR #590, opened 2022-12-14
OnnxConfig now handle the export to encoder / decoder / decoder_with_past themselves

michaelbenayoun:
# What does this PR do?

This handles the export to ONNX in multiple files (for seq2seq models) directly in the relevant OnnxConfig itself, instead of using others.

cc @mht-sharma @fxmarty

### DIFF
--- a/optimum/exporters/onnx/base.py
+++ b/optimum/exporters/onnx/base.py
@@ -529,6 +529,8 @@ class OnnxSeq2SeqConfigWithPast(OnnxConfigWithPast):
         """Override this to specify custom attribute change for a given behavior."""
         if self._behavior is ConfigBehavior.ENCODER:
             self.task = "default"
+            self.use_past_in_inputs = False
+            self.use_present_in_outputs = False
         if self._behavior is ConfigBehavior.DECODER:
             self.use_past_in_inputs = self.use_past
             self.use_present_in_outputs = True
--- a/optimum/exporters/onnx/base.py
+++ b/optimum/exporters/onnx/base.py
@@ -516,93 +516,27 @@ class OnnxSeq2SeqConfigWithPast(OnnxConfigWithPast):
             use_present_in_outputs=use_present_in_outputs,
         )
         self._behavior = behavior
-        self._inputs_to_use = {
-            ConfigBehavior.ENCODER: self.inputs_for_encoder,
-            ConfigBehavior.DECODER: self.inputs_for_decoder,
-        }
+        self.override_attributes_for_behavior()
 
-    def __getattribute__(self, name):
-        if name == "inputs" and self._behavior is not ConfigBehavior.MONOLITH:
-            return self._inputs_to_use[self._behavior]
-        return super().__getattribute__(name)
+    def override_attributes_for_behavior(self):
+        if self._behavior is ConfigBehavior.ENCODER:
+            self.task = "default"
+        if self._behavior is ConfigBehavior.DECODER:
+            self.use_past_in_inputs = self.use_past
+            self.use_present_in_outputs = True
 
-    def _for_version(
-        self,
-        behavior: ConfigBehavior,
-        task: Optional[str] = None,
-        use_past: Optional[bool] = None,
-        use_past_in_inputs: Optional[bool] = None,
-        use_present_in_outputs: Optional[bool] = None,
+    def with_behavior(
+        self, behavior: Union[str, ConfigBehavior], use_past: bool = False
     ) -> "OnnxSeq2SeqConfigWithPast":
-        task = task if task is not None else self.task
-        use_past = use_past if use_past is not None else self.use_past
+        if isinstance(behavior, str) and not isinstance(behavior, ConfigBehavior):
+            behavior = ConfigBehavior(behavior)
         return self.__class__(
             self._config,
-            task=task,
+            task=self.task,
             patching_specs=self._patching_specs,
             use_past=use_past,
-            use_past_in_inputs=use_past_in_inputs,
-            use_present_in_outputs=use_present_in_outputs,
             behavior=behavior,
         )
-
-    def for_encoder(
-        self,
-        task: Optional[str] = None,
-        use_past: Optional[bool] = None,
-        use_past_in_inputs: Optional[bool] = None,
-        use_present_in_outputs: Optional[bool] = None,
-    ) -> "OnnxSeq2SeqConfigWithPast":
-        """
-        Produces the same configuration but adatpted for exporting only the encoder part of the model.
-        """
-        return self._for_version(
-            ConfigBehavior.ENCODER,
-            task=task,
-            use_past=use_past,
-            use_past_in_inputs=use_past_in_inputs,
-            use_present_in_outputs=use_present_in_outputs,
-        )
-
-    def for_decoder(
-        self,
-        task: Optional[str] = None,
-        use_past: Optional[bool] = None,
-        use_past_in_inputs: Optional[bool] = None,
-        use_present_in_outputs: Optional[bool] = None,
-    ) -> "OnnxSeq2SeqConfigWithPast":
-        """
-        Produces the same configuration but adapted for exporting only the decoder part of the model.
-        """
-        return self._for_version(
-            ConfigBehavior.DECODER,
-            task=task,
-            use_past=use_past,
-            use_past_in_inputs=use_past_in_inputs,
-            use_present_in_outputs=use_present_in_outputs,
-        )
-
-    @property
-    @abstractmethod
-    def inputs_for_encoder(self) -> Mapping[str, Mapping[int, str]]:
-        """
-        Mapping containing the axis definition of the input tensors to provide to the encoder part of the model.
-
-        Returns:
-            `Mapping[str, Mapping[int, str]]`: A mapping of each input name to a mapping of axis position to the axes symbolic name.
-        """
-        raise NotImplementedError()
-
-    @property
-    @abstractmethod
-    def inputs_for_decoder(self) -> Mapping[str, Mapping[int, str]]:
-        """
-        Mapping containing the axis definition of the input tensors to provide to the deocder part of the model.
-
-        Returns:
-            `Mapping[str, Mapping[int, str]]`: A mapping of each input name to a mapping of axis position to the axes symbolic name.
-        """
-        raise NotImplementedError()
 
     @property
     def outputs(self) -> Mapping[str, Mapping[int, str]]:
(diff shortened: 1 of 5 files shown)


mht-sharma:
Let's merge this soon as it changes the design of the encoder-decoder export I am working on.

fxmarty:
@michaelbenayoun I'll merge it tomorrow.

optimum/exporters/onnx/base.py
```diff
@@ -473,13 +474,80 @@ def flatten_output_collection_property(self, name: str, field: Iterable[Any]) ->
...
+            use_past_in_inputs=use_past_in_inputs,
+            use_present_in_outputs=use_present_in_outputs,
+        )
+        self._behavior = behavior
+        self.override_attributes_for_behavior()
+
+    def override_attributes_for_behavior(self):
+        """Override this to specify custom attribute change for a given behavior."""
+        if self._behavior is ConfigBehavior.ENCODER:
+            self.task = "default"
```
mht-sharma, line 521: Do we not need to set the both the `use_past_in_inputs` and `use_present_in_outputs` to False in this case?

optimum/exporters/onnx/config.py
```diff
@@ -150,9 +173,47 @@ class AudioOnnxConfig(OnnxConfig):
     DUMMY_INPUT_GENERATOR_CLASSES = (DummyAudioInputGenerator,)
 
 
-class TextAndAudioOnnxConfig(OnnxSeq2SeqConfigWithPast):
+class AudioToTextOnnxConfig(OnnxSeq2SeqConfigWithPast):
```
mht-sharma, line 176: Could we align the names of the config with the `AutoModels`.

For text to text. we have  - `TextSeq2Seq...`
For speech to text, -> `SpeechSeq2Seq`
For Vision to text -> `Vision2Seq`

optimum/exporters/onnx/base.py
```diff
@@ -473,13 +474,80 @@ def flatten_output_collection_property(self, name: str, field: Iterable[Any]) ->
...
     """
 
     PAD_ATTENTION_MASK_TO_MATCH_TOTAL_SEQUENCE_LENGTH = False
 
+    def __init__(
+        self,
+        config: "PretrainedConfig",
+        task: str = "default",
+        patching_specs: Optional[List[PatchingSpec]] = None,
+        use_past: bool = False,
```
fxmarty, line 502: this is somewhat nitpick, but I could not find a docstring for what `use_past`, `use_past_in_inputs`, `use_present_in_outputs` mean and do.

We should add one somewhere in my opinion, otherwise (at least I) you have to guess and try to see what it means. Do you think it is useful? It can be left to an other PR.

optimum/exporters/onnx/base.py
```diff
@@ -473,13 +474,80 @@ def flatten_output_collection_property(self, name: str, field: Iterable[Any]) ->
...
+            task=task,
+            patching_specs=patching_specs,
+            use_past=use_past,
+            use_past_in_inputs=use_past_in_inputs,
+            use_present_in_outputs=use_present_in_outputs,
+        )
+        self._behavior = behavior
+        self.override_attributes_for_behavior()
+
+    def override_attributes_for_behavior(self):
```
fxmarty, line 518: Is there an example where overriding `override_attributes_for_behavior` is useful? I think it's only at the base level for now, but this is meant to be extendible?

optimum/exporters/onnx/config.py
```diff
@@ -125,6 +141,13 @@ def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGen
 
         return dummy_inputs_generators
 
+    def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[str, Any]) -> Mapping[str, Any]:
+        if self._behavior is ConfigBehavior.DECODER:
+            reference_model_inputs["input_ids"] = reference_model_inputs.pop("decoder_input_ids")
+            reference_model_inputs["encoder_hidden_states"] = reference_model_inputs.pop("encoder_outputs")[0]
+            reference_model_inputs["encoder_attention_mask"] = reference_model_inputs.pop("attention_mask")
+        return reference_model_inputs
```
fxmarty, line 149: Thinking out loud: could `torch_to_onnx_input_map` and a loop be used in there?

optimum/exporters/onnx/config.py
```diff
@@ -150,9 +173,47 @@ class AudioOnnxConfig(OnnxConfig):
...
+    def torch_to_onnx_input_map(self) -> Mapping[str, str]:
+        if self._behavior is ConfigBehavior.DECODER:
+            return {
+                "decoder_input_ids": "input_ids",
+                "encoder_outputs": "encoder_hidden_states",
+                "attention_mask": "encoder_attention_mask",
+            }
+        return {}
+
+    def generate_dummy_inputs_for_validation(self, reference_model_inputs: Mapping[str, Any]) -> Mapping[str, Any]:
```
fxmarty, line 214: same comment

optimum/exporters/onnx/utils.py
```diff
@@ -95,20 +95,17 @@ def get_encoder_decoder_models_for_export(
         `Dict[str, Tuple[Union[`PreTrainedModel`, `TFPreTrainedModel`], `OnnxConfig`]: A Dict containing the model and
         onnx configs for the encoder and decoder parts of the model.
     """
-    models_for_export = dict()
+    models_for_export = {}
```
fxmarty, line 98: lol nice catch: https://stackoverflow.com/a/17098017/4370080

optimum/exporters/onnx/utils.py
```diff
@@ -131,7 +128,7 @@ def get_decoder_models_for_export(
...
 
     models_for_export["decoder_model"] = (
         model,
```
fxmarty, line 134: just below I used `config.__class__`, but can we replace it by `config.with_behavior()` now?

fxmarty (approved):
Awesome, the simplifications look great!

optimum/exporters/onnx/base.py
michaelbenayoun: Yes you are right, fixed it!

optimum/exporters/onnx/config.py
michaelbenayoun: For me it is more like `TextAndVision` is because it does not care about the direction, but for instance `AudioToText` means that the encoder side will be audio, and the decoder side will be text.

I think here there is the modality, and the seq2seq part, for instance, we could have:

- `SpeechSeq2Seq` meaning "Speech to speech"
- `Speech2Text` meaning "Speech to text" implying that it's seq2seq

I guess, we can discuss about it on another PR, but I agree that specifying a standard would be good!

optimum/exporters/onnx/base.py
michaelbenayoun: We did not really document those configs in details in terms of their attributes.
I will do that in another PR (I might also try to find some better way than having `use_past` and the other two, which kind of collide).

optimum/exporters/onnx/base.py
michaelbenayoun: It is more to let the possibility to change that whitout having to rewrite the `__init__`.

optimum/exporters/onnx/base.py
```diff
@@ -473,13 +474,80 @@ def flatten_output_collection_property(self, name: str, field: Iterable[Any]) ->
...
+            behavior ([`ConfigBehavior`]):
+                The behavior to use for the new instance.
+            use_past (`bool`, defaults to `False`):
+                Whether or not the new instance should use past.
+
+        Returns:
+            `OnnxSeq2SeqConfigWithPast`
+        """
+        if isinstance(behavior, str) and not isinstance(behavior, ConfigBehavior):
+            behavior = ConfigBehavior(behavior)
```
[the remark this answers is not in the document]
michaelbenayoun: So `ConfigBehavior` is an `Enum` with string values. Doing this will check as you mention. But it also allows to do something like: 

```python
if self._behavior is ConfigBehavior.DECODER
```

Because if `self._behavior` is a string, you need to check for equality instead, with something like:

```python
if self._behavior.value == "decoder"
```

optimum/exporters/onnx/config.py
michaelbenayoun: I get it could, but you're right, let's keep it like that for now as it is more readable.

optimum/exporters/onnx/utils.py
michaelbenayoun: Yes, fixed it!

optimum/exporters/onnx/base.py
fxmarty: awesome!

optimum/exporters/onnx/base.py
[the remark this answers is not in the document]
fxmarty: Yes alright. But since `ConfigBehavior` inherits from `str` the check `ConfigBehavior.MONOLITH == "monolith"` goes just fine.

optimum/exporters/onnx/utils.py
fxmarty: great

optimum/exporters/onnx/config.py
fxmarty: All good

optimum/exporters/onnx/utils.py
```diff
@@ -131,15 +128,12 @@ def get_decoder_models_for_export(
...
         onnx configs for the encoder and decoder parts of the model.
     """
-    models_for_export = dict()
+    models_for_export = {}
 
-    models_for_export["decoder_model"] = (
-        model,
-        config.__class__(model.config, task=config.task, use_past=False, use_present_in_outputs=True),
-    )
+    models_for_export["decoder_model"] = (model, config.with_behavior("decoder", use_past=False))
```
mht-sharma, line 133: With behaviour is for seq2seq models, this will not work for models such as gpt2.

optimum/exporters/onnx/utils.py
```diff
@@ -131,15 +128,12 @@ def get_decoder_models_for_export(
         `Dict[str, Tuple[Union[`PreTrainedModel`, `TFPreTrainedModel`], `OnnxConfig`]: A Dict containing the model and
         onnx configs for the encoder and decoder parts of the model.
     """
-    models_for_export = dict()
+    models_for_export = {}
```
mht-sharma, line 132: nit: The function docstring is incorrect.

optimum/exporters/onnx/config.py
michaelbenayoun: Created the issue => #597

optimum/exporters/onnx/base.py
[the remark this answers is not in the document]
michaelbenayoun: Yes, but not with the `is` operator. Which I think is nice in terms of readability (and is much faster too)

optimum/exporters/onnx/base.py
michaelbenayoun: Created an issue => #595

optimum/exporters/onnx/config.py
mht-sharma: great

optimum/exporters/onnx/utils.py
michaelbenayoun: Solved

mht-sharma (approved):
Thanks @michaelbenayoun looks much clean now

optimum/exporters/onnx/base.py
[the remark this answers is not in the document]
fxmarty: yes yes agreed, just wanted to understand the motive :)

Issue #577, opened 2022-12-12
BetterTransformer for T5ForConditionalGeneration 


R4ZZ3:
### Feature request

Got this error when trying out BetterTransformer:
"The Better Transformers implementation for the model T5ForConditionalGeneration has not been implemented yet. Please open an issue requesting the addition of this model with its BetterTransformerimplementation."

### Motivation

I would like to speed up some text2text models with BetterTransformer

### Your contribution

I can at least test the feature once it is ready

younesbelkada:
Hi @R4ZZ3 
Thanks so much for your interest in BetterTransformer! 
As stated in https://github.com/huggingface/transformers/issues/20372#issuecomment-1324736962 , sadly T5 cannot be supported yet for BetterTransformer

fxmarty:
@younesbelkada Maybe we should add in the doc a section of models that can't be supported.

R4ZZ3:
Yeah, maybe the error message should also be adjusted accordingly

fxmarty:
Hi @R4ZZ3 , more informative error messages are added in https://github.com/huggingface/optimum/pull/594

younesbelkada:
Thanks for addressing this @fxmarty let's close the issue!

PR #594, opened 2022-12-15
Refactor BetterTransformer to be able to raise more informative error messages 

fxmarty:
With this PR:
* the error message if a model type is not supported is raised ahead of time
* a special error message is raised in case the model type *can not* be supported, stating the reason, fixing https://github.com/huggingface/optimum/issues/577

Just need to add tests for this.


tests/bettertransformer/test_bettertransformer_encoder.py
```diff
@@ -80,37 +79,12 @@ def prepare_inputs_for_class(self, model_id=None):
...
-        r"""
-        A test to check if the modified dictionnary is consistent (same number of keys + successfully import
-        the correct `PreTrainedModel` module). This also checks if the list of models to test has the same
-        number of classes than the list of convertable models.
         """
-        for keys in BETTER_TRANFORMER_LAYERS_MAPPING_DICT.keys():
-            self.assertTrue(("Layer" in keys) or ("Block" in keys))
-
-        ALL_SUPPORTED_HF_CLASSES = convert_to_hf_classes(BETTER_TRANFORMER_LAYERS_MAPPING_DICT)
-        self.assertEqual(len(ALL_SUPPORTED_HF_CLASSES.keys()), len(BETTER_TRANFORMER_LAYERS_MAPPING_DICT.keys()))
```
fxmarty, line 113: In my opinion this test is always passing so I removed it. Let me know if I misunderstand it.

optimum/bettertransformer/models/__init__.py
```diff
@@ -28,61 +28,49 @@
...
-}
+    EXCLUDE_FROM_TRANSFORM = {
+        # clip's text model uses causal attention, that is most likely not supported in BetterTransformer
+        "clip": ["text_model"],
+    }
+
+    CAN_NOT_BE_SUPPORTED = {
+        "deberta-v2": "deberta-v2 does not use a regular attention mechanism, which is not suppored in PyTorch's BetterTransformer.",
+        "glpn": "glpn has a convolutional layer present in the FFN network, which is not suppored in PyTorch's BetterTransformer.",
+        "t5": "t5 uses attention bias, which is not suppored in PyTorch's BetterTransformer.",
```
michaelbenayoun, line 72:
```suggestion
        "t5": "T5 uses attention bias, which is not suppored in PyTorch's BetterTransformer.",
```

optimum/bettertransformer/models/__init__.py
```diff
@@ -28,61 +28,49 @@
...
-EXCLUDE_FROM_TRANSFORM = {
-    "clip": ["text_model"],  # text model uses causal attention, that is most likely not supported in BetterTransformer
-}
+    EXCLUDE_FROM_TRANSFORM = {
+        # clip's text model uses causal attention, that is most likely not supported in BetterTransformer
+        "clip": ["text_model"],
+    }
+
+    CAN_NOT_BE_SUPPORTED = {
+        "deberta-v2": "deberta-v2 does not use a regular attention mechanism, which is not suppored in PyTorch's BetterTransformer.",
```
michaelbenayoun, line 70:
```suggestion
        "deberta-v2": "DeBERTa v2 does not use a regular attention mechanism, which is not suppored in PyTorch's BetterTransformer.",
```

optimum/bettertransformer/models/__init__.py
```diff
@@ -28,61 +28,49 @@
...
+        "glpn": "glpn has a convolutional layer present in the FFN network, which is not suppored in PyTorch's BetterTransformer.",
```
michaelbenayoun, line 71:
```suggestion
        "glpn": "GLPN has a convolutional layer present in the FFN network, which is not suppored in PyTorch's BetterTransformer.",
```

optimum/bettertransformer/transformation.py
```diff
@@ -68,12 +69,15 @@ def replace_to_bettertransformer(model, config):
...
-        class_name = module.__class__.__name__
-        is_bt_compatible = class_name in BETTER_TRANFORMER_LAYERS_MAPPING_DICT
-
-        if is_bt_compatible:
-            fast_module = BETTER_TRANFORMER_LAYERS_MAPPING_DICT[class_name](module, config)
-            model._modules[name] = fast_module
+        # replace the module if it is a transformer layer compatible with bettertransformer
+        if (
+            module.__class__.__name__
+            == BetterTransformerManager.BETTER_TRANSFORMER_MODEL_MAPPING[config.model_type][0]
```
michaelbenayoun, line 75: nit: I think this is not super readable, and would put that into an intermediate variable:

```
should_replace_module = module.__class__.__name__ == BetterTransformerManager.BETTER_TRANSFORMER_MODEL_MAPPING[config.model_type][0]
```

optimum/bettertransformer/transformation.py
```diff
@@ -51,8 +51,9 @@ def replace_to_bettertransformer(model, config):
     for name, module in model.named_children():
         if len(list(module.children())) > 0:
             # we may explicitly exclude part of the model to use BetterTransformer
-            if config.model_type not in EXCLUDE_FROM_TRANSFORM or (
-                config.model_type in EXCLUDE_FROM_TRANSFORM and name not in EXCLUDE_FROM_TRANSFORM[config.model_type]
+            if config.model_type not in BetterTransformerManager.EXCLUDE_FROM_TRANSFORM or (
```
michaelbenayoun, line 54: Since this check is made multiple times, what about exposing methods in the `BetterTransformerManager` to check that. 

Like a method called `is_excluded(self, model_type: str) -> bool`.

optimum/bettertransformer/transformation.py
```diff
@@ -117,15 +124,17 @@ def set_last_layer(model: torch.nn.Module):
...
             if isinstance(dict_named_module[key], torch.nn.ModuleList) and all(
                 "LayerBetterTransformer" in module_name for module_name in sort_fn(dict_named_module[key])
             ):
                 setattr(dict_named_module[key][-1], "is_last_layer", True)
                 return
-        raise NotImplementedError(ERROR_MESSAGE.format(model_name=model.__class__.__name__))
+
+    raise Exception(
+        f"The transformation of the model {model.__class__.__name__} to BetterTransformer failed while it should not. Please fill a bug report at https://github.com/huggingface/optimum/issues."
+    )
```
michaelbenayoun, line 137: Maybe also encourage for contribution and opening a PR.

optimum/bettertransformer/transformation.py
```diff
@@ -167,6 +176,19 @@ def transform(
         else:
             load_accelerate = False
 
+        if model.config.model_type in BetterTransformerManager.CAN_NOT_BE_SUPPORTED:
```
michaelbenayoun, line 179:
```suggestion
        if  BetterTransformerManager.cannot_support(config.model_type)
```

optimum/bettertransformer/transformation.py
```diff
@@ -167,6 +176,19 @@ def transform(
...
+            raise ValueError(
+                f"The model type {model.config.model_type} can not be supported to be used with BetterTransformer. The identified reason is:"
+                f" {BetterTransformerManager.CAN_NOT_BE_SUPPORTED[model.config.model_type]}. Currently supported models are:"
+                f" {BetterTransformerManager.BETTER_TRANSFORMER_MODEL_MAPPING.keys()}."
+            )
+        if model.config.model_type not in BetterTransformerManager.BETTER_TRANSFORMER_MODEL_MAPPING:
```
michaelbenayoun, line 185:
```suggestion
        if not BetterTransformerManager.supports(config.model_type):
```

michaelbenayoun (commented):
I left a few comments, most of them being about creating methods for checking if a model type is / can be / cannot be supported. 

We could also create methods to get the information for the name of layer, and the replacement class, instead of accessing the tuple items. In general, I think it is good to hide the underlying structure.

optimum/bettertransformer/transformation.py
fxmarty: Thanks

optimum/bettertransformer/transformation.py
fxmarty: yes

younesbelkada (commented):
Thank you very much for working on this! It looks much cleaner now
This looks good to me, small nit: I think the PR title + description is slightly misleading, in fact this PR proposes also to refactor everything 
As usual just make sure that the GPU tests (with accelerate) work as expected and should be good to merge

tests/bettertransformer/test_bettertransformer_encoder.py
younesbelkada: Correct, I see also that you have added a new test

optimum/bettertransformer/models/__init__.py
```diff
@@ -28,61 +28,69 @@
...
+        "t5": "T5 uses attention bias, which is not suppored in PyTorch's BetterTransformer.",
+    }
+
+    @staticmethod
+    def cannot_support(model_type: str) -> bool:
+        """
+        Returns True if a given model type can not be supported by PyTorch's Better Transformer.
+
+        Args:
+            model_type (str): The model type to check.
```
michaelbenayoun, line 81: nit:
```suggestion
            model_type (`str`):
                The model type to check.
```

optimum/bettertransformer/models/__init__.py
```diff
@@ -28,61 +28,69 @@
...
+        """
+        return model_type in BetterTransformerManager.CAN_NOT_BE_SUPPORTED
+
+    @staticmethod
+    def supports(model_type: str) -> bool:
+        """
+        Returns True if a given model type is supported by PyTorch's Better Transformer, and integrated in Optimum.
+
+        Args:
+            model_type (str): The model type to check.
```
michaelbenayoun, line 91: Same nit

optimum/bettertransformer/transformation.py
```diff
@@ -98,10 +100,13 @@ def set_last_layer(model: torch.nn.Module):
...
                 or (
-                    model.config.model_type in EXCLUDE_FROM_TRANSFORM
-                    and all(name not in key for name in EXCLUDE_FROM_TRANSFORM[model.config.model_type])
+                    model.config.model_type in BetterTransformerManager.EXCLUDE_FROM_TRANSFORM
+                    and all(
+                        name not in key
+                        for name in BetterTransformerManager.EXCLUDE_FROM_TRANSFORM[model.config.model_type]
+                    )
                 )
             )
```
michaelbenayoun, line 111: I do not think that it is easy to read, but feel free to keep it like that!

michaelbenayoun (approved):
Thanks @fxmarty

optimum/bettertransformer/transformation.py
fxmarty: I'll leave as is but I agree

fxmarty:
Thank you! Yes @younesbelkada a bit of refactor was needed to be able to raise errors ahead of time. I'll run the GPU tests (there is only `check_accelerate_compatibility_cpu_gpu` right?) and merge!

younesbelkada:
Thanks! 
Yes, this test should spin 4 tests: https://github.com/huggingface/optimum/blob/main/tests/bettertransformer/test_bettertransformer_encoder.py#L239-L267

Issue #599, opened 2022-12-16
ONNX export of decoder/decoder with past has unwanted output nodes


fxmarty:
### System Info

```shell
Transformers commit: 3ee958207a9830da6afcf2c20772ef9a2159ca30
Optimum commit: 9c71dd52f71f397bd97da2228691a3c7049f4452
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

```
python3 -m optimum.exporters.onnx --model t5-small --task seq2seq-lm-with-past --for-ort t5_onnx
```

![image](https://user-images.githubusercontent.com/9808326/208106672-a1cf1af2-00ba-4057-ac80-28016d9b77ab.png)

### Expected behavior

No unwanted output node

fxmarty:
cc @JingyaHuang I'll look into it

fxmarty:
For decoder with/without past it's the same:

![image](https://user-images.githubusercontent.com/9808326/208107251-a9c389b1-78ba-41c3-a371-927b2d391b2e.png)

fxmarty:
@mht-sharma looks similar to the issue you had no?

mht-sharma:
This is a new issue, have observed this a little back in a few models. 

I have a saved checkpoint of t5 with me but I do not see the node. Some option in the modelling code might be returning the `encoder_hidden_states` with `hidden_states`?

fxmarty:
yes exactly Mohit!

Issue #569, opened 2022-12-09
Seq2Seq ONNX export fails with `ORTModelForConditionalGeneration`


michaelbenayoun:
### System Info

```shell
/
```

### Who can help?

@michaelbenayoun @fxmarty @mht-sharma 

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

This code fails:

```python
from optimum.pipelines import pipeline
from optimum.onnxruntime import ORTModelForSeq2SeqLM

model = ORTModelForSeq2SeqLM.from_pretrained("t5-small", from_transformers=True)

onnx_translation = pipeline("translation_en_to_fr", model=model)

result = onnx_translation("My name is Eustache")
```

But this command-line works:
```bash
python -m optimum.exporters.onnx --model t5-small --task seq2seq-lm-with-past t5_small_onnx
```

### Expected behavior

This should not fail.

mht-sharma:
Is the export part failing? I guess the issue is in the pipeline creation. We need to pass the tokenizer as well.
```python
onnx_translation = pipeline("translation_en_to_fr", model=model, tokenizer=tokenizer)
```

michaelbenayoun:
No i do not think that's failing before (and I added the support for not providing tokenizers when it can be avoided, it will be merged in #567

fxmarty:
In my opinion, the issue is with methods as [`_expand_inputs_for_generation()`](https://github.com/huggingface/transformers/blob/d4bf9ee1ff0e85cb24feec4dd160af39b623d4b9/src/transformers/generation/utils.py#LL1601C44-L1601C73) not being called during the ONNX export, resulting in the wrong shapes for the ONNX export.

fxmarty:
Could also be a bug in `ordered_inputs`:

```
[...]

PR #601, opened 2022-12-16
Add `encoder_last_hidden_state` as an output for encoder-decoder models

fxmarty:
Fixes https://github.com/huggingface/optimum/issues/599

Basically names the last hidden state output:

image

Previously, this output was returned but unnamed (assigned a random name, missing from our `output_names` passed to `torch.onnx.export`) since we trace over `XXXForConditionalGeneration` models, which return the encoder last hidden state.


JingyaHuang (approved):
LGTM! Thanks for fixing this!

optimum/exporters/onnx/base.py
```diff
@@ -148,7 +148,12 @@ class OnnxConfig(ExportConfig, ABC):
             }
         ),
         "semantic-segmentation": OrderedDict({"logits": {0: "batch_size", 1: "num_labels", 2: "height", 3: "width"}}),
-        "seq2seq-lm": OrderedDict({"logits": {0: "batch_size", 1: "decoder_sequence_length"}}),
+        "seq2seq-lm": OrderedDict(
+            {
+                "logits": {0: "batch_size", 1: "decoder_sequence_length"},
+                "encoder_last_hidden_state": {0: "batch_size", 1: "encoder_sequence_length"},
```
mht-sharma, line 154: The issue could be in general to any encoder-decoder model. Could you also check the `speech2text` models also?

optimum/exporters/onnx/config.py
```diff
@@ -75,10 +76,12 @@ class TextSeq2SeqOnnxConfig(OnnxSeq2SeqConfigWithPast):
 
     @property
     def inputs(self) -> Mapping[str, Mapping[int, str]]:
-        common_inputs = {
-            "input_ids": {0: "batch_size", 1: "encoder_sequence_length"},
-            "attention_mask": {0: "batch_size", 1: "encoder_sequence_length"},
-        }
+        common_inputs = OrderedDict(
```
mht-sharma, line 79: You do not need to convert the inputs to `OrderedDict`.  @michaelbenayoun uses the following function [config.ordered_inputs(model)](https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/convert.py#L302) to convert the inputs into `OrderedDict` based on the input signature of the model.

optimum/exporters/onnx/base.py
fxmarty: The tests pass.

optimum/exporters/onnx/config.py
fxmarty: What's a good reason of having `ordered_inputs`? cc @michaelbenayoun  I noticed it failing silently here: https://github.com/huggingface/optimum/issues/569#issuecomment-1348168536

I added it because I would like to enforce the inputs to `add_past_key_values` to be an `OrderedDict`.

optimum/exporters/onnx/__main__.py
```diff
@@ -176,8 +176,8 @@ def main():
             )
         else:
             validate_model_outputs(onnx_config, model, args.output, onnx_outputs, args.atol)
-    except ValueError:
-        logger.error(f"An error occured, but the model was saved at: {args.output.parent.as_posix()}")
+    except ValueError as e:
+        logger.error(f"An error occured, but the model was saved at: {args.output.parent.as_posix()}. Error: {e}")
```
mht-sharma, line 180: Could we also add a suggestion to the user here. Something like, `Try increasing the atol (Absolute difference tolerance) for validation....`

optimum/exporters/onnx/config.py
michaelbenayoun: I makes writing the `inputs` property easier because you dont have to think when to add which input to the dictionary, or what happens when you update it.

optimum/exporters/onnx/base.py
```diff
@@ -523,6 +528,9 @@ def add_past_key_values(self, inputs_or_outputs: Mapping[str, Mapping[int, str]]
             inputs_or_outputs[f"{name}.{i}.encoder.key"] = {0: "batch_size", 2: encoder_sequence}
             inputs_or_outputs[f"{name}.{i}.encoder.value"] = {0: "batch_size", 2: encoder_sequence}
 
+        if direction == "outputs" and "encoder_last_hidden_state" in inputs_or_outputs:
```
michaelbenayoun, line 531: Actually, we do not need to enforce an `OrderedDict` all the time. For the inputs, it is already handled, and outputs are `OrderedDict`s.

optimum/exporters/onnx/model_configs.py
```diff
@@ -77,11 +78,13 @@ class Seq2SeqDecoderOnnxConfig(TextSeq2SeqOnnxConfig):
 
     @property
     def inputs(self) -> Mapping[str, Mapping[int, str]]:
-        common_inputs = {
-            "decoder_input_ids": {0: "batch_size", 1: "past_decoder_sequence_length + sequence_length"},
-            "encoder_outputs": {0: "batch_size", 1: "encoder_sequence_length"},
-            "attention_mask": {0: "batch_size", 1: "encoder_sequence_length"},
-        }
+        common_inputs = OrderedDict(
```
michaelbenayoun, line 81: Not in favor of that, especially because we can get it to work without it.

fxmarty: Okay, I will remove.

optimum/exporters/onnx/config.py
fxmarty: all good

optimum/exporters/onnx/__main__.py
fxmarty: I did an other PR for this

optimum/exporters/onnx/base.py
fxmarty: done

fxmarty:
@mht-sharma @michaelbenayoun Updated, let me know if it's good. I don't check for `OrderedDict` in `add_past_key_values` as we assume the outputs to already be `OrderedDict`, this is not super elegant but I guess fine.

optimum/exporters/onnx/base.py
mht-sharma: Ok lets look into it after the fix

mht-sharma (approved):
LGTM! Thanks @fxmarty

PR #603, opened 2022-12-16
Fix documentation example

echarlaix:


fxmarty (approved):
Looks fine to me to have the doc use pipeline from transformers, arguably the optimum pipeline makes sense if we want to export + load the model on the fly. cc @philschmid 

Just to note, the `from optimum.pipelines import pipeline` is used in many other places, `source/onnxruntime/usage_guides/pipelines.mdx`, `source/onnxruntime/usage_guides/models.mdx`, `source/onnxruntime/usage_guides/gpu.mdx`, `source/bettertransformer/tutorials/convert.mdx`

PR #604, opened 2022-12-18
Improve `use_io_binding` default value for different execution providers

JingyaHuang:
# What does this PR do?

As suggested in the title, there could be no "TensorrtExecutionProvider" in `model.get_providers()` even if the inference session has been created with "TensorrtExecutionProvider" as the provider.

The reason for that is onnxruntime will fallback to CUDA ep when TRT ep has error while creating the inference session.

e,g, `model.get_providers()` returns
```
['CUDAExecutionProvider', 'CPUExecutionProvider']
```
 
The fix ensures that `use_io_binding` is correctly set even when TRT fails.


optimum/onnxruntime/modeling_ort.py
```diff
@@ -428,6 +422,12 @@ def _from_pretrained(
         if model_save_dir is None:
             model_save_dir = new_model_save_dir
 
+        if provider == "TensorrtExecutionProvider" and use_io_binding:
+            logger.warning(
+                "There is no need to do IO binding for TensorrtExecutionProvider, `use_io_binding` is set to False."
```
fxmarty, line 427: I would rather put a `logger.info` here, or better, show a log only if the user explicitly passed `use_io_binding=True, provider="TensorrtExecutionProvider"`.

Couldn't we make `use_io_binding = None` as default, and rather raise an error if it's set and the execution provider is not `CUDAExecutionProvider`?

fxmarty:
An error is raised here https://github.com/huggingface/optimum/issues/610 although the user tries to use CPU. It is because `use_io_bindings` is True even though CUDAExecutionProvider is not used:

```
 self.name_to_np_type = TypeHelper.get_io_numpy_type_map(self.model) if self.use_io_binding else None
```

JingyaHuang:
> An error is raised here #610 although the user tries to use CPU. It is because `use_io_binding` is True even though CUDAExecutionProvider is not used:
> 
> ```
>  self.name_to_np_type = TypeHelper.get_io_numpy_type_map(self.model) if self.use_io_binding else None
> ```
> 
> `use_io_binding` should be True only with CUDAExecutionProvider IMO

@fxmarty, ok, I will set `use_io_binding` default to None and set the value explicitly with EP if it is not given.

But I can't reproduce the error mentioned in #610. When using CPU, `model.get_providers()` shall be `['CPUExecutionProvider']` and  `self.device.type` should be "CPU" isn't it? Although `use_io_binding` was defaulted to `True` it should not have gone with IO binding.

fxmarty:
The error is raised with numpy==1.24.0

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -348,7 +348,7 @@ def __init__(
                 not load the weights associated with the model, only the configuration.
             decoder_with_past_session (`Optional[onnxruntime.InferenceSession]`, *optional*):
                 The ONNX Runtime inference session associated to the decoder with past key values.
-            use_io_binding (`bool`, *optional*, defaults to `True`):
+            use_io_binding (`bool`, *optional*, defaults to `None`):
```
fxmarty, line 351:
```suggestion
            use_io_binding (`Optional[bool]`, defaults to `None`):
```

optimum/onnxruntime/utils.py
```diff
@@ -218,6 +218,28 @@ def validate_provider_availability(provider: str):
...
+
+    if "TensorrtExecutionProvider" in providers:
+        return io_binding_options["TensorrtExecutionProvider"]
+    elif "CUDAExecutionProvider" in providers:
+        return io_binding_options["CUDAExecutionProvider"]
+    elif "CPUExecutionProvider" in providers:
+        return io_binding_options["CPUExecutionProvider"]
+    else:
+        raise KeyError(
+            f"{providers} are not registered execution providers. If you want us to support other ep, please open an issue here: https://github.com/huggingface/optimum/issues."
```
fxmarty, line 239: could we rather have a logger.warning saying that this EP is not tested, use at your own risk?

fxmarty: not sure if we can use something else than CPUExecutionProvider, TensorrtExecutionProvider, CUDAExecutionProvider

optimum/onnxruntime/utils.py
```diff
@@ -218,6 +218,28 @@ def validate_provider_availability(provider: str):
         )
 
 
+def possible_io_binding_for_provider(providers: List[str]) -> bool:
```
fxmarty, line 221: IO binding is only for CUDAExecutionProvider, so do we need this involved logic?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -200,11 +200,15 @@ def __init__(
                 f" Use `ort_model.to()` to send the outputs to the wanted device."
             )
 
-        if "TensorrtExecutionProvider" in self.providers and self.use_io_binding:
-            logger.warning(
-                "There is no need to do IO binding for TensorrtExecutionProvider, `use_io_binding` is set to False."
-            )
-            self.use_io_binding = False
+        io_binding_options = possible_io_binding_for_provider(self.providers)
```
fxmarty, line 203: why not

```python
if provider == "CUDAExecutionProvider" and use_io_binding is None:
    use_io_binding = True
elif provider != "CUDAExecutionProvider" and use_io_binding is True:
    logger.warning("Asked to use IO Binding although the provider used is not CUDAExecutionProvider. IO Binding will not be used.")
    use_io_binding = False
else:
    use_io_binding = False
```

optimum/onnxruntime/utils.py
JingyaHuang: I would prefer that users let us know if it is needed instead of testing and using/dropping it silently.

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -686,7 +686,7 @@ class ORTModelForConditionalGeneration(ORTModel, ABC):
         config ([`PretrainedConfig`]):
             Instance of the configuration associated to the model. Initializing with a config file does
             not load the weights associated with the model, only the configuration.
-        use_io_binding (`bool`):
+        use_io_binding (`bool`, *optional*, defaults to `None`):
```
fxmarty, line 689:
```suggestion
        use_io_binding (`Optional[bool]`, defaults to `None`):
```

aligning on https://huggingface.slack.com/archives/C02P0559X9S/p1669628754896019

optimum/onnxruntime/modeling_ort.py
JingyaHuang: I opened this PR as I don't want to decide the value of `use_io_binding` by `provider`. E.g. one can try loading the inference session with TRT and it falls back to CUDA ep. It would be better to set the io binding depends on what ep are truly used in the end.

optimum/onnxruntime/utils.py
JingyaHuang: If we decide whether to use io binding or not by model.get_provider(), there will be cuda ep in providers when using tensorRT.

optimum/onnxruntime/modeling_ort.py
fxmarty: > E.g. one can try loading the inference session with TRT and it falls back to CUDA ep.

This should not happen if it's an issue with the TRT install, as per https://github.com/huggingface/optimum/pull/514

The possible fallback is only in some nodes in the ONNX.

fxmarty:
> As suggested in the title, there could be no "TensorrtExecutionProvider" in model.get_providers() even if the inference session has been created with "TensorrtExecutionProvider" as the provider.

an error should be raised in this case, if it is not it is that there is a bug in https://github.com/huggingface/optimum/pull/514

optimum/onnxruntime/modeling_ort.py
JingyaHuang: Whether the ep falls back to another ep or not, the error message should not be about io binding. Set the value according to the truly used eps avoid the misleading error message.

fxmarty: there's no such thing as "truly used eps". ONNX Runtime can mix execution providers depending on which EPs support which nodes.

JingyaHuang:
> #514

You put the raising before the inference session, so it won't work if it falls back while instantiating the inference session. 

In #606 nothing was raised while TensorRT was not correctly installed.

fxmarty (approved):
Awesome!

Issue #607, opened 2022-12-18
Validating ONNX model fails for GPT-J


Eichhof:
### System Info

```shell
Optimum: 1.5.1
Python: 3.10.4
Platform: Windows 10
```

### Who can help?

@lewtun @michaelbenayoun 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

I Installed optimum with `pip install optimum[onnxruntime-gpu]`. Then I was running `python -m optimum.exporters.onnx --task causal-lm-with-past --model EleutherAI/gpt-j-6B gptj_onnx/` to transform GPT-J to ONNX.  The output of this call is then as follows:

```
Some weights of the model checkpoint at EleutherAI/gpt-j-6B were not used when initializing GPTJModel: ['lm_head.weight', 'lm_head.bias']
- This IS expected if you are initializing GPTJModel from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing GPTJModel from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
Framework not specified. Using pt to export to ONNX.
Using framework PyTorch: 1.12.1
Overriding 2 configuration item(s)
        - use_cache -> True
        - pad_token_id -> 0
C:\Users\myUsername\Anaconda3\envs\huggingface\lib\site-packages\transformers\models\gptj\modeling_gptj.py:597: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if batch_size <= 0:
C:\Users\myUsername\Anaconda3\envs\huggingface\lib\site-packages\transformers\models\gptj\modeling_gptj.py:177: TracerWarning: torch.tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  mask_value = torch.tensor(mask_value, dtype=attn_weights.dtype).to(attn_weights.device)
[...]

PR #609, opened 2022-12-19
More informative message at ONNX export

fxmarty:
For example, now warn:

```
The ONNX export succeeded with the warning: The maximum absolute difference between the output of the reference model and the ONNX exported model is not within the set tolerance 1e-05:
- present.7.key: max diff = 1.3113021850585938e-05
- present.11.value: max diff = 1.5974044799804688e-05. The exported model was saved at: gpt2_onnx
```

instead of:

```
An error occured, but the model was saved at: gpt2_onnx
```

Fixes https://github.com/huggingface/optimum/issues/607


optimum/exporters/onnx/__main__.py
```diff
@@ -194,10 +202,21 @@ def main():
                 atol=args.atol,
             )
 
-    except ValueError:
-        logger.error(f"An error occured, but the model was saved at: {args.output.parent.as_posix()}")
-        return
-    logger.info(f"All good, model saved at: {args.output.parent.as_posix()}")
+        logger.info(f"The ONNX export succeeded and the exported model was saved at: {args.output.parent.as_posix()}")
+    except ShapeError as e:
+        raise ShapeError(e)
```
michaelbenayoun, line 207: Quick question, not sure about it:
```suggestion
        raise e
```
?

optimum/exporters/onnx/convert.py
```diff
@@ -40,6 +40,18 @@
...
 
+class ShapeError(ValueError):
+    pass
+
+
+class AtolError(ValueError):
+    pass
+
+
+class OutputError(ValueError):
```
michaelbenayoun, line 51: A suggestion, feel free to ignore:
```suggestion
class OutputMatchError(ValueError):
```

michaelbenayoun (approved):
LGTM!

optimum/exporters/onnx/__main__.py
fxmarty: Thank you!

PR #612, opened 2022-12-19
Reimplement get_io_numpy_type_map in IO binding helper

fxmarty:
This allows to avoid an error with numpy==1.24.0 when the parent `get_io_numpy_type_map` is called, and in turn the parents `ort_type_to_numpy_type` are called.

@JingyaHuang  Just to understand, why do we have `"tensor(bool)": np.uint8,` in `ort_type_to_numpy_type` and not `"tensor(bool)": bool`? Because of the dlpack/bool thing support?

Related:
https://github.com/huggingface/optimum/pull/604
https://github.com/microsoft/onnxruntime/pull/14014

### DIFF
--- a/optimum/onnxruntime/io_binding/io_binding_helper.py
+++ b/optimum/onnxruntime/io_binding/io_binding_helper.py
@@ -13,19 +13,20 @@
 #  limitations under the License.
 import logging
 import traceback
-from typing import TYPE_CHECKING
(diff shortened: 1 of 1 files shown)


JingyaHuang:
Thanks for fixing, what about reimplement the whole TypeHelper and make it not rely on onnxruntime.

And for your questions of `"tensor(bool)": np.uint8`, yes it is for dlpack.

PR #613, opened 2022-12-19
Set explictly the device index

JingyaHuang:
# What does this PR do?

IO Bindng needs index information, this set default device index while setting up the ep.

```
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. onnxruntime.capi.onnxruntime_pybind11_state.OrtDevice(arg0: int, arg1: int, arg2: int)

Invoked with: 0, 0, None
```


optimum/onnxruntime/utils.py
```diff
@@ -169,7 +169,7 @@ def get_device_for_provider(provider: str) -> torch.device:
     Gets the PyTorch device (CPU/CUDA) associated with an ONNX Runtime provider.
     """
     return (
-        torch.device("cuda")
+        torch.device("cuda:0")
```
michaelbenayoun, line 172: Is this called when nothing is explicitly specified?
I think this should be set as default, but providing another device should be an option too!

JingyaHuang: @michaelbenayoun, I think that it sets the default value of the device according to the provider.

https://github.com/huggingface/optimum/blob/bc8f33eab9ce9981f229953f6d618ebe66065bc1/optimum/onnxruntime/modeling_ort.py#L178

If users specify the device, the value will be overwrite.

michaelbenayoun (approved):

PR #616, opened 2022-12-19
Use optimum exporter for current weight sharing test

JingyaHuang:
# What does this PR do?

see title.


michaelbenayoun (approved):
LGTM!

I plan on adding "post_export_onnx_optimizations_pass" or something like that to ONNX Config, so that we can perform these optimizations automatically!

Issue #615, opened 2022-12-19
Shall we set diffusers as soft dependency for onnxruntime module?


JingyaHuang:
It seems a little bit strange for me that we need to have diffusers for doing sequence classification.

### System Info

```shell
Dev branch of Optimum
```

### Who can help?

@echarlaix @JingyaHuang 

### Reproduction

```python
from optimum.onnxruntime import ORTModelForSequenceClassification
```

### Error message

```
RuntimeError: Failed to import optimum.onnxruntime.modeling_ort because of the following error (look up to see its traceback):
No module named 'diffusers'
```

### Expected behavior

Be able to do sequence classification without diffusers.

### Contribution

I can open a PR to make diffusers a soft dependency

JingyaHuang:
What do you think about that? @echarlaix

PR #619, opened 2022-12-20
Remove diffusers dependency in onnxruntime code

fxmarty:
# What does this PR do?

As per title.

Fixes https://github.com/huggingface/optimum/issues/615


optimum/exporters/onnx/convert.py
```diff
@@ -51,11 +51,15 @@ def check_dummy_inputs_are_allowed(
...
-    )
+
+    classinfo = (PreTrainedModel,)
+    if is_diffusers_available():
+        classinfo += (ModelMixin,)
+
+    if "diffusers" in str(model.__class__) and not is_diffusers_available():
+        raise ImportError("The pip package `diffusers` is required to export stable diffusion models to ONNX.")
+
+    forward = model.forward if is_torch_available() and issubclass(type(model), classinfo) else model.call
```
fxmarty, line 62: this seems very verbose and redundant, not sure if the check `issubclass` is necessary? @michaelbenayoun @echarlaix

michaelbenayoun: Yes, let's simplify this as you suggest!

optimum/exporters/onnx/convert.py
```diff
@@ -154,17 +158,22 @@ def validate_model_outputs(
...
 
-    framework = (
-        "pt" if is_torch_available() and issubclass(type(reference_model), (PreTrainedModel, ModelMixin)) else "tf"
-    )
+    classinfo = (PreTrainedModel,)
+    if is_diffusers_available():
+        classinfo += (ModelMixin,)
+
+    if "diffusers" in str(reference_model.__class__) and not is_diffusers_available():
+        raise ImportError("The pip package `diffusers` is required to validate stable diffusion ONNX models.")
```
michaelbenayoun, line 166: Maybe make it a function if this is going to be repeated a few times in the functions of this file.

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -623,7 +623,7 @@ def _from_transformers(
         task: Optional[str] = None,
     ) -> "ORTModelDecoder":
         if task is None:
-            task = cls._AUTOMODELS_TO_TASKS[cls.auto_model_class]
+            task = cls._AUTOMODELS_TO_TASKS[cls.auto_model_class.__name__]
```
michaelbenayoun, line 626: Maybe we can make a method that would do that, and avoid this, which is "too aware" of the underlying implementation IMO.

Something like: 
```
cls._auto_models_to_task(auto_model_class)
```

optimum/exporters/onnx/convert.py
fxmarty: Should I just remove the check `issubclass(type(model), classinfo) ` ?

optimum/exporters/onnx/convert.py
fxmarty: I use it only here in the end

fxmarty: It is used only twice here, I'd say it's OK

setup.py
```diff
@@ -35,15 +35,13 @@
...
         "protobuf==3.20.1",
-        "diffusers",
     ],
     "onnxruntime-gpu": [
         "onnx",
         "onnxruntime-gpu>=1.9.0",
         "datasets>=1.2.1",
         "evaluate",
         "protobuf==3.20.1",
-        "diffusers",
```
echarlaix, line 46: Shouldn't we remove this dependency from `"exporters-tf"` and `"exporters"` as well ?

echarlaix (commented):
Thanks for taking care of this @fxmarty !!

setup.py
fxmarty: yes

michaelbenayoun (approved):
LGTM

JingyaHuang (approved):
LGTM

PR #622, opened 2022-12-20
Audio models support with `optimum.exporters.onnx`

michaelbenayoun:
# What does this PR do?

Adds support for most of the audio models supported in Transformers for the ONNX export in `optimum.exporters`:

- Audio Spectrogram Transformer
- Hubert
- SEW
- SEW-D => Not passing yet
- Speech2Text
- UniSpeech
- UniSpeech-SAT
- Wav2Vec2
- Wav2Vec2-Conformer
- WavLM
- Data2Vec Audio

Remains to be done:

- [ ] M-CTC-T: the export could work, but an operation seems to not be supported by ONNX for the moment
- [ ] Speech2Text2 required the support for [Speech Encoder Decoder Models](https://huggingface.co/docs/transformers/v4.25.1/en/model_doc/speech-encoder-decoder#speech-encoder-decoder-models)
- [ ] Add the corresponding `ORTModelForXXX` classes.


optimum/exporters/onnx/model_configs.py
```diff
@@ -728,6 +730,94 @@ def generate_dummy_inputs(self, framework: str = "pt"):
...
+
+class UniSpeechSATOnnxConfig(HubertOnnxConfig):
+    pass
+
+
+class WavLMOnnxConfig(HubertOnnxConfig):
+    pass
+
+
+class ASTDummyAudioInputGenerator(DummyAudioInputGenerator):
```
michaelbenayoun, line 765: It seems that it is often required to overwrite the behavior of `DummyAudioInputGenerator` => Also needed [here](https://github.com/huggingface/optimum/pull/622/files#diff-cf7d52b6c0ee70bc45f40c2c40ac99b49dbf02f2fdb0a141e9483f764126e05eR809).

Maybe we can make it more generic?
WDTY @mht-sharma @fxmarty

mht-sharma: Possible did not look enough models when added for whisper.

AFAIK: There are 2 types of inputs in audio models

1. `input_features` (mel spectogram): Shapes (batch_size, feature_size, sequence_length) & (batch_size, sequence_length, feature_size).
2. `input_values` (raw speech): Shapes (batch_size, sequence_length)


In the current `AudioInputGenerator` we have `nb_max_frames` (terminology only used within feature extractor, docs refer to this as sequence length) for the sequence_length part of the `input_features`. Let's remove that and we can just use the `sequence_length` itself.

For getting the feature size, we can use something like: 
```python
NormalizedAudioConfig.with_args(
        feature_size="num_mel_bins", // or "input_features_per_channel"
    )
```

Handling `input_features` would be tricky as it has 2 shapes for same name. We can make one of them default and create new input generator for other shape when required?

optimum/exporters/onnx/model_configs.py
```diff
@@ -742,6 +744,94 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
...
+class Wav2Vec2ConformerOnnxConfig(HubertOnnxConfig):
+    pass
+
+
+class SEWOnnxConfig(HubertOnnxConfig):
+    pass
+
+
+class SEWDOnnxConfig(HubertOnnxConfig):
+    DEFAULT_ONNX_OPSET = 12
```
fxmarty, line 764: curiosity: why?

optimum/exporters/onnx/model_configs.py
```diff
@@ -742,6 +744,94 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
...
+class UniSpeechOnnxConfig(HubertOnnxConfig):
+    pass
+
+
+class UniSpeechSATOnnxConfig(HubertOnnxConfig):
+    pass
+
+
+class WavLMOnnxConfig(HubertOnnxConfig):
+    DEFAULT_ONNX_OPSET = 12
```
fxmarty, line 776: same question

optimum/exporters/tasks.py
```diff
@@ -517,11 +559,49 @@ class TasksManager:
...
+            "audio-frame-classification",
+            "audio-xvector",
+            onnx="Wav2Vec2OnnxConfig",
+        ),
+        "wav2vec2-conformer": supported_tasks_mapping(
+            "default",
+            "audio-ctc",
+            "audio-classification",
+            "audio-frame-classification",
+            "audio-xvector",
```
fxmarty, line 602: Not super familiar with all of those, but is this fine to have them in this mapping but not in `_TASKS_TO_AUTOMODELS`?

optimum/exporters/onnx/base.py
```diff
@@ -158,7 +158,11 @@ class OnnxConfig(ExportConfig, ABC):
         ),
         "sequence-classification": OrderedDict({"logits": {0: "batch_size"}}),
         "token-classification": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
-        "speech2seq-lm": OrderedDict({"logits": {0: "batch", 1: "sequence"}}),
+        "speech2seq-lm": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "audio-classification": OrderedDict({"logits": {0: "batch_size"}}),
+        "audio-frame-classification": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "audio-ctc": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
```
fxmarty, line 164: why do we need the two? they are the same than `speech2seq-lm`

fxmarty (approved):
looks good as long as the tests pass! Left some understanding questions

optimum/exporters/onnx/model_configs.py
michaelbenayoun: The min (or clip) operator do not support in64 type before this opset.

optimum/exporters/onnx/model_configs.py
michaelbenayoun: The min (or clip) operator do not support in64 type before this opset.

optimum/exporters/onnx/base.py
michaelbenayoun: Are they? I am not sure they are, at least they correspond to different auto classes.

fxmarty: Yes you're right! cf your update & above comment

optimum/exporters/tasks.py
michaelbenayoun: You are right, I forgot but added them!

PR #629, opened 2022-12-21
Fix whisper export

### LINKED ISSUE #600: Whisper model export failure
### System Info

```shell
Optimum - e8d987766c5ff5080c8b1452041427a4b675e701
Transformers - 227f8a708a7f8cf000770bdc3ecc8971d6012b1a
```

### Who can help?

_No response_

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

### Reproduction

The following command fails with error
```bash
python -m optimum.exporters.onnx --model="openai/whisper-tiny.en" --task=speech2seq-lm-with-past  output_whisper
```
**ERROR**
```bash
return self._sess.run(output_names, input_feed, run_options)
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid Feed Input Name:past_key_values.3.encoder.value
```

### Expected behavior

Should not fail

mht-sharma:
# What does this PR do?

Fixes the whisper export

The current export for whisper fails because the cross attention key values are not exported as input in the ONNX model after the new condition introduced in the [transformers@97a51](https://github.com/huggingface/transformers/commit/97a51b0c7d483cdf13ea878a987f9aa1c9eecc91).

The error occurs due to incorrect dummy input generation for cross attention key values for export. The PR fixes the same.

Fixes #600 

## Before submitting

### DIFF
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -751,6 +751,21 @@ class WhisperOnnxConfig(AudioToTextOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedSeq2SeqConfig
     ATOL_FOR_VALIDATION = 1e-3
 
+    def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGenerator"]:
+        dummy_inputs_generators = super()._create_dummy_input_generator_classes(**kwargs)
+
+        if self._behavior is ConfigBehavior.MONOLITH:
+            dummy_seq2seq_past_key_values_generator = self.DUMMY_INPUT_GENERATOR_CLASSES[2](
+                self.task,
+                self._normalized_config,
+                batch_size=dummy_inputs_generators[0].batch_size,
+                encoder_sequence_length=dummy_inputs_generators[0].nb_max_frames // 2,
+                **kwargs,
+            )
+            dummy_inputs_generators[2] = dummy_seq2seq_past_key_values_generator
+
+        return dummy_inputs_generators
+
 
 class MobileNetV1OnnxConfig(VisionOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedVisionConfig
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -753,7 +753,7 @@ class WhisperOnnxConfig(AudioToTextOnnxConfig):
 
     def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGenerator"]:
         dummy_inputs_generators = super()._create_dummy_input_generator_classes(**kwargs)
-
+# The generated encoder_hidden_states for Whisper has dimensions (batch_size, encoder_sequence_length / 2, hidden_size). Therefore, the sequence length is updated to generate the proper cross attention KVS.
         if self._behavior is ConfigBehavior.MONOLITH:
             dummy_seq2seq_past_key_values_generator = self.DUMMY_INPUT_GENERATOR_CLASSES[2](
                 self.task,
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -758,7 +758,6 @@ class WhisperOnnxConfig(AudioToTextOnnxConfig):
             dummy_seq2seq_past_key_values_generator = self.DUMMY_INPUT_GENERATOR_CLASSES[2](
                 self.task,
                 self._normalized_config,
-                batch_size=dummy_inputs_generators[0].batch_size,
                 encoder_sequence_length=dummy_inputs_generators[0].nb_max_frames // 2,
                 **kwargs,
             )
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -753,7 +753,10 @@ class WhisperOnnxConfig(AudioToTextOnnxConfig):
 
     def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGenerator"]:
         dummy_inputs_generators = super()._create_dummy_input_generator_classes(**kwargs)
-# The generated encoder_hidden_states for Whisper has dimensions (batch_size, encoder_sequence_length / 2, hidden_size). Therefore, the sequence length is updated to generate the proper cross attention KVS.
+        # The generated encoder_hidden_states for Whisper has dimensions
+        # (batch_size, encoder_sequence_length / 2, hidden_size). Therefore,
+        # the sequence length is updated to generate the proper cross attention
+        # KVS in monolith case.
         if self._behavior is ConfigBehavior.MONOLITH:
             dummy_seq2seq_past_key_values_generator = self.DUMMY_INPUT_GENERATOR_CLASSES[2](
                 self.task,


mht-sharma:
@michaelbenayoun you mentioned the same error occurs for other model. Are there enough models with the same condition to make it generic in the `AudioInputGenerator`?

michaelbenayoun:
I think it also happens with `Speech2Text` in the PR I am working on. So maybe we should make it part of `AudioToTextOnnxConfig` as you suggest.

mht-sharma:
> LGTM!
> Why is this solving the issue? Basically what was causing it? Asking to know how to fix it if I encounter it.

In transformers a new condition is added `and past_key_value[0].shape[2] == key_value_states.shape[1]` in the line below to use the cross attention key values.

```python
 if ( is_cross_attention
       and past_key_value is not None
       and past_key_value[0].shape[2] == key_value_states.shape[1] 
    ):

# Added link to transformers change in the PR description.
```
Earlier the strict condition was not there so any shape  (though wrong) of key values was able to export the model. 

In the dummy input generator we are reusing the `sequence_length` of encoder for the encoder_sequence_length for `past_key_values` which is true for most models.

But in case of `Whisper`, the encoder reduces the dimension to `sequence_length // 2` and we need to use the same for cross attention kvs. Hence this has to be handled separately.

mht-sharma:
> I think it also happens with `Speech2Text` in the PR I am working on. So maybe we should make it part of `AudioToTextOnnxConfig` as you suggest.

If its just these two models then lets handle them separately. As `AudioToTextOnnxConfig` config is quite generic for audio models

optimum/exporters/onnx/model_configs.py
```diff
@@ -731,3 +731,18 @@ def generate_dummy_inputs(self, framework: str = "pt"):
...
+    def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGenerator"]:
+        dummy_inputs_generators = super()._create_dummy_input_generator_classes(**kwargs)
+
+        if self._behavior is ConfigBehavior.MONOLITH:
+            dummy_seq2seq_past_key_values_generator = self.DUMMY_INPUT_GENERATOR_CLASSES[2](
+                self.task,
+                self._normalized_config,
+                batch_size=dummy_inputs_generators[0].batch_size,
+                encoder_sequence_length=dummy_inputs_generators[0].nb_max_frames // 2,
+                **kwargs,
```
fxmarty, line 744: beware that batch_size can be in `kwargs`

optimum/exporters/onnx/model_configs.py
```diff
@@ -731,3 +731,18 @@ def generate_dummy_inputs(self, framework: str = "pt"):
 class WhisperOnnxConfig(AudioToTextOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedSeq2SeqConfig
     ATOL_FOR_VALIDATION = 1e-3
+
+    def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGenerator"]:
+        dummy_inputs_generators = super()._create_dummy_input_generator_classes(**kwargs)
+
+        if self._behavior is ConfigBehavior.MONOLITH:
```
fxmarty, line 738: why is it in the monolith case? In https://github.com/huggingface/optimum/pull/581/files I think the issue was not only with monolith

fxmarty:
Super cool! I think your solution is cleaner than mine in https://github.com/huggingface/optimum/pull/581/files . I can adapt in a next PR so as not to pollute the `generate_dummy_inputs` itself, but rather the `_create_dummy_input_generator_classes`. What do you think?

optimum/exporters/onnx/model_configs.py
fxmarty: actually I would advise rerunning the tests

optimum/exporters/onnx/model_configs.py
fxmarty: Edit:

Before this PR, have no issue with
```
python3 -m optimum.exporters.onnx --model openai/whisper-tiny.en --task speech2seq-lm-with-past --for-ort out_whisper
python3 -m optimum.exporters.onnx --model openai/whisper-tiny.en --task speech2seq-lm --for-ort out_whisper
python3 -m optimum.exporters.onnx --model openai/whisper-tiny.en --task speech2seq-lm out_whisper
```

but have an issue with
```
python3 -m optimum.exporters.onnx --model openai/whisper-tiny.en --task speech2seq-lm-with-past out_whisper
```

which raises:

```
An error occured with error [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid Feed Input Name:past_key_values.3.encoder.value.
```

optimum/exporters/onnx/model_configs.py
```diff
@@ -751,6 +751,20 @@ class WhisperOnnxConfig(AudioToTextOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedSeq2SeqConfig
     ATOL_FOR_VALIDATION = 1e-3
 
+    def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGenerator"]:
+        dummy_inputs_generators = super()._create_dummy_input_generator_classes(**kwargs)
+
+        if self._behavior is ConfigBehavior.MONOLITH:
```
fxmarty, line 757:
```suggestion
        # in the monolith case, we need to hint that the sequence length is halved in the decoder
        if self._behavior is ConfigBehavior.MONOLITH:
```

or something like this

optimum/exporters/onnx/model_configs.py
```diff
@@ -751,6 +751,20 @@ class WhisperOnnxConfig(AudioToTextOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedSeq2SeqConfig
     ATOL_FOR_VALIDATION = 1e-3
 
+    def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGenerator"]:
+        dummy_inputs_generators = super()._create_dummy_input_generator_classes(**kwargs)
+
```
mht-sharma, line 756:
```suggestion
# The generated encoder_hidden_states for Whisper has dimensions (batch_size, encoder_sequence_length / 2, hidden_size). Therefore, the sequence length is updated to generate the proper cross attention KVS.
```

optimum/exporters/onnx/model_configs.py
```diff
@@ -751,6 +751,20 @@ class WhisperOnnxConfig(AudioToTextOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedSeq2SeqConfig
     ATOL_FOR_VALIDATION = 1e-3
 
+    def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGenerator"]:
+        dummy_inputs_generators = super()._create_dummy_input_generator_classes(**kwargs)
+# The generated encoder_hidden_states for Whisper has dimensions (batch_size, encoder_sequence_length / 2, hidden_size). Therefore, the sequence length is updated to generate the proper cross attention KVS.
+        if self._behavior is ConfigBehavior.MONOLITH:
```
mht-sharma, line 757:
```suggestion
    def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGenerator"]:
        dummy_inputs_generators = super()._create_dummy_input_generator_classes(**kwargs)
        # The generated encoder_hidden_states for Whisper has dimensions
        # (batch_size, encoder_sequence_length / 2, hidden_size). Therefore,
        # the sequence length is updated to generate the proper cross attention
        # KVS in monolith case.
        if self._behavior is ConfigBehavior.MONOLITH:
```

optimum/exporters/onnx/model_configs.py
mht-sharma: Added a comment

optimum/exporters/onnx/model_configs.py
mht-sharma: removed the batch_szie

fxmarty (approved):
LGTM!

Issue #626, opened 2022-12-20
BetterTransformer slower than normal pytorch


nreimers:
### System Info

```shell
optimum==1.5.2
torch==1.13.1
transformers==4.25.1
cuda 11.6
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

I was hoping to get some inference speedup using BetterTransformer following the HF / pytorch blog post when running larger bulk embedding jobs (e.g. encoding 1M docs with 256 sequence length).

But in fact it is slightly slower than normal Pytorch inference on a DGX machine with V100 GPU.

Here the code:
```python
from transformers import AutoModel, AutoTokenizer
from optimum.bettertransformer import BetterTransformer
import torch
from argparse import ArgumentParser
import tqdm
import time

def benchmark_model(model, encoded):
    #Warmup
    for _ in range(10):
        ret = model(**encoded)

    start_time = time.time()
    with torch.inference_mode():
        for _ in tqdm.trange(args.docs):
            ret = model(**encoded)
    total_time = time.time() - start_time
    total_docs = args.batch_size * args.docs
    print("Docs/Second: {:.2f}".format(total_docs/total_time))


def main(args):
    device="cuda"

    model_name = "bert-base-uncased"
    tokenizer = AutoTokenizer.from_pretrained(model_name)

    doc = "This is a doc "*args.max_seq_length
    docs = [doc]*args.batch_size

    encoded = tokenizer(docs, truncation=True, padding=True, return_tensors="pt", max_length=args.max_seq_length).to(device)
    print("Input IDs:", encoded['input_ids'].shape)

    # Normal transformer model
    normal_model = AutoModel.from_pretrained(model_name).to(device).half()

    print("Normal HF model")
    benchmark_model(normal_model, encoded)


    #BetterTransformers
    bt_model = BetterTransformer.transform(normal_model, keep_original_model=False)
    print(bt_model)

    print("BetterTransformer model")
    benchmark_model(bt_model, encoded)
    


if __name__ == "__main__":
    parser = ArgumentParser()
    parser.add_argument("--batch_size", type=int, default=128)
    parser.add_argument("--model", default='bert-base-uncased')
    parser.add_argument("--docs", type=int, default=200) #Number of batches
    parser.add_argument("--max_seq_length", type=int, default=256)
[...]

PR #637, opened 2022-12-22
[`BT`] Fix slowdowns

younesbelkada:
# What does this PR do?

Fixes https://github.com/huggingface/optimum/issues/626

In the current implementation, BetterTransformer keeps the hidden states un-nested when the attention mask is full-ones: https://github.com/huggingface/optimum/blob/d3c282962ecba7a16280eba0770a5479162ce2f4/optimum/bettertransformer/models/encoder_models.py#L212 

This leads to important computational overhead, slowing down BT models by up to 2

This PR addresses this issue, by forcing the hidden states to be always nested

cc @fxmarty @michaelbenayoun


fxmarty:
Fixed. Note that Albert nest/unnest for all layers, and can be slower even with this fix.

Test it with:

```python
TEST_LEVEL=1 pytest tests/bettertransformer/test_gpu.py -s --exitfirst
```

fxmarty (approved):
LGTM

younesbelkada:
Thank you very much @fxmarty for taking care of this! 💪🏻

PR #642, opened 2022-12-23
Fix item access of some _TASKS_TO_AUTOMODELS

fxmarty:
Unspotted before because `task_to_automodel` was used instead of `tasks_to_automodel`, so ctrl+f did not do the job

Will do a patch for this next week


fxmarty:
Will include it in the patch

PR #644, opened 2022-12-23
do not merge bis

fxmarty:
just to check the build PR doc workflow works now

we may expect the workflow to fail here if the build intel doc uses main, which is probably the case

```
doc-builder preview optimum.intel docs/source/
```

works


fxmarty:
all good

Issue #606, opened 2022-12-18
[Bug] tensorrt engine cannot store locally


PoodleWang:
### System Info

```shell
Cannot store the tensorrt engine locally.
I used the PR: https://github.com/huggingface/optimum/pull/586.  I am not sure whether this bug happened on the production.
```

### Who can help?

@JingyaHuang 

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

```Python
model = ORTModelForCausalLM.from_pretrained(
"Salesforce/codegen-6B-mono/",
from_transformers=True,
provider="TensorrtExecutionProvider"
)
model.save_pretrained("./codegen-6B-mono-onnx")
```

above code generate the below structure:
--codegen-6B-mono-onnx
----config.json
----decoder_model.onnx
----decoder_model.onnx_data
----decoder_with_past_model.onnx
----decoder_with_past_model.onnx_data

```Python
provider_options = {
"trt_engine_cache_enable": 1,
"trt_engine_cache_path": "/opt/tiger/genius/checkpoints/codegen-6B-mono/trt",
}
model = ORTModelForCausalLM.from_pretrained(
"/opt/tiger/genius/checkpoints/codegen-6B-mono/onnx",
provider="TensorrtExecutionProvider",
provider_options = provider_options,
)
```

### Expected behavior

LOG:
2022-12-18 13:25:25.887143: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations: AVX2 AVX512F AVX512_VNNI FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2022-12-18 13:25:26.000533: I tensorflow/core/util/port.cc:104] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable TF_ENABLE_ONEDNN_OPTS=0.
[...]

PR #653, opened 2022-12-29
Fix provider options when several providers are passed

fxmarty:
When several providers are passed to the `InferenceSession`, which is the case when `TensorrtExecutionProvider` is chosen, the `provider_options` argument needs to be of the same length than `providers`, otherwise raising:

```
EP Error using ['TensorrtExecutionProvider', 'CUDAExecutionProvider']
Falling back to ['CUDAExecutionProvider', 'CPUExecutionProvider'] and retrying.
```

Reference: https://onnxruntime.ai/docs/api/python/api_summary.html#inferencesession

This was untested up to now. Still need to add a test for this PR.

In a next PR: remove all the code duplication for `load_model()` in `modeling_ort.py`, `modeling_decoder.py`, `modeling_seq2seq.py`. But I won't do it in this PR.

This should fix https://github.com/huggingface/optimum/issues/606


regisss (approved):
LGTM!

Issue #631, opened 2022-12-21
Add tests for error raise in case the TRT/CUDA dependencies are not met


fxmarty:
### Feature request

@JingyaHuang reported that https://github.com/huggingface/optimum/pull/514 may not be throwing errors as expected.

We should add tests for this.

### Motivation

/

### Your contribution

/

fxmarty:
As I used the environment variables set in `capi/_ld_preload.py`, it could be that the checks fail on Windows.

PR #659, opened 2023-01-02
Add test for missing execution providers error messages

fxmarty:
# What does this PR do?

Add test for https://github.com/huggingface/optimum/pull/514

Fixes https://github.com/huggingface/optimum/issues/631

@JingyaHuang I did not find any issue with the messages in https://github.com/huggingface/optimum/pull/514 . Could you give details on the env where the error messages were not showing? Specifically, which error message was not showing?


tests/onnxruntime/test_modeling.py
```diff
@@ -251,6 +254,58 @@ def test_model_on_cpu_str(self):
...
+
+                self.assertTrue(
+                    f"Asked to use {provider}, but `onnxruntime-gpu` package was not found." in str(cm.exception)
+                )
+        else:
+            logger.info("Skipping CUDAExecutionProvider/TensorrtExecutionProvider without `onnxruntime-gpu` test")
+
+        # need to install first onnxruntime-gpu, then onnxruntime for this test to pass,
+        # thus overwritting onnxruntime/capi/_ld_preload.py
+        if is_onnxruntime_installed and is_onnxruntime_gpu_installed:
```
JingyaHuang, line 284: Are you testing that double installation of onnxruntime and onnxruntime-gpu won't work for Cuda ep and TRT ep? Why should we add this test?

tests/onnxruntime/test_modeling.py
```diff
@@ -251,6 +254,58 @@ def test_model_on_cpu_str(self):
...
+        else:
+            logger.info("Skipping double onnxruntime + onnxruntime-gpu install test")
+
+        # LD_LIBRARY_PATH can't be set at runtime,
+        # see https://stackoverflow.com/questions/856116/changing-ld-library-path-at-runtime-for-ctypes
+        # testing only for TensorRT as having ORT_CUDA_UNAVAILABLE is hard
+        if is_onnxruntime_gpu_installed:
+            for provider in ["TensorrtExecutionProvider"]:
+                out = subprocess.run(
+                    f"CUDA_PATH='' LD_LIBRARY_PATH='' python tests/onnxruntime/load_model.py {provider}",
```
JingyaHuang, line 301: What is the goal of this test? And where comes load_model.py?

JingyaHuang (commented):
Thanks for adding the test, just left a few points that I feel confused about.

tests/onnxruntime/test_modeling.py
fxmarty: We check that in case the onnxruntime installation is messed up, a meaningful error message is raised.

JingyaHuang:
What error message are you mentioning? @fxmarty

fxmarty:
@JingyaHuang edited my answer in the thread

tests/onnxruntime/test_modeling.py
fxmarty: Added the missing load_model.py, thanks. This is to test that a meaningful error message is raised in case TRT requirements can not be loaded.

JingyaHuang: Do we really put it in a separate script as it is just a one-liner?

fxmarty: right, will edit!

fxmarty: Should be all good now.

JingyaHuang (approved):
LGTM! Thanks for working on the error message.

PR #660, opened 2023-01-02
Use nvcr.io/nvidia/tensorrt image for GPU tests

fxmarty:
GPU test may need TensorRT, so let's use an image that has TensorRT pre-installed.

Still need to test that everything runs fine before merging.

### DIFF
--- a/.github/workflows/test_onnxruntime_gpu.yml
+++ b/.github/workflows/test_onnxruntime_gpu.yml
@@ -4,6 +4,8 @@ on:
   workflow_dispatch:
   schedule:
     - cron: 0 7 * * * # every day at 7am
+  pull_request:
+    types: [labeled]
   # uncomment to enable on PR merge on main branch:
   #push:
   #  branches:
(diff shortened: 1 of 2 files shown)


fxmarty:
Passing on the DGX.

.github/workflows/test_onnxruntime_gpu.yml
```diff
@@ -2,11 +2,14 @@ name: ONNX Runtime / Test GPU
 
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - main
   schedule:
     - cron: 0 7 * * * # every day at 7am
+  pull_request:
+    types: [labeled]
```
regisss, line 8: It can be any label right?

fxmarty: Yes, I added it so that the test can be triggered on a PR. I don't know if we can specify a specific tag, do you know?

JingyaHuang (approved):
LGTM, thanks for enabling it!

PR #672, opened 2023-01-05
VisualBertOnnxConfig created + inputs implemented

mszsorondo:
# What does this PR do?
Implements VisualBert for Onnx

## Before submitting

### DIFF
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -72,6 +72,26 @@ class BertOnnxConfig(TextEncoderOnnxConfig):
             "token_type_ids": dynamic_axis,
         }
 
+class VisualBertOnnxConfig(BertOnnxConfig):
+    NORMALIZED_CONFIG_CLASS = NormalizedTextAndVisionConfig
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, DummyVisionInputGenerator)
+    ATOL_FOR_VALIDATION = 1e-4
+
+    @property
+    def inputs(self) -> Mapping[str, Mapping[int, str]]:
+        dynamic_axis = {0: "batch_size", 1: "sequence_length"}
+        result = {
+            "input_ids": dynamic_axis,
+            "attention_mask": dynamic_axis,
+            "token_type_ids": dynamic_axis,
+            "visual_embeds": {0: "batch_size", 1:"visual_seq_length", 2: "visual_embedding_dim"},
+            "visual_token_type_ids": {0: "batch_size", 1:"visual_seq_length"},
+            "visual_attention_mask": {0: "batch_size", 1:"visual_seq_length"},
+
+        }
+        if(self.task == "visual-question-answering"):
+            result.update({"region_to_phrase_position": {0:"batch_size", 1:"total_sequence_length"}})
+        return result
 
 class AlbertOnnxConfig(BertOnnxConfig):
     pass


mszsorondo:
Hi @fxmarty sorry for de delay. I have already exported visualbert VQA models, but there are a couple hitches I think that probably should be solved. 

1. This seems to work for vqa checkpoints (``` "uclanlp/visualbert-vqa" ``` ,``` "uclanlp/visualbert-vqa-coco-pre" ```), but I coded it thinking about every VisualBert checkpoint and task, so I introduced inputs and dummy inputs depending on self.task, but the thing is that the script from "Adding support for an unsupported architecture" does not specify how to make this export for varying tasks, so naturally some doubts arise like:
a.  what happens for downstream tasks? how 
b. Should I separately export their checkpoints to ONNX with the same procedure (in this case, visualbert-nlvr2 and visualbert-vcr)? VisualBert has different checkpoints for different tasks. 
c. Should I add visualbert as supported in ``` tests/exporters/test_onnx_export.py ``` after doing so with each checkpoint?

3. The HF docs specify the initialization of downstream VisualBert checkpoints with the VisualBert<downstream task name> class and not with the basic VisualBertModel and the task parameter as with others, this sets the self.task from those downstream models config to 'default'. This implies that VisualBertOnnxConfig and DummyVisualBertInputGenerator will always set the inputs as if the task in hand is 'default' as well. Maybe this is not a problem and it is expected for the downstream model classes to receive default inputs, but I would really appreciate if you throw some light to me here.

4. From the tests that failed so far, I can't find anything related with VisualBERT (or at least not explicitly), this is probably related with the fact that I haven't implemented the tests yet. Is there a sample test suite from other models (say BERT) which I could use? 

Thanks in advance

fxmarty:
Hi @mszsorondo, thank you for working on this! I haven't reviewed yet so just answering your questions:

>  b. Should I separately export their checkpoints to ONNX with the same procedure (in this case, visualbert-nlvr2 and visualbert-vcr)? VisualBert has different checkpoints for different tasks.

No, if possible you should use in the tests a single model for all visualbert modalities, for example https://huggingface.co/hf-internal-testing/tiny-random-VisualBertModel . If it is not possible (as in the link below for perceiver), it is fine to specify specific models for specific tasks where the base models can not be used to initialize fine-tuned random models (in which case the output is garbage, but it's fine since the purpose is just testing if the output match PyTorch), but if possible avoid it

>  c. Should I add visualbert as supported in tests/exporters/test_onnx_export.py after doing so with each checkpoint?

Not sure what you mean exactly here, but if different tasks require different models/architectures, you can do as: https://github.com/huggingface/optimum/blob/bc5e10fb15b04f79ce25dd99823ec9c49b984e1c/tests/exporters/exporters_utils.py#L74-L77

In my opinion for the tests to work you only need to add a checkpoint of visualbert in `tests/exporters/exporters_utils.py`.

> The HF docs specify the initialization of downstream VisualBert checkpoints with the VisualBert<downstream task name> class and not with the basic VisualBertModel and the task parameter as with others, this sets the self.task from those downstream models config to 'default'.

I'm not sure I understand what you mean here. The task can be passed either by specifying `--task` or having it auto-detected: https://github.com/huggingface/optimum/blob/bc5e10fb15b04f79ce25dd99823ec9c49b984e1c/optimum/exporters/onnx/__main__.py#L59-L67 Do you mean that the auto-inferred task is always `default`? It should not be the case I think.

> From the tests that failed so far, I can't find anything related with VisualBert (or at least not explicitly), this is probably related with the fact that I haven't implemented the tests yet. Is there a sample test suite from other models (say BERT) which I could use?

Are you able to run the checks on the PR / are the checks run when you push on the PR? It could very well be that the github workflows are outdated if they are not automatically re-run (as you are an external contributor. I'll look into if we can modify the default so that external contributors get their workflow run). Otherwise you just need to have the `exporters` and `cli` tests pass locally. Since they are somewhat big, if you run locally, you could just run them for visualbert: `pytest tests/exporters/test_*.py -k "visualbert"` should work. I'll upgrade the doc for this, thanks!

Looking at https://github.com/huggingface/optimum/blob/bc5e10fb15b04f79ce25dd99823ec9c49b984e1c/tests/exporters/test_onnx_export.py#L161 (which is a test name), you could try `pytest tests/exporters/test_onnx_export.py -k "bert_question_answering" -s --exitfirst` for an example of bert test. Or if it does not work, maybe `pytest tests/exporters/test_onnx_export.py -k "test_pytorch_export_025_bert_question_answering" -s --exitfirst`

I don't have a laptop at hand so I did not try these commands but hopefully they work.

mszsorondo:
Thank you for the response @fxmarty 

> No, if possible you should use in the tests a single model for all visualbert modalities (tasks), for example https://huggingface.co/hf-internal-testing/tiny-random-VisualBertModel.

So all the testing is simply done by adding it as the value for ```"visual-bert"``` to the ``` PYTORCH_EXPORT_MODELS_TINY ``` dict? I am a bit confused because the [guide](https://huggingface.co/docs/optimum/exporters/onnx/usage_guides/contribute) says:  
> Add your model architecture to the tests in tests/exporters/test_onnx_export.py

Which is sort of ambiguous, my interpretation of this now is that this actually refers to adding the key-value pair previously mentioned at the _exporters_utils.py_ file and that this dictionary was directly stored at _test_onnx_export.py_ by the time the guide was written. 
If I'm correct, I can make a PR for clarifying that sentence from the guide.

> Not sure what you mean exactly here, but if different tasks require different models/architectures, you can do as: 

I was asking exactly for what you answered :)

>Do you mean that the auto-inferred task is always default? It should not be the case I think.

Yes, when using the downstream VisualBert models for constructing VisualBertOnnxConfig the task parameter is always set to "default".

> Are you able to run the checks on the PR / are the checks run when you push on the PR?

they are automatically run at each push, yes 

> you could just run them for visualbert: pytest tests/exporters/test_*.py -k "visualbert" should work. I'll upgrade the doc for this, thanks!

Tried it, but there are lots of collecting and import errors. But after the last push checks end running, you'll see 8 errors. None of them make any reference to visual-bert, all make reference to albert-base-v1, albert-base-v2 and tiny-random-vision_perceiver_conv.

mszsorondo:
Thank you for the response @fxmarty 

> No, if possible you should use in the tests a single model for all visualbert modalities (tasks), for example https://huggingface.co/hf-internal-testing/tiny-random-VisualBertModel.

So all the testing is simply done by adding it as the value for ```"visual-bert"``` to the ``` PYTORCH_EXPORT_MODELS_TINY ``` dict? I am a bit confused because the [guide](https://huggingface.co/docs/optimum/exporters/onnx/usage_guides/contribute) says:  
> Add your model architecture to the tests in tests/exporters/test_onnx_export.py

Which is sort of ambiguous, my interpretation of this now is that this actually refers to adding the key-value pair previously mentioned at the _exporters_utils.py_ file and that this dictionary was directly stored at _test_onnx_export.py_ by the time the guide was written. 
If I'm correct, I can make a PR for clarifying that sentence from the guide.

> Not sure what you mean exactly here, but if different tasks require different models/architectures, you can do as: 

I was asking exactly for what you answered :)

>Do you mean that the auto-inferred task is always default? It should not be the case I think.

Yes, when using the downstream VisualBert models for constructing VisualBertOnnxConfig the task parameter is always set to "default".

> Are you able to run the checks on the PR / are the checks run when you push on the PR?

they are automatically run at each push, yes 

> you could just run them for visualbert: pytest tests/exporters/test_*.py -k "visualbert" should work. I'll upgrade the doc for this, thanks!

I already found some errors now, I'll work on them and pin you when solved. Still, there 9/10 checks are completely unrelated with visual-bert... most of them refer to albert-base-v1, albert-base-v2 or tiny-random-vision_perceiver_conv.

fxmarty:
> Which is sort of ambiguous, my interpretation of this now is that this actually refers to adding the key-value pair previously mentioned at the exporters_utils.py file and that this dictionary was directly stored at test_onnx_export.py by the time the guide was written.
> If I'm correct, I can make a PR for clarifying that sentence from the guide.

Exactly, yes. Sorry for the confusion, feel free to open a PR to clarify!

fxmarty:
Can you add the relevant new inputs in https://github.com/huggingface/optimum/blob/bc5e10fb15b04f79ce25dd99823ec9c49b984e1c/optimum/commands/export/onnx.py#L88-L145 ? Some tests are failing because of this I think.

fxmarty:
Can you add the relevant new tasks in `_TASKS_TO_AUTOMODELS` and `_TASKS_TO_LIBRARY`? Some tests look to be failing due to this. https://github.com/huggingface/optimum/blob/bc5e10fb15b04f79ce25dd99823ec9c49b984e1c/optimum/exporters/tasks.py#L88-L141

mszsorondo:
>Can you add the relevant new inputs in  [image] ... ? Some tests are failing because of this I think. You probably just need to add num_of_detection_patches

Done, but changed "num_of_detection_patches" by "visual_seq_length", which is the actual name of the input shape.

> Can you add the relevant new tasks in _TASKS_TO_AUTOMODELS and _TASKS_TO_LIBRARY? Some tests look to be failing due to this (KeyError: 'Unknown task). 

Added ``` visual-question-answering```, but "natural-language-for-visual-reasoning" and "region-to-phrase-alignment" have no AutoModel yet. I'll open a PR in transformers for this if there's no alternative, it doesn't seem like something big.

fxmarty:
Thank you!

> Added  visual-question-answering, but "natural-language-for-visual-reasoning" and "region-to-phrase-alignment" have no AutoModel yet. I'll open a PR in transformers for this if there's no alternative, it doesn't seem like something big.

Indeed it's an issue. I think we should have your PR merged even without the AutoModel class in transformers, I'll look into how we can solve this. cc @michaelbenayoun

optimum/utils/input_generators.py
```diff
@@ -53,6 +53,7 @@ def wrapper(*args, **kwargs):
     "width": 64,
     "height": 64,
     "num_channels": 3,
+    "visual_seq_length": 5,
     # audio
```
fxmarty, line 57: apparently `num_of_detection_patches` is missing here according to some non-passing tests

mszsorondo:
> apparently num_of_detection_patches is missing here according to some non-passing tests

Yes, this was happening because I forgot to replace it with the correct name in one last place (corrected in last commit). Lots of checks that raised errors are now passing.

>In parallel if you want, I'd suggest to open an issue in transformers and ping @ sgugger to see if it's fine for you to work on it to add the AutoModel for those tasks.

Good, I opened the issue in transformers

There's only one more thing I wanted to ask you about, you mentioned that
> Since they are somewhat big, if you run locally, you could just run them for visualbert: pytest tests/exporters/test_*.py -k "visualbert" should work

Tried it with "visualbert", "visual-bert" and "visual_bert" but all throw ```ERROR: file or directory not found: visualbert_default``` , tell me which alternatives are to try to test locally so that I can debug.

Thanks in advance

mszsorondo:
Ubuntu 20.04, python 3.9, pytest 7.1.1. Bert works fine, in verbose mode it shows how tests pass:
```shell
platform linux -- Python 3.9.12, pytest-7.1.1, pluggy-1.0.0 -- /home/marco/anaconda3/bin/python
cachedir: .pytest_cache
rootdir: /home/marco/Desktop/optimum, configfile: setup.cfg
plugins: anyio-3.5.0, xdist-3.1.0
collected 1140 items / 841 deselected / 299 selected                                                                                                                                                                          

tests/exporters/test_exporters_onnx_cli.py::OnnxExportTestCase::test_exporters_cli_pytorch_000_albert_default PASSED                                                                                                    [  0%]
tests/exporters/test_exporters_onnx_cli.py::OnnxExportTestCase::test_exporters_cli_pytorch_001_albert_masked_lm PASSED 
...
```

And so on.
I somehow managed to stop getting the previous error without changing the code. Now the only issue is that all tests are "deselected". I sense it is because there's no tests associated with the model

fxmarty:
@mszsorondo Right, you actually need to pass `pytest tests/exporters/test_onnx_export.py -k "visual_bert"`

optimum/exporters/tasks.py
```diff
@@ -105,6 +105,8 @@ class TasksManager:
             "audio-ctc": "AutoModelForCTC",
             "audio-xvector": "AutoModelForAudioXVector",
             "stable-diffusion": "StableDiffusionPipeline",
+            "visual-question-answering": "AutoModelForVisualQuestionAnswering",
```
fxmarty, line 108:
```suggestion
            "visual-question-answering": "AutoModelForQuestionAnswering",
```

following https://huggingface.co/uclanlp/visualbert-vqa

optimum/utils/input_generators.py
```diff
@@ -628,3 +629,46 @@ def generate(self, input_name: str, framework: str = "pt"):
...
+        self.visual_seq_length = visual_seq_length
+
+    def generate(self, input_name: str, framework: str = "pt"):
+        # TODO maybe the following should be checked with the model checkpoint_path and checking existance of for example "vqa" substring is better
+        if self.task in ["default", "visual-question-answering", "region-to-phrase-alignment",]:
+            visual_embedding_dim = 2048
+        elif self.task == "multiple-choice":
+            visual_embedding_dim = 512
+        elif self.task == "natural-language-for-visual-reasoning":
+            visual_embedding_dim = 1024
```
fxmarty, line 657: I agree with your comment, I think the input `visual_embedding_dim` should rather be passed in the `__init__` and depend on the model checkpoint.

I think a way to do this is to redefine in `VisualBertOnnxConfig` an attribute like `dynamic_input_shape = {"visual_embedding_dim": self._config.visual_embedding_dim}` in the init (and call the super() init).

Then, in `generate_dummy_inputs`, you could add something like:

```python
if self.hasattr("dynamic_input_shape"):
    for key in self.dynamic_input_shape:
        kwargs[key] = getattr(self, key)
```

optimum/exporters/tasks.py
fxmarty, line 108:
```suggestion
            "visual-question-answering": "VisualBertForQuestionAnswering",
```

We can get away with this for now (since visualbert is the only model supported for those tasks in the onnx export for now). You could do as well for the other tasks waiting for the AutoModel in transformers.

fxmarty:
> Yes, when using the downstream VisualBert models for constructing VisualBertOnnxConfig the task parameter is always set to "default".

It's an issue with the models on the Hub that are wrongly labeled. So it's not important for this PR.

Issue #664, opened 2023-01-03
Test ORTModel.from_pretrained() for all architectures


fxmarty:
### Feature request

Test against all architectures in `tests/onnxruntime/test_modeling.py`.

### Motivation

Sometimes we have issues reported that `ORTModel.from_pretrained(from_transformers=True)` fails. It's a sign our tests are too weak, notably the `MODEL_NAMES` and `SUPPORTED_ARCHITECTURES` in `tests/onnxruntime/test_modeling.py`.

### Your contribution

/

bhavnicksm:
Created an independent issue based on the problem I was facing in comment #555 --> #665

fxmarty:
Same issue https://github.com/huggingface/optimum/pull/669

PR #675, opened 2023-01-05
Test more architectures in ORTModel

fxmarty:
A lot of architectures were untested up to now.

Fixes https://github.com/huggingface/optimum/issues/664


tests/onnxruntime/test_modeling.py
```diff
@@ -1242,7 +1390,20 @@ def test_compare_to_io_binding(self, model_id):
 
 
 class ORTModelForCausalLMIntegrationTest(unittest.TestCase):
-    SUPPORTED_ARCHITECTURES = ("gpt2",)
+    SUPPORTED_ARCHITECTURES = (
+        "gpt2",
+        "bloom",
+        "codegen",
+        "bigbird_pegasus",
+        "bart",
```
JingyaHuang, line 1398: Why you put bart, mbart and marian here? They are seq2seq models.

fxmarty: This was just copy paste, basically because of 

image

As I understand reading `modeling_mbart.py`, there is a `MBartForCausalLM` that does not use the encoder, and a `MBartForConditionalGeneration` that does. I guess that's because of it?

fxmarty:
One remaining issue is that models are exported in many different subtests. I'll fix as well in this PR.

optimum/exporters/tasks.py
```diff
@@ -716,6 +717,17 @@ def get_supported_tasks_for_model_type(
         else:
             return TasksManager._SUPPORTED_MODEL_TYPE[model_type][exporter]
 
+    @staticmethod
+    def get_supported_model_type_for_task(task: str, exporter: str):
```
michaelbenayoun, line 721:
```suggestion
    def get_supported_model_type_for_task(task: str, exporter: str) -> List[str]:
```

tests/onnxruntime/test_modeling.py
```diff
@@ -68,36 +70,120 @@
...
+class ORTModelTestMixin(unittest.TestCase):
+    ARCH_MODEL_MAP = {}
+
+    @classmethod
+    def setUpClass(cls):
+        cls.onnx_model_dirs = {}
+
+    def _setup(self, model_args: Dict):
+        """
+        Export the PyTorch models to ONNX ahead of time to avoid multiple exports during the tests.
```
michaelbenayoun, line 160:
```suggestion
        Exports the PyTorch models to ONNX ahead of time to avoid multiple exports during the tests.
```

tests/onnxruntime/test_modeling.py
```diff
@@ -817,18 +934,47 @@ def test_compare_to_io_binding(self, model_arch):
...
+        "mobilebert",
+        # "perceiver",
+        "roberta",
+        "roformer",
+        "squeezebert",
+        "xlm",
+        "xlm_roberta",
+    ]
+
+    ARCH_MODEL_MAP = {
```
michaelbenayoun, line 970: What's the use of `ARCH_MODEL_MAP`?

tests/onnxruntime/test_modeling.py
```diff
@@ -1402,22 +1656,46 @@ def test_compare_generation_to_io_binding(self, model_arch):
...
+        "mobilenet_v1",
+        "mobilenet_v2",
+        "mobilevit",
+        # "perceiver",
+        "poolformer",
+        "resnet",
+        "segformer",
+        "swin",
+        "vit",
+    ]
```
michaelbenayoun, line 1675: Is it not possible to infer that list automatically?

tests/onnxruntime/test_modeling.py
fxmarty: Some architectures like `perceiver` can't have the same model used for different tasks (e.g. sequence classification and image classification). It is similar to https://github.com/huggingface/optimum/blob/6e1194133cbf25be6d55c6e77f5b9b133a43ec75/tests/exporters/exporters_utils.py#L74-L76

Currently this is not used because perceiver does not work with ORTModel but it can be useful in the future.

tests/onnxruntime/test_modeling.py
fxmarty: @michaelbenayoun I would propose in a next PR to introduce in `optimum/onnxruntime` a dict of officially supported (aka tested) architectures for ORTModel. We would then use in here. Does this sound good to you? I can as well add it to this PR directly if you prefer.

We can't use the ones from `exporters/tasks.py` because some of the models supported in the ONNX export are not in ORTModel.

michaelbenayoun: Alright, perfect!

PR #676, opened 2023-01-05
[`BetterTransformer`] Let's make the transformation invertible

younesbelkada:
This PR adds the `inverse_transform` support for BT-transformed models. 
In the future, PyTorch will support `BetterTransformer` for training. Therefore a user would want to first convert a model, train it with `BetterTransformer` as backend and save the trained model afterwards. Before this PR the keys of the model's state dict are changed, therefore saving the model out of the box would break the model architecture and support with `transformers`. This PR adds a new transformation `invert_transform` that will let users convert back the BT model in its original format so that they can push/save their model.

## TODOs:

- [x] stronger API regarding saving-related methods (block `save_pretrained`, `push_to_hub` etc for converted models)
- [x] leverage copy mechanism (`# Copied from ...`)
- [x] more tests?

cc @fxmarty @michaelbenayoun @HamidShojanazeri

### DIFF
--- a/optimum/bettertransformer/models/base.py
+++ b/optimum/bettertransformer/models/base.py
@@ -56,6 +56,9 @@ class BetterTransformerBaseLayer(nn.Module):
         self.embed_dim = None
         self.num_layers = None
         self.original_layers_mapping = {}
+        # some models does not have some attributes thus needs to be ignored
+        # e.g. whisper does not have self_attn.k_proj.bias but has self_attn.v_proj.bias & self_attn.q_proj.bias
+        self.keys_to_ignore = []
 
         # Get activation function
         for attr in KNOWN_ACTIVATION_ATTRIBUTES:
@@ -162,16 +165,21 @@ class BetterTransformerBaseLayer(nn.Module):
                 # retrieve the current weight
                 current_weight = getattr(self, modified_layer_key_names)
 
-                # split the current weight n chunks
+                # split the current weight n chunks - this is useful to split
+                # the qkv layers into q, k, v layers for example.
                 split_index = current_weight.shape[0] // len(original_layer_key_names)
                 for i, module in enumerate(original_layer_key_names):
-                    self._recurse_setattr(
-                        self.orig_layer, module, nn.Parameter(current_weight[i * split_index : (i + 1) * split_index])
-                    )
+                    if module not in self.keys_to_ignore:
+                        self._recurse_setattr(
+                            self.orig_layer,
+                            module,
+                            nn.Parameter(current_weight[i * split_index : (i + 1) * split_index]),
+                        )
             elif isinstance(original_layer_key_names, str):
-                self._recurse_setattr(
-                    self.orig_layer, original_layer_key_names, getattr(self, modified_layer_key_names)
-                )
+                if module not in self.keys_to_ignore:
+                    self._recurse_setattr(
+                        self.orig_layer, original_layer_key_names, getattr(self, modified_layer_key_names)
+                    )
             else:
                 raise ValueError(
                     f"Invalid type {type(modified_layer_key_names)} for `original_layers_mapping`",
--- a/optimum/bettertransformer/models/base.py
+++ b/optimum/bettertransformer/models/base.py
@@ -17,7 +17,7 @@ from typing import List, Optional
 import torch
 import torch.nn as nn
 
-from ...utils import logging
+from ...utils import logging, recurse_setattr
 
 
 KNOWN_ACTIVATION_ATTRIBUTES = ["hidden_act", "activation", "act_fn", "activation_function"]
@@ -144,16 +144,6 @@ class BetterTransformerBaseLayer(nn.Module):
                 " Please use `model.eval()` before running the model.",
             )
 
-    def _recurse_setattr(self, module, name, value):
-        r"""
-        A wrapper function to recursively set attributes to a module.
-        """
-        if "." not in name:
-            setattr(module, name, value)
-        else:
-            name, rest = name.split(".", 1)
-            self._recurse_setattr(getattr(module, name), rest, value)
-
     def _revert_back_to_original_module(self):
         r"""
         A wrapper function to replace the current layer with the previous non-BetterTransformer
@@ -170,16 +160,14 @@ class BetterTransformerBaseLayer(nn.Module):
                 split_index = current_weight.shape[0] // len(original_layer_key_names)
                 for i, module in enumerate(original_layer_key_names):
                     if module not in self.keys_to_ignore:
-                        self._recurse_setattr(
+                        recurse_setattr(
                             self.orig_layer,
                             module,
                             nn.Parameter(current_weight[i * split_index : (i + 1) * split_index]),
                         )
             elif isinstance(original_layer_key_names, str):
                 if module not in self.keys_to_ignore:
-                    self._recurse_setattr(
-                        self.orig_layer, original_layer_key_names, getattr(self, modified_layer_key_names)
-                    )
+                    recurse_setattr(self.orig_layer, original_layer_key_names, getattr(self, modified_layer_key_names))
             else:
                 raise ValueError(
                     f"Invalid type {type(modified_layer_key_names)} for `original_layers_mapping`",
--- a/optimum/bettertransformer/models/base.py
+++ b/optimum/bettertransformer/models/base.py
@@ -11,10 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from copy import deepcopy
+
 import torch
 import torch.nn as nn
-
-from copy import deepcopy
 
 from ...utils import logging
 
@@ -62,7 +62,7 @@ class BetterTransformerBaseLayer(nn.Module):
             if hasattr(config, attr):
                 self.num_layers = getattr(config, attr)
                 break
-        
+
         if old_layer is not None:
             # Last step, store the old module skeleton by copying the old module and putting
             # it on the `meta` device.
@@ -124,10 +124,12 @@ class BetterTransformerBaseLayer(nn.Module):
                 "Training is not supported for `BetterTransformer` integration.",
                 " Please use `model.eval()` before running the model.",
             )
-    
+
     def _replace_to_original_module(self):
         r"""
         A wrapper function to replace the current layer with the previous non-BetterTransformer
         layer.
         """
-        raise NotImplementedError("Please implement this method in the `BetterTransformerLayer` class to benefit from the `BetterTransformer` inverse transformation.")
+        raise NotImplementedError(
+            "Please implement this method in the `BetterTransformerLayer` class to benefit from the `BetterTransformer` inverse transformation."
+        )
--- a/optimum/bettertransformer/models/base.py
+++ b/optimum/bettertransformer/models/base.py
@@ -48,7 +48,7 @@ class BetterTransformerBaseLayer(nn.Module):
         Args:
             config (`transformers.PretrainedConfig`):
                 The config of the model.
-            orig_layer (`torch.nn.Module`, `optional`):
+            orig_layer (`Optional[torch.nn.Module]`, defaults to `None`):
                 The original layer that needs to be modified. Defaults to `None`.
         """
         super().__init__()
@@ -60,7 +60,7 @@ class BetterTransformerBaseLayer(nn.Module):
         self.embed_dim = None
         self.num_layers = None
         self.original_layers_mapping = {}
-        # some models does not have some attributes thus needs to be ignored
+        # Some models does not have some attributes thus needs to be ignored
         # e.g. whisper does not have self_attn.k_proj.bias but has self_attn.v_proj.bias & self_attn.q_proj.bias
         self.keys_to_ignore = []
 
@@ -154,12 +154,10 @@ class BetterTransformerBaseLayer(nn.Module):
         layer.
         """
         for modified_layer_key_names, original_layer_key_names in self.original_layers_mapping.items():
-            # to deal with qkv layers for example
             if isinstance(original_layer_key_names, list):
-                # retrieve the current weight
                 current_weight = getattr(self, modified_layer_key_names)
 
-                # split the current weight n chunks - this is useful to split
+                # Split the current weight n chunks - this is useful to split
                 # the qkv layers into q, k, v layers for example.
                 split_index = current_weight.shape[0] // len(original_layer_key_names)
                 for i, module in enumerate(original_layer_key_names):
--- a/optimum/bettertransformer/models/base.py
+++ b/optimum/bettertransformer/models/base.py
@@ -55,6 +55,7 @@ class BetterTransformerBaseLayer(nn.Module):
         self.num_heads = None
         self.embed_dim = None
         self.num_layers = None
+        self.original_layers_mapping = {}
 
         # Get activation function
         for attr in KNOWN_ACTIVATION_ATTRIBUTES:
@@ -140,11 +141,41 @@ class BetterTransformerBaseLayer(nn.Module):
                 " Please use `model.eval()` before running the model.",
             )
 
-    def _replace_to_original_module(self):
+    def _recurse_setattr(self, module, name, value):
+        r"""
+        A wrapper function to recursively set attributes to a module.
+        """
+        if "." not in name:
+            setattr(module, name, value)
+        else:
+            name, rest = name.split(".", 1)
+            self._recurse_setattr(getattr(module, name), rest, value)
+
+    def _revert_back_to_original_module(self):
         r"""
         A wrapper function to replace the current layer with the previous non-BetterTransformer
         layer.
         """
-        raise NotImplementedError(
-            "Please implement this method in the `BetterTransformerLayer` class to benefit from the `BetterTransformer` inverse transformation."
-        )
+        for modified_layer_key_names, original_layer_key_names in self.original_layers_mapping.items():
+            # to deal with qkv layers for example
+            if isinstance(original_layer_key_names, list):
+                # retrieve the current weight
+                current_weight = getattr(self, modified_layer_key_names)
+
+                # split the current weight n chunks
+                split_index = current_weight.shape[0] // len(original_layer_key_names)
+                for i, module in enumerate(original_layer_key_names):
+                    self._recurse_setattr(
+                        self.orig_layer, module, nn.Parameter(current_weight[i * split_index : (i + 1) * split_index])
+                    )
+            elif isinstance(original_layer_key_names, str):
+                self._recurse_setattr(
+                    self.orig_layer, original_layer_key_names, getattr(self, modified_layer_key_names)
+                )
+            else:
+                raise ValueError(
+                    f"Invalid type {type(modified_layer_key_names)} for `original_layers_mapping`",
+                    " please use either `str` or `list`.",
+                )
+
+        return self.orig_layer
--- a/optimum/bettertransformer/models/base.py
+++ b/optimum/bettertransformer/models/base.py
@@ -14,6 +14,8 @@
 import torch
 import torch.nn as nn
 
+from copy import deepcopy
+
 from ...utils import logging
 
 
@@ -29,7 +31,7 @@ logger = logging.get_logger(__name__)
 
 
 class BetterTransformerBaseLayer(nn.Module):
-    def __init__(self, config):
+    def __init__(self, config, old_layer=None):
         super().__init__()
         self.norm_first = False
         self.use_gelu = False
@@ -60,6 +62,13 @@ class BetterTransformerBaseLayer(nn.Module):
             if hasattr(config, attr):
                 self.num_layers = getattr(config, attr)
                 break
+        
+        if old_layer is not None:
+            # Last step, store the old module skeleton by copying the old module and putting
+            # it on the `meta` device.
+            self.old_layer = deepcopy(old_layer).to("meta")
+        else:
+            self.old_layer = old_layer
 
     def validate_bettertransformer(self):
         r"""
@@ -115,3 +124,10 @@ class BetterTransformerBaseLayer(nn.Module):
                 "Training is not supported for `BetterTransformer` integration.",
                 " Please use `model.eval()` before running the model.",
             )
+    
+    def _replace_to_original_module(self):
+        r"""
+        A wrapper function to replace the current layer with the previous non-BetterTransformer
+        layer.
+        """
+        raise NotImplementedError("Please implement this method in the `BetterTransformerLayer` class to benefit from the `BetterTransformer` inverse transformation.")
--- a/optimum/bettertransformer/models/base.py
+++ b/optimum/bettertransformer/models/base.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from copy import deepcopy
+from typing import List, Optional
 
 import torch
 import torch.nn as nn
@@ -31,7 +32,21 @@ logger = logging.get_logger(__name__)
 
 
 class BetterTransformerBaseLayer(nn.Module):
-    def __init__(self, config, old_layer=None):
+    def __init__(
+        self,
+        config: "transformers.PretrainedConfig",
+        orig_layer: Optional[nn.Module] = None,
+    ):
+        r"""
+        Base layer for `BetterTransformer` integration. This class is used to wrap all the necessary
+        components for the `BetterTransformer` integration.
+
+        Args:
+            config (`transformers.PretrainedConfig`):
+                The config of the model.
+            orig_layer (`torch.nn.Module`, `optional`):
+                The original layer that needs to be modified. Defaults to `None`.
+        """
         super().__init__()
         self.norm_first = False
         self.use_gelu = False
@@ -63,12 +78,12 @@ class BetterTransformerBaseLayer(nn.Module):
                 self.num_layers = getattr(config, attr)
                 break
 
-        if old_layer is not None:
+        if orig_layer is not None:
             # Last step, store the old module skeleton by copying the old module and putting
             # it on the `meta` device.
-            self.old_layer = deepcopy(old_layer).to("meta")
+            self.orig_layer = deepcopy(orig_layer).to("meta")
         else:
-            self.old_layer = old_layer
+            self.orig_layer = orig_layer
 
     def validate_bettertransformer(self):
         r"""
--- a/optimum/bettertransformer/models/base.py
+++ b/optimum/bettertransformer/models/base.py
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from copy import deepcopy
-from typing import List, Optional
+from typing import TYPE_CHECKING, List, Optional
+
+
+if TYPE_CHECKING:
+    from transformers import PretrainedConfig
 
 import torch
 import torch.nn as nn
@@ -34,7 +38,7 @@ logger = logging.get_logger(__name__)
 class BetterTransformerBaseLayer(nn.Module):
     def __init__(
         self,
-        config: "transformers.PretrainedConfig",
+        config: "PretrainedConfig",
         orig_layer: Optional[nn.Module] = None,
     ):
         r"""
--- a/optimum/bettertransformer/models/encoder_models.py
+++ b/optimum/bettertransformer/models/encoder_models.py
@@ -137,7 +137,7 @@ class BertLayerBetterTransformer(BetterTransformerBaseLayer):
             bert_layer (`torch.nn.Module`):
                 The original BERT Layer where the weights needs to be retrieved.
         """
-        super().__init__(config)
+        super().__init__(config, bert_layer)
         # In_proj layer
         self.in_proj_weight = nn.Parameter(
             torch.cat(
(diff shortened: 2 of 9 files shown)


optimum/bettertransformer/models/base.py
```diff
@@ -61,6 +63,13 @@ def __init__(self, config):
                 self.num_layers = getattr(config, attr)
                 break
 
+        if old_layer is not None:
+            # Last step, store the old module skeleton by copying the old module and putting
+            # it on the `meta` device.
+            self.old_layer = deepcopy(old_layer).to("meta")
+        else:
+            self.old_layer = old_layer
+
```
fxmarty, line 72: The meta allows to avoid x2 in memory, right?

optimum/bettertransformer/transformation.py
```diff
@@ -79,6 +79,31 @@ def replace_to_bettertransformer(model, config):
     return model
 
 
+def invert_to_old_model(bt_model):
```
fxmarty, line 82: Can you add type hint?

optimum/bettertransformer/transformation.py
```diff
@@ -79,6 +79,31 @@ def replace_to_bettertransformer(model, config):
...
+    r"""
+    Replaces the BT-converted model to its old variant, to be able to safely store the weights
+    of a trained model.
+
+    Args:
+        `model` (`torch.nn.Module`):
```
fxmarty, line 88:
```suggestion
        `bt_model` (`torch.nn.Module`):
```

optimum/bettertransformer/transformation.py
```diff
@@ -79,6 +79,31 @@ def replace_to_bettertransformer(model, config):
...
+            The input converted model
```
fxmarty, line 89: I don't understand the description

optimum/bettertransformer/transformation.py
```diff
@@ -79,6 +79,31 @@ def replace_to_bettertransformer(model, config):
...
+    Returns:
+        The invert-converted model
```
fxmarty, line 91: same

tests/bettertransformer/test_bettertransformer_encoder.py
```diff
@@ -290,6 +292,80 @@ def test_logits(self, test_name: str, model_id, padding, max_length=20):
...
+    all_models_to_test = ALL_INVERTIBLE_MODELS_TO_TEST
+
+    def tearDown(self):
+        gc.collect()
+
+    def prepare_inputs_for_class(self, model_id=None):
+        input_dict = {
+            "input_ids": torch.LongTensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 1, 1, 1]]),
+            "attention_mask": torch.LongTensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 0, 0, 0]]),
+        }
```
fxmarty, line 316: Can you test with inputs simulating `padding="max_length"` as well?

tests/bettertransformer/test_bettertransformer_encoder.py
```diff
@@ -56,6 +56,8 @@
     "hf-internal-testing/tiny-random-nllb",
 ]
 
+ALL_INVERTIBLE_MODELS_TO_TEST = ["hf-internal-testing/tiny-random-BertModel"]
```
fxmarty, line 59: Could we rather test on all supported architectures?

tests/bettertransformer/testing_bettertransformer_utils.py
```diff
@@ -177,6 +177,55 @@ def test_conversion(self):
...
 
 
+class BetterTransformersInvertibleTestMixin:
+    r"""
+    `BetterTransformersTestMixin` to wrap the tests for invertible models.
+    """
+    all_models_to_test = []
+
+    def test_save_load_invertible(self):
+        for model in self.all_models_to_test:
```
fxmarty, line 187: I think it is better to use `@parameterized.expand()`

fxmarty:
Would be keen to have the opinion of @michaelbenayoun , but personally, I am not in favor of

> leverage copy mechanism (# Copied from ...)

I would rather prefer something like having a map

```
bettertransformer_attrs = {
    "linear1_weight": "fc1.weight"
    ...
    "norm1_eps": "self_attn_layer_norm.eps"
    "norm1_weight": "self_attn_layer_norm.weight"
}
```

and use this with recursive getattr/setattr/delattr if you need to access several times the items, for example in `_replace_to_original_module`.

It's just a matter of taste, but I think it has the advantage of normalizing everything. And you could write a single `_replace_to_original_module` for all architectures.

optimum/bettertransformer/models/base.py
```diff
@@ -61,6 +63,13 @@ def __init__(self, config):
                 self.num_layers = getattr(config, attr)
                 break
 
+        if old_layer is not None:
+            # Last step, store the old module skeleton by copying the old module and putting
+            # it on the `meta` device.
+            self.old_layer = deepcopy(old_layer).to("meta")
```
michaelbenayoun, line 69: Do we need to make a deepcopy?

optimum/bettertransformer/transformation.py
michaelbenayoun, line 82: Not a big fan of the name of the function, maybe `revert_to_original_model`?

optimum/bettertransformer/models/base.py
```diff
@@ -61,6 +63,13 @@ def __init__(self, config):
                 self.num_layers = getattr(config, attr)
                 break
 
+        if old_layer is not None:
```
michaelbenayoun, line 66: I think we should refer the pre-transformation layer as "original layer" instead of "old layer".
```suggestion
        if orig_layer is not None:
```

optimum/bettertransformer/models/encoder_models.py
```diff
@@ -232,6 +232,53 @@ def forward(self, hidden_states, attention_mask, *_):
             hidden_states = hidden_states.to_padded_tensor(0.0)
         return (hidden_states,)
 
+    def _replace_to_original_module(self):
```
michaelbenayoun, line 235:
```suggestion
    def  _revert_back_to_original_module(self):
```

optimum/bettertransformer/transformation.py
```diff
@@ -79,6 +79,31 @@ def replace_to_bettertransformer(model, config):
...
+    """
+
+    for name, module in bt_model.named_children():
+        if len(list(module.children())) > 0:
+            # we may explicitly exclude part of the model to use BetterTransformer
+            invert_to_old_model(module)
+
+        is_invert_compatible = hasattr(module, "old_layer") and module.old_layer is not None
```
michaelbenayoun, line 99:
```suggestion
        can_be_reversed = getattr(module, "orig_layer", None) is not None
```

optimum/bettertransformer/transformation.py
```diff
@@ -236,3 +261,17 @@ def transform(
         setattr(model_fast, "save_pretrained", warn_uncompatible_save(model_fast.save_pretrained))
 
         return model_fast
+
+    @check_if_pytorch_greater(
+        "1.13.0",
+        "Please upgrade PyTorch following https://pytorch.org/get-started/locally/ in order to use BetterTransformer.",
+    )
+    def inverse_transform(model: torch.nn.Module, **kwargs) -> torch.nn.Module:
```
michaelbenayoun, line 269:
```suggestion
    def reverse(model: torch.nn.Module, **kwargs) -> torch.nn.Module:
```

michaelbenayoun (commented):
Left a few comments, mostly on naming

michaelbenayoun:
> Would be keen to have the opinion of @michaelbenayoun , but personally, I am not in favor of
> 
> > leverage copy mechanism (# Copied from ...)
> 
> I would rather prefer something like having a map
> 
> ```
> bettertransformer_attrs = {
>     "linear1_weight": "fc1.weight"
>     ...
>     "norm1_eps": "self_attn_layer_norm.eps"
>     "norm1_weight": "self_attn_layer_norm.weight"
> }
> ```
> 
> and use this with recursive getattr/setattr/delattr if you need to access several times the items, for example in `_replace_to_original_module`.
> 
> It's just a matter of taste, but I think it has the advantage of normalizing everything. And you could write a single `_replace_to_original_module` for all architectures.

I agree!

optimum/bettertransformer/models/base.py
younesbelkada: To be safe it's better I guess but not sure

optimum/bettertransformer/models/base.py
younesbelkada: yes

optimum/bettertransformer/models/encoder_models.py
younesbelkada: fixed in [8ee8df5](https://github.com/huggingface/optimum/pull/676/commits/8ee8df53d6685110ca2db78b94e9b0c0f49d533e)

optimum/bettertransformer/models/base.py
younesbelkada: fixed in [8ee8df5](https://github.com/huggingface/optimum/pull/676/commits/8ee8df53d6685110ca2db78b94e9b0c0f49d533e)

optimum/bettertransformer/transformation.py
younesbelkada: fixed in [8ee8df5](https://github.com/huggingface/optimum/pull/676/commits/8ee8df53d6685110ca2db78b94e9b0c0f49d533e)

optimum/bettertransformer/models/base.py
```diff
@@ -29,7 +31,7 @@
 
 
 class BetterTransformerBaseLayer(nn.Module):
-    def __init__(self, config):
+    def __init__(self, config, old_layer=None):
```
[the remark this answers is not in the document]
younesbelkada: fixed in [8ee8df5](https://github.com/huggingface/optimum/pull/676/commits/8ee8df53d6685110ca2db78b94e9b0c0f49d533e)

optimum/bettertransformer/transformation.py
younesbelkada: fixed in [39b4a14](https://github.com/huggingface/optimum/pull/676/commits/39b4a14ad6517dcaf91b1854b6a7bb2611d03d91)

optimum/bettertransformer/transformation.py
younesbelkada: Fixed in [cedc1db](https://github.com/huggingface/optimum/pull/676/commits/cedc1dbc93f3249cddf6fe9ca3fc7fa47d410622)

optimum/bettertransformer/transformation.py
younesbelkada: Fixed in [cedc1db](https://github.com/huggingface/optimum/pull/676/commits/cedc1dbc93f3249cddf6fe9ca3fc7fa47d410622)

younesbelkada:
Indeed using dict recusrively is much better @fxmarty @michaelbenayoun and has been addressed in the latest commits to support all models
I propose to change the whole transformation logic (hf_model to bt_model) by that in a follow up PR and keep the reverse transformation as it is now (recusive + dict)

optimum/bettertransformer/models/base.py
```diff
@@ -29,7 +32,21 @@
 
 
 class BetterTransformerBaseLayer(nn.Module):
-    def __init__(self, config):
+    def __init__(
+        self,
+        config: "transformers.PretrainedConfig",
```
fxmarty, line 37: Can you add `from typing import TYPE_CHECKING` block?

optimum/bettertransformer/models/base.py
```diff
@@ -38,6 +55,10 @@ def __init__(self, config):
         self.num_heads = None
         self.embed_dim = None
         self.num_layers = None
+        self.original_layers_mapping = {}
+        # some models does not have some attributes thus needs to be ignored
+        # e.g. whisper does not have self_attn.k_proj.bias but has self_attn.v_proj.bias & self_attn.q_proj.bias
+        self.keys_to_ignore = []
```
fxmarty, line 61: can't we just let the mapping empty then?

optimum/bettertransformer/models/base.py
```diff
@@ -115,3 +143,47 @@ def forward_checker(self, *args, **kwargs):
...
+
+    def _recurse_setattr(self, module, name, value):
+        r"""
+        A wrapper function to recursively set attributes to a module.
+        """
+        if "." not in name:
+            setattr(module, name, value)
+        else:
+            name, rest = name.split(".", 1)
+            self._recurse_setattr(getattr(module, name), rest, value)
```
fxmarty, line 155: can you move this to utils?

optimum/bettertransformer/models/encoder_models.py
```diff
@@ -82,6 +82,23 @@ def __init__(self, albert_layer, config):
...
+            "linear1_bias": "ffn.bias",
+            "linear2_weight": "ffn_output.weight",
+            "linear2_bias": "ffn_output.bias",
+            "norm1_eps": "attention.LayerNorm.eps",
+            "norm1_weight": "attention.LayerNorm.weight",
+            "norm1_bias": "attention.LayerNorm.bias",
+            "norm2_eps": "full_layer_layer_norm.eps",
+            "norm2_weight": "full_layer_layer_norm.weight",
+            "norm2_bias": "full_layer_layer_norm.bias",
+        }
```
fxmarty, line 100: This sounds more like a class attribute to me, I'd put this not in the init, e.g. with all caps after `class AlbertLayerBetterTransformer`. But anyway this may change in the future so I guess it's fine to leave as is.

tests/bettertransformer/test_bettertransformer_audio.py
```diff
@@ -141,3 +143,109 @@ def test_raise_train(self):
...
                 _ = bt_model(**inputs)
+
+    @parameterized.expand([(False,)])
+    def test_invert_modules(self, keep_original_model=False):
+        r"""
+        Test that the inverse converted model and hf model have the same modules
+        Since `Wav2vec2` does not support `deepcopy` we cannot test the transformation
+        with `keep_original_model=True` for this model.
+        """
+        for model in self.all_models_to_test:
```
fxmarty, line 154: Can you rather use `@parameterized.expand`? It allows to run tests individually which is great for debugging.

tests/bettertransformer/test_bettertransformer_audio.py
```diff
@@ -141,3 +143,109 @@ def test_raise_train(self):
...
+
+                self.assertEqual(hf_module_attributes, bt_module_attributes)
+
+    @parameterized.expand([(False,)])
+    def test_save_load_invertible(self, keep_original_model=False):
+        r"""
+        Since `Wav2vec2` does not support `deepcopy` we cannot test the transformation
+        with `keep_original_model=True` for this model.
+        """
+        for model in self.all_models_to_test:
```
fxmarty, line 184: same comment

tests/bettertransformer/test_bettertransformer_audio.py
```diff
@@ -141,3 +143,109 @@ def test_raise_train(self):
...
+                    )
+
+    @parameterized.expand([(False,)])
+    def test_invert_model_logits(self, keep_original_model=False):
+        r"""
+        Test that the inverse converted model and hf model have the same logits.
+        Since `Wav2vec2` does not support `deepcopy` we cannot test the transformation
+        with `keep_original_model=True` for this model.
+        """
+        for model in self.all_models_to_test:
```
fxmarty, line 233: same comment

tests/bettertransformer/test_bettertransformer_vision.py
```diff
@@ -133,3 +135,32 @@ def test_raise_autocast(self, test_name: str, model_id, padding, max_length=20):
...
     def test_raise_train(self, test_name: str, model_id, padding, max_length=20):
         super().test_raise_train([model_id], padding=padding, max_length=max_length)
+
+    @parameterized.expand([(True,), (False,)])
+    def test_invert_model_logits(self, keep_original_model=True, **preprocessor_kwargs):
+        r"""
+        Test that the inverse converted model and hf model have the same logits
+        """
+        # The first row of the attention mask needs to be all ones -> check: https://github.com/pytorch/pytorch/blob/19171a21ee8a9cc1a811ac46d3abd975f0b6fc3b/test/test_nn.py#L5283
+        for model in self.all_models_to_test:
```
fxmarty, line 145: same comment

tests/bettertransformer/testing_bettertransformer_utils.py
```diff
@@ -176,6 +176,104 @@ def test_conversion(self):
             self.assertTrue(isinstance(converted_model, hf_random_model.__class__))
             self.assertTrue(hasattr(converted_model, "generate"))
 
+    # @parameterized.expand([(True,), (False,)])
+    def test_invert_modules(self, keep_original_model=False):
+        r"""
+        Test that the inverse converted model and hf model have the same modules
+        """
+        for model in self.all_models_to_test:
```
fxmarty, line 184: same comment

tests/bettertransformer/testing_bettertransformer_utils.py
```diff
@@ -176,6 +176,104 @@ def test_conversion(self):
...
+                            bt_model_from_load.state_dict()[key],
+                        )
+                    )
+
+    @parameterized.expand([(True,), (False,)])
+    def test_invert_model_logits(self, keep_original_model=True):
+        r"""
+        Test that the inverse converted model and hf model have the same logits
+        """
+        for model in self.all_models_to_test:
```
fxmarty, line 257: same comment

optimum/bettertransformer/transformation.py
```diff
@@ -79,6 +79,33 @@ def replace_to_bettertransformer(model, config):
     return model
 
 
+def revert_to_original_model(
```
fxmarty, line 82:
```suggestion
def _revert_to_original_model(
```

optimum/bettertransformer/transformation.py
```diff
@@ -236,3 +263,17 @@ def transform(
...
+
+    @check_if_pytorch_greater(
+        "1.13.0",
+        "Please upgrade PyTorch following https://pytorch.org/get-started/locally/ in order to use BetterTransformer.",
+    )
+    def reverse(model: torch.nn.Module, **kwargs) -> torch.nn.Module:
+        # Step 1: check if the model has the attribute `use_bettertransformer`
+        if not getattr(model, "use_bettertransformer", False):
+            raise ValueError(
+                "You should inverse_transform a model that has been already transformed to a `BetterTransformer` format."
```
fxmarty, line 275:
```suggestion
                "The method BetterTransformer.reverse() should be used on a model already transformed to the BetterTransformer format, which appears to not be the case."
```

optimum/bettertransformer/models/base.py
```diff
@@ -61,6 +82,13 @@ def __init__(self, config):
                 self.num_layers = getattr(config, attr)
                 break
 
+        if orig_layer is not None:
+            # Last step, store the old module skeleton by copying the old module and putting
+            # it on the `meta` device.
+            self.orig_layer = deepcopy(orig_layer).to("meta")
```
fxmarty, line 88: The `deepcopy(orig_layer)` sounds like it would make a copy in RAM / GPU memory before being set on meta device. I would if `deepcopy(orig_layer.to("meta"))` is not better (if we absolutely need deepcopy)

tests/bettertransformer/test_bettertransformer_audio.py
```diff
@@ -141,3 +143,109 @@ def test_raise_train(self):
...
+
+            # get outputs
+            torch.manual_seed(42)
+            output_bt = bt_model(**inputs)
+
+            torch.manual_seed(42)
+            output_hf = hf_model(**inputs)
+
+            # Assert that the outputs are the same
+            self.assertTrue(torch.allclose(output_bt[0], output_hf[0], atol=1e-3))
```
fxmarty, line 251: since `keep_original_model = False`, it sounds like this test will always pass to True

tests/bettertransformer/test_bettertransformer_audio.py
```diff
@@ -141,3 +143,109 @@ def test_raise_train(self):
...
+            for hf_module, bt_module in zip(hf_modules, bt_modules):
+                self.assertEqual(type(hf_module), type(bt_module))
+                # check the modules have the same methods
+                self.assertEqual(dir(hf_module), dir(bt_module))
+
+                # check the modules have the same attributes
+                hf_module_attributes = [attr for attr in dir(hf_module) if not attr.startswith("_")]
+                bt_module_attributes = [attr for attr in dir(bt_module) if not attr.startswith("_")]
+
+                self.assertEqual(hf_module_attributes, bt_module_attributes)
```
fxmarty, line 176: same comment, it sounds like the test will always pass because `keep_original_model = False`

tests/bettertransformer/test_bettertransformer_audio.py
```diff
@@ -141,3 +143,109 @@ def test_raise_train(self):
...
+                            bt_model_from_load.state_dict()[key],
+                        )
+                    )
+
+                    self.assertTrue(
+                        torch.allclose(
+                            hf_model.state_dict()[key],
+                            bt_model_from_load.state_dict()[key],
+                        )
+                    )
```
fxmarty, line 224: same comment, fuzzy think here because `keep_original_model = False` (we don't test against the original model, are we? It sounds like there are in place modifications)

optimum/bettertransformer/models/base.py
younesbelkada: I tried and didn't worked, I guess it's because `.to` is an in-place operation

optimum/bettertransformer/models/base.py
younesbelkada: Hmm then how would you discard a `k_proj.bias` given the sets: `q_proj`, `k_proj`, `v_proj`. Also it would break the entire logic in https://github.com/younesbelkada/optimum/blob/bt-inverse-transform-bert/optimum/bettertransformer/models/base.py#L164 so I wanted to keep things simple as these are very-corner cases

optimum/bettertransformer/transformation.py
younesbelkada: resolving as fixed in [aefc535](https://github.com/huggingface/optimum/pull/676/commits/aefc535d66dc233ca70c71ff40ebb178ad914c0f)

optimum/bettertransformer/models/base.py
younesbelkada: fixed in [aefc535](https://github.com/huggingface/optimum/pull/676/commits/aefc535d66dc233ca70c71ff40ebb178ad914c0f)

tests/bettertransformer/test_bettertransformer_audio.py
younesbelkada: Good point, I made some modification on the base test, I propose to compare the source code and the signature of the functions `__init__`  and `forward` since it is basically the only components we touch

younesbelkada: Fixed in [34363bf](https://github.com/huggingface/optimum/pull/676/commits/34363bf8f9ad9c29ea0cdfae4580062afebf907d)

tests/bettertransformer/test_bettertransformer_audio.py
younesbelkada: fixed in [34363bf](https://github.com/huggingface/optimum/pull/676/commits/34363bf8f9ad9c29ea0cdfae4580062afebf907d)

tests/bettertransformer/test_bettertransformer_audio.py
younesbelkada: fixed in [34363bf](https://github.com/huggingface/optimum/pull/676/commits/34363bf8f9ad9c29ea0cdfae4580062afebf907d)

optimum/bettertransformer/models/base.py
younesbelkada: fixed in [34363bf](https://github.com/huggingface/optimum/pull/676/commits/34363bf8f9ad9c29ea0cdfae4580062afebf907d)

tests/bettertransformer/test_bettertransformer_audio.py
younesbelkada: It's a personal taste but I perfer to not rely on `parameterized.expand` for expanding on each `model_id` since it will slowdown the entire testing suite (you will call `setUp` and `tearDown` internally as many times as the number of models will delete & re-create again the intermediate variables. I didn't benchmarked it though so if it'll lead to the same speed happy to change !

younesbelkada:
Thanks for your comments @fxmarty , @michaelbenayoun !
Making this PR as ready for review now 💪

tests/bettertransformer/test_bettertransformer_audio.py
```diff
@@ -141,3 +144,135 @@ def test_raise_train(self):
...
+
+                for name, param in bt_model.named_parameters():
+                    self.assertFalse(param.device.type == "meta", f"Parameter {name} is on the meta device.")
+
+                bt_model.save_pretrained(tmpdirname)
+
+                bt_model_from_load = AutoModel.from_pretrained(tmpdirname)
+
+                # check if the state dict is the same
+                # first check if the keys are the same
```
michaelbenayoun, line 204:
```suggestion
```

tests/bettertransformer/test_bettertransformer_vision.py
```diff
@@ -133,3 +135,32 @@ def test_raise_autocast(self, test_name: str, model_id, padding, max_length=20):
...
+            # get hf and bt model
```
michaelbenayoun, line 146:
```suggestion
```

tests/bettertransformer/test_bettertransformer_vision.py
```diff
@@ -133,3 +135,32 @@ def test_raise_autocast(self, test_name: str, model_id, padding, max_length=20):
...
+            hf_model = AutoModel.from_pretrained(model)
+            # get bt model and invert it
```
michaelbenayoun, line 148:
```suggestion
```

tests/bettertransformer/test_bettertransformer_vision.py
```diff
@@ -133,3 +135,32 @@ def test_raise_autocast(self, test_name: str, model_id, padding, max_length=20):
...
+            bt_model = BetterTransformer.transform(hf_model, keep_original_model=keep_original_model)
+            bt_model = BetterTransformer.reverse(bt_model)
+
+            # get inputs
```
michaelbenayoun, line 152:
```suggestion
```

tests/bettertransformer/test_bettertransformer_vision.py
```diff
@@ -133,3 +135,32 @@ def test_raise_autocast(self, test_name: str, model_id, padding, max_length=20):
...
+            if model == "laion/CLIP-ViT-B-32-laion2B-s34B-b79K":
+                inputs = self.prepare_inputs_for_class(model_id=model, padding=True)
+            else:
+                inputs = self.prepare_inputs_for_class(model_id=model, padding="max_length")
+
+            # get outputs
```
michaelbenayoun, line 158:
```suggestion
```

tests/bettertransformer/test_bettertransformer_vision.py
```diff
@@ -133,3 +135,32 @@ def test_raise_autocast(self, test_name: str, model_id, padding, max_length=20):
...
+            torch.manual_seed(42)
+            output_bt = bt_model(**inputs)
+
+            torch.manual_seed(42)
+            output_hf = hf_model(**inputs)
+
+            # Assert that the outputs are the same
```
michaelbenayoun, line 165:
```suggestion
```

tests/bettertransformer/test_bettertransformer_audio.py
younesbelkada: Fixed in [c5179e3](https://github.com/huggingface/optimum/pull/676/commits/c5179e396f023461fd7b11227d8d6121c0f7c2a7)

tests/bettertransformer/test_bettertransformer_audio.py
younesbelkada: fixed in [c5179e3](https://github.com/huggingface/optimum/pull/676/commits/c5179e396f023461fd7b11227d8d6121c0f7c2a7)

tests/bettertransformer/test_bettertransformer_vision.py
younesbelkada: fixed in [c5179e3](https://github.com/huggingface/optimum/pull/676/commits/c5179e396f023461fd7b11227d8d6121c0f7c2a7)

tests/bettertransformer/testing_bettertransformer_utils.py
younesbelkada: Fixed in [c5179e3](https://github.com/huggingface/optimum/pull/676/commits/c5179e396f023461fd7b11227d8d6121c0f7c2a7)

tests/bettertransformer/test_bettertransformer_vision.py
```diff
@@ -133,3 +223,94 @@ def test_raise_autocast(self, test_name: str, model_id, padding, max_length=20):
...
+
+        # get outputs
+        torch.manual_seed(42)
+        output_bt = bt_model(**inputs)
+
+        torch.manual_seed(42)
+        output_hf = hf_model(**inputs)
+
+        # Assert that the outputs are the same
+        self.assertTrue(torch.allclose(output_bt[0], output_hf[0], atol=1e-3))
```
fxmarty, line 279: This test is always passing in case `keep_original_model=False`.

tests/bettertransformer/testing_bettertransformer_utils.py
```diff
@@ -118,7 +119,7 @@ def test_raise_on_save(self):
         Test if the converion properly raises an error if someone tries to save the model using `save_pretrained`.
         """
         for model_id in self.all_models_to_test:
-            with self.assertWarns(UserWarning), tempfile.TemporaryDirectory() as tmpdirname:
+            with self.assertRaises(ValueError), tempfile.TemporaryDirectory() as tmpdirname:
                 hf_model = AutoModel.from_pretrained(model_id).eval()
                 bt_model = BetterTransformer.transform(hf_model, keep_original_model=False)
                 bt_model.save_pretrained(tmpdirname)
```
fxmarty, line 125: No need to change, but just to let you know, you can actually even test for the error message to be the right one: https://github.com/huggingface/optimum/blob/9efb46a38c64eab31324d35d56299730612f93a8/tests/onnxruntime/test_modeling.py#L359-L364

fxmarty (approved):
Left a few more minor comments, but LGTM!

tests/bettertransformer/test_bettertransformer_vision.py
younesbelkada: Thanks for spotting ! 
Fixed in [43fb900](https://github.com/huggingface/optimum/pull/676/commits/43fb9009a66de6de1765ec684ce53395464f9c43)

tests/bettertransformer/test_bettertransformer_vision.py
```diff
@@ -133,3 +223,94 @@ def test_raise_autocast(self, test_name: str, model_id, padding, max_length=20):
...
+            }
+        )
+    )
+    def test_raise_save_pretrained_error(
+        self, test_name: str, model_id, keep_original_model=False, **preprocessor_kwargs
+    ):
+        super().test_raise_save_pretrained_error(model_id=model_id, keep_original_model=keep_original_model)
+
+    @parameterized.expand([(True,), (False,)])
+    def test_invert_model_logits(self, keep_original_model=True, **preprocessor_kwargs):
```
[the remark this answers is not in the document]
younesbelkada: Thanks! Fixed in [43fb900](https://github.com/huggingface/optimum/pull/676/commits/43fb9009a66de6de1765ec684ce53395464f9c43)

optimum/utils/modeling_utils.py
```diff
@@ -0,0 +1,7 @@
+def recurse_setattr(module, name, value):
+    """A wrapper function to recursively set attributes to a module."""
```
michaelbenayoun, line 2: It does not wrap anything.

michaelbenayoun (approved):
LGTM

optimum/utils/modeling_utils.py
younesbelkada: fixed in [91a37bc](https://github.com/huggingface/optimum/pull/676/commits/91a37bcd2d8998373e5acdfd5e5231692208320b)

PR #684, opened 2023-01-08
DetrBetterTransformer added, problem with position_embedding_type=sine

mszsorondo:
# What does this PR do?
It implements BetterTransformer for DETR

# Problem
Raises ``` ValueError: Positional embedding type sine not supported for `BetterTransformer` integration ``` at conversion step. BetterTransformer doesn't support sine positional embeddings.

### DIFF
--- a/optimum/bettertransformer/__init__.py
+++ b/optimum/bettertransformer/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2022 The HuggingFace and Meta Team.  All rights reserved.
+# Copyright 2022 The HuggingFace Team. All rights reserved.
 #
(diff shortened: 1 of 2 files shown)


mszsorondo:
@younesbelkada any idea on how to start solving this? I can give it a try

younesbelkada:
Hi @mszsorondo 
Thanks for your contribution! 
You can modify the condition here: https://github.com/huggingface/optimum/blob/2911a91e6d6212a8c6757999e5723747c4d7590f/optimum/bettertransformer/models/base.py#L80 and make it pass for Detr

Issue #671, opened 2023-01-05
Compute loss inside the training_step for optimum trainer.


AdamLouly:
### Feature request

I would like to request a feature that allows for the modification of the optimum trainer to compute the loss inside the training process. 

### Motivation

We believe that modifying the optimum training process to compute the loss inside the training process will greatly benefit to the memory usage. This is because computing the loss separately is currently consuming more memory, which can be a significant issue for users working with large datasets or on systems with limited memory.

In a proof-of-concept implementation that we conducted; we were able to achieve a 7.8% reduction in memory usage by including the loss calculation inside the training process. 

### Your contribution

We have identified two potential approaches for modifying the optimum training process to compute the loss inside the training process. 

**The first approach** involves leveraging the loss computation function inside the Hugging Face trainer, but we have encountered some limitations related to the default loss function used in this trainer (label_smoother).

**The second approach**, which we have implemented in a proof-of-concept, involves creating a class called 'ModuleWithLoss' that overrides the forward method to compute the loss inside. This approach has shown promising results, but we have encountered issues with the eval_loop and prediction_loop breaking when using this method.

To solve this issue, we propose maintaining two separate models: a training model wrapped inside the 'ModuleWithLoss' class for all training-related tasks, and a normal default model for evaluation and inference. We believe that this approach will be effective in addressing the issues we have encountered and will ultimately lead to the successful implementation of this feature.

we want to know what do you think about 2 approaches and if you have any idea to make the first approach happen, because it looks much cleaner if we can pass the limitations.

if not, we would like to know if maintaining both models is a good approach or is there any suggestions to do this better as you have more context on this than our team.

For reference this is the POC implementation : https://github.com/pengwa/optimum/pull/1

Thank you

AdamLouly:
@JingyaHuang

pengwa:
To more specific, the memory benefits are mainly for onnxruntime ORTTrainer. 

An alternative approach might be: _inner_training_loop wrap the model + loss and did not update the self.model or self.model_wrapped (that would affect evaluation).

JingyaHuang:
Hi @pengwa and @AdamLouly, in `ORTTrainer`, the loss is computed outside the training process only when applying label smoothing. (So if not using `label_smmother`, the loss will be the loss computed in the modeling)

Case 1: w/. `label_smoother` (extra memory usage)
[...]

PR #686, opened 2023-01-11
Compute Loss inside the training step.

AdamLouly:
This PR is the suggested solution for the feature requested in the issues section : https://github.com/huggingface/optimum/issues/671

and it is a completion of this PR: https://github.com/pengwa/optimum/pull/1

**Motivation:** 
Modifying The optimum Trainer to compute the loss inside the training step will benefit from saving the memory allocated during training.

The first solution was to create a Module wrapper, named ModuleWithLoss() which will override the forward function to compute the loss inside the training step, but we faced some limitations when running eval and predict, this solution breaks because on the forward function we're expecting to get input as a dictionary, but there are some cases in eval and predict where the input is passed with **, thus it will be sent as unpacked parameters instead of one dictionary.

to solve this Issue, we have to maintain 2 models, _training_model which will be the wrapped ModuleWithLoss instance, and _inference_model which will be the default model.
and we will switch between the models in train or eval/predict step.

This is the logs for running 2 experiments.
Specs: Tesla V100 - 4 GPU - 16 GB Memory Each.
`python -m torch.distributed.launch --nproc_per_node=4 examples/onnxruntime/training/language-modeling/run_clm.py --model_name_or_path bigscience/bloom-560m --dataset_name wikitext --dataset_config_name wikitext-2-raw-v1 --label_smoothing 0.1 --do_eval --output_dir /tmp/ --overwrite_output_dir --max_steps 400 --logging_steps 20 --fp16 --block_size 256 --per_device_train_batch_size 8 --deepspeed aml_ds_config_zero_1.json`

Using default optimum trainer: 
 
image

Using This Wrapper Solution:
image

We see a difference of almost 2GB between the 2 runs using the same recipe.


AdamLouly:
@JingyaHuang , is there a way to run CI pipelines to make sure this change does not break anything?
is it expected to run do_train and do_predict in the same run?

JingyaHuang:
Hi @AdamLouly, the [CI for ort training](https://github.com/huggingface/optimum/blob/main/.github/workflows/test_onnxruntime_train.yml) is scheduled for every morning. To run the test directly, you can use the following:

```
python -m unittest tests/onnxruntime/nightly_test_trainer.py
```

`do_train`, `do_predict` and `do_eval` are [intended for training/evaluation scripts](https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.do_train). You can definitely run them at the same time if they are in your script (e.g. all three are added in `examples/onnxruntime/training/question-answering/run_qa.py`, but `do_predict` was not added in previous contribution for clm #248)

JingyaHuang:
Hi @AdamLouly, could you rebase your branch? I would like to test the compatibility of the wrapper with other components in `ORTTrainer`. Thanks.

optimum/onnxruntime/trainer.py
```diff
@@ -443,7 +544,7 @@ def _inner_training_loop(
                     RuntimeWarning,
                 )
 
-            self.model = model
+            self.model = self._training_model
```
pengwa, line 547: if deepspeed is not used, should we override also?

pengwa, line 547: once ORTModule wrap training model complete, where is the ortmodule instance stored? How could we get it next time for training?

AdamLouly: OrtModule instance won't be stored, we will wrap the training_model again on every _inner_training_loop called

optimum/onnxruntime/trainer.py
```diff
@@ -260,13 +309,63 @@ def __init__(
...
         self.exported_with_loss = False
         if self.args.local_rank:
             torch.cuda.set_device(self.args.local_rank)
 
+    def _set_signature_columns_if_needed(self):
+        if self._signature_columns is None:
+            # Inspect model forward signature to keep only the arguments it accepts.
+            import inspect
+
+            if self.model == self._training_model:
```
pengwa, line 326: is that better we change to isinstance of ModelWithLoss?

optimum/onnxruntime/trainer.py
```diff
@@ -130,6 +131,54 @@
...
+        # Label smoothing
+        if self.args.label_smoothing_factor != 0:
+            from transformers.trainer_pt_utils import LabelSmoother
+
+            self.label_smoother = LabelSmoother(epsilon=self.args.label_smoothing_factor)
+        else:
+            self.label_smoother = None
+
+    def forward(self, inputs: Dict[str, Union[torch.Tensor, Any]], return_outputs):
+        if self.label_smoother is not None and "labels" in inputs:
```
pengwa, line 148: The logic in the forward are mostly from transformer Trainer's compute_loss.

To avoid the duplicated code, can we bind that transformer Trainer's compute_loss function as member function of ModuleWithLoss, then we just call self.compute_loss(self._original_model ) to get the loss. This would make the maintenance easier I think.

optimum/onnxruntime/trainer.py
```diff
@@ -260,13 +309,63 @@ def __init__(
             optimizers=optimizers,
             preprocess_logits_for_metrics=preprocess_logits_for_metrics,
         )
-
+        self._training_model = ModuleWithLoss(model, args)
```
pengwa, line 312: add a comment explaining how we leverage training_model and inference_model to do train and evaluate?

optimum/onnxruntime/trainer.py
```diff
@@ -260,13 +309,63 @@ def __init__(
...
+        self.model = model
+        self._inferencing_model = model
         self.feature = feature
         self.onnx_model_path = onnx_model_path
         self.exported_with_loss = False
         if self.args.local_rank:
             torch.cuda.set_device(self.args.local_rank)
 
+    def _set_signature_columns_if_needed(self):
+        if self._signature_columns is None:
```
pengwa, line 322: add a note saying we assume training_model and inference_model assume the same forward signature_column. So far, self._signature_columns only store the first-time parsed signature, whatever it is from training path or inference path.

optimum/onnxruntime/trainer.py
```diff
@@ -260,13 +309,63 @@ def __init__(
...
+            self._signature_columns += list(set(["label", "label_ids"] + self.label_names))
+
+    def compute_loss(self, model_with_loss, inputs, return_outputs=False):
+        # Run model forward + loss compute.
+        if self.model == self._training_model:
+            outputs = model_with_loss(inputs, return_outputs)
+            return outputs
+        else:
+
+            if self.label_smoother is not None and "labels" in inputs:
```
pengwa, line 342: we can just call self.compute_loss(self.model, ...) for the line 340 else branch, right?

optimum/onnxruntime/trainer.py
```diff
@@ -785,6 +885,7 @@ def evaluate(
             dictionary also contains the epoch number which comes from the training state.
         """
         # memory metrics - must set up as early as possible
+        self.model = self._inferencing_model
```
pengwa, line 888: add a todo: revisit to enable evaluation/prediction using ORT backend.

pengwa:
As we synced offline, for this PR, something we are curious: 1). memory saving, 2). possible batch size increment 3). any perf impact.

AdamLouly:
adding some extra testing results for reference here:
Specs: Tesla V100 - 4 GPU - 16 GB Memory Each.
`python -m torch.distributed.launch --nproc_per_node=4 examples/onnxruntime/training/language-modeling/run_clm.py --model_name_or_path bigscience/bloom-560m --dataset_name wikitext --dataset_config_name wikitext-2-raw-v1 --do_train --output_dir /tmp/ --overwrite_output_dir --max_steps 400 --logging_steps 2 --bloc_size 2560 --fp16`

image

the max batch we could run using the current trainer is 11, while with this change, we can bring the max batch size to 14.

The percentage of the memory saved is different from and experiment to another based on batch size, block size..

optimum/onnxruntime/trainer.py
AdamLouly: we can't do that because we override the compute_loss function in ORTTrainer class so we don't have access to the compute_loss function from transformers trainer.

`  File "/home/adamlouly/optimum/optimum/onnxruntime/trainer.py", line 337, in compute_loss
    if self.model == self._training_model:
RecursionError: maximum recursion depth exceeded in comparison`

optimum/onnxruntime/trainer.py
askhade: Please also add comments explaining why we need to move from having just model to _training_model, _inferenceing_model and model... Does this mean we have multiple copies of the model now?

AdamLouly:
@JingyaHuang  we added a flag "--loss_in_train" to keep these modifications optional,
if users want to use the ModuleWithLoss Wrapper, they can run the recipes using --loss_in_train flag, otherwise it will run the same as it is right now.

we are planning to remove this in the future.

is there any documentation where we can add this flag so users can be aware of it?
Thank you

optimum/onnxruntime/trainer.py
```diff
@@ -23,6 +23,12 @@
 import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Type, Union
+from tqdm.auto import tqdm
+
+
+# Integrations must be imported before ML frameworks:
+from transformers.integrations import hp_params, is_fairscale_available  # isort: split
```
pengwa, line 30: useless references?

optimum/onnxruntime/trainer.py
```diff
@@ -45,6 +51,7 @@
 from transformers.debug_utils import DebugOption, DebugUnderflowOverflow
 from transformers.deepspeed import deepspeed_init, is_deepspeed_zero3_enabled
 from transformers.dependency_versions_check import dep_version_check
+from transformers.models.auto.modeling_auto import MODEL_FOR_CAUSAL_LM_MAPPING_NAMES
```
pengwa, line 54: not used any more

optimum/onnxruntime/trainer.py
```diff
@@ -134,6 +141,28 @@
 SCALER_NAME = "scaler.pt"
 
 
+class ModuleWithLoss(nn.Module):
+    def __init__(self, model, args) -> None:
+        super().__init__()
+        self._original_model = model
+        self.args = args
+        self.hf_trainer = Trainer(model)
```
pengwa, line 149: shall we add comment explanation why we have created it?

optimum/onnxruntime/trainer.py
```diff
@@ -801,6 +867,8 @@ def evaluate(
             dictionary also contains the epoch number which comes from the training state.
         """
         # memory metrics - must set up as early as possible
+        # TODO: We need to enable evaluation using ORT backend.
```
pengwa, line 870: this TODO is only meant for --loss_in_train flag, right?

optimum/onnxruntime/training_args.py
```diff
@@ -65,6 +65,11 @@ class ORTTrainingArguments(TrainingArguments):
         metadata={"help": "The optimizer to use."},
     )
 
+    loss_in_train: Optional[bool] = field(
+        default=False,
+        metadata={"help": "Use ModuleWithLoss Wrapper to compute loss inside the training loop."},
```
pengwa, line 70: shall we make it clear when we need it? for example, when label smother is NOT none, having this will help save memory for ORTModule runs.

optimum/onnxruntime/trainer.py
```diff
@@ -23,6 +23,12 @@
 import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Type, Union
+from tqdm.auto import tqdm
```
pengwa, line 26: not used?

optimum/onnxruntime/trainer.py
AdamLouly: this was coming from main when I resolved conflict

pengwa (approved):
LGTM, thanks a lot!

optimum/onnxruntime/trainer.py
pengwa: but ideally after merge, this two lines should not appear in the diff.

optimum/onnxruntime/trainer.py
```diff
@@ -134,6 +139,30 @@
...
+        # Label smoothing
+        if self.args.label_smoothing_factor != 0:
+            from transformers.trainer_pt_utils import LabelSmoother
+
+            self.label_smoother = LabelSmoother(epsilon=self.args.label_smoothing_factor)
+        else:
+            self.label_smoother = None
+
+    def forward(self, inputs: Dict[str, Union[torch.Tensor, Any]], return_outputs):
+        return self.hf_trainer.compute_loss(self._original_model, inputs, return_outputs=False)
```
pengwa, line 159: I think this is not correct. doing this way, self.label_smoother won't be used by the compute_loss. What I suggested earlier is to bind the compute_loss of hf Trainer to ModuleWithLoss. Here is the example:
```

class A:
    def __init__(self) -> None:
        self.prop = "A's prop"

    def f(self, x: int) -> int:
        print("A>>f is called, the prop used is: ", self.prop)
        return x + 1

class B:
    def __init__(self) -> None:
        self.prop = "B's prop"

    def main(self, x: int) -> int:
        print("B>>main is called")
        self.f(x)
        return x + 2

    def f(self, x: int) -> int:
        raise NotImplementedError()


b_instance = B()
import types

b_instance.f = types.MethodType(getattr(A,'f'), b_instance)

b_instance.main(1)

```

pengwa: B>>main is called
A>>f is called, the prop used is:  B's prop

optimum/onnxruntime/trainer.py
```diff
@@ -279,12 +308,48 @@ def __init__(
...
+            else:
+                signature = inspect.signature(self.model.forward)
+
+            self._signature_columns = list(signature.parameters.keys())
+            # Labels may be named label or label_ids, the default data collator handles that.
+            self._signature_columns += list(set(["label", "label_ids"] + self.label_names))
+
+    def compute_loss(self, model_with_loss, inputs, return_outputs=False):
+        # Run model forward + loss compute.
+        if self.args.loss_in_train and self.model == self._training_model:
```
pengwa, line 347: have you ever run PyTorch for loss_in_train is enabled, is that working?

optimum/onnxruntime/trainer.py
```diff
@@ -279,12 +308,48 @@ def __init__(
             preprocess_logits_for_metrics=preprocess_logits_for_metrics,
         )
 
+        # We leverage both training_model and inference_model in conjunction with model.
+        # _training_model will be wrapped so it will use ORT and will use the overriden functions in ModuleWithLoss.
+        # _inferencing_model will be storing the default version of the model and we will switch to it in case of eval/test.
+
+        # Only Wrap the model if we pass --loss_in_train flag.
+        if args.loss_in_train:
+            self._training_model = ModuleWithLoss(model, args)
```
pengwa, line 317: nit: maybe we can pass `self` into ModuleWithLoss int its constructor. Then you can get self.model, self.args, and even self.label_smoothers inside the ModuleWithLoss class.

pengwa: here is an example:

```
import types


def create_model_with_loss(self):
    class ModuleWithLoss(nn.Module):
        def __init__(self, model, args, label_smoother):
            super().__init__()
            self._original_model = model
            self.args = args
            self.label_smoother = label_smoother

        def forward(self, inputs: Dict[str, Union[torch.Tensor, Any]], return_outputs):
            # The compute_model_plus_loss_internal is assigned once the class is instantiated.
            # It should have same signature as Trainer.compute_loss()
            return self.compute_model_plus_loss_internal(self._original_model, inputs, return_outputs)

        @property
        def config(self):
            return self._original_model.config

    model_with_loss = ModuleWithLoss(self.model, self.args, self.label_smoother)
    model_with_loss.compute_model_plus_loss_internal = types.MethodType(self.super().compute_loss, model_with_loss)

    return model_with_loss
```

optimum/onnxruntime/trainer.py
JingyaHuang: Will remove this to pass code quality check.

JingyaHuang: Ok, I don't have push access. Can you use ruff styling with the command `make style` and remove the redundant dependencies? Thx!

optimum/onnxruntime/trainer.py
```diff
@@ -134,6 +139,30 @@
 SCALER_NAME = "scaler.pt"
 
 
+class ModuleWithLoss(nn.Module):
```
JingyaHuang, line 142: `ModuleWithLoss` as a wrapper for `torch.nn.module` subclass, can you add a `module` property so that the [`unwrap_model`](https://github.com/huggingface/transformers/blob/bb5a2f2fc30985841289207b9f1f7765d8abc4e0/src/transformers/modeling_utils.py#L3548) could be compatible? I believe [`ORTModule` did the same](https://github.com/microsoft/onnxruntime/blob/6320decf048914009a42528914986eabedb1e900/orttraining/orttraining/python/training/ortmodule/ortmodule.py#L153)

optimum/onnxruntime/trainer.py
```diff
@@ -134,6 +139,30 @@
...
+
+        # Creating an instance of huggingFace Trainer so we can use compute_loss() logic and avoid duplicated code.
+        self.hf_trainer = Trainer(model)
+        # Label smoothing
+        if self.args.label_smoothing_factor != 0:
+            from transformers.trainer_pt_utils import LabelSmoother
+
+            self.label_smoother = LabelSmoother(epsilon=self.args.label_smoothing_factor)
+        else:
+            self.label_smoother = None
```
JingyaHuang, line 156: Why there is a case for not using label smoother? In transformers, unless using label smoother, the loss should be already calculated in the forward pass. C.f. [gpt2](https://github.com/huggingface/transformers/blob/bb5a2f2fc30985841289207b9f1f7765d8abc4e0/src/transformers/models/gpt2/modeling_gpt2.py#L1098-L1104)

optimum/onnxruntime/trainer.py
```diff
@@ -279,12 +308,48 @@ def __init__(
...
 
+        # We leverage both training_model and inference_model in conjunction with model.
+        # _training_model will be wrapped so it will use ORT and will use the overriden functions in ModuleWithLoss.
+        # _inferencing_model will be storing the default version of the model and we will switch to it in case of eval/test.
+
+        # Only Wrap the model if we pass --loss_in_train flag.
+        if args.loss_in_train:
+            self._training_model = ModuleWithLoss(model, args)
+        else:
+            self._training_model = model
```
JingyaHuang, line 319: Do we still need to distinguish training model and inference model, if we have the module property with the wrapper, given that we unwrap self.model for inference here: https://github.com/huggingface/optimum/blob/c45bd53757f2b789139ff0076288d875940699ea/optimum/onnxruntime/trainer.py#L534

JingyaHuang (commented):
Hi @AdamLouly and @pengwa, thanks a lot for the contribution, and sorry for the delay on reviewing.

I just left some comments and questions I have for the PR.

optimum/onnxruntime/trainer.py
JingyaHuang: And with the wrapper, model not using label smoother in the first place shall not have any benefit on memory right?

JingyaHuang:
> is there any documentation where we can add this flag so users can be aware of it?

@AdamLouly , sure! It's import for the visibility.

We can stress the flag both in [trainer's doc](https://github.com/huggingface/optimum/blob/main/docs/source/onnxruntime/usage_guides/trainer.mdx#orttrainingarguments) and [trainer examples' readmes](https://github.com/huggingface/optimum/tree/main/examples/onnxruntime/training/language-modeling#onnx-runtime-training).

optimum/onnxruntime/trainer.py
AdamLouly: FYI @JingyaHuang , In this case the code in the Trainer should be maintained if the compute_loss in hf trainer got changed then it should be changed in the forward pass of ModuleWithLoss as well.

optimum/onnxruntime/trainer.py
pengwa: > And with the wrapper, model not using label smoother in the first place shall not have any benefit on memory right?

This is a good question; I can help answer it. In short, whenever label smoothing factor argument is given or not, we see improvement on memory. The key reason is the target model ORTModule wraps will have one single loss output after the change. 

If label smoothing factor argument is not given, then the CrossEntropyLoss will be done inside the model forward pass. This is true. While there is a minor tricky here: `loss` along with `lm_logits` and other intermediate states are returned in the results. If ORTModule wraps and operates on this model, during model exporting, there will be few outputs besides `loss`, those outputs in training phases are not used later, but exporter don't know it will not be used. In ORT training implementation, though those outputs are not used, but we still fill them with zero during the whole backward propagation phase.

For this case, if we wrap model+loss together, the final output of model+loss (ORTModule wraps) is just loss.

AdamLouly:
The flag got changed to --use_module_with_loss

optimum/onnxruntime/trainer.py
AdamLouly: > Ok, I don't have push access. Can you use ruff styling with the command `make style` and remove the redundant dependencies? Thx!

I did make style and it formatted the trainer.py, but the CI still says it should be formatted.
is there a specific configuration for this?

optimum/onnxruntime/trainer.py
```diff
@@ -279,12 +328,81 @@ def __init__(
...
+            # Labels may be named label or label_ids, the default data collator handles that.
+            self._signature_columns += list(set(["label", "label_ids"] + self.label_names))
+
+    def compute_loss(self, model_with_loss, inputs, return_outputs=False):
+        # Run model forward + loss compute.
+        if isinstance(self.model, ModuleWithLoss):
+            outputs = model_with_loss(inputs, return_outputs)
+            return outputs
+        else:
+            if self.label_smoother is not None and "labels" in inputs:
```
pengwa, line 377: why we revert this super().compute_loss() in else branch?

optimum/onnxruntime/training_args.py
```diff
@@ -60,9 +60,13 @@ class ORTTrainingArguments(TrainingArguments):
...
 
-    optim: Optional[str] = field(
-        default="adamw_hf",
-        metadata={"help": "The optimizer to use."},
+    optim: Optional[str] = field(default="adamw_hf", metadata={"help": "The optimizer to use."})
+
+    use_module_with_loss: Optional[bool] = field(
+        default=False,
+        metadata={
+            "help": "Use ModuleWithLoss Wrapper to compute loss inside the training loop, when label smoother is NOT none having this will help save memory for ORTMOdule Runs."
```
pengwa, line 68: shallwe remove this "when label smoother is NOT none"

optimum/onnxruntime/trainer.py
JingyaHuang: Thanks for the explanation @pengwa, that's very clear.

optimum/onnxruntime/trainer.py
JingyaHuang: Hi @AdamLouly, sorry for the back and forth. I proposed to rewrite the codes as I was considering opening a PR in Transformers to put the label smoother inside forward. If so we don't need a wrapper in Optimum. But as @pengwa explained, a PR in Transformers won't be enough (we can't limit unnecessary outputs in Transformers for the flexibility reason), so we will always need this wrapper in Optimum. 

If so, I agree that we should inherit the `compute_loss()`(as you did before) to ease the maintenance.

optimum/onnxruntime/trainer.py
JingyaHuang: Hi @AdamLouly, yes the CI always uses the latest formatting tools. And whenever the team observe a failure of the check code quality CI,  we would fix it.

Your previous formatting issue could come from the fact that we recently switched from isort to ruff #760.  If you want to be more cautious, you can update your formatter with `pip install -U .[quality]` before `make style`.

optimum/onnxruntime/training_args.py
```diff
@@ -60,9 +60,13 @@ class ORTTrainingArguments(TrainingArguments):
...
 
-    optim: Optional[str] = field(
-        default="adamw_hf",
-        metadata={"help": "The optimizer to use."},
+    optim: Optional[str] = field(default="adamw_hf", metadata={"help": "The optimizer to use."})
+
+    use_module_with_loss: Optional[bool] = field(
+        default=False,
+        metadata={
+            "help": "Use ModuleWithLoss Wrapper to compute loss inside the training loop, having this will help save memory for ORTMOdule Runs."
```
pengwa, line 68: nit: ORTMOdule  -> "ORTModule "

AdamLouly:
@JingyaHuang any updates on reviewing this ?

pengwa:
@AdamLouly  can you fix the CI errors:

"optimum/onnxruntime/trainer.py:17:1: I001 [*] Import block is un-sorted or un-formatted"

JingyaHuang (approved):
<details close>
<summary>All trainer tests passed</summary>

```
====================================== test session starts =======================================
platform linux -- Python 3.8.10, pytest-7.2.2, pluggy-1.0.0
rootdir: /workspace, configfile: pyproject.toml
collected 32 items                                                                               

tests/onnxruntime/nightly_test_trainer.py s...s....................sssss..                 [100%]

======================================== warnings summary ========================================
../usr/lib/python3/dist-packages/requests/__init__.py:89
  /usr/lib/python3/dist-packages/requests/__init__.py:89: RequestsDependencyWarning: urllib3 (1.26.15) or chardet (3.0.4) doesn't match a supported version!
    warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "

../usr/local/lib/python3.8/dist-packages/onnxruntime/capi/onnxruntime_validation.py:114
  /usr/local/lib/python3.8/dist-packages/onnxruntime/capi/onnxruntime_validation.py:114: UserWarning: WARNING: failed to get cudart_version from onnxruntime build info.
    warnings.warn("WARNING: failed to get cudart_version from onnxruntime build info.")

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py:1794: FutureWarning: The first argument to symbolic functions is deprecated in 1.13 and will be removed in the future. Please annotate treat the first argument (g) as GraphContext and use context information from the object instead.
    warnings.warn(

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/onnxruntime/training/ortmodule/_utils.py:250: UserWarning: User Module's attribute name float collides with ORTModule's attribute name. User Module's method may not be called upon invocation through ORTModule.
    warnings.warn(

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/onnxruntime/training/ortmodule/_utils.py:250: UserWarning: User Module's attribute name half collides with ORTModule's attribute name. User Module's method may not be called upon invocation through ORTModule.
    warnings.warn(

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/onnxruntime/training/ortmodule/_utils.py:250: UserWarning: User Module's attribute name to collides with ORTModule's attribute name. User Module's method may not be called upon invocation through ORTModule.
    warnings.warn(

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/transformers/optimization.py:391: FutureWarning: This implementation of AdamW is deprecated and will be removed in a future version. Use the PyTorch implementation torch.optim.AdamW instead, or set `no_deprecation_warning=True` to disable this warning
    warnings.warn(

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/onnxruntime/training/ortmodule/_training_manager.py:192: UserWarning: Fast path enabled - skipping checks. Rebuild graph: True, Execution agent: True, Device check: True
    warnings.warn(

tests/onnxruntime/nightly_test_trainer.py: 12 warnings
  /usr/local/lib/python3.8/dist-packages/transformers/models/distilbert/modeling_distilbert.py:223: TracerWarning: torch.tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
    mask, torch.tensor(torch.finfo(scores.dtype).min)

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/_internal/jit_utils.py:258: UserWarning: The shape inference of org.pytorch.aten::ATen type is missing, so it may result in wrong shape inference for the exported graph. Please consider adding it in symbolic function. (Triggered internally at ../torch/csrc/jit/passes/onnx/shape_type_inference.cpp:1884.)
    _C._jit_pass_onnx_node_shape_type_inference(node, params_dict, opset_version)

tests/onnxruntime/nightly_test_trainer.py: 15 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/_internal/jit_utils.py:258: UserWarning: The shape inference of com.microsoft::SoftmaxCrossEntropyLossInternal type is missing, so it may result in wrong shape inference for the exported graph. Please consider adding it in symbolic function. (Triggered internally at ../torch/csrc/jit/passes/onnx/shape_type_inference.cpp:1884.)
    _C._jit_pass_onnx_node_shape_type_inference(node, params_dict, opset_version)

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/onnxruntime/training/ortmodule/_io.py:66: FutureWarning: 'torch.onnx._patch_torch._graph_op' is deprecated in version 1.13 and will be removed in version 1.14. Please note 'g.op()' is to be removed from torch.Graph. Please open a GitHub issue if you need this functionality..
    return g.op("Identity", self)

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py:687: UserWarning: The shape inference of org.pytorch.aten::ATen type is missing, so it may result in wrong shape inference for the exported graph. Please consider adding it in symbolic function. (Triggered internally at ../torch/csrc/jit/passes/onnx/shape_type_inference.cpp:1884.)
    _C._jit_pass_onnx_graph_shape_type_inference(

tests/onnxruntime/nightly_test_trainer.py: 15 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py:687: UserWarning: The shape inference of prim::Constant type is missing, so it may result in wrong shape inference for the exported graph. Please consider adding it in symbolic function. (Triggered internally at ../torch/csrc/jit/passes/onnx/shape_type_inference.cpp:1884.)
    _C._jit_pass_onnx_graph_shape_type_inference(

tests/onnxruntime/nightly_test_trainer.py: 15 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py:687: UserWarning: The shape inference of com.microsoft::SoftmaxCrossEntropyLossInternal type is missing, so it may result in wrong shape inference for the exported graph. Please consider adding it in symbolic function. (Triggered internally at ../torch/csrc/jit/passes/onnx/shape_type_inference.cpp:1884.)
    _C._jit_pass_onnx_graph_shape_type_inference(

tests/onnxruntime/nightly_test_trainer.py: 25 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py:1178: UserWarning: The shape inference of org.pytorch.aten::ATen type is missing, so it may result in wrong shape inference for the exported graph. Please consider adding it in symbolic function. (Triggered internally at ../torch/csrc/jit/passes/onnx/shape_type_inference.cpp:1884.)
    _C._jit_pass_onnx_graph_shape_type_inference(

tests/onnxruntime/nightly_test_trainer.py: 15 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py:1178: UserWarning: The shape inference of prim::Constant type is missing, so it may result in wrong shape inference for the exported graph. Please consider adding it in symbolic function. (Triggered internally at ../torch/csrc/jit/passes/onnx/shape_type_inference.cpp:1884.)
    _C._jit_pass_onnx_graph_shape_type_inference(

tests/onnxruntime/nightly_test_trainer.py: 15 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py:1178: UserWarning: The shape inference of com.microsoft::SoftmaxCrossEntropyLossInternal type is missing, so it may result in wrong shape inference for the exported graph. Please consider adding it in symbolic function. (Triggered internally at ../torch/csrc/jit/passes/onnx/shape_type_inference.cpp:1884.)
    _C._jit_pass_onnx_graph_shape_type_inference(

tests/onnxruntime/nightly_test_trainer.py: 13 warnings
  /usr/local/lib/python3.8/dist-packages/transformers/models/t5/tokenization_t5_fast.py:155: FutureWarning: This tokenizer was incorrectly instantiated with a model max length of 512 which will be corrected in Transformers v5.
  For now, this behavior is kept to avoid breaking backwards compatibility when padding/encoding with `truncation is True`.
  - Be aware that you SHOULD NOT rely on t5-small automatically truncating your input to 512 when padding/encoding.
  - If you want to encode/pad to sequences longer than 512 you can either instantiate this tokenizer with `model_max_length` or pass `max_length` when encoding/padding.
  - To avoid this warning, please instantiate this tokenizer with `model_max_length` set to your preferred value.
    warnings.warn(

tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp16_ort_inference_2_t5_seq2seq_lm
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp16_pt_inference_2_t5_seq2seq_lm
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp16_pt_inference_3_t5_seq2seq_lm_with_past
  /usr/local/lib/python3.8/dist-packages/transformers/tokenization_utils_base.py:3587: UserWarning: `as_target_tokenizer` is deprecated and will be removed in v5 of Transformers. You can tokenize your labels by using the argument `text_target` of the regular `__call__` method (either in the same call as your input texts if you use the same keyword arguments, or in a separate call.
    warnings.warn(

tests/onnxruntime/nightly_test_trainer.py: 13 warnings
  /usr/local/lib/python3.8/dist-packages/transformers/modeling_utils.py:793: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    if causal_mask.shape[1] < attention_mask.shape[1]:

tests/onnxruntime/nightly_test_trainer.py: 55 warnings
  /usr/local/lib/python3.8/dist-packages/torch/onnx/symbolic_opset9.py:3674: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
    value_t=torch.tensor([fill_value], dtype=scalar_type.dtype()),

tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp16_ort_inference_2_t5_seq2seq_lm
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_5_t5_seq2seq_lm
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_7_t5_seq2seq_lm_with_past
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_with_label_smoothing_5_t5_seq2seq_lm
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_with_label_smoothing_7_t5_seq2seq_lm_with_past
tests/onnxruntime/nightly_test_trainer.py::ORTSeq2SeqTrainerSpecificIntegrationTest::test_predict_with_generate_ort_0_t5_seq2seq_lm
tests/onnxruntime/nightly_test_trainer.py::ORTSeq2SeqTrainerSpecificIntegrationTest::test_predict_with_generate_ort_1_t5_seq2seq_lm_with_past
  /usr/local/lib/python3.8/dist-packages/transformers/models/t5/modeling_t5.py:507: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    elif past_key_value.shape[2] != key_value_states.shape[1]:

tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_0_distilbert_sequence_classification
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_1_distilbert_sequence_classification
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_2_distilbert_token_classification
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_3_distilbert_token_classification
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_4_t5_seq2seq_lm
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_5_t5_seq2seq_lm
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_6_t5_seq2seq_lm_with_past
tests/onnxruntime/nightly_test_trainer.py::ORTTrainerIntegrationTest::test_trainer_fp32_7_t5_seq2seq_lm_with_past
  /usr/local/lib/python3.8/dist-packages/onnxruntime/training/ortmodule/_custom_op_symbolic_registry.py:125: UserWarning: Unsupported diverged input and output types for logits when export cross_entropy_loss.logits type: Float, loss type: Float
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
==================== 25 passed, 7 skipped, 438 warnings in 589.63s (0:09:49) =====================
```
</details>

LGTM, let's get it merged. Thank you both for the awesome contribution @AdamLouly @pengwa , and helping me understand the memory issue!

PR #696, opened 2023-01-16
Avoid initializing unwanted attributes for ORTModel's having several inference sessions

fxmarty:
Attributes as `self.model`, `self.model_name`, `self.model_path` are unwanted and actually may take some RAM for no reason.


michaelbenayoun:
Are they?
I think we use them no?

fxmarty:
@michaelbenayoun I couldn't find these args being used.

fxmarty:
Approved by @michaelbenayoun

PR #698, opened 2023-01-16
`onnxruntime/modeling_ort.py` refactor, part 1

michaelbenayoun:
# What does this PR do?

This PR aims at simplifying and cleaning the `modeling_ort.py` and related files. This will be done in multiple PRs to make it easier to review.

- [x] Remove `__init__` that are not needed
- [x] Make `prepare_io_binding` generic enough so that it is not needed to implement it for every subclass

Left remaining for following PRs:

- [ ] Check dynamic axes for the inputs and outputs of exported models with `exporters.onnx`
   - [ ] Add tests to find problematic models
   - [ ]  `M2M100SinusoidalPositionalEmbedding` makes shape inference fail and creates a dynamic axis for `last_hidden_state` of the encoder, but it is actually static, fix it.
- [ ] Check past key values shape definition in `exporters.onnx` and other related inputs / outputs dynamic axes in decoder and seq2seq models
- [ ] Check which inputs are given to the model according to the `InferenseSession` (for `prepare_io_binding` and regular forward)
- [ ] Fix inference with Whisper, which fails under this PR
- [ ] Fix Marian `.generate`


fxmarty (approved):
LGTM, can you make sure the GPU tests pass locally?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -569,6 +577,64 @@ def from_pretrained(
...
+        dimensions = {}
+        for input_ in self.model.get_inputs():
+            shape = input_.shape
+            for idx, axis in enumerate(shape):
+                if isinstance(axis, str):
+                    dimensions[axis] = input_name_to_tensor[input_.name].shape[idx]
+
+        output_shapes = {}
+        output_buffers = {}
+        for output_node in self.model.get_outputs():
```
fxmarty, line 620: Why not use `self.model_inputs` and `self.model_outputs` here?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -569,6 +577,64 @@ def from_pretrained(
...
+            output_name = output_node.name
+            output_shape = tuple(map(lambda x: dimensions.get(x, x), output_node.shape))
```
JingyaHuang, line 622: How are you going to deal with shapes on dynamic axis? The definition could change according to task, got to be careful.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -250,6 +261,14 @@ def device(self) -> torch.device:
     def device(self, value: torch.device):
         self._device = value
 
+    @property
+    def use_io_binding(self):
+        return check_io_binding(self.providers, self._use_io_binding)
+
+    @use_io_binding.setter
+    def use_io_binding(self, value: bool):
```
michaelbenayoun, line 269: This is actually very important, because the way it was done previously did not handle this case:

```python
my_model = ORTModel(..., providers=["CPUExecutionProvider"]).to("cuda")
```

Before this, we would not use IO binding because `self.use_io_binding` was set at initialization.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -590,6 +610,103 @@ def from_pretrained(
...
+                    ordered_input_names.append(name)
+        input_names = ordered_input_names
+
+        name_to_np_type = TypeHelper.get_io_numpy_type_map(model)
+
+        input_name_to_tensor = {}
+        for idx, tensor in enumerate(model_inputs):
+            if tensor is None:
+                continue
+            # if idx >= len(input_names):
```
michaelbenayoun, line 647: Waiting to remove, to check if this is actually needed.

optimum/onnxruntime/modeling_ort.py
michaelbenayoun: We are getting them when possible, otherwise we can specify them via the `known_output_shapes` argument.

optimum/exporters/onnx/model_configs.py
```diff
@@ -400,6 +406,7 @@ def outputs(self) -> Mapping[str, Mapping[int, str]]:
             common_outputs = super().outputs
         else:
             common_outputs = super(OnnxConfigWithPast, self).outputs
+            common_outputs["encoder_last_hidden_state"] = {0: "batch_size", 1: "sequence_length"}
```
fxmarty, line 409: Why is this necessary?

optimum/exporters/onnx/model_configs.py
```diff
@@ -887,6 +894,24 @@ def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGen
...
+    def outputs(self) -> Mapping[str, Mapping[int, str]]:
+        common_outputs = super().outputs
+        if self._behavior is ConfigBehavior.ENCODER:
+            for name, dyanmic_axes in common_outputs.items():
+                if name == "last_hidden_state":
+                    common_outputs[name][1] = f"{common_outputs[name][1]} / 2"
+                else:
+                    common_outputs[name] = dynamic_axes
+        return common_outputs
+
```
fxmarty, line 914: Why is this necessary?

optimum/onnxruntime/base.py
```diff
@@ -0,0 +1,397 @@
...
+class ORTModelPart:
+    """
+    For multi-file ONNX models, such as encoder-decoder models, represents a part of the model.
+    It has its own `onnxruntime.InferenceSession`, and can perform a forward pass.
+    """
+
+    def __init__(
+        self,
+        session: InferenceSession,
+        parent_model: "ORTModel",
```
fxmarty, line 36: I'm not a huge fan of passing a parent class to its children; it feels like it can be bug prone. Couldn't we simply pass the relevant attributes, and possibly update them when needed (e.g. device)

Also, not sure if the parent model will actually have a `.config` attribute, e.g. see https://huggingface.co/stabilityai/stable-diffusion-2-1/tree/main . But we can leave it to an other PR to make it more general.

optimum/onnxruntime/base.py
```diff
@@ -0,0 +1,397 @@
...
+    def forward(
+        self,
+        input_ids: torch.LongTensor,
+        attention_mask: torch.LongTensor,
+        **kwargs,
+    ) -> BaseModelOutput:
+        if self.device.type == "cuda" and self.parent_model.use_io_binding:
+            model_inputs = [input_ids]
+            if "attention_mask" in self.input_names:
+                model_inputs.append(attention_mask)
```
fxmarty, line 73: This basically makes the attention_mask optional?

optimum/onnxruntime/base.py
```diff
@@ -0,0 +1,397 @@
...
+                if self.expected_key_symbolic_shape is None:
+                    self.expected_key_symbolic_shape = output.shape
+                else:
+                    self.expected_value_symbolic_shape = output.shape
+            if self.expected_key_symbolic_shape is not None and self.expected_value_symbolic_shape is not None:
+                break
+
+        self.key_sequence_length_idx = -2
+        if (
+            isinstance(self.expected_key_symbolic_shape[-1], str)
```
fxmarty, line 140: may `self.expected_key_symbolic_shape` not be None here?

optimum/onnxruntime/base.py
```diff
@@ -0,0 +1,397 @@
...
+
+            outputs_to_not_bind = {name for name in self.output_names if filter_out_output(name)}
+
+            model_inputs = [input_ids]
+
+            if "encoder_attention_mask" in self.input_names:
+                model_inputs.append(encoder_attention_mask)
+
+            if "encoder_hidden_states" in self.input_names:
+                model_inputs.append(encoder_hidden_states)
```
fxmarty, line 314: is this also making those inputs facultative?

optimum/onnxruntime/base.py
```diff
@@ -0,0 +1,397 @@
...
+                encoder_hidden_states,
+                past_key_values=past_key_values,
+            )
+
+            # past_key_values_inputs = past_key_values if past_key_values is not None else [None]
+
+            def filter_out_output(output_name):
+                return not output_name.startswith("present") and output_name not in {"loss", "logits"}
+
+            outputs_to_not_bind = {name for name in self.output_names if filter_out_output(name)}
```
fxmarty, line 306: Why is this needed ?

optimum/exporters/onnx/convert.py
```diff
@@ -46,6 +46,14 @@
 logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
 
 
+class NumberOfInputsMatchError(ValueError):
+    pass
+
+
+class NumberOfOutputsMatchError(ValueError):
+    pass
```
fxmarty, line 54: These are actually not used in `__main__.py` of the ONNX export, is it normal?

tests/onnxruntime/test_modeling.py
```diff
@@ -2217,17 +2210,20 @@ def test_compare_generation_to_io_binding(self, test_name: str, model_arch: str,
...
+            "cuda"
         )
 
         tokenizer = get_preprocessor(model_id)
         tokens = tokenizer("This is a sample output", return_tensors="pt").to("cuda")
         onnx_outputs = onnx_model.generate(**tokens, num_beams=5)
         io_outputs = io_model.generate(**tokens, num_beams=5)
+        import pdb
+
+        pdb.set_trace()
```
fxmarty, line 2226: to remove?

optimum/onnxruntime/base.py
```diff
@@ -0,0 +1,397 @@
...
+        batch_size = input_ids.size(0)
+        num_attention_heads = self.normalized_config.num_attention_heads
+        embed_size_per_head = self.normalized_config.hidden_size // num_attention_heads
+        sequence_length = input_ids.size(1)
+        if past_key_values is not None:
+            sequence_length += past_key_values[0].size(2)
+
+        half_shape = [batch_size, num_attention_heads]
+        if len(self.expected_key_symbolic_shape) == 3:
+            half_shape[0] *= half_shape.pop(1)
```
fxmarty, line 164:
```suggestion
            half_shape[0] = batch_size * num_attention_heads
            del half_shape[1]
```

optimum/exporters/onnx/model_configs.py
```diff
@@ -887,6 +894,24 @@ def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGen
...
+    @property
+    def outputs(self) -> Mapping[str, Mapping[int, str]]:
+        common_outputs = super().outputs
+        if self._behavior is ConfigBehavior.ENCODER:
+            for name, dyanmic_axes in common_outputs.items():
+                if name == "last_hidden_state":
+                    common_outputs[name][1] = f"{common_outputs[name][1]} / 2"
+                else:
+                    common_outputs[name] = dynamic_axes
```
fxmarty, line 912: test ordereddict update

optimum/utils/normalized_config.py
```diff
@@ -154,7 +160,7 @@ class NormalizedConfigManager:
         "bigbird_pegasus": BartLikeNormalizedTextConfig,
         "blenderbot": BartLikeNormalizedTextConfig,
         "blenderbot_small": BartLikeNormalizedTextConfig,
-        "bloom": NormalizedTextConfig.with_args(hidden_size="n_embd", num_layers="n_layer"),
+        "bloom": NormalizedTextConfig.with_args(hidden_size="hidden_size", num_layers="n_layer"),
```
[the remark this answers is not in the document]
michaelbenayoun: Handled.

JingyaHuang (commented):
LGTM, just left two nits.

optimum/exporters/onnx/base.py
```diff
@@ -126,21 +126,22 @@ class OnnxConfig(ExportConfig, ABC):
     _TASK_TO_COMMON_OUTPUTS = {
```
mht-sharma, line 126: small nit: Since this is a refactor, could you also make the following list alphabetical? I guess the audio keys are only misaligned.

optimum/exporters/onnx/base.py
```diff
@@ -587,13 +588,26 @@ def outputs(self) -> Mapping[str, Mapping[int, str]]:
...
-            for axis_idx, name in axes_names.items():
-                if "sequence" in name:
-                    axes_names[axis_idx] = sequence_name
-                # We reset the value as the order in common_outputs (OrderedDict) is lost otherwise
+            if self._behavior is ConfigBehavior.ENCODER:
+                sequence_name = "encoder_sequence_length"
+            else:
+                if "encoder" in name:
+                    sequence_name = "encoder_sequence_length"
                 else:
```
mht-sharma, line 596:
```suggestion
            if self._behavior is ConfigBehavior.ENCODER or "encoder" in name:
                sequence_name = "encoder_sequence_length"
            else:
```

optimum/exporters/onnx/base.py
```diff
@@ -587,13 +588,26 @@ def outputs(self) -> Mapping[str, Mapping[int, str]]:
...
+                    sequence_name = "decoder_sequence_length"
+
+            new_axes_names = {}
+            for axis_idx, axis_name in axes_names.items():
+                if "sequence" in axis_name:
+                    if not self.use_past_in_inputs:
+                        new_axes_names[axis_idx] = sequence_name
+                    else:
+                        # Trick to force it since ONNX sometimes infer a dynamic axis where it's not.
+                        new_axes_names[axis_idx] = "1"
```
mht-sharma, line 606: Does this happen also when we do not put the axis_id in dict?

optimum/exporters/onnx/config.py
```diff
@@ -104,6 +106,7 @@ def inputs(self) -> Mapping[str, Mapping[int, str]]:
                 self.add_past_key_values(common_inputs, direction="inputs")
 
         if self._behavior is ConfigBehavior.DECODER:
+            # TODO: it is actually encoder_last_hidden_state
             common_inputs["encoder_outputs"] = {0: "batch_size", 1: "encoder_sequence_length"}
```
mht-sharma, line 110: It still is `encoder_outputs` when the `def inputs` mean `torch inputs for export`, but would mean `encoder_hidden_state` if considered ONNX inputs.

For me it's the former and the naming is correct. Maybe a comment would help to avoid confusion?

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -336,349 +213,29 @@ class ORTEncoderForWhisper(ORTEncoder):
...
-            self._device.index,
-            self.name_to_np_type["input_features"],
-            tuple(input_features.shape),
-            input_features.data_ptr(),
-        )
-
-        # bind logits
-        output_shape, output_buffer = self.prepare_output_buffer(
-            batch_size=input_features.size(0),
-            sequence_length=input_features.size(2) // 2,
```
mht-sharma, line 359: How are you handling this case?

optimum/exporters/onnx/base.py
michaelbenayoun: Yes, basically I do that because ONNX infers that the axis is dynamic but it is not. I will look into this problem, which is not limited to this case, on another PR.

optimum/onnxruntime/modeling_seq2seq.py
michaelbenayoun: It is not handled. But as soon as this is merged, will work on a solution for that. The idea is to keep this PR as small as possible.

optimum/exporters/onnx/convert.py
michaelbenayoun: Yes, will be used in a PR coming this week.

optimum/onnxruntime/utils.py
```diff
@@ -224,7 +224,8 @@ def check_io_binding(providers: List[str], use_io_binding: Optional[bool] = None
                 "No need to enable IO Binding if the provider used is not CUDAExecutionProvider. IO Binding will be turned off."
             )
         use_io_binding = False
-
+    elif use_io_binding is None:
```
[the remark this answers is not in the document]
michaelbenayoun: About your first question, I do not think so because in the `to` method we update the providers, which can be blocking to have `use_io_binding`. But we actually dont want to use io binding when the initialization was done so that `self._use_io_binding = False`.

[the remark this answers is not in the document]
michaelbenayoun: I think you're right, removed it.

PR #699, opened 2023-01-17
Add saving of diffusion model additional components for onnx export

echarlaix:
Saves the model configuration, feature extractor and scheduler when exporting a stable diffusion model to ONNX using the CLI


fxmarty (approved):
LGTM!

PR #701, opened 2023-01-17
Fix the ORTQuantizer loading from specific file

echarlaix:
The `file_name` parameter of the `ORTQuantizer` `from_pretrained` method is currently discarded when it should be used if given by the users : resulting in errors when quantizing a seq2seq model as done in the current [documentation](https://huggingface.co/docs/optimum/onnxruntime/usage_guides/quantization#quantize-seq2seq-models)


fxmarty (approved):
LGTM!

PR #651, opened 2022-12-29
Support generation config in ORTModel

fxmarty:
This PR adds support for generation config in ORTModel, following https://github.com/huggingface/transformers/pull/20388

Note: we should really add nightly tests tracking on transformers/diffusers main.

PR #703, opened 2023-01-18
Add nightly tests on dependencies dev versions

fxmarty:
As per title.

Don't know if there is a better way.


.github/generate_dev_tests.py
```diff
@@ -0,0 +1,49 @@
...
+        workflox_yml = yaml.load(file, yaml.BaseLoader)
+
+        workflox_yml["name"] = "dev_" + workflox_yml["name"]
+        workflox_yml["on"] = {"schedule": [{"cron": "0 7 * * *"}]}
+
+        for i, step in enumerate(workflox_yml["jobs"]["build"]["steps"]):
+            if "name" in step and step["name"] == "Install dependencies":
+                workflox_yml["jobs"]["build"]["steps"][i][
+                    "run"
+                ] += "pip install -U git+https://github.com/huggingface/transformers@main#egg=evaluate\npip install -U git+https://github.com/huggingface/transformers@main#egg=diffusers\npip install -U git+https://github.com/huggingface/transformers@main#egg=transformers\n"
```
regisss, line 30: Why do we install packages this way with the `#egg`? Not sure what it does exactly (it seems useful for installing extras, but that's not the case here).
I would have expected:
```bash
pip install -U git+https://github.com/huggingface/transformers
pip install -U git+https://github.com/huggingface/diffusers
pip install -U git+https://github.com/huggingface/evaluate
```

fxmarty: I just copy pasted, thanks!

fxmarty:
@regisss They won't pass at least due to https://github.com/huggingface/optimum/pull/651, so it's correct the tests won't probably be super useful for now as some of them will already fail due to this. What do you think?

regisss:
> @regisss They won't pass at least due to #651, so it's correct the tests won't probably be super useful for now as some of them will already fail due to this. What do you think?

Ah yes true, I forgot about this change. For me it's fine to merge as it is since we know why these tests will fail. Better to have failing tests than no tests at all, and they probably don't all rely on `generate`.

PR #705, opened 2023-01-19
Add stable diffusion VAE encoder export

echarlaix:


tests/exporters/test_onnx_export.py
```diff
@@ -375,7 +387,6 @@ def test_pytorch_export_for_stable_diffusion_models(self, model_name):
             validate_models_outputs(
                 models_and_onnx_configs=models_and_onnx_configs,
                 onnx_named_outputs=onnx_outputs,
-                atol=1e-3,
```
fxmarty, line 378: why remove?

optimum/exporters/onnx/utils.py
```diff
@@ -165,14 +165,23 @@ def get_stable_diffusion_models_for_export(
     unet_onnx_config = onnx_config_constructor(pipeline.unet.config)
     models_for_export["unet"] = (pipeline.unet, unet_onnx_config)
 
-    # VAE
-    vae = copy.deepcopy(pipeline.vae)
-    vae.forward = lambda latent_sample: vae.decode(z=latent_sample)
+    # VAE Encoder
+    vae_encoder = copy.deepcopy(pipeline.vae)
+    vae_encoder.forward = lambda sample: {"latent_sample": vae_encoder.encode(x=sample)["latent_dist"].sample()}
```
fxmarty, line 170: I have a hard time reading this but again am not familiar. Could you add a comment as to why it is necessary to override the forward with this?

optimum/exporters/onnx/utils.py
```diff
@@ -165,14 +165,23 @@ def get_stable_diffusion_models_for_export(
...
+        model=vae_encoder, exporter="onnx", task="semantic-segmentation", model_type="vae-encoder"
     )
-    vae_onnx_config = vae_config_constructor(vae.config)
-    models_for_export["vae"] = (vae, vae_onnx_config)
+    vae_onnx_config = vae_config_constructor(vae_encoder.config)
+    models_for_export["vae_encoder"] = (vae_encoder, vae_onnx_config)
+
+    # VAE Decoder
+    vae_decoder = copy.deepcopy(pipeline.vae)
+    vae_decoder.forward = lambda latent_sample: vae_decoder.decode(z=latent_sample)
```
fxmarty, line 179: same comment

optimum/exporters/onnx/model_configs.py
```diff
@@ -632,7 +632,32 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
...
+    @property
+    def outputs(self) -> Mapping[str, Mapping[int, str]]:
+        return {
+            "latent_sample": {0: "batch_size", 1: "num_channels_latent", 2: "height_latent", 3: "width_latent"},
+        }
+
+
+class VaeDecoderOnnxConfig(ViTOnnxConfig):
     ATOL_FOR_VALIDATION = 1e-3
     DEFAULT_ONNX_OPSET = 14
```
fxmarty, line 662: Could you add a comment as to why the opset 13, 12 fail?

optimum/exporters/onnx/model_configs.py
```diff
@@ -632,7 +632,32 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
         return dummy_inputs
 
 
-class VaeOnnxConfig(ViTOnnxConfig):
+class VaeEncoderOnnxConfig(ViTOnnxConfig):
+    ATOL_FOR_VALIDATION = 1e-2
+    DEFAULT_ONNX_OPSET = 14
```
fxmarty, line 637: same comment

optimum/exporters/onnx/utils.py
fxmarty: Are the operations like `latents = 1 / 0.18215 * latents` and `image = (image / 2 + 0.5).clamp(0, 1)` in the onnx vae decoder?

fxmarty (approved):
LGTM, left just a few understanding/comment questions

Edit: actually, the memory duplication is a concern.

optimum/exporters/onnx/__main__.py
```diff
@@ -116,7 +116,12 @@ def main():
         args.for_ort and (model.config.is_encoder_decoder or task.startswith("causal-lm"))
     ):
         if task == "stable-diffusion":
-            output_names = ["text_encoder/model.onnx", "unet/model.onnx", "vae_decoder/model.onnx"]
+            output_names = [
+                "text_encoder/model.onnx",
```
fxmarty, line 120: Is it always necessary? For example, `pipeline_stable_diffusion.py` has no call to `encode()`, while `pipeline_cycle_diffusion.py` and `pipeline_stable_diffusion_img2img.py` do.

optimum/exporters/onnx/__main__.py
```diff
@@ -116,7 +116,12 @@ def main():
...
+                "unet/model.onnx",
+                "vae_encoder/model.onnx",
```
fxmarty, line 122: Is the encoder always necessary? For example `pipeline_stable_diffusion.py` never seems to call `encode()`.

optimum/exporters/onnx/__main__.py
```diff
@@ -116,7 +116,12 @@ def main():
...
+                "vae_decoder/model.onnx",
```
fxmarty, line 123: Aren't we duplicating memory by doing this? I think we should definitely avoid it.

tests/exporters/test_onnx_export.py
echarlaix: It allows to take the atol of each of the indivual component (specified in each onnx config)

optimum/exporters/onnx/utils.py
echarlaix: Yes the VAE should be exported in two different components (encoder and decoder) as they are used independently in the stable diffusion pipeline

optimum/exporters/onnx/model_configs.py
```diff
@@ -632,7 +632,32 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
         return dummy_inputs
 
 
-class VaeOnnxConfig(ViTOnnxConfig):
+class VaeEncoderOnnxConfig(ViTOnnxConfig):
+    ATOL_FOR_VALIDATION = 1e-2
```
michaelbenayoun, line 636: It seems huge no?

optimum/exporters/onnx/model_configs.py
```diff
@@ -632,7 +632,32 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
         return dummy_inputs
 
 
-class VaeOnnxConfig(ViTOnnxConfig):
+class VaeEncoderOnnxConfig(ViTOnnxConfig):
```
michaelbenayoun, line 635: Why inherint from `ViTOnnxConfig` if we do not use much from it?

optimum/exporters/onnx/utils.py
echarlaix: No they are not

optimum/exporters/onnx/utils.py
fxmarty: all good

optimum/exporters/onnx/__main__.py
fxmarty: Actually not, the encoder and decoder are obviously separate. My bad.

optimum/exporters/onnx/model_configs.py
echarlaix: The text encoder (a `CLIPTextModel`) needs a min opset  of 14 to support the [Trilu](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Trilu) operator. Because we export all components together, I've set the same default opset to every components

optimum/exporters/onnx/__main__.py
echarlaix: No you're right it depends on the task but in my opinion it makes sense to export the entire model which can be later used for any tasks

tests/exporters/test_onnx_export.py
echarlaix: This was removed as the VAE encoder has a much bigger atol, I will add it back now that the decoder test is adapted !

optimum/exporters/onnx/model_configs.py
echarlaix: Good catch, it should be `VisionOnnxConfig`

optimum/exporters/onnx/model_configs.py
echarlaix: This is due to https://github.com/huggingface/diffusers/blob/v0.11.1/src/diffusers/models/vae.py#L357, currently I've found that a tolerance of `1e-2` was respected for common models

optimum/exporters/onnx/model_configs.py
fxmarty: Could you add this as a comment?

optimum/exporters/onnx/utils.py
fxmarty: So the `vae_encoder.forward = lambda sample: {"latent_sample": vae_encoder.encode(x=sample)["latent_dist"].sample()}` corresponds to the two lines I linked? Could you add a link as a comment, e.g. https://github.com/huggingface/diffusers/blob/7c82a16fc14840429566aec40eb9e65aa57005fd/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion_inpaint.py#L498 , if that is what the encoder represents?

optimum/exporters/tasks.py
```diff
@@ -623,9 +623,13 @@ class TasksManager:
             "audio-xvector",
             onnx="UniSpeechSATOnnxConfig",
         ),
-        "vae": supported_tasks_mapping(
+        "vae-encoder": supported_tasks_mapping(
             "semantic-segmentation",
```
regisss, line 627: Nit: it sounds weird to have the task "semantic-segmentation" for both the encoder and decoder since they don't perform this task alone (only when they are associated with each other)

regisss (approved):
LGTM :fire: 
I just left one nit.

optimum/exporters/tasks.py
echarlaix: I agree that it's not super clear, we could rename the corresponding key of [`_TASK_TO_COMMON_OUTPUTS`](https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/base.py#L152) to something more general in a future PR (as it's not related to this one)

optimum/exporters/onnx/model_configs.py
```diff
@@ -632,8 +635,37 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
...
+    # operator support, available since opset 14
+    DEFAULT_ONNX_OPSET = 14
+
+    NORMALIZED_CONFIG_CLASS = NormalizedConfig.with_args(
+        num_channels="in_channels",
+        image_size="sample_size",
+        allow_new=True,
+    )
+
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyVisionInputGenerator,)
```
michaelbenayoun, line 650: Can be removed because it is already defined in `VisionOnnxConfig`.

optimum/exporters/tasks.py
michaelbenayoun: Maybe create an issue to not forger about it.

michaelbenayoun (approved):
LGTM

PR #704, opened 2023-01-19
Add test for code examples in the documentation and docstrings

fxmarty:
As per title. This allows to catch failing snippets and to have a more reliable documentation.

Fixes https://github.com/huggingface/optimum/issues/677

PR #713, opened 2023-01-24
Fix documentation for doctest tests to pass

fxmarty:
As per title, this goes along https://github.com/huggingface/optimum/pull/704.

Catching a few code snippets failing along the way.


regisss (approved):
Thanks for fixing all this :fire:

PR #716, opened 2023-01-24
`exporters.tflite` initial support

michaelbenayoun:
# What does this PR do?

This PR adds the `exporters.tflite` subpackage, enabling the export of Transformers models to TFlite.

### DIFF
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-# Copyright 2020 The HuggingFace Team. All rights reserved.
+# Copyright 2022 The HuggingFace Team. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/optimum/exporters/tflite/base.py
+++ b/optimum/exporters/tflite/base.py
@@ -1,0 +1,209 @@
+# coding=utf-8
+# Copyright 2023 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""TensorFlow Lite configuration base classes."""
+
+from ctypes import ArgumentError
+from collections import OrderedDict
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Dict, List, Optional, Union, Any
+
+from transformers.utils import is_tf_available
+# TODO: handle tensorflow dependency.
+import tensorflow as tf
+
+from ..base import ExportConfig
+
+
+if TYPE_CHECKING:
+    from transformers import PretrainedConfig, TFPreTrainedModel
+    from ...utils import DummyInputGenerator
+
+    if is_tf_available():
+        from tensorflow import TensorSpec
+
+
+class MissingMandatoryAxisDimension(ValueError):
+    pass
+
+
+class TFLiteConfig(ExportConfig, ABC):
+    NORMALIZED_CONFIG_CLASS = None
+    DUMMY_INPUT_GENERATOR_CLASSES = ()
+    ATOL_FOR_VALIDATION: Union[float, Dict[str, float]] = 1e-5
+    MANDATORY_AXES = ()
+
+    _TASK_TO_COMMON_OUTPUTS = {
+        "causal-lm": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "default": OrderedDict({"last_hidden_state": {0: "batch_size", 1: "sequence_length"}}),
+        "image-classification": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "image-segmentation": OrderedDict(
+            {
+                "logits": {0: "batch_size", 1: "sequence_length"},
+                "pred_boxes": {0: "batch_size", 1: "sequence_length"},
+                "pred_masks": {0: "batch_size", 1: "sequence_length"},
+            }
+        ),
+        "masked-im": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "masked-lm": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "multiple-choice": OrderedDict({"logits": {0: "batch_size"}}),
+        "object-detection": OrderedDict(
+            {
+                "logits": {0: "batch_size", 1: "sequence_length"},
+                "pred_boxes": {0: "batch_size", 1: "sequence_length"},
+            }
+        ),
+        "question-answering": OrderedDict(
+            {
+                "start_logits": {0: "batch_size", 1: "sequence_length"},
+                "end_logits": {0: "batch_size", 1: "sequence_length"},
+            }
+        ),
+        "semantic-segmentation": OrderedDict({"logits": {0: "batch_size", 1: "num_labels", 2: "height", 3: "width"}}),
+        "seq2seq-lm": OrderedDict(
+            {
+                "logits": {0: "batch_size", 1: "decoder_sequence_length"},
+                "encoder_last_hidden_state": {0: "batch_size", 1: "encoder_sequence_length"},
+            }
+        ),
+        "sequence-classification": OrderedDict({"logits": {0: "batch_size"}}),
+        "token-classification": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "speech2seq-lm": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "audio-classification": OrderedDict({"logits": {0: "batch_size"}}),
+        "audio-frame-classification": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "audio-ctc": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "audio-xvector": OrderedDict({"logits": {0: "batch_size"}, "embeddings": {0: "batch_size"}}),
+    }
+
+    def __init__(
+        self,
+        config: "PretrainedConfig",
+        task: str,
+        batch_size: int = 1,
+        sequence_length: Optional[int] = None,
+        num_choices: Optional[int] = None,
+        width: Optional[int] = None,
+        height: Optional[int] = None,
+        num_channels: Optional[int] = None,
+        feature_size: Optional[int] = None,
+        nb_max_frames: Optional[int] = None,
+    ):
+        self._config = config
+        self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
+        self.task = task
+
+        # self.batch_size = batch_size
+        # self.sequence_length = sequence_length
+        # self.num_choices = num_choices
+        # self.width = width
+        # self.height = height
+        # self.num_channels = num_channels
+        # self.feature_size = feature_size
+        # self.nb_max_frames = nb_max_frames
+
+        self._axes: Dict[str, int] = {}
+
+        self._validate_and_update_mandatory_axes(
+            batch_size=batch_size,
+            sequence_length=sequence_length,
+            num_choices=num_choices,
+            width=width,
+            height=height,
+            num_channels=num_channels,
+            feature_size=feature_size,
+            nb_max_frames=nb_max_frames,
+
+        )
+
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name in self.MANDATORY_AXES:
+            if value is None:
+                raise ValueError(f"Cannot set `None` to {name} because it is a mandatory axis.")
+            self._axes[name] = value
+        return super().__setattr__(name, value)
+
+    def _validate_and_update_mandatory_axes(self, **kwargs):
+        for axis in self.MANDATORY_AXES:
+            if isinstance(axis, tuple):
+                tasks, name = axis
+                if not isinstance(tasks, tuple):
+                    tasks = (tasks,)
+                if self.task not in tasks:
+                    continue
+            else:
+                name = axis
+            axis_dim = kwargs[name]
+            if axis_dim is None:
+                if self._normalized_config.has_attribute(name):
+                    axis_dim =  getattr(self._normalized_config, name)
+                else:
+                    raise MissingMandatoryAxisDimension(
+                        f"The value for the {name} axis is missing, it is needed to perform the export to TensorFlow Lite."
+                    )
+            self._axes[name] = axis_dim
+
+    def _create_dummy_input_generator_classes(self) -> List["DummyInputGenerator"]:
+        return [cls_(self.task, self._normalized_config, **self._axes) for cls_ in self.DUMMY_INPUT_GENERATOR_CLASSES]
+
+
+    @property
+    @abstractmethod
+    def inputs(self) -> List[str]:
+        raise NotImplementedError()
+
+    @property
+    def outputs(self) -> List[str]:
+        return list(self._TASK_TO_COMMON_OUTPUTS[self.task].keys())
+
+    @property
+    def inputs_specs(self) -> List["TensorSpec"]:
+        dummy_inputs_generators = self._create_dummy_input_generator_classes()
+        inputs_specs = []
+
+        for input_name in self.inputs:
+            input_was_inserted = False
+            for dummy_input_gen in dummy_inputs_generators:
+                if dummy_input_gen.supports_input(input_name):
+                    dummy_input = dummy_input_gen.generate(input_name, framework="tf")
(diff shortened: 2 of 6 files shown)


sayakpaul:
Thanks for your work here. If it's not a blocker, can I review it on Monday?

optimum/exporters/tflite/base.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def model_to_signatures(
+        self, model: "TFPreTrainedModel", **model_kwargs: Any
+    ) -> Dict[str, "tf.types.experimental.ConcreteFunction"]:
+        """
+        Creates the signatures that will be used when exporting the model to a `tf.SavedModel`.
+        Each signature can be used to perform inference on the model for a given set of inputs.
+
+        Auto-encoder models have only one signature, decoder models can have two, one for the decoder without
+        caching, and one for the decoder with caching, seq2seq models can have three, and so on.
```
sayakpaul, line 271: * Does this function only exist to aid SavedModel creation? 
* Does `model.save("my_model")` (`model` is an instance of `TFPretrainedModel`) lead to any error? 
* If this helps, TFLite allows us to load specific signatures: https://www.tensorflow.org/lite/guide/signatures

optimum/exporters/tflite/convert.py
```diff
@@ -0,0 +1,183 @@
...
+    model.config.return_dict = True
+
+    # Check if we need to override certain configuration item
+    if config.values_override is not None:
+        logger.info(f"Overriding {len(config.values_override)} configuration item(s)")
+        for override_config_key, override_config_value in config.values_override.items():
+            logger.info(f"\t- {override_config_key} -> {override_config_value}")
+            setattr(model.config, override_config_key, override_config_value)
+
+    signatures = config.model_to_signatures(model)
```
sayakpaul, line 173: We should be able to call `save()` directly on the model to save it as a `SavedModel` resource unless there's an important reason not to. Ccing @Rocketknight1 @gante. 

Also, do we need to run the intermediate save? I can imagine it's likely there for better standardization of the modules but saving a TF model into SavedModel takes a lot of time (because of the compilation involved). 

Could we directly take a `TFPretrainedModel` or even its concrete function and use that for the TFLite export ([here's](https://colab.research.google.com/gist/sayakpaul/6e4f06232b689afa74a43b1b8e2be243/scratchpad.ipynb) a simple example of obtaining concrete function from a Keras model directly)?

optimum/exporters/tflite/convert.py
```diff
@@ -0,0 +1,183 @@
...
+
+    with TemporaryDirectory() as tmp_dir_name:
+        tf.saved_model.save(model, tmp_dir_name, signatures)
+        converter = tf.lite.TFLiteConverter.from_saved_model(tmp_dir_name)
+        tflite_model = converter.convert()
+
+    with open(output, "wb") as fp:
+        fp.write(tflite_model)
+
+    return config.inputs, config.outputs
```
sayakpaul, line 183: There is no optimization involved here. 

TFLite has three post-training optimization offerings through quantization and other things (kernel fusion, for example): https://www.tensorflow.org/lite/performance/post_training_quantization

sayakpaul (commented):
Thanks a lot for working on this!

I have left comments on the TFLite export design. Let me know if anything is unclear.

optimum/exporters/tflite/base.py
michaelbenayoun: - Yes, basically it provides the signatures that will be used for the export of the model.
- I haven't tried, but I think this is equivalent to what we do.
- That's exactly what we use: we export a signature that can be used easily by the `tf.lite.Interpreter`. For decoder and seq2seq models, we will define multiple signatures.

optimum/exporters/tflite/convert.py
michaelbenayoun: Yes, of course this is planned. The idea with this PR is to add the minimal code needed to iterate fast later. That is why there is no other configuration that the one for BERT.

optimum/exporters/tflite/convert.py
sayakpaul: > I could not find a way to provide signatures in that case

Which case? Could you expand a bit more? 

> It should not provide a speedup in the end because it is (or it more precisely it will be) mandatory to provide a trackable object when exporting with a concrete function, which in turn will do the export to a saved model in the backend.

Could you elaborate a bit more on this? Why providing the trackable object is a requirement?

michaelbenayoun: - I could not find a way to provide signatures when exporting from a concrete function.
- When exporting from a concrete function, if you dont provide a trackable object, it kind of works, but a warning tells you that it will be deprecated => you can also see this warning from your [code sample](https://colab.research.google.com/gist/sayakpaul/6e4f06232b689afa74a43b1b8e2be243/scratchpad.ipynb#scrollTo=KLuOmvf5Fm0B&line=2&uniqifier=1)

michaelbenayoun: Yes, that's not an issue, you can specify the signature you want, that's what I did.

Rocketknight1: Ah, sorry, I was looking at an old version of the page and didn't see the up-to-date conversation!

To support signatures when exporting from a concrete function, you need to make one concrete function per signature. That's exactly what a `ConcreteFunction` is in TensorFlow - it's a function with one precisely defined signature.

Also, if you're getting errors about a missing `trackable_obj`, I'm 99% sure the `trackable_obj` you need is just the original `model` object. [Here](https://colab.research.google.com/gist/Rocketknight1/781023e44fd88f62b67d3c46a7cd50f9/scratchpad.ipynb) is an updated code sample that uses multiple concrete functions and doesn't have the error.

Rocketknight1: Ah, sorry, I follow now! Yeah, I don't know of any good way to name the signatures when using concrete functions.

optimum/commands/export/tflite.py
```diff
@@ -0,0 +1,126 @@
...
+from ...exporters import TasksManager
+
+
+def parse_args_tflite(parser):
+    required_group = parser.add_argument_group("Required arguments")
+    required_group.add_argument(
+        "-m", "--model", type=str, required=True, help="Model ID on huggingface.co or path on disk to load model from."
+    )
+    required_group.add_argument(
+        "output", type=Path, help="Path indicating the directory where to store generated ONNX model."
```
mfuntowicz, line 28: ONNX -> TFlite ?

optimum/commands/export/tflite.py
```diff
@@ -0,0 +1,126 @@
...
+        "--sequence_length",
+        type=int,
+        default=None,
+        help=f"Text tasks only. Sequence length {doc_input}",
+    )
+    input_group.add_argument(
+        "--num_choices",
+        type=int,
+        default=None,
+        help=f"Text tasks only. Num choices {doc_input}",
```
mfuntowicz, line 80: Maybe something more detailed? 

What about `Text classification tasks only. How many labels to return` or something like this? wdyt?

optimum/commands/export/tflite.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    input_group.add_argument(
+        "--width",
+        type=int,
+        default=None,
+        help=f"Image tasks only. Width {doc_input}",
```
mfuntowicz, line 86: Image width?

optimum/commands/export/tflite.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    input_group.add_argument(
+        "--height",
+        type=int,
+        default=None,
+        help=f"Image tasks only. Height {doc_input}",
```
mfuntowicz, line 92: Image height?

optimum/commands/export/__init__.py
```diff
@@ -17,12 +17,17 @@
...
 from .. import BaseOptimumCLICommand
 from .onnx import ONNXExportCommand, parse_args_onnx
+from .tflite import TFLiteExportCommand, parse_args_tflite
 
 
 def onnx_export_factory(_):
     return ONNXExportCommand(" ".join(sys.argv[3:]))
 
 
+def tflite_export_factory(_):
```
fxmarty, line 27: Could you add a command line test?

optimum/commands/export/tflite.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    input_group.add_argument(
+        "--num_channels",
+        type=int,
+        default=None,
+        help=f"Image tasks only. Number of channels {doc_input}",
```
mfuntowicz, line 98: Number of channels used to represent the image (RGB = 3, ARGB = 4, GREY =1)?

optimum/commands/export/tflite.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    input_group.add_argument(
+        "--feature_size",
+        type=int,
+        default=None,
+        help=f"Audio tasks only. Feature size {doc_input}",
```
mfuntowicz, line 104: What is feature size here?

optimum/commands/export/tflite.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    optional_group.add_argument("--cache_dir", type=str, default=None, help="Path indicating where to store cache.")
+    optional_group.add_argument(
+        "--trust-remote-code",
+        action="store_true",
+        help="Allow to use custom code for the modeling hosted in the model repository. This option should only be set for repositories you trust and in which you have read the code, as it will execute on your local machine arbitrary code present in the model repository.",
+    )
+
+    input_group = parser.add_argument_group("Input shapes")
+    doc_input = "that the TensorFlow Lite will be able to take as input."
```
fxmarty, line 63:
```suggestion
    doc_input = "that the TensorFlow Lite exported model will be able to take as input."
```

optimum/exporters/onnx/config.py
```diff
@@ -1,5 +1,5 @@
 # coding=utf-8
-# Copyright 2020 The HuggingFace Team. All rights reserved.
+# Copyright 2022 The HuggingFace Team. All rights reserved.
```
mfuntowicz, line 2: 2023 ? 😄

optimum/exporters/tasks.py
```diff
@@ -1011,6 +1013,8 @@ def get_exporter_config_constructor(
                 The model type to retrieve the config for.
             model_name (`Optional[str]`, defaults to `None`):
                 The name attribute of the model object, only used for the exception message.
+            exporter_config_kwargs( `Optional[Dict[str, Any]]`, defaults to `None`):
```
fxmarty, line 1016:
```suggestion
            exporter_config_kwargs(`Optional[Dict[str, Any]]`, defaults to `None`):
```

fxmarty, line 1016: just for my knowledge, why is this needed? I see it is only used in `test_tflite_export.py`.

optimum/exporters/tflite/__main__.py
```diff
@@ -0,0 +1,108 @@
...
+    model = TasksManager.get_model_from_task(
+        task, args.model, framework="tf", cache_dir=args.cache_dir, trust_remote_code=args.trust_remote_code
+    )
+
+    tflite_config_constructor = TasksManager.get_exporter_config_constructor(model=model, exporter="tflite", task=task)
+    # TODO: find a cleaner way to do this.
+    shapes = {name: getattr(args, name) for name in tflite_config_constructor.func.get_mandatory_axes_for_task(task)}
+    tflite_config = tflite_config_constructor(model.config, **shapes)
+
+    # Ensure the requested opset is sufficient
```
fxmarty, line 63: opset is an ONNX thing no?

optimum/exporters/tflite/base.py
```diff
@@ -0,0 +1,288 @@
...
+    from transformers import PretrainedConfig, TFPreTrainedModel
+
+    from ...utils import DummyInputGenerator
+
+    if is_tf_available():
+        from tensorflow import TensorSpec
+
+
+class MissingMandatoryAxisDimension(ValueError):
+    pass
```
fxmarty, line 40: should this be in `error_utils.py`?

optimum/exporters/onnx/__main__.py
```diff
@@ -22,17 +22,10 @@
 from ...commands.export.onnx import parse_args_onnx
 from ...utils import DEFAULT_DUMMY_SHAPES, logging
 from ...utils.save_utils import maybe_save_preprocessors
+from ..error_utils import AtolError, OutputMatchError, ShapeError
 from ..tasks import TasksManager
 from .base import OnnxConfigWithPast
-from .convert import (
-    AtolError,
-    OutputMatchError,
-    ShapeError,
```
fxmarty, line 30: Unless mistaken, those three are used in `__main__.py`. Why remove them?

optimum/exporters/tflite/base.py
```diff
@@ -0,0 +1,288 @@
...
+            "num_choices": num_choices,
+            "width": width,
+            "height": height,
+            "num_channels": num_channels,
+            "feature_size": feature_size,
+            "nb_max_frames": nb_max_frames,
+            "audio_sequence_length": audio_sequence_length,
+        }
+        for name, value in axes_values.items():
+            setattr(self, name, value)
```
fxmarty, line 136: This design seem different than the one for ONNX export.

optimum/exporters/tflite/config.py
```diff
@@ -0,0 +1,33 @@
...
+logger = logging.get_logger(__name__)
+
+
+class TextEncoderTFliteConfig(TFLiteConfig):
+    """
+    Handles encoder-based text architectures.
+    """
+
+    DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator,)
+    MANDATORY_AXES = ("batch_size", "sequence_length", ("multiple-choice", "num_choices"))
```
fxmarty, line 33: What is the motivation for `MANDATORY_AXES` compared to the ONNX export?

optimum/exporters/tflite/convert.py
mfuntowicz: For reference in `optimum.tir` I was generating on the fly concrete definition with custom signatures, maybe it can help (please dont judge me ...):

```python
    def export_model_to_mlir(self, model: TFPreTrainedModel, examples: Optional = None, dynamic_axes: List[int] = None):
        # Define function signature
        if examples is not None:
            serving_signature = {
                "input_ids": tf.TensorSpec((1, examples[0].shape[1]), dtype=tf.dtypes.int32, name="input_ids"),
                "attention_mask": tf.TensorSpec((1, examples[1].shape[1]), dtype=tf.dtypes.int8, name="attention_mask"),
                # "token_type_ids": tf.TensorSpec((1, None), dtype=tf.dtypes.int8, name="token_type_ids")
            }
        else:
            serving_signature = {
                "input_ids": tf.TensorSpec((1, None), dtype=tf.dtypes.int32, name="input_ids"),
                "attention_mask": tf.TensorSpec((1, None), dtype=tf.dtypes.int8, name="attention_mask"),
                # "token_type_ids": tf.TensorSpec((1, None), dtype=tf.dtypes.int8, name="token_type_ids")
            }

        # Traced function to inject at module-level
        @tf.function(input_signature=[serving_signature])
        def forward(self, inputs):
            outputs = self.eager_serving(inputs)
            return model.serving_output(outputs)

        # Bind the function on the module (TODO: my eyes are already crying blood at this stage ...)
        model.forward = forward.__get__(model, type(model))
```

mfuntowicz (approved):
LGTM ! Maybe some wording adjustements to be easier for users to grabs some keyword args, but overall it's really cool!

Thanks a lot for bringing this into optimum!

optimum/commands/export/tflite.py
michaelbenayoun: Fixed!

optimum/exporters/onnx/config.py
michaelbenayoun: Nope, this file was written in 2022!

optimum/commands/export/__init__.py
michaelbenayoun: Done!

optimum/exporters/tasks.py
michaelbenayoun: It is used to provide arguments to the config initalizer when using the `TasksManager`.

optimum/exporters/tflite/base.py
michaelbenayoun: Not for now. maybe later. but this is quite specific to the `TFLiteConfig`.

optimum/exporters/onnx/__main__.py
michaelbenayoun: I import them from the proper file since it moved.

optimum/exporters/tflite/config.py
michaelbenayoun: Mandatory axes are the axes which need to be specified to export to `TFLite` since only static shapes are supported (except the batch siize)

sayakpaul: Just ensuring we are aware:

https://www.tensorflow.org/api_docs/python/tf/lite/Interpreter#resize_tensor_input

optimum/exporters/tasks.py
fxmarty: Why is it required for tflite and not onnx export?

optimum/exporters/tflite/config.py
michaelbenayoun: Yes, I tried that once. I do not think it always works. But I would suggest going that path for now and check for dynamic shapes later.

sayakpaul: +1 to that. Always better to do it progressively.

optimum/exporters/tasks.py
michaelbenayoun: To provide axes values.

Issue #717, opened 2023-01-24
Seq2Seq inference with TensorrtExecutionProvider with `use_cache=True` results in `Segmentation fault (core dumped)`


fxmarty:
### System Info

```shell
optimum dev
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

No issue arises with CUDAExecutionProvider.

```python
from optimum.onnxruntime import ORTModelForCausalLM
from transformers import AutoTokenizer

provider_options = {
    "trt_engine_cache_enable": True,
    "trt_engine_cache_path": "tmp/trt_cache_gpt2_example"
}

print("Initializing ort_model")
ort_model = ORTModelForCausalLM.from_pretrained(
    "gpt2_onnx",
    use_cache=True,
    provider="TensorrtExecutionProvider",
    provider_options=provider_options
)

tokenizer = AutoTokenizer.from_pretrained("gpt2")

print("build engine for a short sequence")
text = ["short"]
encoded_input = tokenizer(text, return_tensors="pt").to("cuda")
output = ort_model(**encoded_input)

print("build engine for a long sequence")
text = [" a very long input just for demo purpose, this is very long" * 10]
encoded_input = tokenizer(text, return_tensors="pt").to("cuda")
output = ort_model(**encoded_input)

text = ["Replace me by any text you'd like."]
encoded_input = tokenizer(text, return_tensors="pt").to("cuda")

print("Generate!")
for i in range(3):
    output = ort_model.generate(**encoded_input)
    print(tokenizer.decode(output[0]))
```

Error:
```
2023-01-24 16:22:51.520935518 [E:onnxruntime:Default, tensorrt_execution_provider.h:58 log] [2023-01-24 15:22:51   ERROR] 4: [graphShapeAnalyzer.cpp::processCheck::587] Error Code 4: Internal Error (IAssertionLayer (Unnamed Layer* 105) [Assertion]: condition[0] is false: 0. For input: 'attention_mask' all named dimensions that share the same name must be equal. Note: Named dimensions were present on the following axes: 1 (name: 'past_sequence_length + sequence_length'), 2 (name: 'past_sequence_length + sequence_length'))
```

### Expected behavior

no error.

Additional tests with TRT execution provider should be added.

fxmarty:
[...]

PR #722, opened 2023-01-25
Add TensorrtExecutionProvider modeling tests

fxmarty:
As per title.

Would have catched https://github.com/huggingface/optimum/issues/717


regisss (approved):
LGTM!

PR #726, opened 2023-01-26
Whisper shape inference fix

michaelbenayoun:
# What does this PR do?

Fixes the shape inference issue for Whisper model.

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -217,6 +217,9 @@ class ORTModel(OptimizedModel):
         AutoConfig.register(self.model_type, AutoConfig)
         self.auto_model_class.register(AutoConfig, self.__class__)
 
+        # Define the pattern here to avoid recomputing it everytime.
+        self.output_shape_inference_pattern = re.compile(r"([a-zA-Z_]+)|([0-9]+)|([+-/*])|([\(\)])")
+
     def __init__(
         self,
         model: ort.InferenceSession,
@@ -632,21 +635,28 @@ class ORTModel(OptimizedModel):
         """
         if isinstance(axis_name, int):
             return axis_name
-        result = dimensions.get(axis_name, None)
-        if result is None:
-            subparts = axis_name.split("+")
-            result = 0
-            for subname in subparts:
-                subname = subname.strip()
-                subresult = dimensions.get(subname, None)
-                if subresult is None:
-                    try:
-                        subresult = int(subname)
-                    except ValueError:
-                        result = None
-                        break
-                result += subresult
-        return axis_name if result is None else result
+
+        tokens = []
+        for idx, match in enumerate(re.finditer(self.output_shape_inference_pattern, axis_name)):
+            groups = match.groups()
+            matched_group = None
+            for idx, group in enumerate(groups):
+                if group is not None:
+                    matched_group = idx
+                    break
+            if matched_group is None:
+                raise ValueError(f"Could not perform shape inference on {axis_name}")
+
+            if matched_group == 0:
+                dim = dimensions.get(groups[0], None)
+                if dim is None or not isinstance(dim, int):
+                    return axis_name
+                tokens.append(str(dim))
+            else:
+                tokens.append(groups[matched_group])
+
+        # Here it should be problematic to use eval since anything not matching the pattern would trigger an exception.
+        return int(eval(" ".join(tokens)))
 
     def _prepare_io_binding(
         self,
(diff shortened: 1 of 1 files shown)


optimum/onnxruntime/modeling_ort.py
```diff
@@ -217,6 +217,9 @@ def shared_attributes_init(
         AutoConfig.register(self.model_type, AutoConfig)
         self.auto_model_class.register(AutoConfig, self.__class__)
 
+        # Define the pattern here to avoid recomputing it everytime.
+        self.output_shape_inference_pattern = re.compile(r"([a-zA-Z_]+)|([0-9]+)|([+-/*])|([\(\)])")
```
fxmarty, line 221: What is `([\(\)])`?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -632,21 +635,32 @@ def _output_shape_inference(self, axis_name: Union[str, int], dimensions: Dict[s
...
+
+            if matched_group == 0:
+                dim = dimensions.get(groups[0], None)
+                if dim is None or not isinstance(dim, int):
+                    return axis_name
+                tokens.append(str(dim))
+            else:
+                tokens.append(groups[matched_group])
+
+        # Here it should not be problematic to use eval since anything not matching the pattern would trigger an
```
fxmarty, line 661: Rather than raising an exception, it's just not captured by `re.finditer(self.output_shape_inference_pattern, axis_name)`, right?

optimum/onnxruntime/modeling_ort.py
```diff
@@ -632,21 +635,32 @@ def _output_shape_inference(self, axis_name: Union[str, int], dimensions: Dict[s
...
-                    except ValueError:
-                        result = None
-                        break
-                result += subresult
-        return axis_name if result is None else result
+        # It is actually covered below, but this is to make things faster.
+        elif axis_name in dimensions:
+            return dimensions[axis_name]
+
+        tokens = []
```
fxmarty, line 642: Reading the code, it is not clear to me what this contains, at first sight. I understand it should be a list of int as strings, from the values of `dimensions`. Is this correct?

optimum/onnxruntime/modeling_ort.py
michaelbenayoun: No, if it's not captured it means the axis name contains something that is not supported, we want to fail here.

optimum/onnxruntime/modeling_ort.py
michaelbenayoun: It contains each subpart of `axis_name`, where each sub axis name as be changed to actual value using `dimensions`.

optimum/onnxruntime/modeling_ort.py
fxmarty: Yes - so basically we enter into the for loop only if the expression matches with the regex, in which case we can expect the `eval()` not to fail, nor to run arbitrary code?

optimum/onnxruntime/modeling_ort.py
fxmarty: I'm not sure I understand why, couldn't find an output named with parenthesis, do you have an example? You could add a comment linking to it, it would be helpful.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -632,21 +635,32 @@ def _output_shape_inference(self, axis_name: Union[str, int], dimensions: Dict[s
...
+            if matched_group is None:
+                raise ValueError(f"Could not perform shape inference on {axis_name}")
+
+            if matched_group == 0:
+                dim = dimensions.get(groups[0], None)
+                if dim is None or not isinstance(dim, int):
+                    return axis_name
+                tokens.append(str(dim))
+            else:
+                tokens.append(groups[matched_group])
```
fxmarty, line 659: this is the kind of code that would benefit from comments for readability

optimum/onnxruntime/modeling_ort.py
michaelbenayoun: Exactly we want to fail if nothing matches to not run arbitrary code.

optimum/onnxruntime/modeling_ort.py
michaelbenayoun: Ok will add some.

michaelbenayoun: Done

optimum/onnxruntime/modeling_ort.py
michaelbenayoun: It's to cover the case where we want to do something like: `(sequence_length + past_length) / 2`. 
Currently we dont have a use for it but it could be needed at some point so added it here.

fxmarty (approved):
LGTM, thank you for the fix, very nice!

Issue #555, opened 2022-12-07
Community contribution -  `optimum.exporters.onnx` support for new models!


michaelbenayoun:
Following what was done by @ChainYo in Transformers, in the [ONNXConfig: Add a configuration for all available models](https://github.com/huggingface/transformers/issues/16308) issue, the idea is to add support for exporting new models in `optimum.exporters.onnx`.

This issue is about the working group specially created for this task. If you are interested in helping out, reply here, take a look at this [organization](https://huggingface.co/OWG), or add  `ChainYo#3610` on discord.

We want to contribute to Hugging Face's ONNX export implementation for all available models on Hugging Face Hub. There are already a lot of architectures implemented for converting PyTorch models to ONNX, but we need more! We need them all!

Feel free to join us in this adventure! Join the org by clicking [here](https://huggingface.co/organizations/OWG/share/TskjfGaGjGnMXXssbPPXrQWEIbosGqZshZ)

Here is a non-exhaustive list of models that all models available:

- [x] Albert
- [x] BART
- [x] BeiT
- [x] BERT
- [ ] BigBird (Critical issue: https://github.com/huggingface/optimum/issues/754#issuecomment-1429830467)
- [ ] BigBirdPegasus (Critical issue: https://github.com/huggingface/optimum/issues/754#issuecomment-1429830467)
- [x] Blenderbot
- [x] BlenderbotSmall
- [ ] BLIP-2
- [x] BLOOM
- [x] CamemBERT
- [ ] CANINE
- [x] CLIP
- [x] CodeGen
- [x] ConvNext
- [x] ConvBert
- [ ] CTRL
- [x] CvT
- [x] Data2VecText
- [x] Data2VecVision
- [x] Deberta
- [x] DebertaV2
- [x] DeiT
- [ ] DecisionTransformer
- [x] DETR
- [x] Distilbert
- [ ] DPR
- [x] DPT
- [x] ELECTRA
- [ ] FNet
- [ ] FSMT
- [x] Flaubert
- [ ] FLAVA
- [ ] Funnel Transformer
- [x] GLPN
- [x] GPT2
- [x] GPTJ
- [x] GPT-Neo
- [x] GPT-NeoX
- [x] Hubert
- [x] I-Bert
- [x] ImageGPT 🛠️ @adit299 
- [ ] LED
- [x] LayoutLM
- [ ] LayoutLMv2 (but 🛠️ in Transformers)
- [x] LayoutLMv3
- [ ] LayoutXLM
[...]

PR #736, opened 2023-02-01
Support Splinter exporters (#555)

Allanbeddouk:
Add Splinter model in exporters.

Contributions related to @michaelbenayoun 's checklist from issue #555


tests/exporters/exporters_utils.py
```diff
@@ -176,6 +176,7 @@
     "roformer": "junnyu/roformer_chinese_base",
     "segformer": "nvidia/segformer-b0-finetuned-ade-512-512",
     "squeezebert": "squeezebert/squeezebert-uncased",
+    "splinter": "hf-internal-testing/tiny-random-SplinterModel",
```
fxmarty, line 179: Can you add it in `PYTORCH_EXPORT_MODELS_TINY` as well? It's actually this dict that is used for testing.

fxmarty (commented):
Thank you for your contribution!

michaelbenayoun (approved):
LGTM once @fxmarty has been addressed and the tests pass.

tests/exporters/exporters_utils.py
Allanbeddouk: Yes you're right, my bad for not adding it there as well. Thank you for your feedback !

fxmarty (approved):
Thank you, the remaining two errors are unrelated, so merging!

Issue #719, opened 2023-01-24
Decoder with cache ONNX export failed after mixed precision training


JingyaHuang:
### System Info

```shell
Optimum dev
onnxruntime 1.13.0
```

### Who can help?

@JingyaHuang 

### Reproduction

Using `evaluate()` after mixed-precision training with `ORTTrainer`. But I think the problem is general for converting PyTorch models to ONNX with FP16 weights.

### Error message

```
Traceback (most recent call last):
  File "test_onnxruntime_train.py", line 279, in test_ort_trainer_decoder
    ort_eval_metrics = trainer.evaluate(inference_with_ort=inference_with_ort)
  File "/workspace/optimum/onnxruntime/trainer.py", line 813, in evaluate
    output = eval_loop(
  File "/workspace/optimum/onnxruntime/trainer.py", line 969, in evaluation_loop_ort
    self._export(onnx_model_path, with_loss=with_loss, device=export_device)
  File "/workspace/optimum/onnxruntime/trainer.py", line 1501, in _export
    _ = export(
  File "/workspace/optimum/exporters/onnx/convert.py", line 607, in export
    return export_pytorch(model, config, opset, output, device=device, input_shapes=input_shapes)
  File "/workspace/optimum/exporters/onnx/convert.py", line 370, in export_pytorch
    onnx_export(
  File "/usr/local/lib/python3.8/dist-packages/torch/onnx/__init__.py", line 350, in export
    return utils.export(
  File "/usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py", line 163, in export
    _export(
  File "/usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py", line 1074, in _export
    graph, params_dict, torch_out = _model_to_graph(
  File "/usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py", line 727, in _model_to_graph
    graph, params, torch_out, module = _create_jit_graph(model, args)
  File "/usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py", line 602, in _create_jit_graph
    graph, torch_out = _trace_and_get_graph_from_model(model, args)
  File "/usr/local/lib/python3.8/dist-packages/torch/onnx/utils.py", line 517, in _trace_and_get_graph_from_model
    trace_graph, torch_out, inputs_states = torch.jit._get_trace_graph(
  File "/usr/local/lib/python3.8/dist-packages/torch/jit/_trace.py", line 1175, in _get_trace_graph
[...]

PR #737, opened 2023-02-01
Patch `ORTTrainer` inference with ONNX Runtime backend

JingyaHuang:
# What does this PR do?

- [x] Enable loss as output in `ORTModelForCausalLM`
- [x] Enable using cache for causal lm task when doing inference in `ORTTrainer`
- [x] Fix the ORT inference of `ORTSeq2SeqTrainer` when `predict_with_generate=True` 

The PR is tested locally, and a PR will be opened for improving the current nightly test.

The caveat #719 always exists, but I won't deal with it because of a lack of bandwidth and need from the community.


optimum/onnxruntime/base.py
```diff
@@ -195,19 +195,19 @@ def forward(
...
         past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None,
+        labels: Optional[torch.LongTensor] = None,
     ) -> CausalLMOutputWithCrossAttentions:
+        known_output_shapes = {}
         # Flatten the past_key_values
         if past_key_values is not None:
             past_key_values = [past_key_value for pkv_per_layer in past_key_values for past_key_value in pkv_per_layer]
 
         if self.device.type == "cuda" and self.parent_model.use_io_binding:
-            past_key_values_shapes = self.compute_past_key_values_output_shapes(
```
fxmarty, line 204: why this variable renaming?

JingyaHuang: Because I then added the shape information of loss inside.

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -774,6 +774,9 @@ def forward(
             encoder_outputs = self.encoder(input_ids=input_ids, attention_mask=attention_mask)
 
         # Decode
+        if decoder_input_ids is None:
+            decoder_input_ids = input_ids
+
```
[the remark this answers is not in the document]
JingyaHuang: Replaced by raising an error as transformers does.

fxmarty (approved):
LGTM, thank you!

Issue #733, opened 2023-01-31
CVE in Protobuf 3.20.1 (CVE-2022-1941)


adriangonz:
### System Info

The `optimum[onnxruntime]` package currently pins the `protobuf` dependency to `3.20.1`:

https://github.com/huggingface/optimum/blob/2dc43439e26843d140562e461ef49693b03850a2/setup.py#L37

However, last year a CVE was published which affects this version of `protobuf` ([CVE-2022-1941](https://www.cve.org/CVERecord?id=CVE-2022-1941)). According to the CVE, `protobuf==3.20.2` should fix this so it would be good to bump the version constrain in Optimum's `setup.py`.

### Who can help?

@JingyaHuang @echarlaix 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

1. Install `optimum[onnxruntime]`
2. Validate that it installs `protobuf==3.20.1`

### Expected behavior

A CVE-free version of `protobuf` gets installed (e.g. `protobuf==3.20.2`)

fxmarty:
Thank you for the report! @mfuntowicz @echarlaix @michaelbenayoun is there any reason we pinned to 3.20.1?

PR #738, opened 2023-02-02
Unpin protobuf 3.20.1

fxmarty:
Fixes https://github.com/huggingface/optimum/issues/733

Feel free to pin in subpackages if necessary.


setup.py
```diff
@@ -34,14 +34,14 @@
         "onnxruntime>=1.9.0",
         "datasets>=1.2.1",
         "evaluate",
-        "protobuf==3.20.1",
+        "protobuf>=3.20.1",
```
regisss, line 37: Shouldn't we put `>=3.20.2` since v3.20.1 is the version with the security issue?

fxmarty: Not sure, maybe other package require 3.20.1.

regisss: Fair enough

regisss (approved):
LGTM!

Issue #740, opened 2023-02-03
A typo in exporters/onnx/model_configs.py


un-certainty:
### System Info

```shell
-
```

### Who can help?

@lewtun @michaelbenayoun 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

`dyanmic_axes` should be `dynamic_axes`

```python
    @property
    def outputs(self) -> Mapping[str, Mapping[int, str]]:
        common_outputs = super().outputs
        if self._behavior is ConfigBehavior.ENCODER:
            for name, dyanmic_axes in common_outputs.items():
                if name == "last_hidden_state":
                    common_outputs[name][1] = f"{common_outputs[name][1]} / 2"
                else:
                    common_outputs[name] = dynamic_axes
        return common_outputs

```

https://github.com/huggingface/optimum/blob/334d3cf007914fee60d401919d5d32ce0edf237b/optimum/exporters/onnx/model_configs.py#L919

### Expected behavior

-

fxmarty:
Thanks for the ping. Quickly testing, it appears in 1.6.3 that the sole output of the encoder is `last_hidden_state`. Why was this control flow needed @mht-sharma @michaelbenayoun ? As for the encoder https://github.com/huggingface/optimum/blob/334d3cf007914fee60d401919d5d32ce0edf237b/optimum/exporters/onnx/base.py#L553-L558 is called, the task is default, and the sole output is `last_hidden_state`?

mht-sharma:
Hi @fxmarty I think this was added because the sequence length in the `last_hidden_state` of the Whisper encoder  is half times that was provided as input to the model. So the dynamic axes is modified accordingly.

fxmarty:
Yes, that's what I was referring to.

PR #741, opened 2023-02-03
Remove dead code in whisper ONNX output

fxmarty:
As per title, fixes https://github.com/huggingface/optimum/issues/740


mht-sharma (approved):
Thanks for the fix. LGTM

PR #742, opened 2023-02-03
Add ORTModelForVision2Seq for VisionEncoderDecoder models inference

mht-sharma:
# What does this PR do?

This PR enables the inference of the `VisionEncoderDecoder` models using ONNXRuntime. The PR adds `ORTModelForVision2Seq` for doing inference similar to `AutoModelForVision2Seq` by changing just a few lines.

# Usage

```diff
>>> from PIL import Image
>>> from transformers import GPT2TokenizerFast, ViTImageProcessor
->>>  from transformers import AutoModelForVision2Seq
+>>>  from optimum.onnxruntime import ORTModelForVision2Seq
>>> import requests

>>> model_name = "nlpconnect/vit-gpt2-image-captioning"
->>> model = AutoModelForVision2Seq.from_pretrained(model_name)
+>>> model = ORTModelForVision2Seq.from_pretrained(model_name, from_transformers=True)
>>> tokenizer = GPT2TokenizerFast.from_pretrained(model_name)
>>> image_processor = ViTImageProcessor.from_pretrained(model_name)

>>> url = "http://images.cocodataset.org/val2017/000000039769.jpg"
>>> image = Image.open(requests.get(url, stream=True).raw)
>>> pixel_values = image_processor(image, return_tensors="pt").pixel_values

>>> generated_ids = model.generate(pixel_values)
>>> generated_text = tokenizer.batch_decode(generated_ids, skip_special_tokens=True)[0]
>>> print(generated_text)
```

## Limitations
- `Donut` model not supported
- `TrOCR` model not supported with `use_cache=True`.
-  IObinding not supported

## Before submitting

### DIFF
--- a/optimum/onnxruntime/base.py
+++ b/optimum/onnxruntime/base.py
@@ -115,10 +115,10 @@ class ORTDecoder(ORTModelPart):
         if len(self.key_value_output_names) == 0:
             self.key_value_output_names = [key for key in self.output_names if "key_values" in key]
 
-        if self.parent_model.use_cache is True or len(self.key_value_output_names) != 0:
-            if len(self.key_value_output_names) == 0:
-                raise RuntimeError("Could not find the past key values in the provided model.")
+        if self.parent_model.use_cache is True and len(self.key_value_output_names) == 0:
+            raise RuntimeError("Could not find the past key values in the provided model.")
 
+        if len(self.key_value_output_names) != 0:
             # Attributes useful when computing the past key/values output shapes.
             self.expected_key_symbolic_shape = None
             self.expected_value_symbolic_shape = None
--- a/optimum/onnxruntime/base.py
+++ b/optimum/onnxruntime/base.py
@@ -114,7 +114,6 @@ class ORTDecoder(ORTModelPart):
             self.key_value_input_names = [key for key in self.input_names if "key_values" in key]
         if len(self.key_value_output_names) == 0:
             self.key_value_output_names = [key for key in self.output_names if "key_values" in key]
-        # from pdb import set_trace; set_trace()
 
         if self.parent_model.use_cache is True or len(self.key_value_output_names) != 0:
             if len(self.key_value_output_names) == 0:
--- a/optimum/onnxruntime/base.py
+++ b/optimum/onnxruntime/base.py
@@ -114,40 +114,42 @@ class ORTDecoder(ORTModelPart):
             self.key_value_input_names = [key for key in self.input_names if "key_values" in key]
         if len(self.key_value_output_names) == 0:
             self.key_value_output_names = [key for key in self.output_names if "key_values" in key]
+        # from pdb import set_trace; set_trace()
 
-        if len(self.key_value_output_names) == 0:
-            raise RuntimeError("Could not find the past key values in the provided model.")
+        if self.parent_model.use_cache is True or len(self.key_value_output_names) != 0:
+            if len(self.key_value_output_names) == 0:
+                raise RuntimeError("Could not find the past key values in the provided model.")
 
-        # Attributes useful when computing the past key/values output shapes.
-        self.expected_key_symbolic_shape = None
-        self.expected_value_symbolic_shape = None
-        for output in self.session.get_outputs():
-            if ".key" in output.name:
-                self.expected_key_symbolic_shape = output.shape
-            elif ".value" in output.name:
-                self.expected_value_symbolic_shape = output.shape
-            # To handle the old case when past_key_values were following the format: past_key_values_{idx}
-            elif "key_values" in output.name:
-                if self.expected_key_symbolic_shape is None:
+            # Attributes useful when computing the past key/values output shapes.
+            self.expected_key_symbolic_shape = None
+            self.expected_value_symbolic_shape = None
+            for output in self.session.get_outputs():
+                if ".key" in output.name:
                     self.expected_key_symbolic_shape = output.shape
-                else:
+                elif ".value" in output.name:
                     self.expected_value_symbolic_shape = output.shape
-            if self.expected_key_symbolic_shape is not None and self.expected_value_symbolic_shape is not None:
-                break
+                # To handle the old case when past_key_values were following the format: past_key_values_{idx}
+                elif "key_values" in output.name:
+                    if self.expected_key_symbolic_shape is None:
+                        self.expected_key_symbolic_shape = output.shape
+                    else:
+                        self.expected_value_symbolic_shape = output.shape
+                if self.expected_key_symbolic_shape is not None and self.expected_value_symbolic_shape is not None:
+                    break
 
-        self.key_sequence_length_idx = -2
-        if (
-            isinstance(self.expected_key_symbolic_shape[-1], str)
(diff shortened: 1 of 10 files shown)


optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -24,18 +24,20 @@
...
 
 import torch
-from transformers import AutoModelForSeq2SeqLM, AutoModelForSpeechSeq2Seq, GenerationConfig
+from transformers import AutoModelForSeq2SeqLM, AutoModelForSpeechSeq2Seq, AutoModelForVision2Seq, GenerationConfig
 from transformers.file_utils import add_start_docstrings_to_model_forward
 from transformers.modeling_outputs import BaseModelOutput, Seq2SeqLMOutput
 
 import onnxruntime as ort
 from huggingface_hub import hf_hub_download
+from onnxruntime import InferenceSession
```
fxmarty, line 33: Can you actually remove `import onnxruntime as ort` and `from onnxruntime import InferenceSession` and just put `import onnxruntime as ort` inside the `TYPE_CHECKING` block? And quote everything in the rest of the code. I think it is used only for type hints.

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -79,6 +81,12 @@
             Mel features extracted from the raw speech waveform. `(batch_size, feature_size, encoder_sequence_length)`.
 """
 
+VISION_ENCODER_INPUTS_DOCSTRING = r"""
+    Args:
+        pixel_values (`torch.FloatTensor`):
+            Featuers extracted from an Image. `(batch_size, num_channels, height, width)`.
```
fxmarty, line 87:
```suggestion
            Features extracted from an image. This tensor should be of shape `(batch_size, num_channels, height, width)`.
```

It's just RGB images no? Not "feature" in the sense of any preprocessing has been applied yet?

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -130,6 +138,22 @@
             `(batch_size, num_heads, encoder_sequence_length, embed_size_per_head)`.
 """
 
+VISION_ENCODER_DECODER_SEQ2SEQ_ONNX_MODEL_DOCSTRING = r"""
+    Args:
+        pixel_values (`torch.FloatTensor`):
+            Featuers extracted from an Image.
```
fxmarty, line 144:
```suggestion
            Features extracted from an image.
```

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -130,6 +138,22 @@
...
+            `(batch_size, num_channels, height, width)`.
```
fxmarty, line 145: same

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -130,6 +138,22 @@
...
+        decoder_input_ids (`torch.LongTensor`):
+            Indices of decoder input sequence tokens in the vocabulary of shape `(batch_size, decoder_sequence_length)`.
+        encoder_outputs (`torch.FloatTensor`):
+            The encoder `last_hidden_state` of shape `(batch_size, encoder_sequence_length, hidden_size)`.
+        past_key_values (`tuple(tuple(torch.FloatTensor), *optional*)`
```
fxmarty, line 150:
```suggestion
        past_key_values (`tuple(tuple(torch.FloatTensor), *optional*)`
```

We should not put `*optional*` in Optimum following this discussion: https://huggingface.slack.com/archives/C02P0559X9S/p1669628754896019

Can rather precise what the default is? Although, for the type: `Tuple[Tuple[torch.FloatTensor]]`

(edit: this is copy paste, so could you edit it as well in the rest?)

tests/onnxruntime/test_modeling.py
```diff
@@ -2742,6 +2747,188 @@ def test_compare_generation_to_io_binding(self, test_name: str, model_arch: str,
...
+    }
+
+    ORTMODEL_CLASS = ORTModelForVision2Seq
+
+    TASK = "vision2seq-lm"
+
+    def exclude_trocr_with_cache(params):
+        if params[0] == "trocr" and params[1] == True:
+            return None
+        return params
```
fxmarty, line 2766: Just for my knowledge, why is this not supported? What will happen if an user tries to use `ORTModelForVision2Seq` with `trocr` with `use_cache = True`?

optimum/utils/normalized_config.py
```diff
@@ -166,7 +173,6 @@ class NormalizedConfigManager:
         'roformer',
         'segformer',
         'squeezebert',
-        'vit',
```
fxmarty, line 169: you can remove `deit` as well above

fxmarty (approved):
LGTM as long as the tests pass, thank you for the addition! Could you also make sure that the code snippet in the documentation is valid?

optimum/onnxruntime/base.py
```diff
@@ -115,39 +115,40 @@ def __init__(
...
-        self.expected_value_symbolic_shape = None
-        for output in self.session.get_outputs():
-            if ".key" in output.name:
-                self.expected_key_symbolic_shape = output.shape
-            elif ".value" in output.name:
-                self.expected_value_symbolic_shape = output.shape
-            # To handle the old case when past_key_values were following the format: past_key_values_{idx}
-            elif "key_values" in output.name:
-                if self.expected_key_symbolic_shape is None:
+        if self.parent_model.use_cache is True or len(self.key_value_output_names) != 0:
```
michaelbenayoun, line 118:
```suggestion
        if len(self.key_value_output_names) != 0:
```
Since the next if raises an error if not?

michaelbenayoun (approved):
LGTM!

What's the status for the pat key / values?

optimum/onnxruntime/modeling_seq2seq.py
mht-sharma: I think the normalization and resize would be applied to the image

fxmarty: right!

tests/onnxruntime/test_modeling.py
mht-sharma: The model currently not output any `past_key_values` when `use_cache=True`.  So need to figure out why. Probably something wrong in the modeling code.

```
| What will happen if an user tries to use ORTModelForVision2Seq with trocr with use_cache = True?
```

For this I have added an error message during export.

optimum/onnxruntime/modeling_seq2seq.py
mht-sharma: Added default to None as in the function sig

mht-sharma:
> LGTM!
> 
> What's the status for the pat key / values?

Past key values is not supported only for TrOCR, would look into this after Donut

optimum/onnxruntime/base.py
mht-sharma: This I added for `TrOCR` because when use_cache = False, currently it does not output the `past_key_values`. So this will skip this and user can do inference with use_cache = False

PR #747, opened 2023-02-06
Replace `--for-ort` option in ONNX export by `--monolith`

fxmarty:
As per title.

Breaking change: the old `--for-ort` behavior is now the default.

Additionally, the tests in `test_onnx_export.py` are now all slow.


michaelbenayoun (approved):
LGTM, thanks for taking care of this!

PR #748, opened 2023-02-06
Make `-with-past` the default in the ONNX export

fxmarty:
As per title.

Breaking change: with-past is now the default.


optimum/exporters/onnx/__main__.py
```diff
@@ -68,6 +68,17 @@ def main():
         task, args.model, framework=args.framework, cache_dir=args.cache_dir, trust_remote_code=args.trust_remote_code
     )
 
+    # make -with-past the default if --task was not explicitely specified
```
michaelbenayoun, line 71:
```suggestion
    # Make -with-past the default if --task was not explicitely specified
```

michaelbenayoun (commented):
Thankls for taking care of that!

I think we should also log a warning when a user could use past and don't (because he might not be aware of it). WDTY?

fxmarty:
> I think we should also log a warning when a user could use past and don't (because he might not be aware of it). WDTY?

I don't think it is necessary as it will be the case only if the user explicitly pass `--task`

docs/source/exporters/onnx/usage_guides/export_a_model.mdx
```diff
@@ -198,7 +198,7 @@ In the ONNX export, the past keys/values are reused by default. This behavior co
 A model exported using past key/values can be reused directly into Optimum's [`~onnxruntime.ORTModel`]:
 
 ```bash
-optimum-cli export onnx --model gpt2 --for-ort --task causal-lm-with-past gpt2_onnx/
+optimum-cli export onnx --model gpt2 --for-ort gpt2_onnx/
```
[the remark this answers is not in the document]
fxmarty: I think so, yes

michaelbenayoun:
Which is mandatory to do for for local models. I think having this warning will let the user know that it exists and it should be used whenever possible.

fxmarty:
Makes sense!

optimum/exporters/onnx/__main__.py
```diff
@@ -74,6 +74,20 @@ def main():
         task, args.model, framework=args.framework, cache_dir=args.cache_dir, trust_remote_code=args.trust_remote_code
     )
 
+    if task + "-with-past" in TasksManager.get_supported_tasks_for_model_type(
+        model.config.model_type.replace("_", "-"), "onnx"
+    ):
+        if args.task == "auto":  # Make -with-past the default if --task was not explicitely specified
+            task = task + "-with-past"
+        elif args.task in ["seq2seq-lm", "causal-lm", "speech2seq-lm", "vision2seq-lm"]:
```
michaelbenayoun, line 82: I think we can infer task that can have a `with-past` flavor dynamically here.

michaelbenayoun (approved):
LGTM once my comment as been discussed / addressed!

optimum/exporters/onnx/__main__.py
fxmarty: fixed thanks

michaelbenayoun (approved):

PR #749, opened 2023-02-06
Support ONNX export on `torch.float16` type

fxmarty:
As per title.

Test still missing. Partly fixes https://discuss.huggingface.co/t/convert-gpt-j-to-fp-16-onnx/30294


optimum/exporters/onnx/utils.py
```diff
@@ -199,3 +199,28 @@ def recursive_to_device(value: Union[Tuple, List, "torch.Tensor"], device: str):
...
+        for i, val in enumerate(value):
+            value[i] = recursive_to_dtype(val, dtype)
+    elif isinstance(value, torch.Tensor):
+        value = value.to(dtype=dtype)
+
+    return value
+
+
+str_dtype_to_torch_dtype = {
+    None: None,
```
JingyaHuang, line 223: None is not a str, I would prefer to alter the name.

optimum/exporters/onnx/utils.py
```diff
@@ -199,3 +199,28 @@ def recursive_to_device(value: Union[Tuple, List, "torch.Tensor"], device: str):
         value = value.to(device)
 
     return value
+
+
+def recursive_to_dtype(value: Union[Tuple, List, "torch.Tensor"], dtype: Optional[torch.dtype]):
+    if dtype is None:
+        return value
+
+    if isinstance(value, tuple):
```
JingyaHuang, line 208: Why do you distinguish between tuple and list, the difference between them and the last case are to check if `value` is iterable no?

JingyaHuang (commented):
Just add two nits, btw is the torch onnx export on fp16 stable?

fxmarty:
> is the torch onnx export on fp16 stable

I've tried on a single model - and did not try to load into an InferenceSession. I will add a test for it, thanks.

optimum/exporters/onnx/utils.py
fxmarty: yes

optimum/exporters/onnx/utils.py
fxmarty: Tuple does not support item assignment.

optimum/exporters/onnx/__main__.py
```diff
@@ -65,13 +66,27 @@ def main():
                 f"The task could not be automatically inferred. Please provide the argument --task with the task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"
             )
 
+    if args.framework == "tf" and args.dtype is not None:
+        raise ValueError("The --dtype option is supported only for PyTorch.")
+
+    if args.dtype is not None and args.device == "cpu":
+        raise ValueError(
+            "The --dtype option is supported on when exporting on GPU. Please pass the option --device cuda."
```
regisss, line 74:
```suggestion
            "The --dtype option is supported only when exporting on GPU. Please pass the option `--device cuda`."
```

optimum/exporters/onnx/convert.py
```diff
@@ -232,7 +238,8 @@ def validate_model_outputs(
         reference_model.to(device)
 
         for key, value in reference_model_inputs.items():
-            reference_model_inputs[key] = recursive_to_device(value=value, device=device)
+            reference_model_inputs[key] = recursive_to_dtype(value=value, dtype=dtype)
+            reference_model_inputs[key] = recursive_to_device(value=reference_model_inputs[key], device=device)
```
regisss, line 242: Nit: since both `recursive_to_dtype` and `recursive_to_device` ultimately rely on `tensor.Tensor.to`, we could actually have a unified `recursive_to_device_dtype` method.

optimum/exporters/onnx/convert.py
```diff
@@ -645,6 +665,3 @@ def export(
         raise RuntimeError(
             "You either provided a PyTorch model with only TensorFlow installed, or a TensorFlow model with only PyTorch installed."
         )
-
-    config.fix_dynamic_axes(output)
```
regisss, line 649: Why removing this? It doesn't work?

fxmarty: It is a mistake, thanks for spotting!

regisss: Okay, so we need to keep the definition of `DynamicAxisNameError` above and `return export_output` here also.

optimum/exporters/onnx/convert.py
michaelbenayoun: `recursive_to` applying to to a nested structure.
It could even be as simple as that:

```python
from torch.utils._pytree import tree_map

def recursive_to(possibly_nested_structure, *args, **kwargs):
     def map_fn(value):
          if isinstance(value, tensor):
              return value.to(*args, **kwargs)
         return value
     return tree_map(map_fn, possibly_nested_structure)
```

This function could be apply to nested structures of tensors.

fxmarty: Thanks for the suggestion, but I'm not super keen on using undocumented private method from pytorch.

optimum/commands/export/onnx.py
```diff
@@ -52,6 +52,13 @@ def parse_args_onnx(parser):
         default="cpu",
         help='The device to use to do the export. Defaults to "cpu".',
     )
+    optional_group.add_argument(
+        "--dtype",
+        type=str,
+        default=None,
+        choices=["float32", "float16", None],
+        help="Experimental option: the dtype of the weights to use during the export. If None, the default dtype will be used. PyTorch-only.",
+    )
```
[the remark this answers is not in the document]
fxmarty: Thanks

regisss (approved):
LGTM!

optimum/commands/export/onnx.py
```diff
@@ -54,6 +54,11 @@ def parse_args_onnx(parser):
         default="cpu",
         help='The device to use to do the export. Defaults to "cpu".',
     )
+    optional_group.add_argument(
+        "--fp16",
+        action="store_true",
+        help="Experimental option: use half precision during the export. PyTorch-only, requires `--device cuda`.",
```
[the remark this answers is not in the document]
fxmarty: I'd say experimental because I haven't thouroughly tested it with ONNX Runtime + CUDAExecutionProvider / TensorrtExecutionProvider, and neither with native TensorRT (though in the validation itself we call InferenceSession on CUDA EP, so it's a good sign it's fine). But the export itself is thoroughly tested.

PR #750, opened 2023-02-06
Raise error on double call to `BetterTransformer.transform()`

fxmarty:
As requested in https://github.com/huggingface/transformers/pull/21259#discussion_r1084146743


younesbelkada (approved):
thanks a lot for this! 💯

PR #751, opened 2023-02-06
Add GPU tests for BetterTransformer

fxmarty:
As requested in https://github.com/huggingface/transformers/pull/21259#issuecomment-1418938913

### DIFF
--- a/tests/bettertransformer/Dockerfile_bettertransformer_gpu
+++ b/tests/bettertransformer/Dockerfile_bettertransformer_gpu
@@ -22,4 +22,5 @@ COPY . /workspace/optimum
 RUN pip install /workspace/optimum[tests]
 
 ENV RUN_SLOW=1
-CMD pytest bettertransformer/test_*.py -s --durations=0
+WORKDIR /workspace/optimum/tests/
+CMD pytest bettertransformer/test_*.py -s --durations=0 -m gpu_test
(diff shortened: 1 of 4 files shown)


tests/bettertransformer/Dockerfile_bettertransformer_gpu
```diff
@@ -0,0 +1,25 @@
...
+RUN apt-get install -y software-properties-common wget apt-utils patchelf git libprotobuf-dev protobuf-compiler cmake \
+    bzip2 ca-certificates libglib2.0-0 libxext6 libsm6 libxrender1 mercurial subversion libopenmpi-dev && \
+    apt-get clean
+RUN unattended-upgrade
+RUN apt-get autoremove -y
+
+RUN python -m pip install -U pip
+
+RUN pip install torch torchvision torchaudio
+RUN pip install transformers accelerate
```
younesbelkada, line 18:
```suggestion
RUN pip install transformers accelerate
```
Maybe we should install `accelerate` from source? No strong opinion but when doing this in `transformers` it helps avoiding to have some issues before `accelerate` releases. Wdyt?

younesbelkada (approved):
Thanks a lot for taking care of this! 💯 
Left a minor comment, otherwise LGTM

tests/bettertransformer/Dockerfile_bettertransformer_gpu
younesbelkada: SGTM !

PR #760, opened 2023-02-08
Update quality tooling for formatting

regisss:
# What does this PR do?

Following the recent changes in Transformers regarding quality tools (see https://github.com/huggingface/transformers/pull/21480 and https://github.com/huggingface/transformers/pull/21493), the package `flake8` is replaced by `ruff`. Files were formatted and corrected accordingly.

## Before submitting

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,11 +23,11 @@ REAL_CLONE_URL = $(if $(CLONE_URL),$(CLONE_URL),$(DEFAULT_CLONE_URL))
 # Run code quality checks
 style_check:
 	black --check .
-	isort --check .
+	ruff .
 
 style:
 	black .
-	isort .
+	ruff . --fix
 
 # Run tests for the library
 test:
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -24,33 +24,29 @@ import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Type, Union
 
-from tqdm.auto import tqdm
-
-
 # Integrations must be imported before ML frameworks:
-from transformers.integrations import (  # isort: split
-    hp_params,
-    is_fairscale_available,
-)
-
 import numpy as np
 import torch
 import torch.distributed as dist
 from torch import nn
 from torch.utils.data import DataLoader, Dataset, RandomSampler
 from torch.utils.data.distributed import DistributedSampler
-from transformers import PreTrainedModel, __version__
+from tqdm.auto import tqdm
 from transformers.data.data_collator import DataCollator
 from transformers.debug_utils import DebugOption, DebugUnderflowOverflow
 from transformers.deepspeed import deepspeed_init, is_deepspeed_zero3_enabled
 from transformers.dependency_versions_check import dep_version_check
 from transformers.file_utils import (
-    CONFIG_NAME,
-    WEIGHTS_NAME,
     is_apex_available,
     is_sagemaker_dp_enabled,
     is_sagemaker_mp_enabled,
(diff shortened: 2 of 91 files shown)


regisss:
@JingyaHuang 
- In `optimum/onnxruntime/trainer.py`, import fairscale is removed because unused. Is that okay?
- In `optimum/onnxruntime/trainer_seq2seq.py` , `autocast` seems unused. Can we remove the conditional statement where it is imported?

fxmarty (approved):
Awesome thank you!

echarlaix (approved):
Thanks for taking care of this @regisss !

mht-sharma (approved):
LGTM! Thanks for the PR

michaelbenayoun (approved):
LGTM!

Is all the reformating done by ruff? It seems very nice!

optimum/onnxruntime/trainer.py
```diff
@@ -24,29 +24,28 @@
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Type, Union
 
-from tqdm.auto import tqdm
-
 
 # Integrations must be imported before ML frameworks:
-from transformers.integrations import (  # isort: split
+# isort: off
```
JingyaHuang, line 29: Do we still need this after replacing isort by ruff?

optimum/onnxruntime/trainer_seq2seq.py
```diff
@@ -51,7 +51,7 @@
 
 
 if version.parse(torch.__version__) >= version.parse("1.8"):
-    from torch.cuda.amp import autocast
+    pass
```
JingyaHuang, line 54: Yeah let's just remove all including in the conditional.

Makefile
```diff
@@ -23,11 +23,11 @@ REAL_CLONE_URL = $(if $(CLONE_URL),$(CLONE_URL),$(DEFAULT_CLONE_URL))
 # Run code quality checks
 style_check:
 	black --check .
-	isort --check .
```
JingyaHuang, line 26: Why do we remove isort? Is ruff ruff a replacement for isort? But why do we still leave isort in the setup?

setup.py
```diff
@@ -25,7 +25,7 @@
 
 TESTS_REQUIRE = ["pytest", "requests", "parameterized", "pytest-xdist", "Pillow", "sacremoses", "diffusers"]
 
-QUALITY_REQUIRE = ["black~=22.0", "flake8>=3.8.3", "isort>=5.5.4"]
+QUALITY_REQUIRE = ["black~=23.1", "ruff>=0.0.241", "isort>=5.5.4"]
```
JingyaHuang, line 28: Do we still need isort?

JingyaHuang (approved):
That's great, thanks for updating the styling tools @regisss! Just left some questions to better understand the change. 💯

optimum/onnxruntime/trainer.py
regisss: Yes, ruff uses the same action comments as isort.

Makefile
regisss: Yes, ruff also manages imports the same way as isort does but it does it much faster.
You're right, I just removed isort from the dependencies!

setup.py
regisss: Nope, thanks for catching this!

regisss:
> LGTM!
> 
> Is all the reformating done by ruff? It seems very nice!

@michaelbenayoun No, the formatting is still done by black. Ruff is a linter that can also do the same job as isort, so it will reorder imports, remove unused variables etc... It's more a replacement for the flake8 package, which we have never used in Optimum. And it's fast!

PR #762, opened 2023-02-08
Fix stable diffusion ONNX export

echarlaix:


optimum/exporters/onnx/__main__.py
```diff
@@ -74,20 +75,17 @@ def main():
         task, args.model, framework=args.framework, cache_dir=args.cache_dir, trust_remote_code=args.trust_remote_code
     )
 
-    if task + "-with-past" in TasksManager.get_supported_tasks_for_model_type(
+    if task != "stable-diffusion" and task + "-with-past" in TasksManager.get_supported_tasks_for_model_type(
         model.config.model_type.replace("_", "-"), "onnx"
     ):
-        if args.task == "auto":  # Make -with-past the default if --task was not explicitely specified
+        if task == "auto":  # Make -with-past the default if --task was not explicitely specified
```
michaelbenayoun, line 81: It will never happen no?
I think what you want here is `args.task`.

optimum/exporters/onnx/base.py
```diff
@@ -378,6 +377,17 @@ def torch_to_onnx_input_map(self) -> Mapping[str, str]:
...
+    @property
+    def torch_to_onnx_output_map(self) -> Mapping[str, str]:
+        """
+        Dictionary of keys to update the ONNX output name for export. Override the function when
+        the output names and the exported ONNX output names need are different.
+
+        Returns:
+            `Mapping[str, str]`: A dictionary specifying the output name to exported ONNX output name map.
+        """
+        return {}
```
michaelbenayoun, line 389: I think you do not need that, you can use the `output_names_for_validation` method.

optimum/exporters/onnx/__main__.py
echarlaix: Good catch !

optimum/exporters/onnx/base.py
echarlaix: It would be less efficient as the `output_names_for_validation` method uses `torch_to_onnx_output_map`

optimum/exporters/onnx/__main__.py
```diff
@@ -74,20 +75,17 @@ def main():
...
         model.config.model_type.replace("_", "-"), "onnx"
     ):
         if args.task == "auto":  # Make -with-past the default if --task was not explicitely specified
             task = task + "-with-past"
         else:
             logger.info(
-                f"The task `{args.task}` was manually specified, and past key values will not be reused in the decoding."
-                f"Please pass `--task {args.task}-with-past` to export using the past key values."
+                f"The task `{task}` was manually specified, and past key values will not be reused in the decoding."
+                f"Please pass `--task {task}-with-past` to export using the past key values."
```
michaelbenayoun, line 86: Same comment, why did it become `task`?

optimum/exporters/onnx/base.py
michaelbenayoun: How is it less efficient? The `output_names_for_validation` method exists and does what you need.

I am not in favor of adding yet a new method. `OnnxConfig` is already packed with methods, and in this specific case there is no real use of this one since you can use what alredy exists.

echarlaix: The `output_names_for_validation` should use a dictionnary to map the torch and onnx model output names in order to be more general instead of returning a list with a fixed order of output names. Having a dictionnary to do the mapping  is needed for both the `ModelPatcher` and the validation step (`validate_model_outputs`), so I would rather remove the `output_names_for_validation` method and use the `torch_to_onnx_output_map` in `validate_model_outputs` directly

optimum/exporters/onnx/__main__.py
echarlaix: `task` and `args.task` are equal in this case, I think it's clearer to only use `args.task` to verify whether the task detection was done automatically and to use `task` for the rest. I don't have a strong opinion on this so I can change it back if needed

optimum/exporters/onnx/base.py
michaelbenayoun: I don't have a strong opinion on which one we should keep, but we should have only one of the two as they are doing mostly the same thing.

optimum/exporters/onnx/__main__.py
fxmarty: sounds good

optimum/exporters/onnx/__main__.py
```diff
@@ -58,6 +58,7 @@ def main():
     # Infer the task
     task = args.task
     if task == "auto":
+        logger.info(f"Automatic task detection to {task}.")
```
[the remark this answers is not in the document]
fxmarty: Yes @echarlaix , this will just print `Automatic task detection to auto`?

I put it below because there is later a `task = task + "-with-past"`.

optimum/exporters/onnx/__main__.py
```diff
@@ -74,20 +75,17 @@ def main():
...
             logger.info(
-                f"The task `{args.task}` was manually specified, and past key values will not be reused in the decoding."
-                f"Please pass `--task {args.task}-with-past` to export using the past key values."
+                f"The task `{task}` was manually specified, and past key values will not be reused in the decoding."
+                f"Please pass `--task {task}-with-past` to export using the past key values."
             )
 
-    if args.task == "auto":
-        logger.info(f"Automatic task detection to {task}.")
-
```
fxmarty, line 90: I think this should be left as is, since it takes into account the `task = task + "-with-past"`

optimum/exporters/onnx/__main__.py
fxmarty, line 61:
```suggestion
```

optimum/exporters/onnx/__main__.py
```diff
@@ -74,20 +75,17 @@ def main():
         task, args.model, framework=args.framework, cache_dir=args.cache_dir, trust_remote_code=args.trust_remote_code
     )
 
-    if task + "-with-past" in TasksManager.get_supported_tasks_for_model_type(
+    if task != "stable-diffusion" and task + "-with-past" in TasksManager.get_supported_tasks_for_model_type(
```
fxmarty, line 78: I think this is not needed, as there is no `stable-diffusion-with-past` anyway.

optimum/exporters/onnx/__main__.py
echarlaix: I see, will put it back !

fxmarty (approved):
LGTM thanks a lot for the fix, if you can just address the comment left. @michaelbenayoun is it fine to merge? I did not look thoroughly at the other thread.

optimum/exporters/onnx/__main__.py
fxmarty: Thanks

optimum/exporters/onnx/base.py
```diff
@@ -379,6 +378,17 @@ def torch_to_onnx_input_map(self) -> Mapping[str, str]:
         """
         return {}
 
+    @property
+    def torch_to_onnx_output_map(self) -> Mapping[str, str]:
+        """
+        Dictionary of keys to update the ONNX output name for export. Override the function when
```
michaelbenayoun, line 384:
```suggestion
        Dictionary mapping output names from the PyTorch model to output names from the exported ONNX model.  Override the function when
```

optimum/exporters/onnx/base.py
```diff
@@ -379,6 +378,17 @@ def torch_to_onnx_input_map(self) -> Mapping[str, str]:
...
+        the output names and the exported ONNX output names need are different.
+
+        Returns:
+            `Mapping[str, str]`: A dictionary specifying the output name to exported ONNX output name map.
```
michaelbenayoun, line 388:
```suggestion
            `Mapping[str, str]`: A dictionary mapping the PyTorch model output names to the exported ONNX model output names.
```

michaelbenayoun (approved):
Left a few nits about the docstring, otherwise LGTM, thanks for the fix!

PR #765, opened 2023-02-09
Fix flaky ONNX Runtime generation test with past key value reuse

fxmarty:
As per title.


echarlaix (approved):
LGTM !

regisss (commented):
Not sure 1.1 is enough, there are tests with speedups below 1.1.
There are even tests with speedups below 1, which is weird.

fxmarty:
Yes, locally I had no issue. Will have a look.

PR #768, opened 2023-02-10
Change type annotation

michaelbenayoun:
# What does this PR do?

This PR changes the type annotation for dictionaries in `exporters.onnx` from [`Mapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) to `Dict`.


fxmarty (approved):
Thanks!

PR #772, opened 2023-02-13
Fix FusionOptions following ORT 1.14 release

fxmarty:
As per title, refer to:

https://github.com/microsoft/onnxruntime/blob/v1.13.1/onnxruntime/python/tools/transformers/fusion_options.py
https://github.com/microsoft/onnxruntime/blob/v1.14.0/onnxruntime/python/tools/transformers/fusion_options.py

### DIFF
--- a/optimum/onnxruntime/configuration.py
+++ b/optimum/onnxruntime/configuration.py
@@ -655,6 +655,19 @@ class OptimizationConfig:
         disable_shape_inference (`bool`, defaults to `False`):
             Whether to disable symbolic shape inference.
             The default value is set to `False` but symbolic shape inference might cause issues sometimes.
+        use_multi_head_attention (`bool`, defaults to `False`):
+            Experimental argument. Use MultiHeadAttention instead of Attention operator, which has merged weights for Q/K/V projection,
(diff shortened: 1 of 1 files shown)


optimum/onnxruntime/configuration.py
```diff
@@ -692,6 +705,13 @@ class OptimizationConfig:
     disable_embed_layer_norm: bool = True
     disable_shape_inference: bool = False
 
+    # ONNX Runtime 1.14.0 arguments
+    use_multi_head_attention = False
+    enable_gemm_fast_gelu = False
+    use_raw_attention_mask = False
```
michaelbenayoun, line 711: So it's different than `use_mask_index` right?

michaelbenayoun (approved):
LGTM

optimum/onnxruntime/configuration.py
fxmarty: Feels a bit redundant to me: https://github.com/microsoft/onnxruntime/blob/6ccaeddefa65ccac402a47fa4d9cad8229794bb2/onnxruntime/python/tools/transformers/fusion_options.py#L92-L95

PR #779, opened 2023-02-14
Improve orttrainer test

JingyaHuang:
# What does this PR do?

- [ ] Improve `ORTTrainer` test.
- [ ] Improve `ORTTrainer` test triggering.


JingyaHuang:
Current constraint, we won't be abled to launch GPU test and ORTTrainer test at the same time, as the workflow will

1. launch both GPU and ORTTrainer tests when adding a label.
2. check the name if not correct skip it immediately

This means that when adding a new label if the previous running test triggered by label is not finished, it will be relaunched and skipped.

As in our use case, we rarely see cases in which we need to check both tests. And even if we do, we can wait for one to finish before launching the other. So I guess it would be fine.

JingyaHuang:
Need to update the CI with onnxruntime 1.14.0 as older numpy has been unpined

```
AttributeError: module 'numpy' has no attribute 'bool'.
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```
A PR will be opened to update ORT training dockerfile.

optimum/exporters/onnx/convert.py
```diff
@@ -540,6 +540,7 @@ def export_models(
     output_names: Optional[List[str]] = None,
     device: str = "cpu",
     input_shapes: Optional[Dict] = None,
+    disable_dynamic_axes_fix: Optional[bool] = False,
```
fxmarty, line 543: Why is this needed (could you precise in the docstring)?

optimum/onnxruntime/trainer.py
```diff
@@ -1500,6 +1514,7 @@ def _export(
                 output_dir=model_path,
                 output_names=output_names,
                 device=device,
+                disable_dynamic_axes_fix=True,  # onnxruntime floating point exception (core dumped)
```
fxmarty, line 1517: Do you know what there is this exception? Can't it be avoided?

optimum/exporters/onnx/base.py
```diff
@@ -803,7 +803,9 @@ class OnnxConfigWithLoss(OnnxConfig, ABC):
         "default": {"labels": {0: "batch_size"}},
         "masked-lm": {"labels": {0: "batch_size", 1: "sequence_length"}},
         "causal-lm": {"labels": {0: "batch_size", 1: "sequence_length"}},
+        "causal-lm-with-past": {"labels": {0: "batch_size"}},
```
fxmarty, line 806: It could be kind of difficult to handle all those cases in a single `OnnxConfigWithLoss` no? Is there a reason for the asymmetry compared to `OnnxConfigWithLoss`?

I am not sure to understand why `OnnxConfigWithLoss` is needed in the first place. It appears to just add labels / loss in the inputs/outputs.

fxmarty (approved):
Thank you! LGTM as long as the tests pass, and `test_trainer.py` passes as well

optimum/exporters/onnx/convert.py
JingyaHuang: Correcting the dynamic axes gave a core dumped error when running onnxruntime.InferenceSession. And it is not necessary for Trainer, so I added the option to not correct it.

optimum/exporters/onnx/base.py
JingyaHuang: As I explained.

optimum/exporters/onnx/convert.py
JingyaHuang: It happens when export a ORT trained model with task `causal-lm-with-past` and only for the with-past decoder.
It tried automatically to fix the dynamic axes name of `logits dim=1`, and error occurs in this line:
https://github.com/huggingface/optimum/blob/4100c384127995de772a7a5b4a22df3b27e27e73/optimum/exporters/onnx/base.py#L335
I doubted it was the past-key-values as it only appeared for with past decoder and I checked,  but the dummy pkv look correct to me. Search a little bit, seems that this type of error could appear when reload model under different device / opset etc... I wonder if it is due to the fact that ORT trained model will have some Microsoft domain op while exporting to ONNX and `SessionOptions` need to precisely configured, but I did not dig further as it is special case and not impacting the ORTTrainer anyway.

optimum/onnxruntime/trainer.py
JingyaHuang: It happens when export a ORT trained model with task `causal-lm-with-past` and only for the with-past decoder.
It tried automatically to fix the dynamic axes name of `logits dim=1`, and error occurs in this line:
https://github.com/huggingface/optimum/blob/4100c384127995de772a7a5b4a22df3b27e27e73/optimum/exporters/onnx/base.py#L335
I doubted it was the past-key-values as it only appeared for with past decoder and I checked,  but the dummy pkv look correct to me. Search a little bit, seems that this type of error could appear when reload model under different device / opset etc... I wonder if it is due to the fact that ORT trained model will have some Microsoft domain op while exporting to ONNX and `SessionOptions` need to precisely configured, but I did not dig further as it is special case and not impacting the ORTTrainer anyway.

PR #764, opened 2023-02-09
Fix wrong shapes used at ONNX export and validation

fxmarty:
**Breaking change:** The export with `--monolith` along with `--task ...-with-past` is no longer supported. It made no sense anyway.

As per title, for GPT2 for causallm, the right shapes should be along the lines of:

```
-----
input_ids: torch.Size([1, 13])
attention_mask torch.Size([1, 13])
encoder_hidden_states None
encoder_attention_mask None
-----
input_ids: torch.Size([1, 1])
past_key_values
    torch.Size([1, 4, 13, 8])
    torch.Size([1, 4, 13, 8])
    torch.Size([1, 4, 13, 8])
    torch.Size([1, 4, 13, 8])
    torch.Size([1, 4, 13, 8])
    torch.Size([1, 4, 13, 8])
    torch.Size([1, 4, 13, 8])
    torch.Size([1, 4, 13, 8])
    torch.Size([1, 4, 13, 8])
    torch.Size([1, 4, 13, 8])
attention_mask torch.Size([1, 14])
encoder_hidden_states None
encoder_attention_mask None
-----
input_ids: torch.Size([1, 1])
past_key_values
    torch.Size([1, 4, 14, 8])
    torch.Size([1, 4, 14, 8])
    torch.Size([1, 4, 14, 8])
    torch.Size([1, 4, 14, 8])
    torch.Size([1, 4, 14, 8])
    torch.Size([1, 4, 14, 8])
    torch.Size([1, 4, 14, 8])
    torch.Size([1, 4, 14, 8])
    torch.Size([1, 4, 14, 8])
    torch.Size([1, 4, 14, 8])
attention_mask torch.Size([1, 15])
encoder_hidden_states None
encoder_attention_mask None
```

For T5 for seq2seq, shapes should be along the lines of:
```
-----
input_ids: None
decoder_input_ids: torch.Size([1, 1])
attention_mask torch.Size([1, 30])
decoder_attention_mask None
encoder_outputs None
-----
input_ids: None
decoder_input_ids: torch.Size([1, 1])
past_key_values
    torch.Size([1, 4, 1, 8])
    torch.Size([1, 4, 1, 8])
    torch.Size([1, 4, 1, 8])
    torch.Size([1, 4, 1, 8])
    torch.Size([1, 4, 1, 8])
    torch.Size([1, 4, 1, 8])
    torch.Size([1, 4, 1, 8])
    torch.Size([1, 4, 1, 8])
    torch.Size([1, 4, 1, 8])
[...]

PR #780, opened 2023-02-14
Fix output shape dimension for OnnxConfigWithPast

fxmarty:
Following https://github.com/huggingface/optimum/pull/764/files#diff-c8b74c7c590ff56a6ecd02908c5f4181afc53fa9bf513e0ac4c7b7c65fecc200L57 , the GPU tests started failing: https://github.com/huggingface/optimum/actions/runs/4139994852/jobs/7158229313

This is because when using CUDAExecutionProvider + IO Binding, the input dynamic axis name `sequence_length` was previously used to infer the axis `sequence_length` of logits that was wrongly registered as having a dynamic length in the case of the with-past model. This PR fixes the issue.

The tests
```
pytest tests/onnxruntime/test_modeling.py -k "test_pipeline_on_gpu and ORTModelForCausalLMIntegrationTest" -s
pytest tests/onnxruntime/test_modeling.py -k "test_compare_generation_to_io_binding and ORTModelForSeq2SeqLMIntegrationTest" -s
```

now go well.

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -759,6 +759,7 @@ class ORTModel(OptimizedModel):
                 output_shape = []
                 for axis_name in output_node.shape:
                     output_shape.append(self._output_shape_inference(axis_name, dimensions))
+
             output_buffer = self._prepare_output_buffer(model, output_shape, output_name)
             io_binding.bind_output(
                 output_name,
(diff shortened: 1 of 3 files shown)


optimum/onnxruntime/modeling_ort.py
```diff
@@ -759,6 +759,7 @@ def _prepare_io_binding(
                 output_shape = []
                 for axis_name in output_node.shape:
                     output_shape.append(self._output_shape_inference(axis_name, dimensions))
+
```
fxmarty, line 762:
```suggestion
```

PR #781, opened 2023-02-15
Pin numpy only for tensorflow export

fxmarty:
As per title, as `tf2onnx` has not updated yet.


regisss (approved):
LGTM!
Some tests didn't pass though, not sure if it is related to this PR.

fxmarty:
Thank you - one was, the others are https://github.com/microsoft/onnxruntime/issues/14663

PR #787, opened 2023-02-15
Update ORT Training to 1.14.0

JingyaHuang:
# What does this PR do?

- [x] Create Dockerfile for onnxruntime-training 1.14.0
- [x] Update trainer documentation
- [x] Complete the fix in #737 (I only did the patch for `evaluation_loop_ort()` but not for `prediction_loop_ort`, but it is the same content)

### DIFF
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -956,6 +956,7 @@ class ORTTrainer(Trainer):
         # Check if there are labels in the dataset
         dummy_inputs = next(iter(dataloader))
         has_labels = all(dummy_inputs.get(k) is not None for k in self.label_names)
+        use_cache = ORTFeaturesManager.do_use_cache(self.feature)
 
         if self.onnx_model_path and (has_labels == self.exported_with_loss):
             logger.info("[INFO] Inference with given ONNX model")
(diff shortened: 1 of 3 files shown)


optimum/onnxruntime/trainer.py
```diff
@@ -1210,14 +1211,10 @@ def prediction_loop_ort(
...
         args = self.args
-        # Temporary fix for decoder, now `use_cache` set to False which cause overhead on re-computing past keys and values.
-        # TODO: Use cache once `ORTModelForCausalLM` supports `loss` as output
         if ort_model_cls is ORTModelForCausalLM:
-            ort_model = ort_model_cls.from_pretrained(
-                model_id=model_id, use_cache=False, provider="CUDAExecutionProvider"
-            )
+            ort_model = ort_model_cls.from_pretrained(model_id=model_id, use_cache=use_cache).to(args.device)
         else:
-            ort_model = ort_model_cls.from_pretrained(model_id=model_id, provider="CUDAExecutionProvider")
```
fxmarty, line 1220: Why this change only in the `else` branch?

fxmarty (approved):
LGTM, thank you!

optimum/onnxruntime/trainer.py
JingyaHuang: you mean `use_cache` or the ep?
EP are changed for both branch, `use_cache` is only needed for decoder cases(if branch) not other cases(else branch).

fxmarty: I meant `provider="CUDAExecutionProvider"` to `.device(args.device)`

JingyaHuang: ? From what I see both branches are changes from precising cuda ep to passing to args.device.

https://github.com/huggingface/optimum/blob/e537404799b478cd890d298ad01516db749efcc6/optimum/onnxruntime/trainer.py#L1214-L1217

PR #794, opened 2023-02-20
ORTOptimizer support ORTModelForCausalLM

fxmarty:
As per title.

Notes:
* `longt5` results in segmentation fault with O4.
* gpt2 / gptj / etc with cache and O4 (fp16) results in:
```
onnxruntime.capi.onnxruntime_pybind11_state.RuntimeException: [ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Non-zero status code returned while running Concat node. Name:'/transformer/h.0/attn/Concat_3' Status Message: /onnxruntime_src/onnxruntime/core/framework/op_kernel.cc:83 virtual OrtValue* onnxruntime::OpKernelContext::OutputMLValue(int, const onnxruntime::TensorShape&) status.IsOK() was false. Shape mismatch attempting to re-use buffer. {1,4,11,8} != {1,4,12,8}. Validate usage of dim_value (values should be > 0) and dim_param (all values with the same string should equate to the same size) in shapes in the model.
```

Did you already witness this @echarlaix @michaelbenayoun @JingyaHuang ?

## Before submitting


optimum/onnxruntime/utils.py
```diff
@@ -116,8 +122,10 @@ def check_supported_model(cls, model_type: str):
             )
 
     @classmethod
-    def check_optimization_supported_model(cls, model_type: str):
-        supported_model_types_for_optimization = ["bert", "gpt2", "bart"]
+    def check_optimization_supported_model(cls, model_type: str, optimization_config):
+        # as of 1.14.O: https://github.com/microsoft/onnxruntime/blob/6ccaeddefa65ccac402a47fa4d9cad8229794bb2/onnxruntime/python/tools/transformers/optimizer.py#L39
+        supported_model_types_for_optimization = ["bert", "gpt2", "bart", "unet"]
```
michaelbenayoun, line 127:
```suggestion
        supported_model_types_for_optimization = _conf.values()
```

michaelbenayoun (approved):
LGTM

fxmarty:
@michaelbenayoun ONNX Runtime have some scripts where they use half precision, I'm not sure if it is with past key values though. I'll have a look.

optimum/onnxruntime/utils.py
fxmarty: No, this is correct.

PR #796, opened 2023-02-20
Use a stateful ordered_input_names in ORTModel

fxmarty:
This allows to avoid repeated regex calls when using IOBinding.

The tests from `pytest tests/onnxruntime/test_modeling.py -k "test_compare_to_io_binding" -s` all pass.

### DIFF
--- a/optimum/onnxruntime/utils.py
+++ b/optimum/onnxruntime/utils.py
@@ -15,8 +15,10 @@
 
 import importlib.util
 import os
+import re
 from enum import Enum
-from typing import Dict, List, Optional, Tuple, Union
+from inspect import signature
+from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import torch
 from transformers.utils import logging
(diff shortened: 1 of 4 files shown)


optimum/onnxruntime/utils.py
```diff
@@ -234,6 +236,30 @@ def check_io_binding(providers: List[str], use_io_binding: Optional[bool] = None
     return use_io_binding
 
 
+def get_ordered_input_names(input_names: Dict[str, int], func: Callable) -> List[str]:
```
michaelbenayoun, line 239:
```suggestion
def get_ordered_input_names(input_names: List[str], func: Callable) -> List[str]:
```
Maybe we should do this because there is no need for the values here.

michaelbenayoun (approved):
LGTM!

optimum/onnxruntime/utils.py
fxmarty: thanks!

PR #816, opened 2023-02-24
Remove feature dimension as dynamic axes for stable diffusion ONNX export

echarlaix:
The feature dimension is static, the dimension shape will be equal to `config.hidden_size` for the `CLIPTextModel` and to  `config.cross_attention_dim` for the `UNet2DConditionModel`.


echarlaix:
> Just to understand, it was never dynamic right?

Yes !

Issue #734, opened 2023-01-31
Whisper not working with optimum pipeline


amazingvince:
### System Info

```shell
Ubuntu 20
pip installed transformers and optimum from git main
```

### Who can help?

@phi

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Seems to work fine using pipeline in huggingface and loading in the bettertransformer model. 

But the optimum pipeline throws this error. I think some code from Transfomers has not been added to support whisper config in optimum pipelines.

```python
from optimum.pipelines import pipeline

w_pipe = pipeline( "automatic-speech-recognition",  model="openai/whisper-small", accelerator="bettertransformer")
```

Error:
ValueError: Unrecognized configuration class <class 'transformers.models.whisper.configuration_whisper.WhisperConfig'> 
for this kind of AutoModel: AutoModelForCTC.
Model type should be one of Data2VecAudioConfig, HubertConfig, MCTCTConfig, SEWConfig, SEWDConfig, UniSpeechConfig, 
UniSpeechSatConfig, Wav2Vec2Config, Wav2Vec2ConformerConfig, WavLMConfig.

### Expected behavior

should be able to load whisper model with pipeline.

michaelbenayoun:
Hi @amazingvince,

You are right, it should work. @younesbelkada could you check that?

JeffreyWardman:
It also isn't up to date with the latest changes in `transformers` that allow for timestamps.

Aka this does not work with `transformers.pipeline` because language, task and return_timestamps are not in the model config:
```pipe_(audio.numpy(), chunk_length_s=30, return_timestamps=True, generate_kwargs={"language": "<|en|>", "task": "transcribe"})
```

fxmarty:
Thanks for the notice, will have a look shortly.

PR #824, opened 2023-02-27
Fix optimum pipeline initialization

fxmarty:
This PR fixes the two issues reported in https://github.com/huggingface/optimum/issues/734

Should add some more tests.


younesbelkada (approved):
Thanks a lot for your work on this!

PR #830, opened 2023-02-28
Fix t5 model type ort optimization

echarlaix:


fxmarty:
@echarlaix as discussed offline: https://github.com/microsoft/onnxruntime/blob/6ccaeddefa65ccac402a47fa4d9cad8229794bb2/onnxruntime/python/tools/transformers/models/t5/t5_helper.py#L235

echarlaix:
> @echarlaix as discussed offline, I used `bert` and not `bart` due to this: https://github.com/microsoft/onnxruntime/blob/6ccaeddefa65ccac402a47fa4d9cad8229794bb2/onnxruntime/python/tools/transformers/models/t5/t5_helper.py#L235

Thanks for the clarification !

PR #834, opened 2023-03-01
Keep intermediary models for ONNX causal-lm

fxmarty:
As per title and discussed internally.

### DIFF
--- a/optimum/onnxruntime/modeling_decoder.py
+++ b/optimum/onnxruntime/modeling_decoder.py
@@ -411,7 +411,7 @@ class ORTModelDecoder(ORTModel):
 
         ort_inference_sessions = cls.load_model(
             decoder_path=decoder_path,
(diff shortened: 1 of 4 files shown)


optimum/onnxruntime/modeling_decoder.py
```diff
@@ -380,57 +402,16 @@ def _from_pretrained(
         if model_path.is_dir():
             new_model_save_dir = model_path
             preprocessors = maybe_load_preprocessors(model_id)
-        else:
```
fxmarty, line 383: This is dead code, right? @michaelbenayoun

fxmarty:
Failing tests are unrelated (Hub down).

optimum/onnxruntime/modeling_decoder.py
fxmarty: This was not (:

PR #841, opened 2023-03-02
Remove redundant code

mht-sharma:
# What does this PR do?

The removed code was removed to export the models when `--monolith` and `use_past=True` are both passed. Since their use is removed the code is no longer required.

Fixes # (issue)

## Before submitting


fxmarty (approved):
LGTM thank you!

PR #843, opened 2023-03-02
Update documentation

echarlaix:


docs/source/onnxruntime/usage_guides/models.mdx
```diff
@@ -111,14 +111,15 @@ prompt = "sailing ship in storm by Leonardo da Vinci"
 image = stable_diffusion(prompt).images[0]
 ```
 
-To load your PyTorch model and convert it to ONNX on-the-fly, simply pass `export=True` to the `from_pretrained` method.
-After this step, don't forget to save your model using the `save_pretrained` method, to be able to load your ONNX model directly.
+To load your PyTorch model and convert it to ONNX on-the-fly, you can set `export=True`.
```
fxmarty, line 114: Is it any different than going through `optimum-cli export onnx`?

fxmarty (approved):
LGTM thank you!

PR #844, opened 2023-03-02
lazy import for task manager

JingyaHuang:


michaelbenayoun (approved):
LGTM!
Do I merge #846 after that?

optimum/exporters/tasks.py
```diff
@@ -76,17 +86,20 @@ def supported_tasks_mapping(
...
-                task, supported_backends_for_task = task
-                if backend not in supported_backends_for_task:
-                    continue
-            if "-with-past" in task:
-                mapping[backend][task] = partial(config_cls.with_past, task=task.replace("-with-past", ""))
-            else:
-                mapping[backend][task] = partial(config_cls, task=task)
+        if is_backend_available(backend):
+            config_cls = getattr(
+                importlib.import_module(f"optimum.exporters.{backend}.model_configs"), config_cls_name
```
JingyaHuang, line 91: I would import backend related module only when the backend's dependencies are available. WDYT? @michaelbenayoun

Issue #847, opened 2023-03-02
BetterTransformer pipeline fails to load on CUDA


vrdn-23:
### System Info

```shell
Optimum version: 1.7.0
Python version: 3.10.1
System: Linux
```

### Who can help?

@fxmarty 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

```python
from optimum.pipelines import pipeline
qa_model = "bert-large-uncased-whole-word-masking-finetuned-squad"
pipe_cuda = pipeline("question-answering",qa_model,accelerator="bettertransformer",device="cuda:0")
```

Error with trace ending pops up:
```
  985                 return t.to(device, dtype if t.is_floating_point() or t.is_complex() else None,
    986                             non_blocking, memory_format=convert_to_format)
--> 987             return t.to(device, dtype if t.is_floating_point() or t.is_complex() else None, non_blocking)
    988 
    989         return self._apply(convert)

NotImplementedError: Cannot copy out of meta tensor; no data!
```
Problem is only with latest release. Works fine when loading on CPU. 
Notebook for reproducible behavior: https://colab.research.google.com/drive/1aVy_jJREDQD3ll14CRZg4WfesE-HXb6i?usp=sharing

### Expected behavior

Pipeline loads without error.

fxmarty:
@vrdn-23 I had a short look, it is indeed critical. I'll probably do a patch release to disable https://github.com/huggingface/optimum/pull/676 temporarily that introduced the issue. We need to improve the tests as well.

fxmarty:
Hi @vrdn-23 , can you run `pip install -U optimum` to install the patch 1.7.1? It should fix the issue.

Thank you for notifying!

jturner116:
Upgrading to 1.7.1 fixed the issue for me, thanks for the speedy update!

PR #849, opened 2023-03-03
Disable reverse BetterTransformer

fxmarty:
Fixes https://github.com/huggingface/optimum/issues/847

cc @younesbelkada as discussed offline


younesbelkada (approved):
Thanks for fixing!

PR #852, opened 2023-03-03
Adding support for Facebook's OPT models

hivaze:
# What does this PR do?

I noticed that there is still no support for exporting to ONNX `facebook/opt-{x}` models when I needed them. Therefore, I propose very simple changes that add this support.

I checked the ability to export models with 350M, 1.3B and 6.7B parameters on A100 40GB, as you can see from the results, they can be used. Although, not all outputs for 1.3B and 6.7B models are successfully checked for closeness, in my experience, this is not something terrible.

I am [attaching](https://pastebin.com/xtRgtRYF) the export log of the 1.3B model with the `causal-lm-with-past` task.

_Note: 80GB of free RAM is not enough to export 13B OPT model._

### DIFF
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -192,6 +192,11 @@ class GPTNeoXOnnxConfig(TextDecoderOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
 
 
+class OPTOnnxConfig(TextDecoderOnnxConfig):
+    DEFAULT_ONNX_OPSET = 13
+    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
+
+
 class BloomDummyPastKeyValuesGenerator(DummyPastKeyValuesGenerator):
     def generate(self, input_name: str, framework: str = "pt"):
         past_key_shape = (
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -416,15 +416,6 @@ class TasksManager:
             "causal-lm-with-past",
             onnx="GPTNeoXOnnxConfig",
         ),
-        "opt": supported_tasks_mapping(
-            "default",
-            "default-with-past",
-            "causal-lm",
-            "causal-lm-with-past",
-            "question-answering",
-            "sequence-classification",
-            onnx="OPTOnnxConfig",
-        ),
         "groupvit": supported_tasks_mapping(
             "default",
             onnx="GroupViTOnnxConfig",
@@ -572,6 +563,15 @@ class TasksManager:
         #     "zero-shot-object-detection",
         #     onnx="OwlViTOnnxConfig",
         # ),
+        "opt": supported_tasks_mapping(
+            "default",
+            "default-with-past",
+            "causal-lm",
+            "causal-lm-with-past",
+            "question-answering",
+            "sequence-classification",
+            onnx="OPTOnnxConfig",
+        ),
         "pegasus": supported_tasks_mapping(
             "default",
             "default-with-past",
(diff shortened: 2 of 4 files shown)


regisss:
Hi @hivaze! Could you add en entry for OPT [here](https://github.com/huggingface/optimum/blob/main/tests/exporters/exporters_utils.py)?
[trl-internal-testing/tiny-random-OPTForCausalLM](https://huggingface.co/trl-internal-testing/tiny-random-OPTForCausalLM) for the tiny one and [facebook/opt-350m](https://huggingface.co/facebook/opt-350m) for the large one maybe?

Then you can make sure tests pass with:
```
pip install -e .[tests,exporters]
pytest tests/exporters/onnx/test_*.py -k "opt" -s
```
And for the large model:
```
pip install -e .[tests,exporters]
RUN_SLOW=1 pytest tests/exporters/onnx/test_*.py -k "opt" -s -m "run_slow"
```

optimum/exporters/tasks.py
```diff
@@ -416,6 +416,15 @@ class TasksManager:
             "causal-lm-with-past",
             onnx="GPTNeoXOnnxConfig",
         ),
+        "opt": supported_tasks_mapping(
```
fxmarty, line 419: Could you move this down below to keep it ordered alphabetically?

optimum/exporters/onnx/model_configs.py
```diff
@@ -192,6 +192,11 @@ class GPTNeoXOnnxConfig(TextDecoderOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
 
 
+class OPTOnnxConfig(TextDecoderOnnxConfig):
+    DEFAULT_ONNX_OPSET = 13
```
fxmarty, line 196: Is there any specific reason for this opset?

hivaze: Oh, hi, fxmarty, happy to see you here! So, speaking in general, all transformer models with `LayerNormalization` operator must have opset version equal or higher than 17, its according to https://github.com/onnx/onnx/blob/main/docs/Operators.md. Moreover, after ONNXRuntime graph optimization for transformers, this operator will appear even in models that didn't contain it before optimization, so in my opinion all supported transformer models in optimum must have `DEFAULT_ONNX_OPSET=17` or higher. 

However, in fact, when directly inferencing the model, inappropriate opset is not a big problem, at least in my experience, but checking the ONNX model using `onnx.checker.check_model()` will throw this exception: `ValidationError: No Op registered for LayerNormalization with domain_version of 13`. This might create some pitfalls or confuse people who don't have my level of tolerance for ONNX issues :)

I assigned version 13, simply because it was in some other models in the config file, and in general the model works with it. I can change it to whatever you want here, as you can see.

fxmarty: Nice to see you here as well! Thank you for the detail! I hit the exact same issue as you in: https://github.com/microsoft/onnxruntime/issues/14768 My understanding is that the `LayerNormalization` from ONNX Runtime optimization comes actually from `com.microsoft` and not the native ONNX operators. So for the issue with `onnx.checker.check_model()`, I have just ignored it for now as even with opset 17 I was getting `No Op registered` for some other custom ORT nodes.

Leaving opset 13 is fine, I was just being curious!

By the way, off topic but I've had a try at the solution you proposed to use a single decoder by hacking a fake pkv + attention mask. It works nicely. One issue I hit was that some models have a `position_ids` different whether pkv are passed or not, so you actually need this as an input to the ONNX if you want to have the exact same logits (gpt2 and co for example). I want to have a second look at it at some point - it could be the current solution with an If node has some limitations, see: https://github.com/NVIDIA/TensorRT/issues/2723

PR #853, opened 2023-03-03
Add scaled_dot_product_attention support for decoder models

fxmarty:
Not fully ready yet. The case `batch_size>1` was a bit tricky, as well as `padding="max_length"`.

This PR refactors the tests a bit as well.

Remaining to do:
* more GPU tests
* autocast error to fix
* check on pytorch version


optimum/bettertransformer/models/__init__.py
```diff
@@ -100,6 +112,48 @@ def supports(model_type: str) -> bool:
...
+    @staticmethod
+    def requires_nested_tensor(model_type: str) -> bool:
+        """
+        Returns True if the BetterTransformer implementation for a given architecture uses nested tensors, False otherwise.
+
+        Args:
+            model_type (`str`):
+                The model type to check.
+        """
+        if model_type in ["codegen", "gpt2", "gptj", "gpt_neo", "gpt_neox", "opt"]:
```
younesbelkada, line 124: nit: this is all the decoders no? can we just add something like `is_decoder`? feel free to ignore though this is a nitpicking

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -0,0 +1,291 @@
...
+
+        if batch_size == 1:
+            if query.shape[2] > 1:
+                sdpa_result = torch.nn.functional.scaled_dot_product_attention(
+                    query, key, value, attn_mask=None, dropout_p=0.0, is_causal=True
+                )
+            else:
+                sdpa_result = torch.nn.functional.scaled_dot_product_attention(
+                    query, key, value, attn_mask=None, dropout_p=0.0, is_causal=False
+                )
```
younesbelkada, line 53: Just to understand, why this check is necessary?

younesbelkada (approved):
Thanks a mile for working on this and put decoder support into reality 🚀 ! Love the new way you have refactored the tests + integration! 
I left few open questions for learning purpose
Also we can add this in the documentation in a follow up PR after the benchmarking etc

optimum/bettertransformer/models/__init__.py
fxmarty: I won't do it now, but yes, the three added methods are very similar, that should be done!

optimum/bettertransformer/models/decoder_models.py
younesbelkada: Thanks a lot for explaining!

maxiek0071:
Hi, you added a check to see if the user has PyTorch version >=2.0. According to PyTorch [website ](https://pytorch.org/get-started/pytorch-2.0/#requirements) this PyTorch version is called `2.0.0.dev20230202+cu116`.
The check you have in code under  `optimum/bettertransformer/transformation.py` in line 237 doesn't pass this version, although it is 2.0.
image

Is my statement above correct, or do I have an incorrect PyTorch version installed? Which one do you use?

fxmarty:
Thank you! Could you run:
```python
import packaging
from packaging.version import parse
import sys
import torch

print(sys.version)
print(packaging.__version__)
print(parse(torch.__version__) < parse("2.0"))
```

and give me the output?

maxiek0071:
Certainly!
image

fxmarty:
Thank you, will fix! For some reason I have `2.1.0.dev20230306+cu117` so did not hit the problem.

maxiek0071:
Great, I appreciate your help!

fxmarty:
This should be fixed in https://github.com/huggingface/optimum/pull/862, could you try again? Note that this has been added just yesterday - fp16 and training support is coming soon.

maxiek0071:
Yes, it works now, if I install from this branch. Thanks!

PR #854, opened 2023-03-03
Quantization with TFLite

michaelbenayoun:
# What does this PR do?

This PR adds support for quantization with TFlite in `optimum.exporters.tflite`:

- It refactors `optimum/utils/preprocessing`, `TasksProcessors` now offer a `load_dataset` method, which wraps the `Datasets.load_dataset` function (with a few additional options) and returns a **processed version of the dataset for the considered task**. Currently 4 tasks are supported:
    - Text classification
    - Token classification
    - Question answering
    - Image classification
 - The following quantization approaches are supported:
    - Dynamic INT8 quantization
    - Static INT8 quantization (requires a calibration dataset): it is possible to allow fallback to float kernels and to request for end-to-end integer computation (inputs and outputs are also integers).
    - Static INT8x16 quantization (requires a calibration dataset): the weights are in int8 and activation are in int16
    - FP16 quantization

Usage:
```bash
optimum-cli export tflite -m csarron/bert-base-uncased-squad-v1 --task question-answering --sequence_length 384 --quantize int8 --num_calibration_samples 100 bert_int8
```
 
Heavy testing is done, and the tests are decoupled in multiple github actions to make things faster.

To do in later PRs:

- [ ] Document how to use quantization
- [ ] Extend the `TasksProcessors` to all the supported tasks
- [ ] Add a validation step after quantization (still needs to be determined how)


optimum/commands/export/tflite.py
```diff
@@ -116,6 +117,97 @@ def parse_args_tflite(parser):
         help=f"Audio tasks only. Audio sequence length {doc_input}",
     )
 
+    quantization_group = parser.add_argument_group("Quantization")
+    quantization_group.add_argument(
+        "--quantize",
+        choices=[e.value for e in QuantizationApproach],
+        type=str,
+        default=None,
+        help="The method of quantization to perform. No quantization will happen if left unspecified.",
```
fxmarty, line 126: Maybe detail the choices here or have a link to a reference

optimum/exporters/tflite/convert.py
```diff
@@ -132,10 +137,39 @@ def validate_model_outputs(
...
+    calibration_dataset_name_or_path: Optional[Union[str, Path]] = None,
+    calibration_dataset_config_name: Optional[str] = None,
+    preprocessor: Optional[Preprocessor] = None,
+    num_calibration_samples: int = 200,
+    calibration_split: Optional[str] = None,
+    primary_key: Optional[str] = None,
+    secondary_key: Optional[str] = None,
+    question_key: Optional[str] = None,
+    context_key: Optional[str] = None,
+    image_key: Optional[str] = None,
```
fxmarty, line 172: Are all of these arguments related to quantization? Couldn't we separate that into a dataclass, a dict or something? Here some arguments are task specific (question_key, context_key), which should probably be avoided.

optimum/exporters/tflite/convert.py
```diff
@@ -173,6 +244,118 @@ def export(
     with TemporaryDirectory() as tmp_dir_name:
         model.save(tmp_dir_name, signatures=signatures)
         converter = tf.lite.TFLiteConverter.from_saved_model(tmp_dir_name)
+
+        if quantization is not None:
```
fxmarty, line 248: For readability, the quantization case should IMO belong to an other function, but it is taste

tests/exporters/tflite/test_exporters_tflite_cli.py
```diff
@@ -65,8 +66,13 @@ def _get_models_to_test(export_models_dict: Dict):
                         exporter_config_kwargs=DEFAULT_DUMMY_SHAPES,
                     )
 
+                    if task == "question-answering":
+                        default_shapes = dict(DEFAULT_DUMMY_SHAPES)
+                        default_shapes["sequence_length"] = 384
+                    else:
+                        default_shapes = DEFAULT_DUMMY_SHAPES
```
fxmarty, line 73: why?

optimum/exporters/tflite/base.py
```diff
@@ -40,6 +41,13 @@ class MissingMandatoryAxisDimension(ValueError):
     pass
 
 
+class QuantizationApproach(str, Enum):
+    INT8_DYNAMIC = "int8-dynamic"
+    INT8 = "int8"
+    INT8x16 = "int8x16"
+    FP16 = "fp16"
```
fxmarty, line 48: Is there a reference in TFLite documentation where fp16 is considered as quantization? I don't think this is a common name.

Also, this could be confusing with the ONNX export that just uses `--fp16`.

optimum/exporters/tflite/convert.py
```diff
@@ -173,6 +244,118 @@ def export(
...
+            # Handling the calibration dataset:
+            # - Either loading the default dataset if no calibration dataset was provided or the required dataset,
+            # - Splitting the dataset with the provided a dataset split or with the first split if none is provided.
+            # - Shuffling the split.
+            # - Selecting num_calibration_samples in the dataset split.
+            # - Batching the dataset.
+            # - Converting it to the TensorFlow format.
+            from ...exporters import TasksManager
+
+            task = TasksManager.get_task_from_model(model)
```
fxmarty, line 264: Is this really needed? Couldn't we rather pass the task as an argument, and avoid adding `get_task_from_model` in the TasksManager?

optimum/exporters/tasks.py
```diff
@@ -1078,6 +1097,48 @@ def get_model_from_task(
                 model = model_class.from_pretrained(model_name_or_path, **kwargs)
         return model
 
+    @classmethod
+    def get_task_from_model(
```
fxmarty, line 1101: Apart from using it in `export()`, is there any other use?

fxmarty:
I think extending the documentation would be great!

optimum/exporters/tflite/convert.py
michaelbenayoun: I can make a dataclass, but I think it will become less readable on what should the function take. 
Disagree on the second part, we support many tasks and need to be able to provide an argument for each of them. This is not problematic to have those arguments living together here.

optimum/commands/export/tflite.py
michaelbenayoun: Done!

tests/exporters/tflite/test_exporters_tflite_cli.py
michaelbenayoun: I am trying to avoid that, currently preprocessing fails for some reason (`stride < max_len`) and related errors.

optimum/exporters/tflite/base.py
michaelbenayoun: That's how they call it in TFLite => https://www.tensorflow.org/lite/performance/post_training_float16_quant?hl=en

But we could have a `--fp16` as well, which would be synonymous to `--quantization fp16`,

optimum/exporters/tflite/convert.py
michaelbenayoun: We could, but I think this is better (it is normalized, the user does not need to know if he should use `text-classification` or `sequence-classification` for instance). It is also used in ORT now since we had a similar need there.

optimum/exporters/tasks.py
michaelbenayoun: https://github.com/huggingface/optimum/pull/854/files#diff-4fa3987b008533d0913c0cb0c3dade3fa591f47937b3baa7f41696a17beac945R185

michaelbenayoun:
Yes, will do it in later PRs, otherwise this PR will become huge.

optimum/exporters/tflite/convert.py
michaelbenayoun: Made it a dataclass.

optimum/exporters/tasks.py
fxmarty: Yes, but it was working before so I'm not sure why this change is necessary.

optimum/exporters/tflite/base.py
fxmarty: Great if it's the terminology of TFLite let's keep it!

optimum/exporters/tasks.py
michaelbenayoun: It basically does the same thing.

fxmarty: Yes, so I'm not sure there's a point adding this method, normally I would think you would get the task from the model id and work from there, not load the model and then try to get the task.

optimum/utils/preprocessing/base.py
```diff
@@ -1,101 +1,224 @@
...
+                    f"{num_samples} examples."
+                )
+            dataset = dataset.select(range(num_samples))
+
+        column_names = dataset.column_names
+        if isinstance(column_names, dict):
+            column_names = list(set(itertools.chain.from_iterable(column_names.values())))
+
+        if data_keys is None:
+            data_keys = self.try_to_guess_data_keys(column_names)
```
fxmarty, line 158: A warning should be raised here I think - just to avoid unexpected behavior

optimum/utils/preprocessing/base.py
```diff
@@ -1,101 +1,224 @@
...
+                    "Data keys need to be specified manually since they could not be guessed from "
+                    f"{', '.join(column_names)}"
+                )
+        elif not set(data_keys.keys()) <= self.ALLOWED_DATA_KEY_NAMES:
+            raise ValueError(
+                f"data_keys contains unallowed keys {set(data_keys.keys())}, allowed_keys: {self.ALLOWED_DATA_KEY_NAMES}."
+            )
+
+        if ref_keys is None:
+            ref_keys = self.try_to_guess_ref_keys(column_names)
```
fxmarty, line 170: same, I think a warning here would be good

optimum/utils/preprocessing/image_classification.py
```diff
@@ -1,114 +1,115 @@
...
-        # `metric.compute` may return a dict or a number
-        if not isinstance(metrics_res, dict):
-            metrics_res = {metric.name: metrics_res}
-
-        return metrics_res
-
-    def get_pipeline_kwargs(self):
-        return {}
+        # TODO: do we want to do that here?
+        # eval_dataset = eval_dataset.align_labels_with_mapping(self.config.label2id, self.ref_keys[0])
```
fxmarty, line 114: As long as you don't evaluate it's fine. If you do evaluation, this is needed as some models on the Hub don't have there label2id matching the datasets label2id (different ordering).

pytest.ini
fxmarty, line 1: I would rather not have this file and put it in pyproject.toml

setup.py
```diff
@@ -17,6 +17,7 @@
     "sympy",
     "transformers[sentencepiece]>=4.26.0",
     "torch>=1.9",
+    "torchvision",
```
fxmarty, line 20: why?

tests/exporters/tflite/test_exporters_tflite_cli.py
```diff
@@ -83,20 +88,73 @@ class TFLiteCLIExportTestCase(unittest.TestCase):
...
+                command.append(f"--image_key={image_key}")
+            command.append(f"--num_calibration_samples={num_calibration_samples}")
+
+            tflite_config = tflite_config_constructor(AutoConfig.from_pretrained(model_name))
+            mandatory_axes = tflite_config.get_mandatory_axes_for_task(task)
+            shapes = [f"--{name}={getattr(tflite_config, name)}" for name in mandatory_axes]
+            command += shapes
+            command.append(tmpdir)
+
+            proc = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
fxmarty, line 146: No need to do it now, but in a following PR we could remove subprocess calls as well.

.github/workflows/test_export_tflite_cli.yml
```diff
@@ -31,4 +31,4 @@ jobs:
     - name: Test with unittest
       working-directory: tests
       run: |
-        pytest exporters/tflite/test_exporters_tflite_cli.py -s --durations=0
+        pytest exporters/tflite/test_exporters_tflite_cli.py -m "not quantization" -s --durations=0
```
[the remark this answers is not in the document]
michaelbenayoun: I created a `pytest.ini` for that, did not know (or forgot) that we had that in `pyproject.toml`, doing it now.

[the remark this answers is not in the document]
michaelbenayoun: Actually I think this is better in `pytest.ini` since you can add a comment on the marker?

optimum/exporters/tflite/convert.py
```diff
@@ -132,10 +138,180 @@ def validate_model_outputs(
...
+        # Handling the calibration dataset:
+        # - Either loading the default dataset if no calibration dataset was provided or the required dataset,
+        # - Splitting the dataset with the provided a dataset split or with the first split if none is provided.
+        # - Shuffling the split.
+        # - Selecting num_calibration_samples in the dataset split.
+        # - Batching the dataset.
+        # - Converting it to the TensorFlow format.
+        from ...exporters import TasksManager
+
+        task = TasksManager.get_task_from_model(model)
```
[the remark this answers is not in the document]
michaelbenayoun: This also solves the issue with `text-classification` vs `sequence-classification`.
Again, I do not see why making it explicit is better in any way. We should be able to infer it since it is an auto-class.

optimum/exporters/tflite/convert.py
```diff
@@ -132,10 +138,180 @@ def validate_model_outputs(
...
+            "roberta",
+        }:
+            preprocessor.add_prefix_space = True
+
+        load_smallest_split = quantization_config.calibration_split is None
+        if load_smallest_split:
+            logger.warning(
+                "Since no calibration split was provided for the calibration dataset, the smallest split will be "
+                "used if the dataset contains multiple splits."
+            )
```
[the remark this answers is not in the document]
michaelbenayoun: We are not validating yet. When we are validating I will make sure it will not be the case.
Without this, the user would need to pass the split for a "default" dataset (when none is provided, we use a default one), which is annoying.

optimum/utils/preprocessing/base.py
michaelbenayoun: Done

optimum/utils/preprocessing/base.py
michaelbenayoun: Done

setup.py
michaelbenayoun: Needed for preprocessing. 
I think this is not an issue as long as we have pytorch as a requirement. When we removed pytorch (which we should), we can handle this case as well.

.github/workflows/test_export_tflite_cli.yml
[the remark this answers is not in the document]
fxmarty: Comments can also be put in pyproject.toml

optimum/exporters/tflite/convert.py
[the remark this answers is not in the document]
fxmarty: Yes, I think it would be better if we force the user to pass a split if e.g. the `train` split is not in the dataset. Or maybe at least log which split is auto-selected (maybe it is already done)?

optimum/exporters/tflite/convert.py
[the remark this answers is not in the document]
fxmarty: The task is already known in the export script - here we reinfer the task, and you have to be confident in that `get_task_from_model` does its job and yields the right and task as the one you got during the export.

It's not a big deal but just from design aesthetic pov, we do twice the same thing in two different methods and just hope they are both well implemented - instead of hopping one is well implemented, which is easier to maintain.

But this is just taste, it's fine to me as is.

optimum/exporters/tflite/convert.py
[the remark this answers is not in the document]
michaelbenayoun: Added a log.

fxmarty (approved):
LGTM, looks really good!

PR #856, opened 2023-03-06
Support scaled_dot_product_attention for t5

fxmarty:
This is only usable for the decoder, with [`T5LayerCrossAttention`](https://github.com/huggingface/transformers/blob/820c46a707ddd033975bc3b0549eea200e64c7da/src/transformers/models/t5/modeling_t5.py#L647) as discussed offline, due to the position bias present in `T5LayerSelfAttention`.


younesbelkada (approved):
Thanks a lot for this! 🔥

PR #857, opened 2023-03-06
[`BT`] add decoder benchmark script

younesbelkada:
# What does this PR do?

This PR adds the benchmark script adapted for decoder-based models as well, to benchmark the speedup we obtain with `torch.sdpa` and `transformers` models

cc @fxmarty

### DIFF
--- a/tests/benchmark/benchmark_bettertransformer.py
+++ b/tests/benchmark/benchmark_bettertransformer.py
@@ -1,7 +1,7 @@
 import argparse
 
 import torch
-from transformers import AutoModel
+from transformers import AutoModel, AutoModelForCausalLM, GenerationConfig
 
 from optimum.bettertransformer import BetterTransformer
 
@@ -56,6 +56,16 @@ def get_parser():
         "--use-mask",
         action="store_true",
     )
+    parser.add_argument(
+        "--is_decoder",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--max_token",
+        type=int,
+        default=100,
+        help="",
+    )
     return parser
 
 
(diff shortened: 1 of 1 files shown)


tests/benchmark/benchmark_bettertransformer.py
```diff
@@ -90,41 +100,46 @@ def get_batch(batch_size, avg_seqlen, max_sequence_length, seqlen_stdev, vocab_s
...
-    _ = bt_model(input_ids, masks)
-    torch.cuda.synchronize()
-
-    total_hf_time = timing_cuda(hf_model, num_batches, input_ids, masks)
-    total_bt_time = timing_cuda(bt_model, num_batches, input_ids, masks)
+    if is_decoder:
+        gen_config = GenerationConfig(
+            do_greedy=True,
+            max_new_tokens=max_token,
+            use_cache=False,
```
fxmarty, line 123: Why `use_cache = False`? I'd rather benchmark with the cache enabled.

Also, can we rather set `min_length=length, max_length=length`? So that we can generate a specific length.

tests/benchmark/benchmark_bettertransformer.py
```diff
@@ -90,41 +100,46 @@ def get_batch(batch_size, avg_seqlen, max_sequence_length, seqlen_stdev, vocab_s
...
+        )
+        _ = hf_model.generate(input_ids, generation_config=gen_config)
```
fxmarty, line 125: the `input_ids` are not defined no?

tests/benchmark/benchmark_bettertransformer.py
```diff
@@ -90,41 +100,46 @@ def get_batch(batch_size, avg_seqlen, max_sequence_length, seqlen_stdev, vocab_s
...
 
-def timing_cuda(model, num_batches, input_ids, masks):
+def timing_cuda(model, num_batches, input_ids, masks, is_decoder, generation_config=None):
     start_event = torch.cuda.Event(enable_timing=True)
     end_event = torch.cuda.Event(enable_timing=True)
     start_event.record()
     for _ in range(num_batches):
-        _ = model(input_ids, masks)
+        if is_decoder:
+            _ = model.generate(input_ids, generation_config=generation_config)
```
fxmarty, line 109: the `input_ids` is not defined no?

tests/benchmark/benchmark_bettertransformer.py
fxmarty: Does it make sense to pass `attention_mask` as well? Not sure

tests/benchmark/benchmark_bettertransformer.py
younesbelkada: I think it's because `max_length` is deprecated, yes we can add `min_length` I think let me try

fxmarty (approved):
Feel free to merge once you think it's good. I think we should use `use_cache=True` to benchmark.

PR #859, opened 2023-03-07
[`BT`] Add fp16 support

younesbelkada:
# What does this PR do?

Currently on the `main` branch the fp16 inference for BetterTransformer decoder models is not supported, this PR aims to fix this

## TODO

- [x] add tests

cc @fxmarty

### DIFF
--- a/optimum/bettertransformer/models/decoder_models.py
+++ b/optimum/bettertransformer/models/decoder_models.py
@@ -52,6 +52,9 @@ class GPT2AttentionLayerBetterTransformer(BetterTransformerBaseLayer):
         raise_on_head_mask(head_mask)
         batch_size = query.shape[0]
 
+        mask_value = torch.finfo(value.dtype).min
+        attention_mask = torch.full([], mask_value, dtype=value.dtype).to(value.device)
+
         if batch_size == 1 and attention_mask is not None and attention_mask[0, 0, 0, -1] < -1:
             raise ValueError("BetterTransformer does not support padding='max_length' with a batch size of 1.")
 
@@ -74,13 +77,11 @@ class GPT2AttentionLayerBetterTransformer(BetterTransformerBaseLayer):
                     torch.bool
                 )
 
-                causal_mask = torch.where(causal_mask, 0, self._mask_value)
+                causal_mask = torch.where(causal_mask, 0, self._mask_value).to(value.dtype)
 
                 # torch.Tensor.expand does no memory copy
                 causal_mask = causal_mask.expand(batch_size, -1, -1, -1)
                 attention_mask = causal_mask + attention_mask
-
-            attention_mask = attention_mask.to(torch.bool)
 
             query = query.to(value.dtype)
             key = key.to(value.dtype)
@@ -126,6 +127,10 @@ class GPTNeoAttentionLayerBetterTransformer(BetterTransformerBaseLayer):
         raise_on_head_mask(head_mask)
         query = query * self.scale
         batch_size = query.shape[0]
+
+        mask_value = torch.finfo(value.dtype).min
+        attention_mask = torch.full([], mask_value, dtype=value.dtype).to(value.device)
+
         if batch_size == 1 and attention_mask is not None and attention_mask[0, 0, 0, -1] < -1:
             raise ValueError("BetterTransformer does not support padding='max_length' with a batch size of 1.")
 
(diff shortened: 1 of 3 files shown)


younesbelkada:
thanks @fxmarty for the heads up, will try now

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -121,6 +127,10 @@ def wrapped_scaled_dot_product(
         raise_on_head_mask(head_mask)
         query = query * self.scale
         batch_size = query.shape[0]
+
+        mask_value = torch.finfo(value.dtype).min
+        attention_mask = torch.full([], mask_value, dtype=value.dtype).to(value.device)
```
younesbelkada, line 132: This will probably break the logis tests

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -74,12 +76,15 @@ def wrapped_scaled_dot_product(
                     torch.bool
                 )
 
-                causal_mask = torch.where(causal_mask, 0, self._mask_value)
+                causal_mask = torch.where(causal_mask, 0, self._mask_value).to(value.dtype)
```
fxmarty, line 79: This cast is bad. Can we instead move the definition of `mask_value` in the forward?

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -74,12 +76,15 @@ def wrapped_scaled_dot_product(
...
 
                 # torch.Tensor.expand does no memory copy
                 causal_mask = causal_mask.expand(batch_size, -1, -1, -1)
                 attention_mask = causal_mask + attention_mask
 
+            query = query.to(value.dtype)
+            key = key.to(value.dtype)
```
fxmarty, line 86: Is this really needed? It should already be of the same dtype no?

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -38,8 +38,20 @@ def __init__(self, gpt_layer: "nn.Module", config: "PretrainedConfig"):
...
+        elif config.model_type == "gpt_neox":
+            target_dtype = self.gpt_layer.dense.weight.dtype
+        # gpt-j
+        else:
+            target_dtype = self.gpt_layer.out_proj.weight.dtype
+
+        self.downcast_qk = config.model_type in ["gptj", "gpt_neox"]
+
+        mask_value = torch.finfo(target_dtype).min
+        self._mask_value = torch.full([], mask_value, dtype=target_dtype)
```
fxmarty, line 54: This will IMO not work because the user may use `model = model.to(torch.float16)` after initializing the model. Here, `self._mask_value` would still be on fp32. I think we really need it in the forward.

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -52,6 +64,8 @@ def wrapped_scaled_dot_product(
         raise_on_head_mask(head_mask)
         batch_size = query.shape[0]
 
+        # attention_mask = attention_mask.to(value.device, value.dtype)
+
```
younesbelkada, line 68:
```suggestion
```

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -98,8 +118,9 @@ def __init__(self, gpt_layer: "nn.Module", config: "PretrainedConfig"):
         self.gpt_layer = gpt_layer
         self.gpt_layer._attn = self.wrapped_scaled_dot_product
 
-        mask_value = torch.finfo(torch.float32).min
-        self._mask_value = torch.full([], mask_value, dtype=torch.float32)
+        target_dtype = self.gpt_layer.k_proj.weight.dtype
+        mask_value = torch.finfo(target_dtype).min
+        self._mask_value = torch.full([], mask_value, dtype=target_dtype)
```
fxmarty, line 121: same

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -74,12 +76,18 @@ def wrapped_scaled_dot_product(
                     torch.bool
                 )
 
-                causal_mask = torch.where(causal_mask, 0, self._mask_value)
+                causal_mask = torch.where(causal_mask, 0, self._mask_value.to(value.dtype))
```
fxmarty, line 79: This is not equivalent:
```
import torch

mask_value = torch.finfo(torch.float32).min
mask_value = torch.full([], mask_value, dtype=torch.float32)
casted = mask_value.to(torch.float16)

mask_value = torch.finfo(torch.float16).min
mask_value = torch.full([], mask_value, dtype=torch.float16)
assert torch.equal(casted, mask_value)
```

younesbelkada:
@fxmarty  btw `self._mask_value` seems to be not used for T5 and OPT, shall we remove them?

tests/bettertransformer/test_decoder.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 import unittest
 
+import pytest
```
fxmarty, line 17:
```suggestion
import pytest
from optimum.utils.testing_utils, require_torch_gpu
```

fxmarty (approved):
LGTM thank you for iterating on this!

PR #860, opened 2023-03-07
[`BT`] Add decoder training support

younesbelkada:
# What does this PR do?

This PR adds training support for decoder-based models with `BT`

## TODO

- [ ] add tests

cc @fxmarty

### DIFF
--- a/optimum/bettertransformer/transformation.py
+++ b/optimum/bettertransformer/transformation.py
@@ -246,7 +246,6 @@ class BetterTransformer(object):
             remove_hook_from_module(model, recurse=True)
 
         if keep_original_model:
-            model = model.requires_grad_(False)
             try:
                 model_fast = deepcopy(model)
             except RuntimeError:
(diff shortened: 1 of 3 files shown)


tests/bettertransformer/testing_utils.py
```diff
@@ -175,6 +175,24 @@ def _test_raise_train(self, model_id: str, **kwargs):
...
+
+        bt_model = BetterTransformer.transform(hf_random_model, keep_original_model=True)
+        bt_model.train()
+        output = bt_model(**inputs)[0]
+        loss = output.sum()
+        loss.backward()
+
+        # check if gradients are not None
+        for param in bt_model.parameters():
+            self.assertIsNotNone(param.grad)
```
fxmarty, line 194: Looks fine to me for now - we will still need to have a thorough convergence test at some point, but it is fine for now. See https://arxiv.org/pdf/2205.14135.pdf#figure.caption.19

optimum/bettertransformer/transformation.py
```diff
@@ -246,7 +246,6 @@ def transform(
             remove_hook_from_module(model, recurse=True)
 
         if keep_original_model:
-            model = model.requires_grad_(False)
```
fxmarty, line 249: is this fine to remove this for encoder models?

younesbelkada: actually no, let me propose a fix

PR #863, opened 2023-03-07
Bart support scaled_dot_product_attention

fxmarty:
As per title.

### DIFF
--- a/optimum/bettertransformer/models/decoder_models.py
+++ b/optimum/bettertransformer/models/decoder_models.py
@@ -332,7 +332,6 @@ class OPTAttentionLayerBetterTransformer(BetterTransformerBaseLayer):
                 f" {attn_output.size()}"
             )
 
-        # attn_output = attn_output.view(bsz, self.opt_layer.num_heads, tgt_len, self.opt_layer.head_dim)
         attn_output = attn_output.transpose(1, 2)
 
         # Use the `embed_dim` from the config (stored in the class) rather than `hidden_state` because `attn_output` can be
@@ -568,7 +567,6 @@ class BartAttentionLayerBetterTransformer(BetterTransformerBaseLayer):
                 f" {attn_output.size()}"
             )
 
-        # attn_output = attn_output.view(bsz, self.layer.num_heads, tgt_len, self.layer.head_dim)
         attn_output = attn_output.transpose(1, 2)
 
         # Use the `embed_dim` from the config (stored in the class) rather than `hidden_state` because `attn_output` can be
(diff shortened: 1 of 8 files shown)


optimum/bettertransformer/models/decoder_models.py
```diff
@@ -476,10 +475,6 @@ def project(hidden_states, proj_layer, key_value_states, past_key_value):
                 attn_weights, p=self.layer.dropout, training=self.layer.training
             )  # (batch_size, n_heads, seq_length, key_length)
 
-            # Mask heads if we want to
```
younesbelkada, line 479: Why these have been removed?

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -491,3 +486,94 @@ def project(hidden_states, proj_layer, key_value_states, past_key_value):
         if output_attentions:
             outputs = outputs + (attn_weights,)
         return outputs
+
+
+class BartAttentionLayerBetterTransformer(BetterTransformerBaseLayer):
+    def __init__(self, layer: "nn.Module", config: "PretrainedConfig"):
+        super().__init__(config, layer)
+
+        self.layer = layer
```
younesbelkada, line 495:
```suggestion
        self.layer = layer
        self.is_decoder = True
```
no?

tests/bettertransformer/testing_utils.py
```diff
@@ -89,32 +89,34 @@ def _test_fp16_inference(self, model_id: str, use_to_operator=False, **preproces
...
 
+        length = 50
         with torch.no_grad():
             r"""
             Make sure the models are in eval mode! Make also sure that the original model
             has not been converted to a fast model. The check is done above.
             """
             torch.manual_seed(0)
-            output_hf = hf_random_model.generate(**inputs)
+            output_hf = hf_random_model.generate(**inputs, min_length=length, max_length=length)
```
younesbelkada, line 112: I don't really see the advantage of using `auto_model_class` since here we always call `generate`

younesbelkada (approved):
Thanks a lot for this! 🚀

optimum/bettertransformer/models/decoder_models.py
fxmarty: This is not supported and checked for above

younesbelkada: Thanks makes sense!

tests/bettertransformer/testing_utils.py
younesbelkada: Makes sense! Thanks for clarifying

PR #864, opened 2023-03-07
[`BT`] add `accelerate_test` markers

younesbelkada:
# What does this PR do?

as per title for a more convenient way of running accelerate only tests

cc @fxmarty


fxmarty (approved):
Thanks can you add this (and fp16 as well) in https://github.com/huggingface/optimum/blob/f24195cb1e5cb750d2f42904ab9b8f61ea51942c/pyproject.toml#L34 ?

PR #865, opened 2023-03-07
Mbart, pegasus, blenderbot, marian, m2m_100 support scaled_dot_product_attention

fxmarty:
Most probably wav2vec2, whisper, hubert would support as well, but I'll leave it to an other PR once we have benchmarked with what we already have.


younesbelkada (approved):
Thanks a lot for your great work on this!

ArthurZucker:
Cool, but the other models you mentioned are more used and more downloaded so maybe more important ! 🚀 Can't wait for Whisper

PR #868, opened 2023-03-08
Add bettertransformer reverse transform

fxmarty:
Should be good now.

Wav2vec2 and Hubert can not be loaded on meta device due to a bug in PyTorch.


younesbelkada (approved):
Thanks a mile for working on this!

Issue #869, opened 2023-03-08
ORT whisper on CUDAExecutionProvider is slower than PyTorch


yufenglee:
### System Info

```shell
optimum 1.7.1
```

### Who can help?

@lewtun , @michaelbenayoun 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

The exported whisper ONNX decoder model has encoder.value as outputs. Actually encoder.value are constant in the decoding stage. Those copies with Identiy are very heavy and make the performance much worse.

```
from optimum.onnxruntime import ORTModelForSpeechSeq2Seq
session_options = onnxruntime.SessionOptions()
session_options.enable_profiling = True
model_ort = ORTModelForSpeechSeq2Seq.from_pretrained(whisper_model_name, from_transformers=True, use_io_binding=True, session_options=session_options)
generator_ort = pipeline(
    task="automatic-speech-recognition",
    model=model_ort,
    feature_extractor=processor.feature_extractor,
    tokenizer=processor.tokenizer,
    # batch_size=2,
    # device=0,
    chunk_length_s=30,
    stride_length_s=(1,1), # must have with chunk_length_s
    generate_kwargs={"max_new_tokens": 1024},
)
```

![image](https://user-images.githubusercontent.com/30486710/223828429-0235d14c-2537-41d6-846d-bb65cb88ae0d.png)

### Expected behavior

Don't output encoder value in decoder model.

fxmarty:
Thank you for the hint @yufenglee, will fix shortly!

xenova:
I think this might be connected to an issue I am currently having with onnxruntime-web: https://discuss.huggingface.co/t/when-exporting-seq2seq-models-with-onnx-why-do-we-need-both-decoder-with-past-model-onnx-and-decoder-model-onnx/33354

For some reason, if I use an empty tensor of the correct shape (e.g., [batch, heads, 0, dim]) - which is completely valid for the PyTorch implementation - it returns an empty tensor for the present key values (as opposed to the PyTorch implementation which produces the correct output. (Let me see if I can get a minimal example working)

yufenglee:
> Thank you for the hint @yufenglee, will fix shortly!

Thanks!

hannan72:
[...]

PR #872, opened 2023-03-09
Remove constant outputs from decoder with past ONNX model for encoder-decoder architectures

fxmarty:
This PR removes useless ONNX outputs in decoder with past models for encoder-decoder architectures.

Note: the naming `present.0.decoder.value` and `present.0.encoder.value` is wrong: all past key values are used in the decoder, some happen to be used for self attention, other for cross attention. This should be fixed in an other PR.

**Breaking change:**
* ONNX export does not include cross attention past key values in `decoder_with_past_model.onnx` outputs
* ONNX Runtime models do not support `decoder_with_past_model.onnx` that have cross attention past key values outputs

@michaelbenayoun @mht-sharma @JingyaHuang @regisss @echarlaix Is this too big of a breaking change?

Potential issue: Since the `decoder_model.onnx` and `decoder_with_past_model.onnx` will not have the same set of outputs, it will be harder to merge them with a `If` node.

Fixes https://github.com/huggingface/optimum/issues/869


optimum/onnxruntime/base.py
```diff
@@ -140,6 +140,11 @@ def __init__(
         if self.parent_model.use_cache is True and len(self.key_value_output_names) == 0:
             raise RuntimeError("Could not find the past key values in the provided model.")
 
+        if len(self.key_value_input_names) > 0:
+            self.use_past = True
+        else:
+            self.use_past = False
```
michaelbenayoun, line 146:
```suggestion
         self.use_past = len(self.key_value_input_names) > 0
```

optimum/onnxruntime/base.py
```diff
@@ -403,7 +420,8 @@ def compute_past_key_values_output_shapes(
         past_key_values_shapes = {}
         for idx, name in enumerate(self.key_value_output_names):
             is_self_attn = idx % 4 < 2
-            past_key_values_shapes[name] = self_attn_shape if is_self_attn else cross_attn_shape
+            # decoder with past does not ouput cross attention key/values as they are constants
```
michaelbenayoun, line 423:
```suggestion
            # Decoder with past does not ouput cross attention key/values as they are constants.
```

optimum/onnxruntime/base.py
```diff
@@ -552,4 +560,19 @@ def filter_out_output(output_name):
                 if use_torch:
                     loss = torch.from_numpy(loss).to(self.device)
 
-        return Seq2SeqLMOutput(loss=loss, logits=logits, past_key_values=past_key_values)
+        # TODO: this is extremely ugly and unreadable. What if cross-attention k/v change?
+        # Tuple of tuple of length `n_layers`, with each tuple of length equal to:
+        # * 4 for the decoder without cache (k/v of self-attention + k/v of cross-attention)
+        # * 2 for the decoder with cache (k/v of self-attention as cross-attention cache is constant)
+        if self.use_past is False:
```
michaelbenayoun, line 567: Not easily readable imo

hannan72:
I'm facing such a problem on whisper model.
Is this just due to decoder_with_past  model?

optimum/onnxruntime/base.py
fxmarty: Agreed, this can be improved later. Let me know if the comment can be improved.

Currently some pkv inputs and outputs are exactly the same, so not useful and they add an unnecessary overhead.

optimum/onnxruntime/base.py
fxmarty: feels less readable

fxmarty:
Merging as approved internally

PR #881, opened 2023-03-13
Remove positional-only parameters not support by python < v3.8

echarlaix:
The [syntax](https://peps.python.org/pep-0570/) for specifying positional-only parameters in python function was introduced in python v3.8 and will throw error for lower version (for `v3.7` in our case)


regisss (approved):
LGTM!

younesbelkada (approved):
Thanks a lot for this!

PR #883, opened 2023-03-15
Automatic framework detection in TasksManager for large models

fxmarty:
Fixes https://github.com/huggingface/optimum/issues/867

### DIFF
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -912,38 +912,40 @@ class TasksManager:
         if framework is not None:
             return framework
 
-        framework_map = {"pt": "PyTorch", "tf": "TensorFlow"}
-
         full_model_path = Path(model_name_or_path) / subfolder
         if full_model_path.is_dir():
-            if (full_model_path / WEIGHTS_NAME).is_file():
-                framework = "pt"
-            elif (full_model_path / TF2_WEIGHTS_NAME).is_file():
-                framework = "tf"
-            else:
-                raise FileNotFoundError(
-                    "Cannot determine framework from given checkpoint location."
-                    f" There should be a {WEIGHTS_NAME} for PyTorch"
-                    f" or {TF2_WEIGHTS_NAME} for TensorFlow."
-                )
-            logger.info(f"Local {framework_map[framework]} model found.")
+            all_files = [
+                os.path.relpath(os.path.join(dirpath, file), full_model_path)
+                for dirpath, _, filenames in os.walk(full_model_path)
+                for file in filenames
+            ]
         else:
             if not isinstance(model_name_or_path, str):
                 model_name_or_path = str(model_name_or_path)
-            try:
-                url = huggingface_hub.hf_hub_url(model_name_or_path, WEIGHTS_NAME, subfolder=subfolder)
-                huggingface_hub.get_hf_file_metadata(url)
-                framework = "pt"
-            except Exception:
-                pass
+            all_files = huggingface_hub.list_repo_files(model_name_or_path, repo_type="model")
 
-            if framework is None:
-                try:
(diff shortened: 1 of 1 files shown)


optimum/exporters/tasks.py
```diff
@@ -912,38 +912,40 @@ def determine_framework(
...
-                framework = "tf"
-            else:
-                raise FileNotFoundError(
-                    "Cannot determine framework from given checkpoint location."
-                    f" There should be a {WEIGHTS_NAME} for PyTorch"
-                    f" or {TF2_WEIGHTS_NAME} for TensorFlow."
-                )
-            logger.info(f"Local {framework_map[framework]} model found.")
+            all_files = [
+                os.path.relpath(os.path.join(dirpath, file), full_model_path)
```
michaelbenayoun, line 918: I am more in favor to use pathlib here but that's just my tastes:

- os.walk => path.glob
- os.path.relpath => path.relative_to

optimum/exporters/tasks.py
```diff
@@ -912,38 +912,40 @@ def determine_framework(
...
-
-            if framework is None:
-                try:
-                    url = huggingface_hub.hf_hub_url(model_name_or_path, TF2_WEIGHTS_NAME, subfolder=subfolder)
-                    huggingface_hub.get_hf_file_metadata(url)
-                    framework = "tf"
-                except Exception:
-                    pass
+            all_files = huggingface_hub.list_repo_files(model_name_or_path, repo_type="model")
+
```
michaelbenayoun, line 926:
```suggestion
```

optimum/exporters/tasks.py
```diff
@@ -912,38 +912,40 @@ def determine_framework(
...
+            all_files = [file for file in all_files if file.startswith(subfolder)]
+
+        if any(
+            [
+                file.startswith(Path(WEIGHTS_NAME).stem) and file.endswith(Path(WEIGHTS_NAME).suffix)
+                for file in all_files
+            ]
```
michaelbenayoun, line 933: No need to define a list for `any`, using a generator is better because it is lazy.
Also I would defined both generators above instead than in the if statements to make things easier to read:

```python
weight_name = Path(WEIGHT_NAME).stem
weight_extension = Path(WEIGHT_NAME).suffix
is_pt_weight_file = (file.startswith(weight_name) and file.endswith(weight_extension) for file in all_files)
if any(is_pt_weight_file):
    ...
```

Same thing for TF

michaelbenayoun (approved):
Left a few nits, but LGTM besides that!

optimum/exporters/tasks.py
fxmarty: Thanks, this is indeed cleaner!

optimum/exporters/tasks.py
```diff
@@ -912,38 +912,37 @@ def determine_framework(
...
+        if any(is_pt_weight_file):
+            framework = "pt"
+        elif any(is_tf_weight_file):
+            framework = "tf"
+        else:
+            raise FileNotFoundError(
+                "Cannot determine framework from given checkpoint location."
+                f" There should be a {Path(WEIGHTS_NAME).stem}*{Path(WEIGHTS_NAME).suffix} for PyTorch"
+                f" or {Path(TF2_WEIGHTS_NAME).stem}*{Path(TF2_WEIGHTS_NAME).suffix} for TensorFlow."
+            )
```
regisss, line 945: I think we should log which framework has been chosen, something like:

```
logger.info(f"No framework specified so {framework} has been automatically chosen.")
```

Because here PT will have the priority over TF when there are checkpoints for both frameworks in the repo (which is fine), like BERT for instance: https://huggingface.co/bert-base-uncased/tree/main

fxmarty: This is already logged in the CLI: `Framework not specified. Using pt to export to ONNX.` If passing `export=True` this is indeed not logged.

PR #886, opened 2023-03-15
Automatic support for large ONNX models in ORTOptimizer

fxmarty:
Previously we had errors as

```
  File "/home/user/miniconda3/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/user/miniconda3/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/user/scripts/optimum/optimum/exporters/onnx/__main__.py", line 255, in <module>
    main()
  File "/home/user/scripts/optimum/optimum/exporters/onnx/__main__.py", line 212, in main
    optimizer.optimize(save_dir=args.output, optimization_config=optimization_config, file_suffix="")
  File "/home/user/scripts/optimum/optimum/onnxruntime/optimization.py", line 192, in optimize
    optimizer.save_model_to_file(output_path.as_posix(), use_external_data_format, one_external_file)
  File "/home/user/miniconda3/lib/python3.10/site-packages/onnxruntime/transformers/models/gpt2/../../onnx_model.py", line 994, in save_model_to_file
    OnnxModel.save(self.model, output_path, use_external_data_format, all_tensors_to_one_file)
  File "/home/user/miniconda3/lib/python3.10/site-packages/onnxruntime/transformers/models/gpt2/../../onnx_model.py", line 984, in save
    save_model(model, output_path)
  File "/home/user/miniconda3/lib/python3.10/site-packages/onnx/__init__.py", line 231, in save_model
    s = _serialize(proto)
  File "/home/user/miniconda3/lib/python3.10/site-packages/onnx/__init__.py", line 75, in _serialize
    raise ValueError(
ValueError: The proto size is larger than the 2 GB limit. Please use save_as_external_data to save tensors separately from the model file.
```

because the argument `use_external_data_format` was not automatically set automatically for `optimizer.save_model_to_file`. Here we use a hopefully decent heuristic that will help to avoid errors in the optimization & export for large models. One slight issue is that `optimizer.save_model_to_file` saves as `.data` the external data instead of `.onnx_data`, the convention we took in the ONNX export. I believe this is a non-issue, just style.

### DIFF
--- a/optimum/onnxruntime/optimization.py
+++ b/optimum/onnxruntime/optimization.py
@@ -46,18 +46,21 @@ class ORTOptimizer:
     Handles the ONNX Runtime optimization process for models shared on huggingface.co/models.
     """
 
-    def __init__(self, onnx_model_path: List[os.PathLike], config: "PretrainedConfig"):
+    def __init__(self, onnx_model_path: List[os.PathLike], config: "PretrainedConfig", from_ortmodel: bool = False):
         """
         Args:
             onnx_model_path (`List[os.PathLike]`):
                 The paths of the onnx models to optimize.
             config ([`~transformers.PretrainedConfig`]):
                 An instance of the configuration associated to the model to optimize.
+            from_ortmodel (`bool`, defaults to `False`):
+                Whether the model being optimized is already loaded into an ORTModel, or if it was passed from disk.
         """
         super().__init__()
         self.onnx_model_path = onnx_model_path
         self.config = config
         self.model_type = self.config.model_type
+        self.from_ortmodel = from_ortmodel
 
         try:
             self.normalized_config = NormalizedConfigManager.get_normalized_config_class(self.model_type)(self.config)
@@ -84,6 +87,7 @@ class ORTOptimizer:
         onnx_model_path = []
         config = None
         if isinstance(model_or_path, ORTModel):
+            from_ortmodel = True
             if isinstance(model_or_path, ORTModelForSeq2SeqLM):
                 onnx_model_path += [
                     model_or_path.encoder_model_path,
@@ -107,6 +111,7 @@ class ORTOptimizer:
                 onnx_model_path.append(model_or_path.model_path)
             config = model_or_path.config
         elif os.path.isdir(model_or_path):
+            from_ortmodel = False
             file_names = [ONNX_WEIGHTS_NAME] if file_names is None else file_names
             model_or_path = Path(model_or_path)
             if CONFIG_NAME not in os.listdir(model_or_path):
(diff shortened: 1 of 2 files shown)


optimum/onnxruntime/optimization.py
```diff
@@ -115,14 +122,14 @@ def from_pretrained(
...
-        return cls(onnx_model_path, config=config)
+        return cls(onnx_model_path, config=config, from_ortmodel=from_ortmodel)
 
     def optimize(
         self,
         optimization_config: OptimizationConfig,
         save_dir: Union[str, os.PathLike],
         file_suffix: Optional[str] = "optimized",
-        use_external_data_format: bool = False,
+        use_external_data_format: bool = None,
```
michaelbenayoun, line 132:
```suggestion
        use_external_data_format: Optional[bool] = None,
```

optimum/onnxruntime/optimization.py
```diff
@@ -136,31 +143,48 @@ def optimize(
             file_suffix (`str`, defaults to `"optimized"`):
                 The file suffix used to save the optimized model.
             use_external_data_format (`bool`, defaults to `False`):
-                Whether to use external data format to store model of size >= 2Gb.
+                Whether to use external data format to store model of size >= 2Gb. This argument is deprecated.
```
michaelbenayoun, line 146: It defaults to `None` now.

optimum/onnxruntime/optimization.py
```diff
@@ -189,15 +213,25 @@ def optimize(
...
+
+            # TODO: ORT save_model_to_file will save as `.data` although we save as `.onnx_data` in the export
+            optimizer.save_model_to_file(
+                output_path.as_posix(),
+                use_external_data_format=model_uses_external_data,
+                all_tensors_to_one_file=one_external_file,
+            )
+
+            # if loading from disk and saving in the same repository, remove previous external data
+            if Path(model_path.as_posix() + "_data").is_file() and self.from_ortmodel is False:
```
michaelbenayoun, line 225: The second part of the if statement is ugly IMO.

Also it does not read the same way, why not doing that then if the plan is to be very verbose?
```python
if Path(model_path.as_posix() + "_data").is_file() is True and self.from_ortmodel is False:
```

michaelbenayoun: Personally I'm in favor of a more pythonic way:
```python
if Path(model_path.as_posix() + "_data").is_file() is True and not self.from_ortmodel:
```

michaelbenayoun (approved):
Left a few nits, but LGTM!

optimum/onnxruntime/optimization.py
fxmarty: `is_file()` is defined in pathlib, and I expect pathlib to respect its signatures given in documentation. `self.from_ortmodel` may be passed / changed by the user/ourselves, and I don't expect the user to respect always signatures.

michaelbenayoun: Then I would also check that `model_path` is actually a `Path`, not ` os.PathLike` as annotated in the `__init__ before trying `model_path.as_posix()` because the user can pass this value. 

Anyway, just nitpicking but I truly think it's not readable. Feel free to ignore ;)

PR #888, opened 2023-03-16
Do not use scaled_dot_product_attention for stable diffusion onnx export

fxmarty:


optimum/exporters/onnx/utils.py
```diff
@@ -164,6 +167,7 @@ def get_stable_diffusion_models_for_export(
         model=pipeline.unet, exporter="onnx", task="semantic-segmentation", model_type="unet"
     )
     unet_onnx_config = onnx_config_constructor(pipeline.unet.config)
+    pipeline.unet.set_attn_processor(CrossAttnProcessor())
```
sayakpaul, line 170: Maybe accompany it with a comment as to why this explicit setup is needed?

echarlaix (approved):
Thanks a lot for taking care of this @fxmarty

fxmarty:
Failing tests are unrelated.

PR #891, opened 2023-03-16
Drop python 3.7 support

fxmarty:
As PyTorch did: https://github.com/pytorch/pytorch/releases/tag/v2.0.0

### DIFF
--- a/.github/workflows/test_bettertransformer.yml
+++ b/.github/workflows/test_bettertransformer.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


echarlaix:
TensorFlow, as well as Transformers and Diffusers, support python 3.7 so I don't think we should drop it

fxmarty:
@echarlaix yes right. Still think we should disable the CI on 3.7, as done in transformers. WDYT?

regisss:
> @echarlaix yes right. Still think we should disable the CI on 3.7, as done in transformers. WDYT?

Souns good to me

PR #895, opened 2023-03-17
Update bug-report.yml

lewtun:
As I'm no longer actively working on `optimum`, perhaps it makes sense to remove me from the default issue tags?

Of course, happy to help if something in particular is needed on my end!

# What does this PR do?

Fixes # (issue)

## Before submitting


regisss:
Actually we should update this list since we have TFLite also now. I'll push a commit to do it!

PR #898, opened 2023-03-17
Registering mechanism for the `TasksManager`

michaelbenayoun:
# What does this PR do?

Adds a registering mechanism for the `TasksManager`, allowing to dynamically register configs for any (backend, model, task) triplet.

### DIFF
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -54,14 +54,7 @@ def is_backend_available(backend):
     return backend_availablilty[backend]
 
 
-def make_backend_config_constructor_for_task(
-    backend: str, config_cls: Type, task: Union[str, Tuple[str, Tuple[str, ...]]]
-) -> Optional[ExportConfigConstructor]:
-    constructor = None
-    if isinstance(task, tuple):
-        task, supported_backends_for_task = task
-        if backend not in supported_backends_for_task:
-            constructor = None
+def make_backend_config_constructor_for_task(config_cls: Type, task: str) -> ExportConfigConstructor:
     if "-with-past" in task:
         if not hasattr(config_cls, "with_past"):
             raise ValueError(f"{config_cls} does not support tasks with past.")
@@ -109,9 +102,11 @@ def supported_tasks_mapping(
             )
             mapping[backend] = {}
             for task in supported_tasks:
-                config_constructor = make_backend_config_constructor_for_task(backend, config_cls, task)
-                if config_constructor is None:
-                    continue
+                if isinstance(task, tuple):
+                    task, supported_backends_for_task = task
+                    if backend not in supported_backends_for_task:
+                        continue
+                config_constructor = make_backend_config_constructor_for_task(config_cls, task)
                 mapping[backend][task] = config_constructor
     return mapping
 
@@ -843,7 +838,7 @@ class TasksManager:
                         )
                     if not overwrite_existing and task in mapping_backend:
                         continue
-                    mapping_backend[task] = make_backend_config_constructor_for_task(backend, config_cls, task)
+                    mapping_backend[task] = make_backend_config_constructor_for_task(config_cls, task)
                 mapping[backend] = mapping_backend
                 cls._SUPPORTED_MODEL_TYPE[model_type] = mapping
                 return config_cls
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -54,6 +54,23 @@ def is_backend_available(backend):
     return backend_availablilty[backend]
 
 
+def make_backend_config_constructor_for_task(
+    backend: str, config_cls: Type, task: Union[str, Tuple[str, Tuple[str, ...]]]
+) -> Optional[ExportConfigConstructor]:
+    constructor = None
+    if isinstance(task, tuple):
+        task, supported_backends_for_task = task
+        if backend not in supported_backends_for_task:
+            constructor = None
+    if "-with-past" in task:
+        if not hasattr(config_cls, "with_past"):
+            raise ValueError(f"{config_cls} does not support tasks with past.")
+        constructor = partial(config_cls.with_past, task=task.replace("-with-past", ""))
+    else:
+        constructor = partial(config_cls, task=task)
+    return constructor
+
+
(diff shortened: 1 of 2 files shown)


optimum/exporters/tasks.py
```diff
@@ -792,6 +804,54 @@ class TasksManager:
...
+            ```python
+            >>> register_for_new_backend = create_register("new-backend")
+
+            >>> @register_for_new_backend("bert", "sequence-classification", "token-classification")
+            >>> class BertNewBackendConfig(NewBackendConfig):
+            >>>     pass
+            ```
+        """
+
+        def wrapper(model_type: str, *supported_tasks: str) -> Callable[[Type], Type]:
```
JingyaHuang, line 834: Will the supported tasks differ a lot among different backend? It could be a long list if I register every task for every model in this manner.

tests/exporters/common/test_tasks_manager.py
```diff
@@ -63,3 +66,91 @@ def test_all_onnx_models_are_registered(self):
...
+        )
+
+        # Case 3: Registering an unknown task.
+        with self.assertRaisesRegex(ValueError, "The TasksManager does not know the task called"):
+
+            @register_for_onnx("bert", "this is a wrong name for a task")
+            class UnknownTask(BertOnnxConfig):
+                pass
+
+        # Case 4: Registering for a new backend.
```
JingyaHuang, line 122: For unknown backend, will that really work as it is not validated by `is_backend_available`:
https://github.com/huggingface/optimum/blob/7ecc5cac5dd16e65bfdb63e54a8751fc3ac522f7/optimum/exporters/tasks.py#L49-L54

JingyaHuang:
And another question about the usage. Each time we use the TasksManager with a new backend, we must register it again.  
So to say to use a registered TaskManager for the CLI, we need to do the registration in the CLI script.   And if we want to use it elsewhere, we will need to do the process again? If so, does it make more sense to have an extended `TasksManager` for subpackages?

optimum/exporters/tasks.py
```diff
@@ -74,7 +91,7 @@ def supported_tasks_mapping(
         exporters (`Dict[str, str]`):
             The export backend name -> config class name mapping. For instance:
             ```python
-            >>> kwargs = {  # doctest: +SKIP
+            >>> exporters = {  # doctest: +SKIP
```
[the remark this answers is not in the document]
michaelbenayoun: It's the name of the argument, it has been like that for a while.

optimum/exporters/tasks.py
michaelbenayoun: Maybe.
It will be like that:

```python
@register("bert", "sequence-classification", "token-classification", "multiple-choice", "question-answering")
class BertNeuronConfig(...):
...
```

But you can also define global variables and re-use them:
```python
COMMON_TEXT_TASKS = ("sequence-classification", "token-classification", "multiple-choice", "question-answering")

@register("bert", *COMMON_TEXT_TAKS)
class BertNeuronConfig(...):
...
```

tests/exporters/common/test_tasks_manager.py
michaelbenayoun: For a custom backend we cannot check for sure. But since to register it you need to import the classes, it actually means that the subpackage is installed.

michaelbenayoun:
I do not understand your question. You need to register the config once, then it is usable everywhere.

fxmarty:
For the CLI, I think it may be enough to have a check for `optimum-neuron` being install. The neuron commands may be in optimum main repo, but since it's a simple call to e.g. `main_export` from optimum-neuron and all the code is deported there, it is imo fine to have minimal code referencing it in optimum main.

For the ONNX export, it is as light as: https://github.com/fxmarty/optimum/blob/avoid-subprocess-calls-onnx-export/optimum/commands/export/onnx.py

For this PR, it feels like `@register` decorator makes things less readable than having a clear dict of the supported tasks. I also wonder why we could not have different TasksManager for different exporter? And shared methods at the top level?

michaelbenayoun:
We will not have anything in the main Optimum repo most likely. Just as we have a registering mechanism for the `TasksManager` we can do the same thing for the CLI.

The register method is intendend to be used by external code (= other people, and other subpackages). So that is why we register on a case by case basis. The whole goal of the TasksManager was to centralize what is needed for each exporters. And we do not need different ones for now, I do not see any issue as it works perfectly well with already 2 exporters.

michaelbenayoun:
> > Just as we have a registering mechanism for the TasksManager we can do the same thing for the CLI
> 
> What's the point if an if/else and check if optimum-neuron is installed is enough? How would this register mechanism look like?
> 
> > The whole goal of the TasksManager was to centralize what is needed for each exporters
> 
> But there is no point sharing dict as `_SUPPORTED_MODEL_TYPE` if this is only usable for a fraction of the exporters. I don't see the issue with having shared methods in the TasksManager and e.g. inherited subclasses for each exporter defining what is supported and what is not, with this code residing in subpackages as optimum-neuron.
> 
> I don't understand the need to add an extra layer of complexity with decorators, partial functions and such. If possible, it would make sense to me to keep it simple: you look at the code, you know what is supported and what is not, you don't have to understand what decorators do to understand the code.
> 
> It feels like we are overengineering something that does not need to be overengineered.

This allows anyone to extend the code as it pleases. It is also not complex at all. It is a register mechanism, like many exist. 
Anytime you have some factory (which the `TasksManager` kinda is), it is nice to be able to dynamically extends the code.

Example in our ecosystem: the "auto" classes in Transformers all support registering. [For example `AutoModel`](https://huggingface.co/docs/transformers/v4.27.1/en/model_doc/auto#extending-the-auto-classes).

fxmarty:
The difference with the Auto registering in transformers is that the Auto is designed for external users to extend it - here we are using a registering mechanic for an officially supported feature in Optimum, and making what we support explicit / less hacky is IMO better.

michaelbenayoun:
Both are orthogonal, this PR is adding a registering mechanism. About us making things explicit is something else. 
We could imagine having a `NeuronTasksManager` and still have this registering feature.

We can still make things explicit, would a method "get_supported_models_and_tasks_for_backend" help?

Issue #905, opened 2023-03-21
Whisper pipeline raises error when using `return_timestamps` (ValueError: The following `model_kwargs` are not used by the model: ['return_timestamps'])


xenova:
### System Info

```shell
optimum: 1.7.1
Python: 3.8.3
transformers: 4.27.2
platform: Windows 10
```

### Who can help?

@philschmid 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction

This is a working example using the transformers pipeline function:
```python
from transformers import pipeline

transcriber = pipeline('automatic-speech-recognition', 'openai/whisper-tiny.en')

text = transcriber(
    'https://xenova.github.io/transformers.js/assets/audio/ted_60.wav',

    return_timestamps=True,
    chunk_length_s=30,
    stride_length_s=5
)

print(f'{text=}')
# outputs correctly
```

After converting to onnx using this command:
```
python -m optimum.exporters.onnx --model openai/whisper-tiny.en whisper_onnx/
```

and running the equivalent code:
```python

import onnxruntime
from transformers import pipeline, AutoProcessor
from optimum.onnxruntime import ORTModelForSpeechSeq2Seq

whisper_model_name = './whisper_onnx/'
processor = AutoProcessor.from_pretrained(whisper_model_name)
session_options = onnxruntime.SessionOptions()

model_ort = ORTModelForSpeechSeq2Seq.from_pretrained(
    whisper_model_name,
    use_io_binding=True,
    session_options=session_options
)
generator_ort = pipeline(
    task="automatic-speech-recognition",
    model=model_ort,
    feature_extractor=processor.feature_extractor,
    tokenizer=processor.tokenizer,
)

out = generator_ort(
    'https://xenova.github.io/transformers.js/assets/audio/ted_60.wav',

    return_timestamps=True,
    chunk_length_s=30,
    stride_length_s=5
)

print(f'{out=}')
```

I get the error:
```
ValueError: The following `model_kwargs` are not used by the model: ['return_timestamps'] (note: typos in the generate arguments will also show up in this     
list)
```

### Expected behavior

The code which uses the ONNX model should work the same as the transformers version (and not throw an error).

fxmarty:
@xenova Thank you, will look into it asap!

Issue #906, opened 2023-03-21
Optimum export of whisper raises ValueError: There was an error while processing timestamps, we haven't found a timestamp as last token. Was WhisperTimeStampLogitsProcessor used?


xenova:
### System Info

```shell
optimum: 1.7.1
Python: 3.8.3
transformers: 4.27.2
platform: Windows 10
```

### Who can help?

@philschmid @michaelbenayoun 

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

1. Convert the model to ONNX:
```
python -m optimum.exporters.onnx --model openai/whisper-tiny.en whisper_onnx/
```

2. Due to [another bug in the pipeline function](https://github.com/huggingface/optimum/issues/905), you may need to comment out the lines in the generate function which raises an error for unused model kwargs:

https://github.com/huggingface/transformers/blob/48327c57182fdade7f7797d1eaad2d166de5c55b/src/transformers/generation/utils.py#L1104-L1108

3. Try to transcribe longer audio clip:
```python

import onnxruntime
from transformers import pipeline, AutoProcessor
from optimum.onnxruntime import ORTModelForSpeechSeq2Seq

whisper_model_name = './whisper_onnx/'
processor = AutoProcessor.from_pretrained(whisper_model_name)
session_options = onnxruntime.SessionOptions()

model_ort = ORTModelForSpeechSeq2Seq.from_pretrained(
    whisper_model_name,
    use_io_binding=True,
    session_options=session_options
)
generator_ort = pipeline(
    task="automatic-speech-recognition",
    model=model_ort,
    feature_extractor=processor.feature_extractor,
    tokenizer=processor.tokenizer,
)

out = generator_ort(
    'https://xenova.github.io/transformers.js/assets/audio/ted_60.wav',

    return_timestamps=True,
    chunk_length_s=30,
    stride_length_s=5
)

print(f'{out=}')
```

4. This raises the error:
```python
│   878 │   │   if return_timestamps:                                                              │
│   879 │   │   │   # Last token should always be timestamps, so there shouldn't be                │
│   880 │   │   │   # leftover                                                                     │
│ ❱ 881 │   │   │   raise ValueError(                                                              │
│   882 │   │   │   │   "There was an error while processing timestamps, we haven't found a time   │
[...]

PR #919, opened 2023-03-24
Fix ORTModel MRO for whisper

fxmarty:
As per title, fixes https://github.com/huggingface/optimum/issues/905, https://github.com/huggingface/optimum/issues/906

This bug is introduced by https://github.com/huggingface/transformers/pull/21252 that moves `generate()` implementation in `WhisperForConditionalGeneration`, hence breaking the inheritance order in transformers.

### DIFF
--- a/optimum/exporters/onnx/__main__.py
+++ b/optimum/exporters/onnx/__main__.py
@@ -405,6 +405,8 @@ def main_export(
 
         # Saving the model config and preprocessor as this is needed sometimes.
         model.config.save_pretrained(output)
+        if hasattr(model, "generation_config"):
+            model.generation_config.save_pretrained(output)
         maybe_save_preprocessors(model_name_or_path, output)
 
     if task == "stable-diffusion":
--- a/optimum/exporters/onnx/__main__.py
+++ b/optimum/exporters/onnx/__main__.py
@@ -405,7 +405,7 @@ def main_export(
 
         # Saving the model config and preprocessor as this is needed sometimes.
         model.config.save_pretrained(output)
-        if hasattr(model, "generation_config"):
+        if hasattr(model, "generation_config") and model.generation_config is not None:
             model.generation_config.save_pretrained(output)
         maybe_save_preprocessors(model_name_or_path, output)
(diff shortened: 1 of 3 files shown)


optimum/exporters/onnx/__main__.py
```diff
@@ -405,6 +405,8 @@ def main_export(
 
         # Saving the model config and preprocessor as this is needed sometimes.
         model.config.save_pretrained(output)
+        if hasattr(model, "generation_config") and model.generation_config is not None:
+            model.generation_config.save_pretrained(output)
```
michaelbenayoun, line 409: Quick tip: I usually use `getattr` in this situation because I do both checks and access at the same time.
```suggestion
        generation_config = getattr(model, "generation_config")
        if generation_config is not None:
            generation_config.save_pretrained(output)
```

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -1046,6 +1054,53 @@ def can_generate(self):
...
+            + (WhisperForConditionalGeneration,)
+            + ORTModelForSpeechSeq2Seq.__mro__[top_inheritance_index:]
+        )
+
+
+class _ORTModelForWhisper(
+    ORTModelForSpeechSeq2Seq, WhisperForConditionalGeneration, metaclass=MetaClassRemoveParentsAndReorder
+):
+    """
+    Whisper implements its own generate() method
```
michaelbenayoun, line 1089:
```suggestion
    Whisper implements its own generate() method.
```

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -1046,6 +1054,53 @@ def can_generate(self):
...
+    """
+
+    def generate(self, **kwargs):
+        return super().generate(**kwargs)
```
michaelbenayoun, line 1093: Not needed no?

fxmarty: yes

PR #927, opened 2023-03-27
`optimun-cli onnxruntime quantize / optimize` output argument is now required

michaelbenayoun:
# What does this PR do?

This PR makes the `-o` / `--output` argument for `optimum-cli onnxruntime optimize / quantize` mandatory and does not allow it to be the same as the original directory.

Still wondering whether we should raise an error or just a warning.
While saving everything under the same directory is completely working here, it makes it unusable with `ORTModel`s because multiple ONNX files are in the directory. While we could provide the filename explicitly, I feel like keeping a easy to use workflow is better.


michaelbenayoun:
I think it's better to enforce it for now, for instance by creating `_optimized` directory, we would expect without looking that we can use the same directory, but no. Can be reverted if needed.

PR #928, opened 2023-03-28
Register mechanism for the Optimum CLI

michaelbenayoun:
# What does this PR do?

This PR both refactors how the Optimum CLI was implemented and provides a way to register commands to the Optimum CLI for subpackages.

## Copy-pasting `optimum/commands/register/README.md` to explain how it works:

It is possible to register a command in the Optimum CLI, either as a command or a subcommand of an already existing command.

Steps to follow:

1. Create a command as a subclass of `optimum.commands.BaseOptimumCLICommand`.
2. Create a Python file under `optimum/commands/register/`, and define a `REGISTER_COMMANDS` list variable there.
3. Fill the `REGISTER_COMMANDS` as folows:

```python
# CustomCommand1 and CustomCommand2 could also be defined in this file actually.
from ..my_custom_commands import CustomCommand1, CustomCommand2
from ..export import ExportCommand
REGISTER_COMMANDS = [
  # CustomCommand1 will be registered as a subcommand of the root Optimum CLI. 
  CustomCommand1, 
  # CustomCommand2 will be registered as a subcommand of the `optimum-cli export` command. 
  (CustomCommand2, ExportCommand) # CustomCommand2 will be registered
]

### DIFF
--- a/optimum/commands/base.py
+++ b/optimum/commands/base.py
@@ -74,12 +74,10 @@ class BaseOptimumCLICommand(ABC):
                 raise ValueError(f"Subcommands must be instances of CommandInfo, but got {type(subcommand)} here.")
             self.register_subcommand(subcommand)
 
-
         self.args = args
 
     @property
     def defaults_factory(self):
-        
         def defaults_factory_func(args):
             return self.__class__(self.subparsers, args, command=self.COMMAND, from_defaults_factory=True)
 
@@ -95,7 +93,7 @@ class BaseOptimumCLICommand(ABC):
                     self.parser.set_defaults(func=self.defaults_factory)
             else:
                 self._subparsers = None
-        return self._subparsers 
+        return self._subparsers
 
     @subparsers.setter
     def subparsers(self, subparsers: Optional["_SubParsersAction"]):
@@ -112,7 +110,6 @@ class BaseOptimumCLICommand(ABC):
         pass
 
     def register_subcommand(self, command_info: CommandInfo):
-
         command_info.is_subcommand_info_or_raise()
         self.SUBCOMMANDS = self.SUBCOMMANDS + (command_info,)
         self.registered_subcommands.append(command_info.subcommand_class(self.subparsers, command=command_info))
@@ -124,12 +121,7 @@ class BaseOptimumCLICommand(ABC):
 class RootOptimumCLICommand(BaseOptimumCLICommand):
     COMMAND = CommandInfo(name="root", help="optimum-cli root command")
 
-    def __init__(
-        self,
-        cli_name: str,
-        usage: Optional[str] = None,
-        args: Optional["Namespace"] = None
-    ):
+    def __init__(self, cli_name: str, usage: Optional[str] = None, args: Optional["Namespace"] = None):
         self.parser = ArgumentParser(cli_name, usage=usage)
         self.subparsers = self.parser.add_subparsers()
         self.args = None
--- a/optimum/commands/base.py
+++ b/optimum/commands/base.py
@@ -55,6 +55,21 @@ class BaseOptimumCLICommand(ABC):
         command: Optional[CommandInfo] = None,
         from_defaults_factory: bool = False,
     ):
+        """
+        Initializes the instance.
+
+        Args:
+            subparsers (`Optional[_SubParsersAction]`):
+                The parent subparsers this command will create its parser on.
+            args (`Optional[Namespace]`, defaults to `None`):
+                The arguments that are going to be parsed by the CLI.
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

**Review by fxmarty — APPROVED** [review]
A bit involved but why not - adding commands is still simple which is nice. Thank you for the refactor!

Out of curiosity - is this registering done ahead of time or on the fly? In which cases it may involve a lot of logic and may add an overhead when running a command (which is already a bit slow), have you tested that the change is not overly slow?

File: optimum/commands/base.py [hunk 1]
```diff
@@ -0,0 +1,129 @@
...
+        If `self` does not have any subcommand, no subparsers should be created or it will mess with the command.
+        This property ensures that we create subparsers only if needed.
+        """
+        subparsers = getattr(self, "_subparsers", None)
+        if subparsers is None:
+            if self.SUBCOMMANDS:
+                if self.parser is not None:
+                    self._subparsers = self.parser.add_subparsers()
+                else:
+                    self._subparsers = None
```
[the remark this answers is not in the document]
michaelbenayoun: It is a bit of an overkill optimization, but we create the object also with the `set_defaults` method, in which case we do not want to recreate a parser that is not needed. Check [here](https://github.com/huggingface/optimum/pull/928/files#diff-b3d991129997f88806266d32867d33d9f135ce90e77c4dfd3fe789cab72ac62dR61-R73) for more details.

tests/cli/test_cli.py
```diff
@@ -77,3 +92,29 @@ def test_quantize_commands(self):
...
+        succeeded = self._run_command_and_check_content(custom_command, command_content)
+        self.assertTrue(succeeded, "The command should succeed here since it is registered.")
+        REGISTERED_CLI_WITH_CUSTOM_COMMAND_PATH.unlink()
+
+        shutil.copy(CLI_WIH_CUSTOM_COMMAND_PATH, REGISTERED_CLI_WITH_CUSTOM_COMMAND_PATH)
+        os.environ["TEST_REGISTER_COMMAND_WITH_SUBCOMMAND"] = "true"
+        custom_command = "optimum-cli export blablabla"
+        succeeded = self._run_command_and_check_content(custom_command, command_content)
+        self.assertTrue(succeeded, "The command should succeed here since it is registered.")
+        REGISTERED_CLI_WITH_CUSTOM_COMMAND_PATH.unlink()
```
JingyaHuang, line 120: What's the difference of these two tests. Why do we need both?

optimum/commands/register/README.md
```diff
@@ -0,0 +1,22 @@
+# Register commands in the Optimum CLI from a subpackage
+
+It is possible to register a command in the Optimum CLI, either as a command or a subcommand of an already existing command.
+
+Steps to follow:
+
+1. Create a command as a subclass of `optimum.commands.BaseOptimumCLICommand`.
+2. Create a Python file under `optimum/commands/register/`, and define a `REGISTER_COMMANDS` list variable there.
+3. Fill the `REGISTER_COMMANDS` as folows:
```
JingyaHuang, line 9:
```suggestion
3. Fill the `REGISTER_COMMANDS` as follows:
```

tests/cli/test_cli.py
JingyaHuang: Ok, I see...

optimum/commands/base.py
```diff
@@ -0,0 +1,144 @@
...
+    @property
+    def is_subcommand_info(self):
+        return self.subcommand_class is not None
+
+    def is_subcommand_info_or_raise(self):
+        if not self.is_subcommand_info:
+            raise ValueError(f"The command info must define a subcommand_class attribute, but got: {self}.")
+
+
+class InvalidCLICommand(Exception):
```
JingyaHuang, line 43: When and how this class is used?

michaelbenayoun: It was used before but not anymore, removed it.

optimum/commands/base.py
```diff
@@ -0,0 +1,129 @@
...
+class BaseOptimumCLICommand(ABC):
+    COMMAND: CommandInfo
+    SUBCOMMANDS: Tuple[CommandInfo, ...] = ()
+
+    def __init__(
+        self,
+        subparsers: Optional["_SubParsersAction"],
+        args: Optional["Namespace"] = None,
+        command: Optional[CommandInfo] = None,
+        from_defaults_factory: bool = False,
```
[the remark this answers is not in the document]
fxmarty: `from_defaults_factory: bool = False,`

fxmarty:
> It cannot be done not on the fly for commands outside main optimum. It should not add a big overhead though.

And in main optimum? Is it on the fly?

PR #930, opened 2023-03-28
Support merged seq2seq models in ORTModel

fxmarty:
Up to now we were supporting only decoder-only merged decoders, not for encoder-decoder architectures.

### DIFF
--- a/optimum/onnxruntime/base.py
+++ b/optimum/onnxruntime/base.py
@@ -177,6 +177,7 @@ class ORTDecoder(ORTModelPart):
             ):
                 self.value_sequence_length_idx = -1
 
+    # TODO: this likely fails for seq2seq models
     def prepare_inputs_for_merged(
         self, input_ids: Optional[torch.LongTensor] = None, past_key_values: Optional[List[torch.FloatTensor]] = None
     ):
@@ -403,7 +404,11 @@ class ORTDecoderForSeq2Seq(ORTDecoder):
             self.num_pkv = 4
 
     def compute_past_key_values_output_shapes(
-        self, input_ids, encoder_hidden_states, past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None
+        self,
+        input_ids: torch.Tensor,
+        encoder_hidden_states: torch.Tensor,
+        use_cache_branch: Optional[bool],
+        past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None,
     ) -> Dict[str, int]:
         batch_size = input_ids.size(0)
         num_attention_heads = self.normalized_config.num_attention_heads
@@ -411,7 +416,9 @@ class ORTDecoderForSeq2Seq(ORTDecoder):
 
         sequence_length = input_ids.size(1)
         encoder_sequence_length = encoder_hidden_states.size(1)
-        if past_key_values is not None:
+        if past_key_values is not None and use_cache_branch is not False:
+            # Here, use_cache_branch may be None in the case of separate decoder without/with past, or True if the with past branch
+            # of a merged decoder is used
             sequence_length += past_key_values[0].size(2)
 
         self_attn_shape = (batch_size, num_attention_heads, sequence_length, embed_size_per_head)
@@ -431,6 +438,7 @@ class ORTDecoderForSeq2Seq(ORTDecoder):
         encoder_attention_mask: Optional[torch.LongTensor] = None,
         past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None,
         labels: Optional[torch.LongTensor] = None,
+        use_cache_branch: None = None,
     ) -> Seq2SeqLMOutput:
         use_torch = isinstance(input_ids, torch.Tensor)
         self.parent_model.raise_on_numpy_input_io_binding(use_torch)
@@ -444,6 +452,7 @@ class ORTDecoderForSeq2Seq(ORTDecoder):
             known_output_shapes = self.compute_past_key_values_output_shapes(
                 input_ids,
                 encoder_hidden_states,
+                use_cache_branch=use_cache_branch.item() if use_cache_branch is not None else None,
                 past_key_values=past_key_values,
             )
 
@@ -462,6 +471,9 @@ class ORTDecoderForSeq2Seq(ORTDecoder):
 
             if past_key_values is not None:
                 model_inputs += past_key_values
+
+            if use_cache_branch is not None:
+                model_inputs.append(use_cache_branch)
 
             if "labels" in self.input_names:
                 model_inputs.append(labels)
@@ -517,6 +529,9 @@ class ORTDecoderForSeq2Seq(ORTDecoder):
                 if "labels" in self.input_names:
                     # TODO: Any preprocessing like  `self._shift_right(labels)`?
                     onnx_inputs["labels"] = labels.cpu().detach().numpy()
+
+                if self.parent_model.use_merged is True:
+                    onnx_inputs["use_cache_branch"] = use_cache_branch.cpu().detach().numpy()
             else:
                 onnx_inputs = {
                     "input_ids": input_ids,
@@ -538,6 +553,9 @@ class ORTDecoderForSeq2Seq(ORTDecoder):
                 if "labels" in self.input_names:
                     # TODO: Any preprocessing like  `self._shift_right(labels)`?
                     onnx_inputs["labels"] = labels
+
+                if self.parent_model.use_merged is True:
+                    onnx_inputs["use_cache_branch"] = use_cache_branch
 
             # Run inference
             outputs = self.session.run(None, onnx_inputs)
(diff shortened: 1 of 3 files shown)


optimum/onnxruntime/base.py
```diff
@@ -178,21 +179,28 @@ def __init__(
                 self.value_sequence_length_idx = -1
 
     def prepare_inputs_for_merged(
-        self, input_ids: Optional[torch.LongTensor] = None, past_key_values: Optional[List[torch.FloatTensor]] = None
+        self,
+        input_ids: Union[None, torch.LongTensor, np.ndarray] = None,
+        past_key_values: Optional[Union[Tuple[torch.FloatTensor], Tuple[np.ndarray]]] = None,
```
michaelbenayoun, line 184: IMO make a choice here `Union[None, ...]` then `Optional`.

fxmarty:
This seem to introduce a large latency overhead - at least with whisper on CUDA EP + IO Binding. Have to check.

optimum/onnxruntime/base.py
```diff
@@ -178,21 +179,28 @@ def __init__(
...
+        use_torch: bool = False,
     ):
+        constructor = torch if use_torch is True else np
```
michaelbenayoun, line 187: :)

optimum/onnxruntime/base.py
```diff
@@ -178,21 +179,28 @@ def __init__(
...
         if past_key_values is None and self.parent_model.use_merged:
             # Uses "no past" branch of a merged decoder
-            use_cache_branch = torch.full((1,), False).to(self.device)
+            use_cache_branch = constructor.full((1,), False)
         elif past_key_values is not None and self.parent_model.use_merged:
             # Uses "with past" branch of a merged decoder
-            use_cache_branch = torch.full((1,), True).to(self.device)
+            use_cache_branch = constructor.full((1,), True)
```
michaelbenayoun, line 193: Can we merge this and make the second argument equal to `past_key_values is not None`?

optimum/onnxruntime/base.py
```diff
@@ -397,31 +418,44 @@ def __init__(
     ):
         super().__init__(session, parent_model)
 
-        if self.use_past:
+        if self.parent_model.use_merged is False and self.use_past:
```
michaelbenayoun, line 421: Same make a choice here

optimum/onnxruntime/base.py
```diff
@@ -397,31 +418,44 @@ def __init__(
...
             self.num_pkv = 2
         else:
+            # when using a merged model, we always have the same number of output whether we use past key values or not,
```
michaelbenayoun, line 424:
```suggestion
            # When using a merged model, we always have the same number of output whether we use past key values or not,
```

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -365,30 +367,30 @@ def _from_pretrained(
...
-                    use_auth_token=use_auth_token,
-                    revision=revision,
-                )
-            except FileNotFoundError as e:
-                if use_cache is True and use_merged is False:
-                    raise FileNotFoundError(
-                        "The parameter `use_cache=True` was passed to ORTModelForCausalLM.from_pretrained()"
-                        " but no ONNX file using past key values could be found in"
-                        f" {str(Path(model_id, subfolder))}, with the error: {e}"
+        if use_cache is True and use_merged is False:
```
michaelbenayoun, line 370: :'(

optimum/onnxruntime/base.py
fxmarty: yes good catch!

optimum/onnxruntime/base.py
fxmarty: thanks

optimum/onnxruntime/base.py
fxmarty: Do you mean `Optional[Union[None, Tuple[torch.FloatTensor], Tuple[np.ndarray]]]`? What's the difference?

fxmarty: Oh nvm got it, you were referencing the two type annotations!

optimum/onnxruntime/base.py
```diff
@@ -401,64 +419,103 @@ def __init__(
...
 
-        self.legacy_outputs = any(
+        self.past_key_values_cross_attention_output_names = set()
+        for output_name in self.output_names:
+            if output_name.startswith("present") and "encoder" in output_name:
+                self.past_key_values_cross_attention_output_names.add(output_name)
+
+        self.use_legacy_outputs = self.parent_model.use_merged is False and any(
             "encoder" in output_name and "present" in output_name for output_name in self.output_names
         )
```
michaelbenayoun, line 437: Instead of rechecking here with the `any(....`), you could just do:

```python
self.use_legacy_outputs = not self.parent_model.use_merged and self.past_key_values_cross_attention_output_names
```

optimum/onnxruntime/base.py
```diff
@@ -401,64 +419,103 @@ def __init__(
...
 
         past_key_values_shapes = {}
         for idx, name in enumerate(self.key_value_output_names):
             is_self_attn = idx % 4 < 2
             # decoder with past does not ouput cross attention key/values as they are constants
-            past_key_values_shapes[name] = self_attn_shape if (is_self_attn or self.use_past) else cross_attn_shape
+            past_key_values_shapes[name] = self_attn_shape if (is_self_attn or self.num_pkv == 2) else cross_attn_shape
         return past_key_values_shapes
 
+    def get_outputs_not_to_bind(self, use_merged_cache):
```
michaelbenayoun, line 471:
```suggestion
    def get_outputs_not_to_bind(self, use_merged_cache: bool) -> Set[str]:
```

optimum/onnxruntime/base.py
```diff
@@ -401,64 +419,103 @@ def __init__(
...
+        result = {
+            output_name
+            for output_name in self.output_names
+            if (not output_name.startswith("present") and output_name not in {"loss", "logits"})
+        }
+        if use_merged_cache is True:
+            # When using a merged decoder and the use cache branch, we output 0-dim tensors that IO Binding do not support.
+            # Therefore, we do not bind them.
+            result = result.union(self.past_key_values_cross_attention_output_names)
```
michaelbenayoun, line 480: FYI you also have the `|` operator for the union:
```suggestion
            result = result | self.past_key_values_cross_attention_output_names
```

optimum/onnxruntime/base.py
```diff
@@ -401,64 +419,103 @@ def __init__(
...
+        return result
+
     def forward(
         self,
         input_ids: torch.LongTensor,
         encoder_hidden_states: torch.FloatTensor,
         encoder_attention_mask: Optional[torch.LongTensor] = None,
         past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None,
         labels: Optional[torch.LongTensor] = None,
+        use_cache_branch: None = None,
```
michaelbenayoun, line 490:
```suggestion
        use_cache_branch: Optional[bool] = None,
```

optimum/onnxruntime/base.py
```diff
@@ -401,64 +419,103 @@ def __init__(
...
     ) -> Seq2SeqLMOutput:
+        # adding use_cache_branch in the signature here is just a hack for IO Binding
```
michaelbenayoun, line 492:
```suggestion
        # Adding use_cache_branch in the signature here is just a hack for IO Binding
```

optimum/onnxruntime/base.py
```diff
@@ -496,13 +556,49 @@ def filter_out_output(output_name):
...
+                        out_past_key_values[i : i + self.num_pkv]
+                        + past_key_values[2 * i + 2 : 2 * i + 2 + self.num_pkv]
+                        for i in range(0, len(out_past_key_values), self.num_pkv)
+                    )
+                elif self.num_pkv == 4:
+                    # despite num_pkv being 4, we did not bind the cross-attention output
+                    out_past_key_values = tuple(
+                        out_past_key_values[i : i + 2] + past_key_values[2 * i + 2 : 2 * i + 4]
+                        for i in range(0, len(out_past_key_values), 2)
+                    )
```
michaelbenayoun, line 599: I feel like it is hard to read. But I guess we can keep it like that.

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -601,6 +633,89 @@ def _from_pretrained(
     ):
         model_path = Path(model_id)
 
+        decoder_merged_path = None
+        if use_merged is not False:
```
michaelbenayoun, line 637: Maybe specify that `is not False` does not mean `is True` since it can also be `None`.

michaelbenayoun (approved):
LGTM

optimum/onnxruntime/base.py
fxmarty: No it is really `None`

PR #587, opened 2022-12-14
Merge ONNX decoder models

JingyaHuang:
# What does this PR do?

- [x] Merge `decoder` and `decoder_with_past` in a single model proto
- [x] Check what opset to use
- [x] Apply ONNXSIM to simplify subgraphs
- [x] Cross-subgraphs initializers deduplication for the merged model

Next steps:

- [ ] Need to check if it works for large proto, and how shall we export it(but the support will be added in another PR)
- [ ] Need to adapt ORTModels to be able to take benefits(New input `use_cache`, dummy inputs for `past_key_values`...)
- [ ] Need to check if merged ONNX model works for IOBinding (introduces new input `use_cache`, but dlpack doesn't support `dtype=bool` ) 
- [ ] Open discussion: Should it be applied automatically after export or before inference? (Guess no, optimizer won't be able to due with cross-subgraphs optimization)

Fixes #530 

--------------------------------------

Snippets to test the merge:

```python
# Load decoders
decoder = onnx.load(decoder_path)
decoder_with_past = onnx.load(decoder_with_past_path)

merged_model = merge_decoders(decoder, decoder_with_past)
onnx.save(merged_model, merged_model_path)
```

And test inference:
```python
from transformers import AutoConfig, AutoTokenizer
from optimum.onnxruntime import ORTModelForCustomTasks

model_id = "gpt2"
tokenizer = AutoTokenizer.from_pretrained(model_id)
payload = tokenizer("I love burritos!", return_tensors="pt")
onnx_inputs = {
    "input_ids": payload.input_ids.cpu().detach().numpy(),
    "attention_mask": payload.attention_mask.cpu().detach().numpy(),
    "use_cache": np.full(1, True),
}
shape = (1, 12, 0, 64)
for i in range(24):
    onnx_inputs[f"past_key_values_{i}"] = np.random.uniform(low=0.5, high=13.3, size=shape).astype(np.float32)

sess = ort.InferenceSession(merged_model_path, providers=["CPUExecutionProvider"])
pred = sess.run(None, onnx_inputs)
```

--------------------------------------
After the merge, the merged decoder shall have almost the same size as decoder or decoder with past

* GPT2
image

* T5 small

image

PR #932, opened 2023-03-29
Hash tensor data during deduplication

VikParuchuri:
# What does this PR do?

When exporting a causal lm model to onnx, the past key version and regular version are merged.  The merge loops through each tensor in the graph and stores the values as dict keys.  This requires massive memory (>100GB for an 11GB model), and prevents merging from happening.

This is due to the `tensor_data` in `_find_duplicate_initializers` being converted from `bytes` to a `tuple`.  I assume there is a good reason to convert it to a tuple (maybe for comparison across dtypes).

As such, this PR doesn't remove that conversion, but instead converts the tuple to a string and hashes it.  It uses SHA-512 to reduce the probability of a hash collision.  This uses very little memory to do the deduplication.

You could also:
-  just convert to `str` to remove the possibility of collisions - this will take more memory than hashing
- remove the initial `tuple` conversion entirely - also more memory than hashing
- Trade off memory for compute - loop through each pair of tensors to compare them - much slower than hashing

I have done some basic manual testing, and this method seems to work.

## Before submitting

### DIFF
--- a/optimum/onnx/transformations_utils.py
+++ b/optimum/onnx/transformations_utils.py
@@ -14,6 +14,7 @@
 
 from collections import defaultdict
 from typing import DefaultDict, Dict, List, Set, Tuple
+import hashlib
 
 import numpy as np
 
@@ -29,7 +30,7 @@ logger.setLevel(logging.INFO)
 
 def _find_duplicate_initializers(
     models: List[ModelProto],
-) -> DefaultDict[Tuple[int, bytes, Tuple], Set[Tuple[str, int]]]:
+) -> DefaultDict[Tuple[int, str, Tuple], Set[Tuple[str, int]]]:
     """
     Creates a map (unique data) --> set of (initializer name, model id)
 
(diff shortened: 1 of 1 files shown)


fxmarty:
Thanks a lot for the PR, it's an issue I've had with large models, they surely will help!

fxmarty:
@VikParuchuri This step `str(tensor_data).encode("utf-8")` is quite slow. I wonder if there is anything better we could hash than this?

Hashing simply `tensor_data` does not seem to work, getting `object supporting the buffer API required`. This is kind of surprising as it's of type `<class 'bytes'>`, and what you get after the encode is also a byte? Any idea @michaelbenayoun?

The reason we put that in a tuple is that some initializers have names as `onnx::MatMul_3741`, that may be different between the two graphs. So we need to look into the actual data to check whether they are the same.

optimum/onnx/transformations_utils.py
```diff
@@ -45,13 +46,19 @@ def _find_duplicate_initializers(
                     if tensor_data:
                         tensor_data = tuple(tensor_data)
                         break
-                duplicates[(initializer.data_type, tensor_data, tensor_dims)].add((initializer.name, i))
+
+                # Hash tensor data to avoid storing large amounts of data in memory
+                hashed = hashlib.sha512()
+                hashed.update(str(tensor_data).encode("utf-8"))
```
fxmarty, line 52: Here's a proposal that seem to work fine and is much faster:
```suggestion
                hashed.update(numpy_helper.to_array(initializer))
```

Could you give it a try?

michaelbenayoun, line 52: A small nit because I think it actually never happens but could we do a check that `tensor_data` is None?

fxmarty:
For reference: https://github.com/huggingface/optimum/pull/587/files#r1152974762. I think it's safe to keep as a tuple + use `to_array` for now.

michaelbenayoun:
@fxmarty I am not talking about the tuple in the dictionary but `tensor_data`.

VikParuchuri:
@fxmarty Your method seems to work!  I updated the PR.

fxmarty:
@VikParuchuri Great! Could you do `pip install -e .[quality]` and `make style`?

VikParuchuri:
Done

fxmarty (approved):
LGTM thank you for the fix!

PR #933, opened 2023-03-30
Ensure backward compatibility of ORTModel

fxmarty:
Following https://github.com/huggingface/optimum/pull/872, we make sure we have no breaking change for the existing exported models.

Thank you @echarlaix for pointing it out!

### DIFF
--- a/optimum/onnxruntime/base.py
+++ b/optimum/onnxruntime/base.py
@@ -588,7 +588,7 @@ class ORTDecoderForSeq2Seq(ORTDecoder):
                     out_past_key_values[i : i + self.num_pkv] for i in range(0, len(out_past_key_values), self.num_pkv)
                 )
             else:
-                # grab the cross attention key/values from the inputs
+                # Grab the cross attention key/values from the inputs
                 out_past_key_values = tuple(
                     out_past_key_values[i : i + self.num_pkv] + past_key_values[2 * i + 2 : 2 * i + 2 + self.num_pkv]
                     for i in range(0, len(out_past_key_values), self.num_pkv)
--- a/optimum/onnxruntime/base.py
+++ b/optimum/onnxruntime/base.py
@@ -22,7 +22,11 @@ from transformers.modeling_outputs import BaseModelOutput, CausalLMOutputWithCro
 from onnxruntime import InferenceSession
 
 from ..utils import NormalizedConfigManager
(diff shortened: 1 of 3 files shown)


optimum/onnxruntime/base.py
```diff
@@ -569,10 +577,21 @@ def filter_out_output(output_name):
                 out_past_key_values[i : i + self.num_pkv] for i in range(0, len(out_past_key_values), self.num_pkv)
             )
         else:
-            # grab the cross attention key/values from the inputs
-            out_past_key_values = tuple(
-                out_past_key_values[i : i + self.num_pkv] + past_key_values[2 * i + 2 : 2 * i + 2 + self.num_pkv]
-                for i in range(0, len(out_past_key_values), self.num_pkv)
-            )
+            if self.legacy_outputs is True:
```
michaelbenayoun, line 580: :)

optimum/onnxruntime/base.py
```diff
@@ -569,10 +577,21 @@ def filter_out_output(output_name):
...
+                    "For the decoder with past, using ONNX models outputting cross attention past key values"
+                    " is deprecated and the support will be removed in optimum 2.0. We recommend exporting again the model"
+                    " with optimum>=1.7.3."
+                )
+                warn_once(logger, msg=msg)
+                out_past_key_values = tuple(
+                    out_past_key_values[i : i + self.num_pkv] for i in range(0, len(out_past_key_values), self.num_pkv)
+                )
+            else:
+                # grab the cross attention key/values from the inputs
```
michaelbenayoun, line 591:
```suggestion
                # Grab the cross attention key/values from the inputs
```

optimum/utils/logging.py
```diff
@@ -262,3 +267,8 @@ def reset_format() -> None:
 
     for handler in handlers:
         handler.setFormatter(None)
+
+
+@lru_cache(None)
```
michaelbenayoun, line 272:
```suggestion
@lru_cache
```
is enough no?

michaelbenayoun (approved):
LGTM

optimum/utils/logging.py
fxmarty: No, the default is 128: https://docs.python.org/3/library/functools.html#functools.lru_cache

PR #934, opened 2023-03-30
Remove getattr overhead in spda

fxmarty:
We go from:
image

to:
image


sanchitintel:
Hi @fxmarty, I know it's pretty late for me to be asking this question, but do you know why GPT2 can't use `is_causal` as `True` in `torch.nn.functional.scaled_dot_product_attention`? Thanks!

fxmarty:
@sanchitintel do you mean why there is this controlflow? https://github.com/huggingface/optimum/blob/9ff5ea8fbde06afdafef8db9876b124091b8f44d/optimum/bettertransformer/models/attention.py#L57-L64

It is because of https://github.com/pytorch/pytorch/issues/108108

sanchitintel:
Thanks a lot for the info, @fxmarty! :)

PR #936, opened 2023-03-30
Turn command import into optional

JingyaHuang:
# What does this PR do?

as per title.

### DIFF
--- a/optimum/commands/__init__.py
+++ b/optimum/commands/__init__.py
@@ -12,8 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from transformers.utils import is_tf_available
+
+from ..utils import is_onnx_available, is_onnxruntime_available
 from .base import BaseOptimumCLICommand, CommandInfo, RootOptimumCLICommand
 from .env import EnvironmentCommand
-from .export import ExportCommand, ONNXExportCommand, TFLiteExportCommand
-from .onnxruntime import ONNXRuntimeCommand, ONNXRuntimmeOptimizeCommand, ONNXRuntimmeQuantizeCommand
+from .export import ExportCommand
 from .optimum_cli import register_optimum_cli_subcommand
+
+
+if is_onnx_available():
+    from .export import ONNXExportCommand
+
+if is_tf_available():
+    from .export import TFLiteExportCommand
+
+if is_onnxruntime_available():
+    from .onnxruntime import ONNXRuntimeCommand, ONNXRuntimmeOptimizeCommand, ONNXRuntimmeQuantizeCommand
--- a/optimum/commands/export/__init__.py
+++ b/optimum/commands/export/__init__.py
@@ -12,7 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from transformers.utils import is_tf_available
 
+from ...utils import is_onnx_available
 from .base import ExportCommand
-from .onnx import ONNXExportCommand
-from .tflite import TFLiteExportCommand
+
+
+if is_onnx_available():
+    from .onnx import ONNXExportCommand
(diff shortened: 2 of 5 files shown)


fxmarty:
I am personally not in favor of this - basically some commands will not show depending on whether you have installed a package or not? Ideally you would want to see all possible commands, with errors raised if you miss a package for one.

JingyaHuang:
@michaelbenayoun WDYT?

@fxmarty What about my initial suggestion on not setting any default subcommand, and add them by register? Like registering onnx, tflite and onnxruntime in a `register.py` and in optimum-neuron override it with a same `register.py` but only register neuron associated commands. In this way when using the cli of optimum message will be related to submodule in optimum main, and in neuron, it will only be neuron related requirements. (although they are all optimum-cli, IMO it's fine not poping out onnx info whenusing optimum-cli under neuron env)

michaelbenayoun:
I agree with @fxmarty, what about doing this instead: each `BaseOptimumCliCommand` can define a check_requirements method, that is being run at in the right time, to fail if the command is ran without the requirement, and nothing happens otherwise?

optimum/commands/export/tflite.py
```diff
@@ -237,6 +239,11 @@ def __init__(
     def parse_args(parser: "ArgumentParser"):
         return parse_args_tflite(parser)
 
+    def check_requirements(self):
+        if not is_tf_available():
+            raise ImportError("TensorFlow is not installed. Please install TensorFlow first.")
+
     def run(self):
+        self.check_requirements()
```
michaelbenayoun, line 247: I would rather add the check in `optimum/commands/optimum_cli.py`
Just after creating the service and before actually running it:

```python
      service = args.func(args)
      service.check_requirements()
      service.run()
```

The reason for this is that I do not like when we assume a method will be called in another method. It also adds less code !

michaelbenayoun: Check slack!

optimum/commands/onnxruntime/quantize.py
```diff
@@ -70,7 +69,14 @@ class ONNXRuntimmeQuantizeCommand(BaseOptimumCLICommand):
     def parse_args(parser: "ArgumentParser"):
         return parse_args_onnxruntime_quantize(parser)
 
+    def check_requirements(self):
+        if not is_onnxruntime_available():
+            raise ImportError("Onnxruntime is not installed. Please install Onnxruntime first.")
+
```
fxmarty, line 75: Why is this needed? Can't we just run the code in `run()`, and if a dependency is needed an importerror will be raised anyway no?

michaelbenayoun: Both can be done that is right, I do not have a strong opinion on which way is best. 
I would say that with the method being called it is more explicit that the command writer needs to be the one checking that everything is installed at runtime. But since it's mostly for internal use as long as we know...

JingyaHuang: I prefer not adding extra check. The error log would be more comprehensive. Check removed.

JingyaHuang:
Can I merge this one? @michaelbenayoun @fxmarty

optimum/commands/export/tflite.py
```diff
@@ -238,5 +238,6 @@ def parse_args(parser: "ArgumentParser"):
         return parse_args_tflite(parser)
 
     def run(self):
+        self.check_requirements()
```
michaelbenayoun, line 241: I guess to be removed?

michaelbenayoun (approved):
LGTM (as long as the tests pass)

optimum/commands/export/tflite.py
JingyaHuang: Good catch!

PR #947, opened 2023-04-03
Update documentation

echarlaix:


fxmarty:
https://github.com/huggingface/optimum/pull/935

docs/source/onnxruntime/usage_guides/models.mdx
```diff
@@ -30,16 +30,15 @@ pred = onnx_qa(question, context)
...
->>> from optimum.onnxruntime import ORTModelForSequenceClassification
+from optimum.onnxruntime import ORTModelForSequenceClassification
 
->>> # Load the model from the hub and export it to the ONNX format
->>> model = ORTModelForSequenceClassification.from_pretrained(
-...     "distilbert-base-uncased-finetuned-sst-2-english", from_transformers=True
-... )
+# Load the model from the hub and export it to the ONNX format
+model_id = "distilbert-base-uncased-finetuned-sst-2-english"
+model = ORTModelForSequenceClassification.from_pretrained(model_id, export=True)
```
fxmarty, line 41: Why remove the `>>>` and `...`? They are the necessary for https://docs.python.org/3/library/doctest.html test to pass (which probably maybe does not...)

See: https://docs.python.org/3/library/doctest.html#how-are-docstring-examples-recognized

echarlaix:
> #935

Didn't see  #935, will close #947 then

docs/source/onnxruntime/usage_guides/models.mdx
echarlaix: This was removed for consistency with https://github.com/huggingface/optimum/blob/v1.7.3/docs/source/onnxruntime/usage_guides/models.mdx?plain=1#L112-L118, and because I think `>>>` makes the documentation harder to read, will close this PR as it was fixed in #935

PR #954, opened 2023-04-04
Auto-set use_merged to False if use_cache is passed as False

fxmarty:


optimum/onnxruntime/modeling_decoder.py
```diff
@@ -320,7 +320,18 @@ def _from_pretrained(
     ):
         model_path = Path(model_id)
 
+        # We do not implement the logic for use_cache=False, use_merged=True
+        if use_cache is False:
+            if use_merged is True:
```
nenkoru, line 325:
```suggestion
        if use_cache is False and use_merged is True:
```

PR #960, opened 2023-04-07
Fix issue with doc build sometimes failing silently in GH workflows

regisss:
# What does this PR do?

This PR aims at avoiding silent errors when building the documentation in a Github workflow (see [here](https://github.com/huggingface/optimum/actions/runs/4531898837/jobs/7982610819#step:12:2524) for an example).

It seems that
```yaml
defaults:
  run:
    shell: bash -l {0}
```
that is used in the main documentation build workflow makes some errors silent. Removing it since it is not necessary for building the doc.

## Before submitting


fxmarty (approved):
LGTM

PR #965, opened 2023-04-11
Adds ONNX Export Support for Timm Models

mht-sharma:
# What does this PR do?

This Pull Request introduces the capability to export `Timm` models to the ONNX format.

## Changes Included
- Added support for exporting Timm models to the ONNX format.
- Specifically, the PR includes support for `ResNext` model variants.

## Usage Example
Here's an example of how to use this newly added feature:
```bash
optimum-cli export onnx --model timm/resnext101_64x4d.c1_in1k out
```

## Before submitting


optimum/exporters/onnx/convert.py
```diff
@@ -264,17 +265,27 @@ def validate_model_outputs(
...
     ref_outputs_dict = {}
 
+    if isinstance(ref_outputs, dict):
+        outputs = ref_outputs.items()
+    elif isinstance(ref_outputs, (list, tuple)):
+        outputs = zip(list(config.outputs.keys()), ref_outputs)
+    else:
+        name = list(config.outputs.keys())[0]
+        ref_outputs_dict[name] = ref_outputs
+        outputs = []
```
fxmarty, line 275: same comment as in model_patcher.py, can you comment add a comment as to why the if/elif/else is needed?

fxmarty (commented):
Looks great!

optimum/exporters/tasks.py
```diff
@@ -201,6 +203,7 @@ class TasksManager:
         "stable-diffusion": "diffusers",
         "zero-shot-image-classification": "transformers",
         "zero-shot-object-detection": "transformers",
+        "timm-image-classification": "timm",
```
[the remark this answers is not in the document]
fxmarty: Well, models for the very same task (aka pipeline tag) may rely on a different library:
* task `image-classification`, library `timm`: https://huggingface.co/timm/resnetaa101d.sw_in12k
* task `image-classification`, library `transformers`: https://huggingface.co/microsoft/resnet-50

Given that we have a way to distinguish whether the model repo is from timm or transformers automatically, I don't see the need for the `_TASKS_TO_LIBRARY` map, nor for adding `image-classification-timm` (which is not a task, but a couple task, library).

But currently we have `_TASKS_TO_AUTOMODELS` that maps one to one. This would not be the case anymore, so either we could have it as `Dict[str, Union[str, Tuple[str]]]` like I did in https://github.com/huggingface/optimum/pull/967 with `"automatic-speech-recognition": ("AutoModelForSpeechSeq2Seq", "AutoModelForCTC")`, delaying the complexity in later functions, or we would need an additional map library, task -> loading class that would allow to distinguish (in the case of transformers/timm, but unfortunately not in the ASR case, I rather rely on the `_model_mapping` from automodel).

WDYT @mht-sharma ?

optimum/exporters/onnx/model_patcher.py
```diff
@@ -78,11 +78,20 @@ def patched_forward(*args, **kwargs):
...
+                        allow_past_in_outputs and k.startswith("past_key_values")
+                    ):
+                        filterd_outputs[k] = v
+            elif isinstance(outputs, (list, tuple)):
+                outputs_list = list(config.outputs.keys())
+                for k, v in zip(outputs_list, outputs):
                     filterd_outputs[k] = v
+            else:
+                k = list(config.outputs.keys())[0]
+                filterd_outputs[k] = outputs
```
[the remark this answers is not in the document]
mht-sharma: done

optimum/exporters/tasks.py
```diff
@@ -1174,6 +1185,99 @@ def infer_task_from_model(
...
+                return "diffusers"
+            if "config.json" in all_files:
+                with open("config.json") as fp:
+                    cfg = json.load(fp)
+                if "pretrained_cfg" in cfg:
+                    return "timm"
+                elif "_diffusers_version" in cfg:
+                    return "diffusers"
+
+            return "transformers"
```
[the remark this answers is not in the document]
mht-sharma: updated

optimum/exporters/onnx/model_patcher.py
```diff
@@ -112,15 +112,28 @@ def patched_forward(*args, **kwargs):
...
+                        or (allow_past_in_outputs and name.startswith("past_key_values"))
+                        or any(key.startswith(onnx_output_name) for key in config.outputs.keys())
+                    ):
+                        filterd_outputs[name] = value
+            elif isinstance(outputs, (list, tuple)):
+                outputs_list = list(config.outputs.keys())
+                dict(zip(outputs_list, outputs))
+            else:
+                name = list(config.outputs.keys())[0]
+                filterd_outputs[name] = outputs
```
fxmarty, line 136: Shouldn't we raise an error here if config.outputs has more than one key?

optimum/exporters/tasks.py
```diff
@@ -770,6 +770,12 @@ class TasksManager:
         "resnet": supported_tasks_mapping(
             "feature-extraction", "image-classification", onnx="ResNetOnnxConfig", tflite="ResNetTFLiteConfig"
         ),
+        "resnext26ts": supported_tasks_mapping("image-classification", onnx="TimmResNextOnnxConfig"),
+        "resnext50-32x4d": supported_tasks_mapping("image-classification", onnx="TimmResNextOnnxConfig"),
+        "resnext50d-32x4d": supported_tasks_mapping("image-classification", onnx="TimmResNext50d_32x4dOnnxConfig"),
```
fxmarty, line 775: So this is the only one that requires static shapes / channels?

optimum/exporters/tasks.py
```diff
@@ -1093,7 +1099,11 @@ def _validate_framework_choice(framework: str):
 
     @staticmethod
     def get_model_class_for_task(
-        task: str, framework: str = "pt", model_type: Optional[str] = None, model_class_name: Optional[str] = None
+        task: str,
+        framework: str = "pt",
+        model_type: Optional[str] = None,
+        model_class_name: Optional[str] = None,
+        library: Optional[str] = "transformers",
```
fxmarty, line 1106: Why Optional? The code below will fail if library is None

optimum/exporters/tasks.py
```diff
@@ -1400,9 +1424,8 @@ def infer_task_from_model(
             subfolder (`str`, *optional*, defaults to `""`):
                 In case the model files are located inside a subfolder of the model directory / repo on the Hugging
                 Face Hub, you can specify the subfolder name here.
-            revision (`Optional[str]`, *optional*):
+            revision (`Optional[str]`, *optional*, defaults to `None`):
```
fxmarty, line 1427: Feel free to remove the *optional* here, cf https://huggingface.slack.com/archives/C02P0559X9S/p1669628754896019

optimum/exporters/tasks.py
```diff
@@ -1421,6 +1444,100 @@ def infer_task_from_model(
...
+        Updates the model for export. This function is suitable to make required changes to the models from different
+        libraries to follow transformers style.
+
+        Args:
+            model_name_or_path (`Union[str, Path]`):
+                Can be either the model id of a model repo on the Hugging Face Hub, or a path to a local directory
+                containing a model.
+            model (`Union[PreTrainedModel, TFPreTrainedModel]`):
+                The instance of the model.
+            subfolder (`str`, *optional*, defaults to `""`):
```
fxmarty, line 1510:
```suggestion
            subfolder (`str`, defaults to `""`):
```

optimum/exporters/tasks.py
```diff
@@ -1421,6 +1444,100 @@ def infer_task_from_model(
...
+                In case the model files are located inside a subfolder of the model directory / repo on the Hugging
+                Face Hub, you can specify the subfolder name here.
+            revision (`Optional[str]`, *optional*, defaults to `None`):
```
fxmarty, line 1513:
```suggestion
            revision (`Optional[str]`, defaults to `None`):
```

optimum/exporters/tasks.py
```diff
@@ -1421,6 +1444,100 @@ def infer_task_from_model(
...
+                    library_name = "diffusers"
+                else:
+                    library_name = "transformers"
+        else:
+            library_name = huggingface_hub.model_info(model_name_or_path, revision=revision).library_name
+
+        return library_name
+
+    @classmethod
+    def patch_model_for_export(
```
fxmarty, line 1493: Maybe we could find an other name in order not to confuse with the model patcher from the ONNX export. Something like `standardize_model_attributes` for example

tests/exporters/exporters_utils.py
```diff
@@ -95,6 +95,12 @@
     "poolformer": "hf-internal-testing/tiny-random-PoolFormerModel",
     "regnet": "hf-internal-testing/tiny-random-RegNetModel",
     "resnet": "hf-internal-testing/tiny-random-resnet",
+    "resnext26ts": "timm/resnext26ts.ra2_in1k",
+    "resnext50-32x4d": "timm/resnext50_32x4d.tv2_in1k",
+    "resnext50d-32x4d": "timm/resnext50d_32x4d.bt_in1k",
+    "resnext101-32x4d": "timm/resnext101_32x4d.gluon_in1k",
+    "resnext101-32x8d": "timm/resnext101_32x8d.tv_in1k",
+    "resnext101-64x4d": "timm/resnext101_64x4d.c1_in1k",
```
fxmarty, line 103: These models are too large. Are there no tiny models on the Hub. If not, we should not run the CI for TIMM export on every PR commit. Maybe we could run it only on push to main or triggered manually.

optimum/exporters/onnx/model_configs.py
```diff
@@ -650,6 +650,25 @@ class DonutSwinOnnxConfig(ViTOnnxConfig):
     pass
 
 
+class TimmResNextOnnxConfig(ViTOnnxConfig):
+    ATOL_FOR_VALIDATION = 1e-3
+
+    def rename_ambiguous_inputs(self, inputs):
+        #  The input name in the model signature is `x, hence the export input name is updated.
+        model_inputs = {}
+        model_inputs["x"] = inputs["pixel_values"]
```
fxmarty, line 659: This is more in an effort of standardization rather than removing ambiguity right? For example if we are to use the exported models with ORTModel later.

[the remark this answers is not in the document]
mht-sharma: In `timm` the forward function only has input `x`. So keeping it one.

```python
def forward(self, x):
```

optimum/exporters/onnx/base.py
```diff
@@ -406,6 +406,16 @@ def torch_to_onnx_output_map(self) -> Dict[str, str]:
         """
         return {}
 
+    def rename_ambiguous_inputs(self, inputs) -> Dict[str, Dict[int, str]]:
```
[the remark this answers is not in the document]
mht-sharma: Done

optimum/exporters/onnx/base.py
```diff
@@ -406,6 +406,16 @@ def torch_to_onnx_output_map(self) -> Dict[str, str]:
...
+        """
+        Updates the input names of the model to export.
+        Override the function when the model input names are ambiguous or too generic.
+
```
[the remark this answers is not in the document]
mht-sharma: Done

optimum/exporters/onnx/model_patcher.py
mht-sharma: Raised an error, however, we should not catch this if the config is correct.

optimum/exporters/onnx/model_patcher.py
```diff
@@ -112,15 +112,28 @@ def patched_forward(*args, **kwargs):
...
+                    onnx_output_name = config.torch_to_onnx_output_map.get(name, name)
+                    if (
+                        onnx_output_name in config.outputs
+                        or (allow_past_in_outputs and name.startswith("past_key_values"))
+                        or any(key.startswith(onnx_output_name) for key in config.outputs.keys())
+                    ):
+                        filterd_outputs[name] = value
+            elif isinstance(outputs, (list, tuple)):
+                outputs_list = list(config.outputs.keys())
+                dict(zip(outputs_list, outputs))
```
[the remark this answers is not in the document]
mht-sharma: Yeah, probably a mistake during rebased. Fixed.

optimum/exporters/tasks.py
mht-sharma: Yes, at least for the resnext models.

optimum/exporters/tasks.py
mht-sharma: True, removed

optimum/exporters/tasks.py
mht-sharma: removed

optimum/exporters/tasks.py
mht-sharma: Done

optimum/exporters/tasks.py
mht-sharma: done

optimum/exporters/onnx/model_patcher.py
fxmarty: Raising an error may help to detect that a config is not correct!

.github/workflows/test_export_onnx.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     - name: Test with unittest
       working-directory: tests
       run: |
-        pytest exporters/onnx/test_onnx_*.py -s -n auto -m "not tensorflow_test" --durations=0
+        pytest exporters/onnx/test_onnx_*.py -s -n auto -m "not tensorflow_test and not timm_test" --durations=0
```
fxmarty, line 34: Can you add `timm_test` to https://github.com/huggingface/optimum/blob/fb57f9f01e9cf8bfcf4eb02ecf9beefd6b47f8b3/pyproject.toml#L33 ?

fxmarty (approved):
LGTM that's great! Could you add relevant information in the documentation about the support?

optimum/exporters/onnx/model_configs.py
mht-sharma: Yeah intention was both

fxmarty (approved):
Once the above is addressed it is good to me, don't know if @michaelbenayoun wants to rereview.

michaelbenayoun:
LGTM

PR #967, opened 2023-04-11
Allow to use a custom class in TasksManager & use canonical tasks names

fxmarty:
Some architectures do not map to an auto class in transformers (e.g. as pix2struct and VisualBert). Thus we introduce the `_CUSTOM_CLASSES` in TasksManager to handle those.

Moreover, when possible, we use `model_info.pipeline_tag` instead of `model_info.transformersInfo.pipeline_tag` to infer the task, as the later may be wrong. For reference: https://huggingface.slack.com/archives/C02EMARJ65P/p1681215560263539

For example, now `task  = TasksManager.infer_task_from_model("google/pix2struct-base")` rightfully returns `image-to-text`, while it was returning `text2text-generation` before (due to the issue explained on slack).

Moreover, `model = TasksManager.get_model_from_task("image-to-text", "google/pix2struct-base")` successfully loads the model as `Pix2StructForConditionalGeneration`, although it is not registered in `AutoModelForVision2Seq`.

Moreover, we align with the pipelines/tasks names from https://huggingface.co/datasets/huggingface/transformers-metadata/blob/main/pipeline_tags.json & https://docs.google.com/spreadsheets/d/1zvGwDwo-efnAhqepIleYLsALO70dn6dxRAF2ih8mCH8/edit#gid=0 instead of defining our own. For reference: https://huggingface.slack.com/archives/C03AHFPCWHM/p1680025770248079?thread_ts=1680020763.420539&cid=C03AHFPCWHM Backward compatibility is kept in the CLI.

Note: using `-with-past` to denote the case where past key values are used should probably not be in the task name.

Backward compatibility can be tested with `RUN_SLOW=1 pytest tests/exporters/onnx/test_exporters_onnx_cli.py -k "test_legacy_tasks_backward_compatibility" -s`

Edit: remaining issues:
- [x] (fixed) automatic-speech-recognition maps to both AutoModelForSpeechSeq2Seq and AutoModelForCTC, so we can not force with the task automatic-speech-recognition one or the other.


optimum/commands/export/tflite.py
```diff
@@ -189,7 +189,7 @@ def parse_args_tflite(parser: "ArgumentParser"):
         default=None,
         help=(
             "The name of the column in the dataset containing the main data to preprocess. "
-            "Only for sequence-classification and token-classification. "
+            "Only for text-classification and token-classification. "
```
regisss, line 192: Why replacing "sequence" by "text" here while the task `sequence-classification` is still documented?

optimum/exporters/tasks.py
```diff
@@ -925,8 +949,10 @@ def get_supported_model_type_for_task(task: str, exporter: str) -> List[str]:
         ]
 
     @staticmethod
-    def format_task(task: str) -> str:
-        return task.replace("-with-past", "")
+    def map_from_legacy(task: str) -> str:
+        if task in TasksManager._LEGACY_TASK_MAP:
+            task = TasksManager._LEGACY_TASK_MAP[task]
```
regisss, line 954: Should we log a warning here telling the user the non-legacy task name to use?

regisss (commented):
Great PR to have more consistency between libraries :fire: 
I left a couple of comments, I think the doc still contains some occurrences of `sequence-classification` while it should now be `text-classification` if I'm not mistaken.

optimum/exporters/tasks.py
fxmarty: Done!

optimum/commands/export/tflite.py
fxmarty: Fixed in c8b9abb03b84e91c892b316538148d1c899c102e

optimum/exporters/tasks.py
```diff
@@ -925,8 +949,14 @@ def get_supported_model_type_for_task(task: str, exporter: str) -> List[str]:
         ]
 
     @staticmethod
-    def format_task(task: str) -> str:
-        return task.replace("-with-past", "")
+    def map_from_legacy(task: str) -> str:
+        if task in TasksManager._LEGACY_TASK_MAP:
+            logger.warning(
+                f"The task name {task} is deprecated, and will not be supported in"
+                f" optimum 2.0. Please use {TasksManager._LEGACY_TASK_MAP[task]} instead."
```
regisss, line 956:
```suggestion
                f" Optimum 2.0. Please use {TasksManager._LEGACY_TASK_MAP[task]} instead."
```

regisss (approved):
Great PR :rocket:

optimum/exporters/onnx/__main__.py
```diff
@@ -146,14 +146,8 @@ def main_export(
...
-    # Infer the task
-    if task == "auto":
-        try:
-            task = TasksManager.infer_task_from_model(model_name_or_path)
-        except KeyError as e:
-            raise KeyError(
-                f"The task could not be automatically inferred. Please provide the argument --task with the task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"
-            )
+
+    task = TasksManager.map_from_legacy(task)
```
michaelbenayoun, line 150: Insteqd of having this, I think we should define a `"task_synonyms"` dictionary in the `TasksManager`, allowing to specify either `sequence-classification` or `text-classification` for instance, not make it legacy but rather synonyms.

optimum/exporters/tasks.py
```diff
@@ -140,50 +141,70 @@ class TasksManager:
...
             "audio-frame-classification": "TFAutoModelForAudioFrameClassification",
-            "audio-ctc": "TFAutoModelForCTC",
             "audio-xvector": "TFAutoModelForAudioXVector",
-            "vision2seq-lm": "TFAutoModelForVision2Seq",
+            "image-to-text": "TFAutoModelForVision2Seq",
             "zero-shot-image-classification": "TFAutoModelForZeroShotImageClassification",
             "zero-shot-object-detection": "TFAutoModelForZeroShotObjectDetection",
         }
 
+    _LEGACY_TASK_MAP = {
```
michaelbenayoun, line 179:
```suggestion
    _TASK_SYNONYMS_MAP = {
```

optimum/exporters/tasks.py
```diff
@@ -192,54 +213,53 @@ class TasksManager:
...
+    # TODO: some models here support text-generation export but are not supported in ORTModelForCausalLM
     # Set of model topologies we support associated to the tasks supported by each topology and the factory
     _SUPPORTED_MODEL_TYPE = {
         "audio-spectrogram-transformer": supported_tasks_mapping(
-            "default",
+            "feature-extraction",
             "audio-classification",
             onnx="ASTOnnxConfig",
         ),
         "albert": supported_tasks_mapping(
```
michaelbenayoun, line 234: I would make it so that `supported_tasks_mapping` can accept synonyms as well. But not primordial either.

optimum/exporters/tasks.py
```diff
@@ -925,8 +949,14 @@ def get_supported_model_type_for_task(task: str, exporter: str) -> List[str]:
         ]
 
     @staticmethod
-    def format_task(task: str) -> str:
-        return task.replace("-with-past", "")
+    def map_from_legacy(task: str) -> str:
```
michaelbenayoun, line 952:
```suggestion
    def map_from_synonym(task: str) -> str:
```

optimum/exporters/tasks.py
michaelbenayoun, line 956: We should always accept both IMO.
Basically the previous naming was due because it is aligned with Auto classes for many models. So accepting both makes sense. 
```suggestion
```

optimum/exporters/tasks.py
```diff
@@ -942,33 +972,55 @@ def _validate_framework_choice(framework: str):
...
+        task: str, framework: str = "pt", model_type: Optional[str] = None, model_class_name: Optional[str] = None
+    ) -> Type:
         """
         Attempts to retrieve an AutoModel class from a task name.
 
         Args:
             task (`str`):
                 The task required.
-            framework (`str`, *optional*, defaults to `"pt"`):
+            framework (`Optional[str]`, defaults to `"pt"`):
```
michaelbenayoun, line 984:
```suggestion
            framework (`str`, defaults to `"pt"`):
```

tests/exporters/onnx/test_exporters_onnx_cli.py
```diff
@@ -270,3 +275,32 @@ def test_export_on_fp16(
...
+            ["sequence-classification", "bert"],
+            ["masked-lm", "bert"],
+            ["default", "blenderbot"],
+            ["default-with-past", "blenderbot"],
+            ["audio-ctc", "wav2vec2-conformer"],
+        ]
+    )
+    @slow
+    @pytest.mark.run_slow
+    def test_legacy_tasks_backward_compatibility(self, task: str, model_type: str):
```
michaelbenayoun, line 297:
```suggestion
    def test_task_synonyms(self, task: str, model_type: str):
```

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -11,7 +11,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-"""Classes handling causal-lm related architectures in ONNX Runtime."""
+"""Classes handling text-generation related architectures in ONNX Runtime."""
```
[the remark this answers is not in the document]
fxmarty: Yes I agree, for reference the thread on slack: https://huggingface.slack.com/archives/C02EMARJ65P/p1681229504030279?thread_ts=1681215560.263539&cid=C02EMARJ65P

optimum/exporters/tasks.py
pcuenca: Good call, that'd be interesting if external users are writing their own mappings. Is that something that happens frequently?

Issue #972, opened 2023-04-15
Add output of  `output_hidden_states` for onnx model export.


KexinFeng:
### Feature request

Currently, the merged decoder in  ORTModelForCausalLM only output a List of 25 OrtNDArrays, 1 for the output logits + 24 from past keys and values. 
Could you add the output of `hidden_states`? It doesn't look like much work; only need to set the key `output_hidden_states=True` in [`GPT2LMHeadModel`](https://huggingface.co/docs/transformers/model_doc/gpt2#transformers.GPT2LMHeadModel). But this is a very important feature, which enables one of the SOTA contrastive search: https://huggingface.co/blog/introducing-csearch

Thanks!

### Motivation

This is a very important feature, which enables the contrastive search: https://huggingface.co/blog/introducing-csearch.

### Your contribution

I don't know how this works yet. But it doesn't seems hard. 

In the existing contrastive search implementation, this output of hidden_states is simply enabled by setting a key output_hidden_states=True in 
/transformers/generation/utils.py L1834

Relavant past PRs https://github.com/huggingface/optimum/pull/647/files, https://github.com/huggingface/optimum/pull/930/files.

fxmarty:
Thank you, I will add an option for this!

PR #984, opened 2023-04-18
add position_ids

jiqing-feng:
Some models like gpt2 and llama have "position_ids" in their generation inputs. We can add "position_ids" in the model's config to fix it.

@echarlaix @fxmarty Would you please help to review it? Thanks!

### DIFF
--- a/optimum/onnxruntime/modeling_decoder.py
+++ b/optimum/onnxruntime/modeling_decoder.py
@@ -641,6 +641,7 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
         input_ids: torch.LongTensor = None,
         attention_mask: Optional[torch.FloatTensor] = None,
         past_key_values: Optional[Tuple[Tuple[torch.Tensor]]] = None,
+        position_ids: Optional[torch.LongTensor] = None,
         labels: Optional[torch.LongTensor] = None,
         **kwargs,
     ) -> CausalLMOutputWithCrossAttentions:
@@ -649,6 +650,7 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
                 input_ids=input_ids,
                 attention_mask=attention_mask,
                 past_key_values=past_key_values,
+                position_ids=position_ids,
                 labels=labels,
             )
         elif self.use_merged is True:
@@ -656,12 +658,14 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
                 input_ids=input_ids[:, -1:],
                 past_key_values=past_key_values,
                 attention_mask=attention_mask,
+                position_ids=position_ids,
             )
         else:
             outputs = self.decoder_with_past(
                 input_ids=input_ids[:, -1:],
                 past_key_values=past_key_values,
                 attention_mask=attention_mask,
+                position_ids=position_ids,
                 labels=labels,
             )
 
@@ -675,12 +679,13 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
 
         attention_mask = kwargs.get("attention_mask", None)  # input_ids.new_ones(input_ids.shape)
         use_cache = kwargs.get("use_cache", None)
+        position_ids = kwargs.get("position_ids", None)
 
         return {
             "input_ids": input_ids,
             "past_key_values": past_key_values,
             "use_cache": use_cache,
-            "position_ids": None,
+            "position_ids": position_ids,
             "attention_mask": attention_mask,
         }
 
--- a/optimum/exporters/onnx/base.py
+++ b/optimum/exporters/onnx/base.py
@@ -539,7 +539,7 @@ class OnnxConfigWithPast(OnnxConfig, ABC):
                     if (
                         self.use_past is True
                         and self.use_cache_branch is not False
-                        and input_name in ["decoder_input_ids", "input_ids"]
+                        and input_name in ["decoder_input_ids", "input_ids", "position_ids"]
                     ):
                         sequence_length = dummy_input_gen.sequence_length
                         if "sequence_length" in kwargs and kwargs["sequence_length"] != 1:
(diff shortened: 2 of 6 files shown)


fxmarty:
Hi @jiqing-feng , just to be sure, is this concerning the ONNX export or ONNX Runtime integration? Or both? In the ONNX export, we should give the option, yes.

echarlaix:
Currently `position_ids` is ignored during ORT inference for causal LM https://github.com/huggingface/optimum/blob/v1.8.2/optimum/onnxruntime/modeling_decoder.py#L683, so when enabling it as inputs for the ONNX export we should enable it for `ORTModelForCausalLM` as well

jiqing-feng:
Hi @echarlaix @fxmarty thanks for your comments. I have added `position_ids` on `ORTModelForCausalLM`. Would you please help to review it? Thanks.

fxmarty:
Hi @jiqing-feng , could you comment a bit on your use case? I'm not sure what's the benefit of this in the ORT integration.

While `attention_mask` [is updated in `_update_model_kwargs_for_generation`](https://github.com/huggingface/transformers/blob/aab14120d4bbc64f0181c95434b64b92985ffd16/src/transformers/generation/utils.py#L761), `position_ids` is not. Thus, in my understanding, if we are to pass a `position_ids` to the `generate()` method, it would be kept as a constant at each `prepare_inputs_for_generation` call in the generation, e.g. [here](https://github.com/huggingface/transformers/blob/aab14120d4bbc64f0181c95434b64b92985ffd16/src/transformers/generation/utils.py#L1940). Which is something we don't wish.

Am I misunderstanding something @gante ?

gante:
@fxmarty What I write next is from a `transformers` point of view :)

At the moment, `position_ids` is a required input to support left padding. For legacy reasons, we consider it an optional input the same way `attention_mask` is an optional input -- if not passed, we assume all tokens are valid, and `position_ids` is a `torch.arange`. In `.generate()`, we delegate its creation to the model's `prepare_inputs_for_generation`, where it is computed from the `attention_mask` (e.g. [here](https://github.com/huggingface/transformers/blob/3080fb714f11d28d8d3468548d4fa8a41e671a11/src/transformers/models/gpt2/modeling_gpt2.py#L1018)). 

This means that while `position_ids` is never passed into `.generate()`, is is passed from `.generate()` to the model and is updated at each iteration.

fxmarty:
> In .generate(), we delegate its creation to the model's prepare_inputs_for_generation, where it is computed from the attention_mask (e.g. [here](https://github.com/huggingface/transformers/blob/3080fb714f11d28d8d3468548d4fa8a41e671a11/src/transformers/models/gpt2/modeling_gpt2.py#L1018)).

So this means that executing this controlflow is absolutely necessary, right?

KexinFeng:
> Hi @jiqing-feng , could you comment a bit on your use case? I'm not sure what's the benefit of this in the ORT integration.
> 
> While `attention_mask` [is updated in `_update_model_kwargs_for_generation`](https://github.com/huggingface/transformers/blob/aab14120d4bbc64f0181c95434b64b92985ffd16/src/transformers/generation/utils.py#L761), `position_ids` is not. Thus, in my understanding, if we are to pass a `position_ids` to the `generate()` method, it would be kept as a constant at each `prepare_inputs_for_generation` call in the generation, e.g. [here](https://github.com/huggingface/transformers/blob/aab14120d4bbc64f0181c95434b64b92985ffd16/src/transformers/generation/utils.py#L1940) (since we'd go into [this piece of controlflow](https://github.com/huggingface/transformers/blob/9417c924af539be5f941c8a709a96b60dfe29eb3/src/transformers/models/gpt2/modeling_gpt2.py#L1024-L1025)). Which is something we don't wish.
> 
> Am I misunderstanding something @gante ?

@fxmarty Yeah, I was about to request for exactly the same feature. As @gante mentioned above, the `position_ids` (along with `attention_mask`) are necessary inputs to transformer based models like gpt2 to deal with left-padded input.  More specifically, my feature request comes from the context described in https://github.com/huggingface/optimum/issues/972. 


> > In .generate(), we delegate its creation to the model's prepare_inputs_for_generation, where it is computed from the attention_mask (e.g. [here](https://github.com/huggingface/transformers/blob/3080fb714f11d28d8d3468548d4fa8a41e671a11/src/transformers/models/gpt2/modeling_gpt2.py#L1018)).
> 
> So this means that executing this controlflow is absolutely necessary, right?

I don't think the controlflow of computing the `position_ids` from `attention_mask` is necessary, if this is what you referred to. As long as the traced transformer model (*.onnx) can take `position_ids` as an effective input, it should be good enough.

fxmarty:
cc @echarlaix @michaelbenayoun

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -649,19 +650,22 @@ def forward(
                 input_ids=input_ids,
                 attention_mask=attention_mask,
                 past_key_values=past_key_values,
+                position_ids=position_ids,
```
michaelbenayoun, line 653: `self.decoder` is an `ORTDecoder` right?
If so, we also need to update its forward method to handle `position_ids`.

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -649,19 +650,22 @@ def forward(
...
                 past_key_values=past_key_values,
                 attention_mask=attention_mask,
+                position_ids=position_ids,
             )
         else:
             outputs = self.decoder_with_past(
                 input_ids=input_ids[:, -1:],
                 past_key_values=past_key_values,
                 attention_mask=attention_mask,
+                position_ids=position_ids,
```
michaelbenayoun, line 668: Same comment.

michaelbenayoun (commented):
On my side, I am ok with adding this as long as we make sure we do not:

- Add any break changes
- Do not break `.generate`, it should work with and without the `position_ids` (except for the left padding case of course)

jiqing-feng:
Hi @michaelbenayoun  , thanks for your comment. I have updated position_ids in ORTdecoder forward. Could you please review it? Thanks! cc @fxmarty @gante

jiqing-feng:
> Hi everyone, @jiqing-feng @fxmarty @gante is there any updates on this feature?

cc @michaelbenayoun

fxmarty:
@jiqing-feng @KexinFeng I now realize I misunderstood your argument and motivation for the PR. This appears to me to be a major bug in Optimum which should be fixed ASAP.

Issue #894, opened 2023-03-17
Merging of ONNX decoder >2GB fails


fxmarty:
### System Info

```shell
optimum main
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

`optimum-cli export onnx --model gpt2-large gpt2_onnx`

Traceback:
```
(fx) felix@hf-dgx-01:~/optimum$ optimum-cli export onnx --model gpt2-large gpt2_onnx
Framework not specified. Using pt to export to ONNX.
Automatic task detection to causal-lm-with-past.
use_past = False is different than use_present_in_outputs = True, the value of use_present_in_outputs value will be used for the outputs.
Using framework PyTorch: 2.1.0.dev20230306+cu117
Overriding 2 configuration item(s)
        - use_cache -> True
        - pad_token_id -> 0
/home/felix/miniconda3/envs/fx/lib/python3.9/site-packages/transformers/models/gpt2/modeling_gpt2.py:794: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if batch_size <= 0:
======= Diagnostic Run torch.onnx.export version 2.1.0.dev20230306+cu117 =======
verbose: False, log level: Level.ERROR
======================= 0 NONE 0 NOTE 0 WARNING 0 ERROR ========================

Saving external data to one file...
Using framework PyTorch: 2.1.0.dev20230306+cu117
Overriding 2 configuration item(s)
        - use_cache -> True
        - pad_token_id -> 0
Asked a sequence length of 16, but a sequence length of 1 will be used with use_past == True for `input_ids`.
======= Diagnostic Run torch.onnx.export version 2.1.0.dev20230306+cu117 =======
verbose: False, log level: Level.ERROR
======================= 0 NONE 0 NOTE 0 WARNING 0 ERROR ========================

Saving external data to one file...
Asked a sequence length of 16, but a sequence length of 1 will be used with use_past == True for `input_ids`.
Traceback (most recent call last):
  File "/home/felix/optimum/optimum/exporters/onnx/config.py", line 111, in post_process_exported_models
    merge_decoders(
[...]

PR #988, opened 2023-04-20
Fix protobuf max allowed size

fxmarty:
Use 2000000000 as onnx (for some reason) instead of 2147483648.

Fixes https://github.com/huggingface/optimum/issues/894


JingyaHuang (approved):
LGTM, thanks for fixing it.

<sub>Do you know why the number?</sub>

fxmarty:
Not really, this is hardcoded in checker.py from onnx repo.

PR #989, opened 2023-04-20
Add SAM bettertransformer support

fxmarty:
As per title


fxmarty:
The gain will be marginal as <20% of the time is spent in attention

PR #992, opened 2023-04-21
remove encoder_sequence_length in seq2seq config so users can pass it

jiqing-feng:
Hi, @fxmarty @echarlaix . The param of `encoder_sequence_length` does not need to be specified here. First, the parameter will be set to `sequence_length` if it is None, see [DummySeq2SeqPastKeyValuesGenerator](https://github.com/huggingface/optimum/blob/main/optimum/utils/input_generators.py#L449). Second, users can specify the `encoder_sequence_length` by themselves if we remove this code. Would you please help to review it? Thanks!

### DIFF
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -194,7 +194,6 @@ class TextSeq2SeqOnnxConfig(OnnxSeq2SeqConfigWithPast):
         dummy_seq2seq_past_key_values_generator = self.DUMMY_INPUT_GENERATOR_CLASSES[2](
             self.task,
             self._normalized_config,
-            encoder_sequence_length=dummy_text_input_generator.sequence_length,
             **kwargs,
         )
         dummy_inputs_generators = [


fxmarty:
Hi @jiqing-feng , thank you for the PR. I'm not sure to understand what's the benefit of this. No matter the `encoder_sequence_length` passed, the controlflow captured should be the same by the PyTorch ONNX export, is that not?

jiqing-feng:
> Hi @jiqing-feng , thank you for the PR. I'm not sure to understand what's the benefit of this. No matter the `encoder_sequence_length` passed, the controlflow captured should be the same by the PyTorch ONNX export, is that not?
> 
> Moreover, I believe the `encoder_outputs.last_hidden_states` and the past key values length for cross-attention should be the same. Thus I believe it is better to hard code the cross-attention pkv length.

Hi @fxmarty , thanks for your comment. I notice that the [DummySeq2SeqPastKeyValuesGenerator](https://github.com/huggingface/optimum/blob/main/optimum/utils/input_generators.py#L449) works for encoder-decoder models. The `past_key_values` is only the input of the decoder, so I think the `encoder_sequence_length` depends on the output of the encoder. If I just want to get the decoder inputs, the param of `encoder_sequence_length` would be better to manipulate by myself and it should be the same as the input sequence length. The `decoder_sequence_length` is always `1` for the decoder if we set `use_cache=True` in generation tasks because generation only uses the last token and the previous past_key_values. So for the decoder model, the `encoder_sequence_length` and the `decoder_sequence_length` could be different. I'm not sure I'm making it clear and would appreciate your advice. Thanks!

jiqing-feng:
Hi @fxmarty thanks for your clarification. However, the `encoder_outputs length` should be the same as `input_ids length`, but in the function I see that both the length of `input_ids` and `decoder_input_ids` are set to 1. There are two facts: `input_ids_length = encoder_outputs_length = encoder_sequence_length` and `decoder_ids_length = decoder_sequence_length`. If I set `sequence_length=5` in a generation task with past, I will get `input_ids: torch.Size([1, **1**]); decoder_input_ids torch.Size([1, **1**]); past_key_values[0][0,1,2,3]: past_key_values torch.Size([1, 8, **5**, 64])` which is against this fact.

jiqing-feng:
Thank you @fxmarty for the correction. I tried to set `decoder_sequence_length` and `sequence_length` to different values, but the `encoder_sequence_length` and the `decoder_sequence_length` still keep the same.

jiqing-feng:
> @jiqing-feng Yes, this option is missing!

@fxmarty Thanks
I think I know what you mean. If the `encoder_sequence_length` cannot be manipulated by users, could I add another PR to let the `decoder_sequence_length` can be manipulated by users?

fxmarty:
cc @echarlaix

michaelbenayoun:
Hi @jiqing-feng ,
You would like to be able to specify the `decoder_sequence_length` where?

jiqing-feng:
> Hi @jiqing-feng , You would like to be able to specify the `decoder_sequence_length` where?

In my example, I wish I could pass a value to control `decoder_sequence_length` so I can get different `decoder_sequence_length` and `encoder_sequence_length` in the dummy inputs.

Issue #956, opened 2023-04-04
Adding newer models to Optimum (bloom, alpaca)


AbhayGoyal:
### Feature request

Hi,
I just wanted to know if you could add Bloom and Alpaca for better transformers

### Motivation

We need to be able to see how the big models run on smaller systems

### Your contribution

I could help in making this happen but I do not have a basic understanding of optimum yet,

ingo-m:
Bloom models are supported (right?) but do not work as expected: https://github.com/huggingface/optimum/issues/1171

fxmarty:
Hi @vincentmin, @AbhayGoyal Bloom support with SDPA was added in https://github.com/huggingface/optimum/pull/1221. Alpaca is based on Llama so should be working already.

Feel free to open a new issue if you encounter any problem!

PR #998, opened 2023-04-21
Support llama bettertransformer

fxmarty:
As per title, let me benchmark a bit. Fixes https://github.com/huggingface/optimum/issues/956


younesbelkada (approved):
Thanks a lot for this great addition! 🔥

PR #1003, opened 2023-04-21
Fix _versions.yml upload in doc build

regisss:
# What does this PR do?

This PR fixes an issue where the file `_versions.yml` is not updated in the `build_main_documentation` workflow for new releases.

This issue comes from the fact that the file `_versions.yml` is downloaded in `build_dir/optimum` (see [here](https://github.com/huggingface/optimum/blob/ccf4b4dbb6d5f4421551ed0d83e0eb07b0261257/.github/workflows/build_main_documentation.yml#L74)) while the doc is generated in `optimum/optimum-doc-build/optimum` (see [here](https://github.com/huggingface/optimum/blob/ccf4b4dbb6d5f4421551ed0d83e0eb07b0261257/.github/workflows/build_main_documentation.yml#L101)) and that's when the version file is updated if it exists (see [here](https://github.com/huggingface/doc-builder/blob/3f6a2fd4c0e88be4ef55d74930b2517f50cf24cb/src/doc_builder/commands/build.py#L111)). Since it doesn't exist in `optimum/optimum-doc-build/optimum`, it is not updated and thus at the end of the workflow the non-updated version file of `build_dir/optimum` is uploaded.

## Before submitting

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -67,12 +67,6 @@ jobs:
           pip install black
           cd ..
 
-      - name: Create build directory
-        run: |
-          mkdir -p build_dir/optimum
-          cd build_dir/optimum
-          wget https://huggingface.co/datasets/hf-doc-build/doc-build/raw/main/optimum/_versions.yml
-
       - name: Make Graphcore documentation
         run: |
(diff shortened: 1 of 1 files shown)


fxmarty (approved):
LGTM thank you!

PR #995, opened 2023-04-21
Add sam ONNX export

fxmarty:
As per title. For now only support `input_points` as an input, will the rest really be useful @younesbelkada ?

We will need an other PR to be merged in transformers https://github.com/huggingface/transformers/pull/22915, and we will need to wait for transformers to release.

Yon can try with: `optimum-cli export onnx --model fxmarty/sam-vit-tiny-random sam_onnx`

Validation on various shapes pass: `RUN_SLOW=1 pytest tests/bettertransformer/test_encoder.py -k "test_accelerate_compatibility_cpu_gpu" -s -vvvvv`

Fixes https://github.com/huggingface/optimum/issues/990

PR #1025, opened 2023-05-02
Sam ONNX export support

michaelbenayoun:
# What does this PR do?

Builds on what @fxmarty did in #995 .

### DIFF
--- a/optimum/exporters/onnx/convert.py
+++ b/optimum/exporters/onnx/convert.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 """ONNX model check and export functions."""
 
-import os
+import os # noqa: I001
 from inspect import signature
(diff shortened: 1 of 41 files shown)


optimum/exporters/onnx/convert.py
```diff
@@ -508,12 +508,12 @@ def export_tensorflow(
     # This is needed to import onnx and tf2onnx because onnx is also the name of the current directory.
     import sys
 
+    import onnx
```
regisss, line 511: Do we need this import here? `onnx` is already imported at line 24 of this file

regisss: Geez I read the comment but didn't understand it that way :rofl:

PR #1037, opened 2023-05-07
Set onnx max dependecy

echarlaix:
Set onnx max dependency to make it compatible with ONNX Runtime inference


setup.py
```diff
@@ -41,21 +41,21 @@
...
-        "onnx",
+        "onnx<1.14.0",
         "onnxruntime-gpu>=1.9.0",
         "datasets>=1.2.1",
         "evaluate",
         "protobuf>=3.20.1",
     ],
-    "exporters": ["onnx", "onnxruntime", "timm"],
-    "exporters-gpu": ["onnx", "onnxruntime-gpu", "timm"],
+    "exporters": ["onnx<1.14.0", "onnxruntime", "timm"],
```
echarlaix, line 57: To make it compatible if needed for ONNX Runtime (cc @regisss)

regisss (approved):
LGTM!

Weird that a few tests failed because. It seems PIP constraint solver had an issue in those tests.

PR #1042, opened 2023-05-08
Add ONNX / ONNXRuntime support for StarCoder

JingyaHuang:
# What does this PR do?

- [x] Add `gpt-bigcode` support in ONNX exporter
- [x] Add `gpt-bigcode` inference support for ONNX Runtime
    - [x] CPU inference
    - [x] GPU inference (IO Binding)
- [x] Tests
---------------
## Snippets

* Export starcoder with `optimum-cli`
```bash
optimum-cli export onnx --model bigcode/starcoderbase-1b starcoder_onnx/
```

* Inference with ONNX Runtime
```python

checkpoint = "starcoder_onnx/"
device = "cuda"

tokenizer = AutoTokenizer.from_pretrained(checkpoint)
model = ORTModelForCausalLM.from_pretrained(checkpoint, use_io_binding=True, provider="CUDAExecutionProvider")

inputs = tokenizer.encode("def print_hello_world():", return_tensors="pt").to(device)
outputs = model.generate(inputs)

print(tokenizer.decode(outputs[0]))
```


dulalbert:
Hi, any news for the support ?
Thanks

optimum/onnxruntime/utils.py
```diff
@@ -109,6 +109,7 @@ class ORTConfigManager:
         "distilbert": "bert",
         "electra": "bert",
         "gpt2": "gpt2",
+        "gpt-bigcode": "gpt2",
```
liulhdarks, line 112: `gpt-bigcode` should be changed to `gpt_bigcode`

optimum/exporters/onnx/model_configs.py
```diff
@@ -268,6 +269,45 @@ def add_past_key_values(self, inputs_or_outputs: Dict[str, Dict[int, str]], dire
...
+        else:
+            decoder_sequence_name = "past_sequence_length + 1"
+            name = "present"
+
+        for i in range(self._normalized_config.num_layers):
+            # No dim for `n_head` when using multi-query attention
+            inputs_or_outputs[f"{name}.{i}.key_value"] = {
+                0: "batch_size",
+                1: decoder_sequence_name,
+            }
```
michaelbenayoun, line 305: Maybe it's worth creating a `TextDecoderWithMQAOnnxConfig` or something like that, that handles that for all the necessary models.

michaelbenayoun (approved):
LGTM

fxmarty:
hi, which one is up to date between https://github.com/huggingface/optimum/pull/1236 & https://github.com/huggingface/optimum/pull/1042 ?

JingyaHuang:
> hi, which one is up to date between #1236 & #1042 ?

@fxmarty This one.

optimum/exporters/onnx/model_configs.py
JingyaHuang: We can do that when adding other models using mqa.

optimum/exporters/onnx/base.py
```diff
@@ -596,7 +596,8 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
             and self.use_cache_branch is not False
             and "attention_mask" in dummy_inputs
         ):
-            past_length = dummy_inputs["past_key_values"][0][0].shape[2]
+            past_length = dummy_inputs["past_key_values"][0][0].shape[-2]
```
JingyaHuang, line 599: @fxmarty here I adapt the indexing for fetching the past seq len, will it impact bloom?

JingyaHuang: ok, it's due to the fact that for bloom the shape of key and value are transposed in the last two dims.

optimum/exporters/onnx/model_configs.py
fxmarty: Actually not sure whether llama 2 can be exported with our current code?

optimum/onnxruntime/base.py
```diff
@@ -227,6 +235,14 @@ def prepare_inputs_for_merged(
                 past_key_values = tuple(
                     key_or_value for _ in range(len(self.key_value_input_names) // 2) for key_or_value in [key, value]
                 )
+            elif self.parent_model.config.model_type in MULTI_QUERY_ATTN_MODELS:
```
fxmarty, line 238: This kind of controlflow that we have more and more (for bloom, now models that use multi-query attention) and that is run in the forward makes me think we should somehow subclass, and have `ORTBloomForCausalLM`, `ORTModelForCausalLMWithMQA`.

The solution in https://github.com/huggingface/optimum/pull/1152 is very ugly, but somehow I still think we should be more modular (not in this PR obviously!)

What do you think?

fxmarty (approved):
LGTM

tests/onnxruntime/test_modeling.py
```diff
@@ -1946,6 +1946,7 @@ class ORTModelForCausalLMIntegrationTest(ORTModelTestMixin):
         "bloom",
         "codegen",
         "gpt2",
+        "gpt_bigcode",
```
fxmarty, line 1949: Have you run the tests on GPU?

JingyaHuang: Just triggered the CI on GPU. Only tested on GPU for bigcode on my local.

optimum/onnxruntime/base.py
JingyaHuang: Yes, I also thought of that. It would be better to have separate decoder modeling other than adding these control flows. I just wanted to be align with the case of bloom. We can target a refactoring for that!

optimum/exporters/onnx/model_configs.py
JingyaHuang: @fxmarty I don't get it. Isn't llama already supported by onnx exporter in optimum? How is it related to this PR?

fxmarty: llama 2 (34b and 70b) uses grouped-query attention, and I wondered whether the export would work with the current implementation. But fairly confident nobody will try to export llama 2 70b to ONNX.

JingyaHuang: Ah ha, it would be great to try if so! (however, I haven't got access to the weight of llama2 yet...

BBerabi:
Hi @JingyaHuang ,

Thanks for adding this PR. I am trying to run `bigcode/starcoderbase-1b` with ONMX optimizations under `transformers.Pipeline` and running into several different errors. 

I looked at your PR and the code in detail and identified several potential problems. I would love to hear your thoughts. 

I will write comments directly on the code

optimum/onnxruntime/base.py
```diff
@@ -288,6 +304,24 @@ def compute_past_key_values_output_shapes(
...
+        use_cache_branch: Optional[bool],
+        past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None,
+    ) -> Dict[str, List[int]]:
+        batch_size = input_ids.size(0)
+        num_attention_heads = self.normalized_config.num_attention_heads
+        embed_size_per_head = self.normalized_config.hidden_size // num_attention_heads
+
+        sequence_length = input_ids.size(1)
+        if past_key_values is not None and use_cache_branch is not False:
+            sequence_length += past_key_values[0].size(-2)
```
BBerabi, line 319: This line gives me the error:

`tuple object has no attribute size`. This is expected because as the typing in the function signature says, `past_key_values` is a tuple of tuples. Hence, `past_key_values[0]` is a tuple as well, not a `torch.Tensor`.

You probably copy pasted this signature and the code from the existing function `compute_past_key_values_output_shapes` but before calling that function we flatten `past_key_values` [here](https://github.com/huggingface/optimum/pull/1042/files#diff-b2296ed4846e14e948034c77f89f02dce3eabf97cddb47eac16b29ada232a06dR337). Therefore, the parameter type there is tuple of tensors. Note that the typing in function signature is wrong, because it says `Tuple[Tuple[Tensor]`. 

I think the correct code here should be: 

`sequence_length += past_key_values[0][0].size(-2)`

What do you say?

BBerabi:
After fixing the dimension issue, I ran into another problem. So, [here](https://github.com/huggingface/optimum/pull/1042/files#diff-b2296ed4846e14e948034c77f89f02dce3eabf97cddb47eac16b29ada232a06dR368-R369) we add `past_key_values` to `model_inputs`

But when the model type is query attention, we do not flatten the `past_key_values`. It is of type `Tuple[Tuple[Tensor]` when concatenating with `+operator`, python removes the outer tuple. So concatenating a list of tensor (model_inputs) with a Tuple[Tuple[Tensor]] gives:

[Tensor, Tensor, ..., Tuple[Tensor], Tuple[Tensor],...]. 

I do not think this is what we want because later when binding inputs to gpu, we actually require everything inside `model_inputs` to be a tensor! See [here](https://github.com/huggingface/optimum/blob/main/optimum/onnxruntime/modeling_ort.py#L725-L729).

So, here we end up calling `contiguous` on a tuple and then again it crashes for me with the error: 
`tuple object has no attribute contiguous`. 

I am not sure how to fix this because I do not know how I can apply binding for a tuple given that one of the function argument requires a memory pointer. This is possible to achieve with torch.Tensor via `data_ptr` but I don't know how we can get a pointer to a built-int python tuple. 

What would be the correct fix here?

BBerabi:
Finally, I tried to run this with io_binding disabled. That also does not work, due to the same reason. 

[This part](https://github.com/huggingface/optimum/blob/main/optimum/onnxruntime/base.py#L435-L436) crashes again due to the same reason. 

`past_key_values` is not a tuple of torch because it was not flattened. It is still of type tuple[tuple[tensor]. So, inside the loop, the loop variable `past_key_value` is not a tensor but a tuple of tensors. And tuple of tensors do not have a member function called `cpu`. 


What do you say?

PR #1045, opened 2023-05-09
Optional `attention_mask` in `ORTModelForxxx`

### LINKED ISSUE #1043: optional `attention_mask` in `ORTModelForxxx`
### Feature request

The behavior of some `ORTModelForxxx` differs from that of `AutoModelForxxx`.
For example, `ORTModelForSequenceClassification` doesn't accept `None` as `attention_mask` even though it's an optional argument.
There's also some missing masks (`attention_mask`, `pixel_mask` or `head_mask` for example in audio and vision models) but I think those are intentionally omitted for now since they're not generated in exporters.

### Motivation

Drop in replacement when using `ORTModelForxxx` instead of `AutoModelForxxx`.

### Your contribution

I will submit a PR.

IlyasMoutawwakil:
# What does this PR do?

Fixes #1043 

## Before submitting

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -890,7 +890,7 @@ class ORTModelForFeatureExtraction(ORTModel):
         if self.device.type == "cuda" and self.use_io_binding:
             if attention_mask is None:
                 attention_mask = torch.ones_like(input_ids)
-            
+
             io_binding, output_shapes, output_buffers = self.prepare_io_binding(
                 input_ids,
                 attention_mask,
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -888,6 +888,9 @@ class ORTModelForFeatureExtraction(ORTModel):
         self.raise_on_numpy_input_io_binding(use_torch)
 
         if self.device.type == "cuda" and self.use_io_binding:
+            if attention_mask is None:
+                attention_mask = torch.ones_like(input_ids)
+            
             io_binding, output_shapes, output_buffers = self.prepare_io_binding(
                 input_ids,
                 attention_mask,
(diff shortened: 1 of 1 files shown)


IlyasMoutawwakil:
For `token_type_ids` since they're model dependent (`bert` vs `distilbert`) I guess they should be left the way they are.

michaelbenayoun (approved):
LGTM!
Quick question: it is the same behavior as what's happening in transformers right?

IlyasMoutawwakil:
@michaelbenayoun Yes here's an example:
```python
from optimum.utils import DummyTextInputGenerator, NormalizedTextConfig
from transformers import AutoConfig

MODEL = "distilbert-base-uncased-finetuned-sst-2-english"

dummy_generator = DummyTextInputGenerator(
    task="sequence-classification",
    normalized_config=NormalizedTextConfig(
        AutoConfig.from_pretrained(MODEL)
    )
)

input_ids = dummy_generator.generate('input_ids')
```
this will pass:
```python
from transformers import AutoModelForSequenceClassification

model = AutoModelForSequenceClassification.from_pretrained(MODEL)

model(input_ids=input_ids, attention_mask=None)
```
this will raise a `AttributeError: 'NoneType' object has no attribute 'cpu'` error on `cpu` and a `KeyError: 'attention_mask'` error on `cuda`:
```python
from optimum.onnxruntime import ORTModelForSequenceClassification

ort_model = ORTModelForSequenceClassification.from_pretrained(
    MODEL, 
    export=True,
    use_io_binding=False
).to(DEVICE)

ort_model(input_ids=input_ids, attention_mask=None)
```

`transformers` also generates `token_type_ids`, with `bert-base-uncased` for example, when it's not given. 
Which is practical in some cases (one sentence classification) and misleading in others (two sentences classification).
So it might be worth adding, what do you think ?

IlyasMoutawwakil:
Actually there are many cases handled by this part here:
https://github.com/huggingface/transformers/blob/3335724376319a0c453049d0cd883504f530ff52/src/transformers/models/bert/modeling_bert.py#L965-L1004

michaelbenayoun (approved):
Perfect, thanks for taking care of this @IlyasMoutawwakil !

Issue #1046, opened 2023-05-10
Make torchvision optional?


BramVanroy:
### Feature request

Currently torchvision is a required dependency

https://github.com/huggingface/optimum/blob/22e4fd6de3ac5e7780571570f962947bd8777fd4/setup.py#L20

### Motivation

I only work on text so I don't need vision support

### Your contribution

I am sure the change would be more difficult than just "remove the line from the setup.py" file but if you have other suggestions how to tackle the removal, I am happy to help.

JingyaHuang:
I think we can move `torchvision` from `REQUIRED_PKGS` to `TESTS_REQUIRE` as we did for diffusers, WDYT? Are there any constraints from our subpackages? @regisss @michaelbenayoun @echarlaix

JingyaHuang:
Just checked, should also be ok for the neuron side, `torchvision` is installed explicitly for tests. Not seems to break for intel either. 

@BramVanroy Would you like to submit a PR for it?

PR #1052, opened 2023-05-12
remove torchvision requirement

BramVanroy:
# What does this PR do?

As discussed [here](https://github.com/huggingface/optimum/issues/1046#issuecomment-1545663717) the torchvision is not always needed. As such it can be removed from the requirements although it does need to be available for the tests, so it is now included in `TESTS_REQUIRE`.

Fixes https://github.com/huggingface/optimum/issues/1046#issuecomment-1545663717


JingyaHuang (approved):
LGTM, thanks a lot for contributing!

PR #986, opened 2023-04-19
Add Whisper to ORT optimizer configuration

kunal-vaishnavi:
# What does this PR do?

This PR adds Whisper to the ORT optimizer configuration. This [ORT PR](https://github.com/microsoft/onnxruntime/pull/15473) shows how to use Whisper with ORT.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1068, opened 2023-05-22
added ORTModelForSpeechSeq2Seq support to optimizer

IlyasMoutawwakil:
# What does this PR do?

ORT optimization support for `whisper` was added here #986 but `ORTOptimizer` didn't support `ORTModelForSpeechSeq2Seq`.

@regisss I can add tests if necessary.

## Before submitting

### DIFF
--- a/optimum/onnxruntime/optimization.py
+++ b/optimum/onnxruntime/optimization.py
@@ -31,7 +31,7 @@ from ..utils.save_utils import maybe_save_preprocessors
 from .configuration import OptimizationConfig, ORTConfig
 from .modeling_decoder import ORTModelForCausalLM
 from .modeling_ort import ORTModel
-from .modeling_seq2seq import ORTModelForSeq2SeqLM
+from .modeling_seq2seq import ORTModelForSeq2SeqLM, ORTModelForSpeechSeq2Seq
 from .utils import ONNX_WEIGHTS_NAME, ORTConfigManager
 
 
@@ -89,7 +89,7 @@ class ORTOptimizer:
         config = None
         if isinstance(model_or_path, ORTModel):
             from_ortmodel = True
-            if isinstance(model_or_path, ORTModelForSeq2SeqLM):
+            if isinstance(model_or_path, (ORTModelForSeq2SeqLM, ORTModelForSpeechSeq2Seq)):
                 onnx_model_path += [
                     model_or_path.encoder_model_path,
                     model_or_path.decoder_model_path,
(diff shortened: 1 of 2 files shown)


regisss:
@IlyasMoutawwakil Yes let's add some tests for this. Maybe you can take seq2seq models as a template for the tests: https://github.com/huggingface/optimum/blob/a6951c17c3450e1dea99617aa842334f4e904392/tests/onnxruntime/test_optimization.py#L208

tests/onnxruntime/test_optimization.py
```diff
@@ -322,6 +324,122 @@ def test_optimization_levels_gpu(self, test_name: str, model_arch: str, use_cach
...
+            }
+        )
+    )
+    @require_torch_gpu
+    @pytest.mark.gpu_test
+    def test_optimization_levels_gpu(self, test_name: str, model_arch: str, use_cache: bool, optimization_level: str):
+        for use_io_binding in [False, True]:
+            # TODO: investigate why marian with IO Binding fails
+            if model_arch == "marian" and use_io_binding is True:
+                continue
```
regisss, line 431: I think we can remove this as Marian is not a speech model. And move `use_io_binding` into the `grid_parameters` above.

regisss (approved):
LGTM! Also adding @JingyaHuang to this review as she knows this part of the code better than me :slightly_smiling_face:

optimum/onnxruntime/optimization.py
```diff
@@ -89,7 +89,7 @@ def from_pretrained(
         config = None
         if isinstance(model_or_path, ORTModel):
             from_ortmodel = True
-            if isinstance(model_or_path, ORTModelForSeq2SeqLM):
+            if isinstance(model_or_path, (ORTModelForSeq2SeqLM, ORTModelForSpeechSeq2Seq)):
```
JingyaHuang, line 92: It shall be the same situation for `ORTModelForVision2Seq` isn't it? Why not just use the `ORTModelForConditionalGeneration` base case to check if it is a seq2seq case?

IlyasMoutawwakil: Yes that makes sense! I forgot they share a common base. We should probably also test some vision models and add them to [`ORTConfigManager._conf`](https://github.com/huggingface/optimum/blob/369f44edd38670c68414aa954b9fcd493fa27d7e/optimum/onnxruntime/utils.py#L97).

JingyaHuang (approved):
LGTM, thanks for the fix.

PR #1071, opened 2023-05-24
Add ONNX clip-vision-model support

pakelley:
# What does this PR do?

This PR adds clip-vision-model/ONNX support. I've tested this locally, but haven't gotten docs/tests added yet. Would be happy to get feedback as I get to that over the next couple of days.
Also, this seems to work, but I haven't tested it extensively (I've just saved an ONNX model, loaded it to a `ORTModelForCustomTasks`, and generated embeddings. Not sure what constitutes an appropriate level of testing in this case). Happy to add any tests that seem appropriate here though.

Fixes # (issue)

## Before submitting
- [not yet] Did you make sure to update the documentation with your changes?
- [not yet] Did you write any new necessary tests?

### DIFF
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -670,6 +670,27 @@ class CLIPTextOnnxConfig(TextEncoderOnnxConfig):
         return dummy_inputs
 
 
+class CLIPVisionOnnxConfig(VisionOnnxConfig):
+    ATOL_FOR_VALIDATION = 1e-3
+    # The ONNX export of this architecture needs the Trilu operator support, available since opset 14
+    DEFAULT_ONNX_OPSET = 14
+
+    NORMALIZED_CONFIG_CLASS = NormalizedVisionConfig
+
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        return {
+            "pixel_values": {0: "image_batch_size", 1: "num_channels", 2: "height", 3: "width"},
+        }
+
+    @property
+    def outputs(self) -> Dict[str, Dict[int, str]]:
+        return {
+            "last_hidden_state": {0: "batch_size", 1: "sequence_length"},
+            "pooler_output": {0: "batch_size"},
+        }
+
+
 class UNetOnnxConfig(VisionOnnxConfig):
     ATOL_FOR_VALIDATION = 1e-3
     # The ONNX export of a CLIPText architecture, an other Stable Diffusion component, needs the Trilu
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -373,6 +373,10 @@ class TasksManager:
             "feature-extraction",
             onnx="CLIPTextOnnxConfig",
         ),
+        "clip-vision-model": supported_tasks_mapping(
+            "feature-extraction",
+            onnx="CLIPVisionOnnxConfig",
+        ),
         "codegen": supported_tasks_mapping(
             "feature-extraction",
             "feature-extraction-with-past",
@@ -896,7 +900,7 @@ class TasksManager:
             onnx="YolosOnnxConfig",
         ),
     }
-    _UNSUPPORTED_CLI_MODEL_TYPE = {"unet", "vae-encoder", "vae-decoder", "clip-text-model", "trocr"}
+    _UNSUPPORTED_CLI_MODEL_TYPE = {"unet", "vae-encoder", "vae-decoder", "clip-text-model", "clip-vision-model", "trocr"}
     _SUPPORTED_CLI_MODEL_TYPE = set(_SUPPORTED_MODEL_TYPE.keys()) - _UNSUPPORTED_CLI_MODEL_TYPE
 
     @classmethod


pakelley:
I resolved the conflicts here, fwiw.
Similar to [my PR here](https://github.com/huggingface/optimum/pull/1263), I'm having trouble getting the tests to run (getting a segfault for `exporters/onnx/test_onnx_config_loss.py::TestOnnxConfigWithLoss::test_onnx_config_with_loss` and `ValueError: Unrecognized model in philschmid/unit_test_model.` for `test_modeling_base.py::TestOptimizedModel::test_load_model_from_hub`, even on `main`.)
I don't really see docs for the `clip_text_model` model type, so do I need to add docs for this? If there's a natural place to add them I'm happy to do so.

Also, haven't gotten much attention here but I'm not sure who to @... if anyone happens to know who'd be appropriate lmk, otherwise I'll just find someone who seems active.

regisss (commented):
Hi @pakelley! What is the model and the command you use to test your PR?

optimum/exporters/onnx/model_configs.py
```diff
@@ -723,6 +723,27 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
...
+        return {
+            "pixel_values": {0: "image_batch_size", 1: "num_channels", 2: "height", 3: "width"},
+        }
+
+    @property
+    def outputs(self) -> Dict[str, Dict[int, str]]:
+        return {
+            "last_hidden_state": {0: "batch_size", 1: "sequence_length"},
+            "pooler_output": {0: "batch_size"},
+        }
```
regisss, line 744:
```suggestion
class CLIPVisionOnnxConfig(ViTOnnxConfig):
    pass
```
Just inheriting from ViT's ONNX config seems to work according to the discussion in this issue: https://github.com/huggingface/optimum/issues/1186#issuecomment-1637641760

pakelley:
Just updated to inherit from the `ViTOnnxConfig`, and include the with-projection version of the model.
The `ATOL_FOR_VALIDATION` seems to be necessary for the validation to run, but I'm not sure about the `DEFAULT_ONNX_OPSET`. I carried that over from the other CLIP models, but I'm happy to remove it if it isn't necessary.

I've tested (both with and without projection) using the `openai/clip-vit-base-patch32` model.
I used this to write/validate the model:
```
from pathlib import Path

from optimum.exporters.onnx.convert import export_models, validate_models_outputs
from optimum.exporters.onnx.model_configs import CLIPVisionWithProjectionOnnxConfig, CLIPVisionOnnxConfig
from transformers import CLIPVisionModelWithProjection

BASE_MODEL = "openai/clip-vit-base-patch32"
OUTPUT_DIR = Path("clip-image-model").expanduser()

model = CLIPVisionModelWithProjection.from_pretrained(BASE_MODEL)
models_and_onnx_configs = {"model": (model, CLIPVisionWithProjectionOnnxConfig(model.config))}

# save model
_, onnx_outputs = export_models(
    models_and_onnx_configs=models_and_onnx_configs,
    output_dir=OUTPUT_DIR,
)

# validate model
validate_models_outputs(
    models_and_onnx_configs=models_and_onnx_configs,
    onnx_named_outputs=onnx_outputs,
    output_dir=OUTPUT_DIR,
    onnx_files_subpaths=["model.onnx"],
    device="cpu",
)
```

And then this to run the model:
```
import torch
from optimum.onnxruntime import ORTModelForCustomTasks

ort_model = ORTModelForCustomTasks.from_pretrained(
    OUTPUT_DIR, config=model.config
)

pixel_values = torch.zeros((10, 3, 224, 224))
ort_model(pixel_values=pixel_values)
```
Maybe there's a better way than using `ORTModelForCustomTasks`, but that seems to work fine for me 🙂 
Happy to clarify anything else!

pakelley:
I've also tested optimizing these image models, but the changes are dependent on #1263 . Depending on which gets merged first, I can add those changes to either PR, if that sounds good to you.

optimum/exporters/tasks.py
```diff
@@ -392,6 +392,10 @@ class TasksManager:
             "feature-extraction",
             onnx="CLIPTextOnnxConfig",
         ),
+        "clip-vision-model": supported_tasks_mapping(
```
regisss, line 395: Maybe you can add `clip-vision-with-projection` since you also added an ONNX config for the model with projection. Like `clip-text-with-projection` right below.

optimum/exporters/tasks.py
```diff
@@ -952,6 +956,7 @@ class TasksManager:
         "vae-decoder",
         "clip-text-model",
         "clip-text-with-projection",
+        "clip-vision-model",
```
regisss, line 959: Same here

optimum/exporters/onnx/model_configs.py
```diff
@@ -723,6 +723,35 @@ def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
         return dummy_inputs
 
 
+class CLIPVisionWithProjectionOnnxConfig(ViTOnnxConfig):
+    ATOL_FOR_VALIDATION = 1e-3
+    # The ONNX export of this architecture needs the Trilu operator support, available since opset 14
+    DEFAULT_ONNX_OPSET = 14
```
regisss, line 729: Regarding `DEFAULT_ONNX_OPSET`, if you're not sure, you can quickly try smaller values and see if that works.

regisss (commented):
@pakelley Thanks for updating the PR! I left a few minor comments :slightly_smiling_face: 

The code you used for testing the export looks good to me :+1:

Let's merge this PR first and then we'll check in #1263 if optimized model runs properly.

pakelley:
I added the `clip-vision-with-projection` entries, but keep running into issues with trying to install onnx 1.8 (to test onnx opset 13).
Is there an easier way to test a different onnx opset than installing an onnx version where that's the max supported opset?

pakelley:
@regisss any chance you know of a simple way to test a different onnx opset without installing an older version of onnx? Or would you by any chance be able to test this on an older version, if you have an existing environment built?
If not, I can try again to set up an env, but I'm hoping to avoid spending a few hours debugging env/dependency issues 😅 
Aside from that, let me know if there's anything else you'd like changed in this PR, happy to clean anything else up.

regisss:
So setting `DEFAULT_ONNX_OPSET` to 13 doesn't work?

pakelley:
My script still seems to work if I set `DEFAULT_ONNX_OPSET` to 13, I'm probably just misunderstanding.
I was assuming that wouldn't do anything to change the opset being used, since the version of onnx I have installed uses opset 14.
If simply changing that value is enough to test I think we're safe to get rid of the `DEFAULT_ONNX_OPSET`, but I'll do one more test without it to make sure.

echarlaix:
Clip vision model ONNX export https://github.com/huggingface/optimum/pull/1920 so closing, feel free to re-open if I missed something

Issue #775, opened 2023-02-14
ORTModelForCausalLM batch inference is not consistent with infernece with batch size of 1


snehith-r:
I've a finetuned model of gpt2 to perform automated named entity recognition. For improving the inference speed I've used optimum to convert the model to onnx and use CUDAEXECUTION provide with IO bindings. With a batch size of 1 the attribute-value pairs are being extracted for each input, but If I use a batch size of 32 only attributes of a sequence are being generated. When I use batch size of 1 with the same 32 sequences it's extracting attribute-value pairs for each of the sequences. Idk what's causing failure in the inference of batch data. Any help would be appreciated.

TIA

snehith-r:
It was due to the pad token argument in the generate function. But I've observed that onnx converted model is faster than pytorch with accelerate only if the batch size is 1 and pytorch with accelerate is outperforming the onnx inference on cuda if batch size is 32. Will share the code to reproduce once I check if it's the same with the pretrained Gpt2 provided in transformers library, for the scenario I mentioned it's a finetuned version trained on personal dataset.

snehith-r:
@fxmarty  The code for comparing inference time, with increase in batch size(>1) onnx model is slower than pytorch
`from transformers import AutoTokenizer, GPT2TokenizerFast
from transformers import GPT2LMHeadModel,AdamW , GPT2Tokenizer, OpenAIGPTConfig

from transformers import AutoTokenizer

from optimum.onnxruntime import ORTModelForCausalLM
import torch

device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")

tokenizer = AutoTokenizer.from_pretrained("gpt2-large")
tokenizer.padding_side = "left"
tokenizer.pad_token = '<pad>'

sentneces=[ 'The people of the State of California do enact as follows:\n\n\nSECTION 1.\nSection 10295.35 is added to the Public Contract Code, to read:\n10295.35.\n(a) (1)', 
'Notwithstanding any other law, a state agency shall not enter into any contract for the acquisition of goods or services in the amount of one hundred thousand dollars ($100,000) or more with a contractor that, in the provision of benefits, discriminates between employees on the basis of an employees or dependent’s actual or perceived gender identity, including, but not limited to,' ,
[...]

PR #1076, opened 2023-05-31
Use mask for seq2seq ONNX decoder models 

fxmarty:
As per title. I am not sure why the encoder attention mask for seq2seq models was not used up to now. The bug may have been introduced in https://github.com/huggingface/optimum/pull/698.

Fixes https://github.com/huggingface/optimum/issues/775


michaelbenayoun (approved):
LGTM!
Thanks for fixing

PR #1082, opened 2023-06-05
Add a task detection class method in `OptimizedModel`

JingyaHuang:
# What does this PR do?

as per title.

### DIFF
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -24,6 +24,7 @@ from typing import TYPE_CHECKING, Optional, Union
 from huggingface_hub import HfApi, HfFolder
 from transformers import AutoConfig, add_start_docstrings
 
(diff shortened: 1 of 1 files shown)


optimum/modeling_base.py
```diff
@@ -370,3 +371,10 @@ def from_pretrained(
             trust_remote_code=trust_remote_code,
             **kwargs,
         )
+
+    @classmethod
+    def _auto_model_to_task(cls, auto_model_class):
+        """
+        Get the task corresponding to a class (for example AutoModelForXXX in transformers).
+        """
+        return TasksManager.infer_task_from_model(auto_model_class)
```
fxmarty, line 380: How different is this classmethod different from directly calling `TasksManager.infer_task_from_model(auto_model_class)`?

PR #1087, opened 2023-06-07
Enable use_io_binding = True on CPU

yihonglyu:
# What does this PR do?

KV cache is large for large prompt (~1GB in total for prompt 1024) but onnxruntime python binding copies all outputs from ORT memory to Python memory. It is resolved with IO binding. With this patch, we can use code like
```
    model_path,
    use_io_binding = True,
)
```
to enable IO binding with CPU

## Before submitting

### DIFF
--- a/optimum/onnxruntime/base.py
+++ b/optimum/onnxruntime/base.py
@@ -311,7 +311,7 @@ class ORTDecoder(ORTModelPart):
             input_ids, past_key_values, use_torch=use_torch
         )
 
-        if self.device.type == "cuda" and self.parent_model.use_io_binding:
+        if (self.device.type == "cuda" or self.device.type == "cpu") and self.parent_model.use_io_binding:
             known_output_shapes = self.compute_past_key_values_output_shapes(
                 input_ids,
                 use_cache_branch=use_cache_branch_tensor.item() if use_cache_branch_tensor is not None else None,
--- a/optimum/onnxruntime/base.py
+++ b/optimum/onnxruntime/base.py
@@ -344,10 +344,11 @@ class ORTDecoder(ORTModelPart):
                 ordered_input_names=self._ordered_input_names,
             )
 
-            if self.device.type != "cpu":
+            if self.device.type == "cpu":
+                self.session.run_with_iobinding(io_binding)
+            else:
                 io_binding.synchronize_inputs()
-            self.session.run_with_iobinding(io_binding)
-            if self.device.type != "cpu":
+                self.session.run_with_iobinding(io_binding)
                 io_binding.synchronize_outputs()
 
             # Tuple of length equal to : number of layer * number of past_key_value per decoder layer(2)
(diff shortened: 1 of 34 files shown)


fxmarty:
I indeed see a slight speedup, but I tested with a model not really large (gpt2-large, 1000 token context length, batch size = 1, about ~7% speedup). I am kind of surprised we would see any speedup because of this: https://github.com/microsoft/onnxruntime/issues/10000#issuecomment-991220117 

It could be that there is some kind of overhead in the non-IO Binding path in Optimum that is not present in the IOBinding one.

What kind of speedup do you see?

fxmarty:
Hi @tianleiwu, do you have any thoughts on this?

tianleiwu:
@fxmarty, 

IO Binding for CPU can help performance when input/output tensor is huge (like past_key_value for long sequence length). Without IO Binding, the interop of python and ONNX Runtime need copy of tensor (between numpy tensor and OrtValue), and memory copy is slow when tensor size is large.

optimum/onnxruntime/base.py
```diff
@@ -311,7 +311,7 @@ def forward(
             input_ids, past_key_values, use_torch=use_torch
         )
 
-        if self.device.type == "cuda" and self.parent_model.use_io_binding:
+        if (self.device.type == "cuda" or self.device.type == "cpu") and self.parent_model.use_io_binding:
```
tianleiwu, line 314: For CPU, it does not need synchronize_inputs() synchronize_outputs in line 343 below.

fxmarty:
Thank you for the details @tianleiwu!

optimum/onnxruntime/base.py
yihonglyu: Exclude synchronize_inputs() for CPU

optimum/onnxruntime/utils.py
```diff
@@ -252,7 +252,7 @@ def check_io_binding(providers: List[str], use_io_binding: Optional[bool] = None
     """
     if use_io_binding is None and providers[0] == "CUDAExecutionProvider":
         use_io_binding = True
-    elif providers[0] != "CUDAExecutionProvider":
+    elif providers[0] != "CPUExecutionProvider" and providers[0] != "CUDAExecutionProvider":
         if use_io_binding is True:
             logger.warning(
                 "No need to enable IO Binding if the provider used is not CUDAExecutionProvider. IO Binding will be turned off."
```
tianleiwu, line 258: update message to include CPUExecutionProvider

yihonglyu: Updated the message.

tianleiwu (approved):

yihonglyu:
@fxmarty Could you give me some review comments?

fxmarty:
For some reason I have shape errors with batch size > 1 with CPU EP + IO Binding with this PR, while fine on CUDA EP. Will investigate and get back to you shortly.

fxmarty (approved):
LGTM, actually made it the default for ORTModelForCausalLM. Could be made default for encoder-decoder models as well.

Issue #1088, opened 2023-06-08
from_pretained typo and can not use config.json in optimum-cli quantize


YooSungHyun:
### System Info

```shell
optimum==1.8.6
```

### Who can help?

@michaelbenayoun @jin

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction

Use optimum-cli, and must use config

example
optimum-cli onnxruntime quantize --onnx_model ~~~ -o ~~~ -c ~~~/quantize_config.json

then error accur
Traceback (most recent call last):
  File "/usr/local/bin/optimum-cli", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/dist-packages/optimum/commands/optimum_cli.py", line 163, in main
    service.run()
  File "/usr/local/lib/python3.10/dist-packages/optimum/commands/onnxruntime/quantize.py", line 97, in run
    qconfig = ORTConfig.from_pretained(self.args.config).quantization
AttributeError: type object 'ORTConfig' has no attribute 'from_pretained'. Did you mean: 'from_pretrained'?

see this line. it is typo
https://github.com/huggingface/optimum/blob/main/optimum/commands/onnxruntime/quantize.py#L97

from_pretained -> from_pretrained

### Expected behavior

success quantize, using config json

YooSungHyun:
and i think `ORTConfig.from_pretrained(self.args.config).quantization` is just dict. not QuantizationConfig. so, if `qconfig` input `quantize` function, it error occur (AttributeError: 'dict' object has no attribute 'is_static')

YooSungHyun:
how about this way?

optimum.onnxruntime.configuration.py
``` python3
@staticmethod
def from_config(config: dict) -> QuantizationConfig:
    """
    Creates a [`~onnxruntime.QuantizationConfig`] from input config json file (default: ort_config.json)
    Args:
        config (`dict`):
            Input ONNX Runtime config file's (e.g. ort_config.json) `quantization` attribute
    """
    format, mode = default_quantization_parameters(is_static=config["is_static"])
    if config["activations_dtype"].lower() in ["u8", "quint8", "uint8"]:
        activations_dtype = QuantType.QUInt8
    else:
        activations_dtype = QuantType.QInt8
[...]

PR #1089, opened 2023-06-08
Update: fix typo `from_pretained` to `from_pretrained`

YooSungHyun:
# What does this PR do?
Fix typo `from_pretained` to `from_pretrained`

Fixes # (issue)
https://github.com/huggingface/optimum/issues/1088

## Before submitting

chk plz! @michaelbenayoun @jin

### DIFF
--- a/optimum/commands/onnxruntime/quantize.py
+++ b/optimum/commands/onnxruntime/quantize.py
@@ -97,7 +97,7 @@ class ONNXRuntimmeQuantizeCommand(BaseOptimumCLICommand):
         else:
             ort_config = ORTConfig.from_pretrained(self.args.config)
             use_external_data_format = ort_config.use_external_data_format
-            qconfig = AutoQuantizationConfig.from_config(config=ort_config.quantization)
+            qconfig = ort_config.quantization
 
         for q in quantizers:
             q.quantize(
--- a/optimum/commands/onnxruntime/quantize.py
+++ b/optimum/commands/onnxruntime/quantize.py
@@ -94,7 +94,7 @@ class ONNXRuntimmeQuantizeCommand(BaseOptimumCLICommand):
         elif self.args.tensorrt:
             qconfig = AutoQuantizationConfig.tensorrt(per_channel=self.args.per_channel)
         else:
-            qconfig = ORTConfig.from_pretained(self.args.config).quantization
+            qconfig = ORTConfig.from_pretrained(self.args.config).quantization
 
         for q in quantizers:
             q.quantize(save_dir=save_dir, quantization_config=qconfig)
--- a/optimum/commands/onnxruntime/quantize.py
+++ b/optimum/commands/onnxruntime/quantize.py
@@ -83,6 +83,7 @@ class ONNXRuntimmeQuantizeCommand(BaseOptimumCLICommand):
             for model in self.args.onnx_model.glob("*.onnx")
         ]
 
+        use_external_data_format = False
         if self.args.arm64:
             qconfig = AutoQuantizationConfig.arm64(is_static=False, per_channel=self.args.per_channel)
         elif self.args.avx2:
@@ -91,10 +92,12 @@ class ONNXRuntimmeQuantizeCommand(BaseOptimumCLICommand):
             qconfig = AutoQuantizationConfig.avx512(is_static=False, per_channel=self.args.per_channel)
         elif self.args.avx512_vnni:
             qconfig = AutoQuantizationConfig.avx512_vnni(is_static=False, per_channel=self.args.per_channel)
-        elif self.args.tensorrt:
-            qconfig = AutoQuantizationConfig.tensorrt(per_channel=self.args.per_channel)
         else:
-            qconfig = ORTConfig.from_pretrained(self.args.config).quantization
+            ort_config = ORTConfig.from_pretrained(self.args.config)
+            use_external_data_format = ort_config.use_external_data_format
+            qconfig = AutoQuantizationConfig.from_config(config=ort_config.quantization)
 
         for q in quantizers:
-            q.quantize(save_dir=save_dir, quantization_config=qconfig)
+            q.quantize(
+                save_dir=save_dir, quantization_config=qconfig, use_external_data_format=use_external_data_format
+            )
--- a/optimum/commands/onnxruntime/quantize.py
+++ b/optimum/commands/onnxruntime/quantize.py
@@ -92,6 +92,8 @@ class ONNXRuntimmeQuantizeCommand(BaseOptimumCLICommand):
             qconfig = AutoQuantizationConfig.avx512(is_static=False, per_channel=self.args.per_channel)
         elif self.args.avx512_vnni:
             qconfig = AutoQuantizationConfig.avx512_vnni(is_static=False, per_channel=self.args.per_channel)
+        elif self.args.tensorrt:
+            qconfig = AutoQuantizationConfig.tensorrt(per_channel=self.args.per_channel)
         else:
             ort_config = ORTConfig.from_pretrained(self.args.config)
             use_external_data_format = ort_config.use_external_data_format
--- a/optimum/onnxruntime/configuration.py
+++ b/optimum/onnxruntime/configuration.py
@@ -19,7 +19,7 @@ import warnings
 from dataclasses import asdict, dataclass, field
 from enum import Enum
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple, Union, Any
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from datasets import Dataset
 from packaging.version import Version, parse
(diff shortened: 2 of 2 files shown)


YooSungHyun:
"maybe this is lookin small, but can made big wave...! 😂"

michaelbenayoun:
@YooSungHyun Let's talk about the issues here instead than #1088.

About quantization attribute not being a `QuantizationConfig` but a dictionary, you are right, and this is bad because it is actually documented as a `QuantizationConfig`.

You can also fix this here, by doing in `optimum/onnxruntime/configuration.py`:

```python
def __init__(                                                                                                                                                                           
    self,                                                                                                                                                                                         
    opset: Optional[int] = None,                                                                                                                                                                  
    use_external_data_format: bool = False,                                                                                                                                             
    one_external_file: bool = True,                                                                                                                                                     
    optimization: Optional[OptimizationConfig] = None,                                                                                                                                  
    quantization: Optional[QuantizationConfig] = None,                                                                                                                                  
    **kwargs,                                                                                                                                                                           
      ): -> None                                                                                                                                                                              
          super().__init__()                                                                                                                                                                  
          self.opset = opset                                                                                                                                                   
          self.use_external_data_format = use_external_data_format                                                                                                                   
          self.one_external_file = one_external_file                                                                                                                      
          self.optimization = optimization                                                                                                                      
          self.quantization = quantization                                                                                                                     
          self.optimum_version = kwargs.pop("optimum_version", None)    
```

Basically the only difference here is that we do not call the `dataclass_to_dict` method.

Then to be able to serialize `ORTConfig`, you will need to override the `to_dict` method:

```python
def to_dict(self) -> Dict[str, Any]:
    clone = copy.deepcopy(self) : Self@ORTConfig                                                                                                                                        
    clone.optimization = self.dataclass_to_dict(clone.optimization)
    clone.quantization = self.dataclass_to_dict(clone.quantization) : dict[Unknown, Unknown]                                                                                            
    return BaseConfig.to_dict(clone)
```

YooSungHyun:
@michaelbenayoun you mean, convert to `OptimizationConfig` and `QuantizationConfig` in `__init__():` and if when i needs some `save_pretrained()` kind of thing, then `ORTConfig().to_dict()` -> `save_pretrained()`??

michaelbenayoun:
No I mean:

1. No transforming the `optimization` and `quantization` attributes to dictionaries in `ORTConfig.__init__`.
2. Doing that will make the `ORTConfig` not JSON-serializable, to make it serializable, just add the `to_dict` method I shared to the `ORTConfig` class.

YooSungHyun:
first of all, i commit first. so let we talk with `code conversation`..!

YooSungHyun:
@michaelbenayoun 
first, i don't speak english as well. so i may not understand it very well..but, I want to contribute this code intensely..! so.... plz bear with me🤗

YooSungHyun:
i will convert source code about `ORTConfig.from_pretrained()` ->  `ORTConfig.from_dict()`
i will override `PretrainedConfig.from_dict()` in `ORTConfig`. so, It will just call `ORTConfig.from_pretrained()` then, `optimization` and `quantization` attribute will be each `OptimizationConfig` and `Quantizationconfig`

optimum/onnxruntime/configuration.py
```diff
@@ -612,6 +613,40 @@ def tensorrt(
...
+            weights_symmetric=config["weights_symmetric"],
+            per_channel=config["per_channel"],
+            reduce_range=config["reduce_range"],
+            nodes_to_quantize=config["nodes_to_quantize"] or [],
+            nodes_to_exclude=config["nodes_to_exclude"] or [],
+            operators_to_quantize=config["operators_to_quantize"] or ["MatMul", "Add"],
+            qdq_add_pair_to_weight=config["qdq_add_pair_to_weight"],
+            qdq_dedicated_pair=config["qdq_dedicated_pair"],
+        )
+
```
michaelbenayoun, line 649: Why?

YooSungHyun:
@michaelbenayoun  Ta-da! i've changed all my code.
i think, before version source-code is two-step necessary.
1. load all ort config
2. make QuantizationConfig
and this way, if someone needs OptimizationConfig, he or she must write down OptimizationConfig's load code. but, QuantizationConfig is not have any side-effect

but, look at after version...
just override `from_dict` for `BaseConfig.from_pretrained()`, it can will be load all each config classes (OptimizationConfig and QuantizationConfig) at once.
and also, using `getattr`, we can load exactly type (if is none, will be default value)
if, someone needs OptimizationConfig, he or she just write down under the `if config.optimization:`, or maybe all loaded well.

finally, at `quantize.py` it is more scalable, because just using ORTConfig is ORTConfig, not any additional change (what if using additional config value)

i deleted `super.__init__()` that is make not used additional parameter
and i think `pruned_heads` and some like `label` kind of things is not using ORTConfig, so deleted too.

this way, is using `override`, so i think have some side-effect problem. so code review is necessary

i tested it well done, and when if config json key missing, it will set default value...
how about it?

optimum/onnxruntime/configuration.py
```diff
@@ -965,3 +965,66 @@ def dataclass_to_dict(config) -> dict:
...
+                if key != "torch_dtype":
+                    to_remove.append(key)
+        for key in to_remove:
+            kwargs.pop(key, None)
+
+        logger.info(f"ORTConfig: {config.to_dict()}")
+        if return_unused_kwargs:
+            return config, kwargs
+        else:
+            return config
```
michaelbenayoun, line 1030: I think a better approach would be to do:

```suggestion
    def from_dict(cls, config_dict: Dict[str, Any], **kwargs) -> "PretrainedConfig":
         config = super().from_dict(config_dict, **kwargs)
         if config.quantization:
             config.quantization = QuantizationConfig(**config.quantization)
         else:
             config.quantization = None
         # Same thing for optimization  
```

This way we duplicate less code.

YooSungHyun:
[69ba1c7](https://github.com/huggingface/optimum/pull/1089/commits/69ba1c76dec1676c997559d4e0bb61fad5bfd30c)

is good i think. how about that?

@michaelbenayoun

YooSungHyun:
@fxmarty @michaelbenayoun 🤗

fxmarty:
Thank you @YooSungHyun and apologizes for the delay, I'll give a review on Monday!

fxmarty:
Can you run `make style`?

optimum/commands/onnxruntime/quantize.py
```diff
@@ -83,6 +83,7 @@ def run(self):
             for model in self.args.onnx_model.glob("*.onnx")
         ]
 
+        use_external_data_format = False
```
fxmarty, line 86: Why is it hard-coded here?

optimum/onnxruntime/configuration.py
```diff
@@ -343,6 +344,13 @@ def ensure_valid_data_type_or_raise(
             "activations_dtype = QuantType.QInt8 with weights_dtype = QuantType.QUInt8."
         )
 
+def set_quantization_parameters(is_static, format, mode, activations_dtype, weights_dtype):
+    format = getattr(QuantFormat, str(format or ""), None),
```
fxmarty, line 348: Why `str(format or "")` (here and the following)?

optimum/onnxruntime/configuration.py
```diff
@@ -343,6 +344,13 @@ def ensure_valid_data_type_or_raise(
             "activations_dtype = QuantType.QInt8 with weights_dtype = QuantType.QUInt8."
         )
 
+def set_quantization_parameters(is_static, format, mode, activations_dtype, weights_dtype):
```
fxmarty, line 347: It could be nice to have type hints here.

optimum/onnxruntime/configuration.py
```diff
@@ -943,12 +951,11 @@ def __init__(
         quantization: Optional[QuantizationConfig] = None,
```
fxmarty, line 951: So are these type hints correct in the end? `optimization: Optional[OptimizationConfig] = None,` and `quantization: Optional[QuantizationConfig] = None,`? According to https://github.com/huggingface/optimum/pull/1089#issuecomment-1582233708 they are not?

optimum/onnxruntime/configuration.py
```diff
@@ -344,6 +345,41 @@ def ensure_valid_data_type_or_raise(
         )
 
 
+def set_quantization_parameters(
+    is_static: bool, format: str = "", mode: str = "", activations_dtype: str = "", weights_dtype: str = ""
+):
+    """
+    Get each QuantFormat, QuantizationMode, QuantType to str in ort_config.json
```
michaelbenayoun, line 352:
```suggestion
    Converts `QuantFormat`, `QuantizationMode`, `QuantType` to string for serialization.
```

optimum/onnxruntime/configuration.py
```diff
@@ -344,6 +345,41 @@ def ensure_valid_data_type_or_raise(
...
+    So, each onnx runtime config parameters is had to name each classes's class attribute names
```
michaelbenayoun, line 353: I do not understand.

optimum/onnxruntime/configuration.py
```diff
@@ -344,6 +345,41 @@ def ensure_valid_data_type_or_raise(
...
+
+    Example) onnxruntime/onnxruntime/python/tools/quantization/quant_utils.py
```
michaelbenayoun, line 355:
```suggestion
    Example: onnxruntime/onnxruntime/python/tools/quantization/quant_utils.py
```

optimum/onnxruntime/configuration.py
```diff
@@ -344,6 +345,41 @@ def ensure_valid_data_type_or_raise(
...
+        - QuantFormat.QOperator         = "QOperator"
+        - QuantFormat.QDQ               = "QDQ"
+        - QuantizationMode.IntegerOps   = "IntegerOps"
+        - QuantizationMode.QLinearOps   = "QLinearOps"
+        - QuantType.QInt8               = "QInt8"
+        - QuantType.QUInt8              = "QUInt8"
+    Args:
```
michaelbenayoun, line 362:
```suggestion
        - QuantType.QUInt8              = "QUInt8"
        
    Args:
```

optimum/onnxruntime/configuration.py
```diff
@@ -965,3 +1000,34 @@ def dataclass_to_dict(config) -> dict:
...
+                quant_config["mode"],
+                quant_config["activations_dtype"],
+                quant_config["weights_dtype"],
+            ) = set_quantization_parameters(
+                quant_config["is_static"],
+                quant_config.get("format", ""),
+                quant_config.get("mode", ""),
+                quant_config.get("activations_dtype", ""),
+                quant_config.get("weights_dtype", ""),
+            )
```
michaelbenayoun, line 1030: Not very readable IMO

YooSungHyun:
@fxmarty

YooSungHyun:
@fxmarty @michaelbenayoun  plz check...!😅

optimum/onnxruntime/configuration.py
fxmarty: bump

PR #1097, opened 2023-06-09
Fix transformers version 

echarlaix:
Limit `transformers` version to avoind incompatibility issues and add warning log for `diffusers>=0.17.0` (as some subpackage like openvino)

### DIFF
--- a/optimum/exporters/onnx/utils.py
+++ b/optimum/exporters/onnx/utils.py
@@ -45,7 +45,7 @@ if is_diffusers_available():
     TMP_DIFFUSERS_MAX_VERSION = "0.17.0"
     if check_if_diffusers_greater(TMP_DIFFUSERS_MAX_VERSION):
         logger.warning(
-            f"We found an older version of diffusers {_diffusers_version} but we require diffusers to be < {TMP_DIFFUSERS_MAX_VERSION}."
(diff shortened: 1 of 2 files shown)


optimum/exporters/onnx/utils.py
```diff
@@ -26,18 +26,27 @@
...
 if is_diffusers_available():
     if not check_if_diffusers_greater(DIFFUSERS_MINIMUM_VERSION.base_version):
         raise ImportError(
             f"We found an older version of diffusers {_diffusers_version} but we require diffusers to be >= {DIFFUSERS_MINIMUM_VERSION}. "
             "Please update diffusers by running `pip install --upgrade diffusers`"
         )
+    TMP_DIFFUSERS_MAX_VERSION = "0.17.0"
+    if check_if_diffusers_greater(TMP_DIFFUSERS_MAX_VERSION):
+        logger.warning(
+            f"We found an older version of diffusers {_diffusers_version} but we require diffusers to be < {TMP_DIFFUSERS_MAX_VERSION}."
```
regisss, line 48:
```suggestion
            f"We found an newer version of diffusers {_diffusers_version} but we require diffusers to be < {TMP_DIFFUSERS_MAX_VERSION}."
```

regisss (approved):
Thanks @echarlaix!
I just left one nit.

PR #1099, opened 2023-06-12
Adamlouly/fix unwrap model eval

AdamLouly:
# Fixing unwrap model when using ModuleWithLoss.

When using the --module_with_loss flag, if we run training and then evaluation right after it in t5-base, it will fail with this error:
```
AttributeError: 'ModuleWithLoss' object has no attribute 'main_input_name'
```
The error occurs because it tries to execute the following part:
```
packages/optimum/onnxruntime/trainer_seq2seq.py", line 675, in prediction_step
    generation_inputs = inputs[self.model.main_input_name]
 ```
The failure indicates that the model hasn't switched back from the ModuleWithLoss class to Module, which causes the failure.

The reason is that we used _unwrap_model, thinking it would convert ModuleWithLoss back to Module, but that's not the case. The unwrap_model function is intended to recursively unwrap a model from potential containers that might be used during distributed training.

Therefore, in this case, we need to use:
```
model = model._original_model
```
to fix this issue.

### DIFF
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -907,7 +907,8 @@ class ORTTrainer(Trainer):
         """
         # memory metrics - must set up as early as possible
         # TODO: We need to enable evaluation using ORT backend.
-        self.model = self.model._original_model
+        if self.args.use_module_with_loss:
+            self.model = self.model._original_model
         self._memory_tracker.start()
 
         eval_dataloader = self.get_eval_dataloader(eval_dataset)
@@ -1000,7 +1001,8 @@ class ORTTrainer(Trainer):
               labels).
         """
         # TODO: We need to enable evaluation using ORT backend.
-        self.model = self.model._original_model
+        if self.args.use_module_with_loss:
+            self.model = self.model._original_model
 
         # memory metrics - must set up as early as possible
         self._memory_tracker.start()
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -907,7 +907,7 @@ class ORTTrainer(Trainer):
         """
         # memory metrics - must set up as early as possible
         # TODO: We need to enable evaluation using ORT backend.
-        self.model = unwrap_model(self.model)
+        self.model = self.model._original_model
         self._memory_tracker.start()
 
         eval_dataloader = self.get_eval_dataloader(eval_dataset)
@@ -1000,7 +1000,7 @@ class ORTTrainer(Trainer):
               labels).
         """
         # TODO: We need to enable evaluation using ORT backend.
-        self.model = unwrap_model(self.model)
+        self.model = self.model._original_model
 
         # memory metrics - must set up as early as possible
         self._memory_tracker.start()
--- a/docs/source/onnxruntime/package_reference/modeling_ort.mdx
+++ b/docs/source/onnxruntime/package_reference/modeling_ort.mdx
@@ -48,13 +48,13 @@ The following ORT classes are available for the following natural language proce
 
 [[autodoc]] onnxruntime.ORTModelForMultipleChoice
 
-## Computer vision
-
-The following ORT classes are available for the following computer vision tasks.
-
 ### ORTModelForQuestionAnswering
 
 [[autodoc]] onnxruntime.ORTModelForQuestionAnswering
+
+## Computer vision
+
+The following ORT classes are available for the following computer vision tasks.
 
 ### ORTModelForImageClassification
 

### CODE REVIEW

**Comment by AdamLouly** [comment]
@JingyaHuang  quick fix here, Thank you!

File: optimum/onnxruntime/trainer.py [hunk 1]
```diff
@@ -907,7 +907,7 @@ def evaluate(
         """
         # memory metrics - must set up as early as possible
         # TODO: We need to enable evaluation using ORT backend.
```
pengwa, line 909: in the PR description, it mentions "packages/optimum/onnxruntime/trainer_seq2seq.py", line 675, in prediction_step
    generation_inputs = inputs[self.model.main_input_name]"

but why we need fix trainer.py instead of trainer_seq2seq.py?

optimum/onnxruntime/trainer.py
```diff
@@ -907,7 +907,7 @@ def evaluate(
...
-        self.model = unwrap_model(self.model)
```
pengwa, line 910: Another question is, why only t5 model evaluation fails, but other models are good to evaluate?

optimum/onnxruntime/trainer.py
AdamLouly: we fixed trainer.py in optimum.
because that's where we have the ModuleWithLoss class.

this is the whole stack trace :
```
***** train metrics *****
  epoch                    =        1.0
  train_loss               =     0.0979
  train_runtime            = 0:08:14.83
  train_samples            =     610320
  train_samples_per_second =   1233.374
  train_steps_per_second   =       4.82
[INFO] Evaluating with PyTorch backend. If you want to use ONNX Runtime for the evaluation, set `trainer.evaluate(inference_with_ort=True)`.
'ModuleWithLoss' object has no attribute 'main_input_name'
  File "run_translation.py", line 680, in <module>
    main()
  File "run_translation.py", line 612, in main
    metrics = trainer.evaluate(
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/optimum/onnxruntime/trainer_seq2seq.py", line 92, in evaluate
    return super().evaluate(
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/optimum/onnxruntime/trainer.py", line 957, in evaluate
    output = eval_loop(
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/transformers/trainer.py", line 3213, in evaluation_loop
    loss, logits, labels = self.prediction_step(model, inputs, prediction_loss_only, ignore_keys=ignore_keys)
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/optimum/onnxruntime/trainer_seq2seq.py", line 675, in prediction_step
    generation_inputs = inputs[self.model.main_input_name]
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1614, in __getattr__
    raise AttributeError("'{}' object has no attribute '{}'".format(
AttributeError: 'ModuleWithLoss' object has no attribute 'main_input_name'
0543863ffef24c09a57b3fc5ce0f584b000000:93:290 [0] NCCL INFO [Service thread] Connection closed by localRank 0
0543863ffef24c09a57b3fc5ce0f584b000000:93:93 [0] NCCL INFO comm 0x5b711860 rank 0 nranks 8 cudaDev 0 busId 100000 - Abort COMPLETE
ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: 1) local_rank: 0 (pid: 93) of binary: /opt/conda/envs/ptca/bin/python
  File "/opt/conda/envs/ptca/bin/torchrun", line 8, in <module>
    sys.exit(main())
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 346, in wrapper
    return f(*args, **kwargs)
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/torch/distributed/run.py", line 794, in main
    run(args)
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/torch/distributed/run.py", line 785, in run
    elastic_launch(
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/torch/distributed/launcher/api.py", line 134, in __call__
    return launch_agent(self._config, self._entrypoint, list(args))
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/torch/distributed/launcher/api.py", line 250, in launch_agent
    raise ChildFailedError(
torch.distributed.elastic.multiprocessing.errors.ChildFailedError: 
```

optimum/onnxruntime/trainer.py
AdamLouly: Based on the assumption that T5 is a sequence-to-sequence (seq2seq) model, the reason for its failure can be attributed to the fact that when the trainer falls back to the Seq2SeqTrainer class, it encounters an error.

optimum/onnxruntime/trainer.py
```diff
@@ -907,7 +907,7 @@ def evaluate(
...
+        self.model = self.model._original_model
```
pengwa, line 910: don't we need a check module is loss is enabled? otherwise, how thye have ._original model????

JingyaHuang: @AdamLouly Can we add a checker here for the case without the `ModuleWithLoss` wrapper?

JingyaHuang (approved):
Hi @AdamLouly, thanks for the fix, it looks good to me.

Left a nit for the case that `use_module_with_loss` is disabled. Can you also run the [trainer test](https://github.com/huggingface/optimum/blob/main/tests/onnxruntime/nightly_test_trainer.py) to ensure the change won't break other features? Thanks!

```
RUN_SLOW=1 pytest -v -rs onnxruntime/nightly_test_trainer.py --durations=0
```

optimum/onnxruntime/trainer.py
AdamLouly: Good catch, we have to check here.

pengwa (approved):

JingyaHuang:
The PR is merged, thanks again for the fix and review 🙏 ! @AdamLouly @pengwa

PR #1102, opened 2023-06-13
Fix optimum model inference compatibility with transformers >= v4.30.0

echarlaix:
In this PR we add a workaround to fix pipeline inference with optimum models compatibility with `transformers>=v4.30.0`. Coming from [`infer_framework`](https://github.com/huggingface/transformers/blob/70c79940957fb25b54bd1b106935c756b90345eb/src/transformers/utils/generic.py#L568)  which should be modified to be compatible with optimum models

### DIFF
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -70,7 +70,7 @@ FROM_PRETRAINED_START_DOCSTRING = r"""
 
 
 # workaround to enable compatibility between optimum models and transformers pipelines
-class PreTrainedModel(ABC): # noqa: F811
+class PreTrainedModel(ABC):  # noqa: F811
     pass
 
 
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -68,6 +68,7 @@ FROM_PRETRAINED_START_DOCSTRING = r"""
             the Hub on your local machine.
 """
 
+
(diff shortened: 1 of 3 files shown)


fxmarty:
A test appears to fail:

`2023-06-13T11:46:36.1199562Z FAILED exporters/onnx/test_exporters_onnx_cli.py::OnnxCLIExportTestCase::test_trust_remote_code - AssertionError: 0 is not true : 1`

fxmarty (approved):
LGTM otherwise!

echarlaix:
> A test appears to fail:
> 
> `2023-06-13T11:46:36.1199562Z FAILED exporters/onnx/test_exporters_onnx_cli.py::OnnxCLIExportTestCase::test_trust_remote_code - AssertionError: 0 is not true : 1`

This is not related to the fix added for Optimum's model inference but directly from https://github.com/huggingface/transformers/pull/23959 and our [test]( https://github.com/huggingface/optimum/blob/40b3ceca69f1714159e3f1c4845fb13441262b29/tests/exporters/onnx/test_exporters_onnx_cli.py#L234) which checks that loading this [model](https://huggingface.co/fxmarty/tiny-testing-gpt2-remote-code) shouldn't be possible without setting the option `trust_remote_code=True` (now possible since https://github.com/huggingface/transformers/pull/23959)

Issue #1104, opened 2023-06-13
onnx exporter does NOT use cache_dir with task=auto


escorciav:
### System Info

```shell
Python 3.10.11
optimum==1.8.6
Ubuntu x86_64
```

### Who can help?

@michaelbenayoun

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

1. Disable/Block any internet connection (e.g., disconnect wifi, or block internet connections via Docker, etc.)
2.  `optimum-cli export onnx --model t5-base checkpoints/t5-base_onnx/ --framework pt --optimize O3 --batch_size 1 --sequence_length 512 --atol 0.0001 --cache_dir ~/.cache/huggingface/hub`

Ask `task='auto'` (default), the code gets stuck [here sending a request to HF servers](https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/__main__.py#L185) to infer something from online catalog.

IMHO, a patch is needed.

Feel free to close the issue if out of the scope. More details are provided [here](https://discuss.huggingface.co/t/how-to-use-the-cache-dir-along-with-optimum-cli-export/42911/3)

### Expected behavior

The script should run without checking the record in HF server. Especially, as HF blocks IPs.

fxmarty:
Hi, should be fixed in https://github.com/huggingface/optimum/pull/1109

escorciav:
When no passing `--task` I got the following error message. The error message does not provide any clue about `task=auto` & `cache_dir` is ignored. Did setup optimum incorrectly?

I used a fresh conda environment with Python=3.11 + pip
```bash
pip install onnx 
pip install git+https://github.com/huggingface/optimum.git@main
```

```
$ optimum-cli export onnx --model t5-base favor-hf-delete-me --framework pt --optimize O3 --batch_size 1 --sequence_length 512 --atol 0.0001 --cache_dir ~/.cache/huggingface/hub
Traceback (most recent call last):
  File "/apps/install/bin/miniconda3/envs/hf-optimum/lib/python3.11/site-packages/huggingface_hub/utils/_errors.py", line 259, in hf_raise_for_status
    response.raise_for_status()
  File "/apps/install/bin/miniconda3/envs/hf-optimum/lib/python3.11/site-packages/requests/models.py", line 1021, in raise_for_status
[...]

PR #1109, opened 2023-06-14
Improve the offline support for the ONNX/TFLite export

fxmarty:
Fixes https://github.com/huggingface/optimum/issues/1104

### DIFF
--- a/optimum/exporters/onnx/__main__.py
+++ b/optimum/exporters/onnx/__main__.py
@@ -17,6 +17,7 @@
 import argparse
 from pathlib import Path
 
+from requests.exceptions import ConnectionError as RequestsConnectionError
 from transformers import AutoTokenizer
 from transformers.utils import is_torch_available
 
@@ -166,6 +167,19 @@ def main_export(
         )
 
     torch_dtype = None if fp16 is False else torch.float16
+
+    if task == "auto":
+        try:
+            task = TasksManager.infer_task_from_model(model_name_or_path)
+        except KeyError as e:
+            raise KeyError(
+                f"The task could not be automatically inferred. Please provide the argument --task with the relevant task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"
+            )
+        except RequestsConnectionError as e:
+            raise RequestsConnectionError(
+                f"The task could not be automatically inferred as this is available only for models hosted on the Hugging Face Hub. Please provide the argument --task with the relevant task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"
+            )
+
     model = TasksManager.get_model_from_task(
         task,
         model_name_or_path,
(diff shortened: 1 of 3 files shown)


optimum/exporters/tasks.py
```diff
@@ -1161,11 +1164,26 @@ def determine_framework(
                 for file in filenames
             ]
         else:
-            if not isinstance(model_name_or_path, str):
-                model_name_or_path = str(model_name_or_path)
-            all_files = huggingface_hub.list_repo_files(model_name_or_path, repo_type="model")
-            if subfolder != "":
-                all_files = [file[len(subfolder) + 1 :] for file in all_files if file.startswith(subfolder)]
+            object_id = model_name_or_path.replace("/", "--")
```
regisss, line 1167: Not super familiar with the hub cache, are we sure that replacing "/" by "--" always works?

regisss (approved):
LGTM!

optimum/exporters/tasks.py
fxmarty: In the "new" caching system I guess yes, I just followed https://github.com/huggingface/huggingface_hub/blob/2c598c484687d586a0ad358301364974912e0262/src/huggingface_hub/file_download.py#L1466-L1467

optimum/exporters/onnx/__main__.py
```diff
@@ -17,6 +17,7 @@
 import argparse
 from pathlib import Path
 
+from requests.exceptions import ConnectionError as RequestsConnectionError
```
michaelbenayoun, line 20: Why changing the name?

optimum/exporters/tasks.py
```diff
@@ -1161,11 +1164,26 @@ def determine_framework(
...
+            full_model_path = Path(cache_dir, f"models--{object_id}")
+            if full_model_path.is_dir():  # explore the cache first
+                # Resolve refs (for instance to convert main to the associated commit sha)
+                revision_file = Path(full_model_path, "refs", "main")
+                if revision_file.is_file():
+                    with open(revision_file) as f:
+                        revision = f.read()
+                cached_path = Path(full_model_path, "snapshots", revision, subfolder)
+                all_files = [
+                    os.path.relpath(os.path.join(dirpath, file), cached_path)
```
michaelbenayoun, line 1177: nit: We should stick to the pathlib API here instead of mixing pathlib and os.path IMO

michaelbenayoun (approved):
LGTM

optimum/exporters/onnx/__main__.py
fxmarty: It collides with python native ConnectionError

optimum/exporters/tasks.py
fxmarty: agreed

Issue #1086, opened 2023-06-07
Support for Document Question Answering


mgoin:
### Feature request

It appears there is no support for the [`document-question-answering`](https://huggingface.co/tasks/document-question-answering) task. I am curious if this is expected to be particularly difficult or planned for future work?

### Motivation

It would be useful to accelerate and this is the most common finetuned task for LayoutLM and Donut, which are supported for `feature-extraction` already.

### Your contribution

I found this issue by attempting to export a finetuned DocVQA model using latest main:
```
optimum-cli export onnx -m naver-clova-ix/donut-base-finetuned-docvqa donut-base-finetuned-docvqa/
Framework not specified. Using pt to export to ONNX.
Traceback (most recent call last):
  File "/Users/mgoin/venv/bin/optimum-cli", line 8, in <module>
    sys.exit(main())
  File "/Users/mgoin/venv/lib/python3.10/site-packages/optimum/commands/optimum_cli.py", line 163, in main
    service.run()
  File "/Users/mgoin/venv/lib/python3.10/site-packages/optimum/commands/export/onnx.py", line 203, in run
    main_export(
  File "/Users/mgoin/venv/lib/python3.10/site-packages/optimum/exporters/onnx/__main__.py", line 169, in main_export
    model = TasksManager.get_model_from_task(
  File "/Users/mgoin/venv/lib/python3.10/site-packages/optimum/exporters/tasks.py", line 1357, in get_model_from_task
    if TasksManager._TASKS_TO_LIBRARY[task.replace("-with-past", "")] == "transformers":
KeyError: 'document-question-answering'
```

fxmarty:
Thank you, for sure we should add support for it! I don't expect it to be too much work, I can have a look next week.

fxmarty:
Hi @mgoin, the export of vision-encoder-decoder models should work for the document-question-answering task following https://github.com/huggingface/optimum/pull/1110. LayoutLM should still be done, although I don't know if it used that much.

mgoin:
Thank you for picking it up so quickly @fxmarty ! I'll give it a try

PR #1110, opened 2023-06-14
Support document-question-answering ONNX export for vision-encoder-decoder

fxmarty:
Fixes https://github.com/huggingface/optimum/issues/1086

Fixes as well a bug in the ONNX export validation where gradients would be computed for the PyTorch model.

edit: actually moved the gradient bug to https://github.com/huggingface/optimum/pull/1111 to have it merged right away

### DIFF
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -316,7 +316,8 @@ class EncoderDecoderOnnxConfig(OnnxSeq2SeqConfigWithPast):
                     "past key values."
                 )
 
-            self._decoder_onnx_config = decoder_onnx_config_constructor(config.decoder, **kwargs)
+            _decoder_onnx_config = decoder_onnx_config_constructor(config.decoder, **kwargs)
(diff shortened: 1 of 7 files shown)


optimum/exporters/onnx/config.py
```diff
@@ -360,8 +361,19 @@ def generate_dummy_inputs_for_validation(self, reference_model_inputs: Dict[str,
...
+                """
+
+            # TODO: likely breaking e.g. for t5 as decoder
+            if "encoder_outputs" in reference_model_inputs:
+                if self.use_past_in_inputs is False or self.is_merged:
+                    # ONNX without past uses encoder_hidden_states even when we don't outputing them
+                    reference_model_inputs["encoder_hidden_states"] = reference_model_inputs.pop("encoder_outputs")[0]
+                else:
+                    # ONNX with past does not use encoder_hidden_states when we don't output them
+                    reference_model_inputs.pop("encoder_outputs")
```
fxmarty, line 376: A proper solution should be found for this

fxmarty: fixed

PR #1116, opened 2023-06-16
Fix stable diffusion ONNX export following diffusers breaking change

fxmarty:
As per title and discussed on slack, slightly more minimal than https://github.com/huggingface/optimum/pull/1108 given that SDPA ONNX export will be supported from PyTorch 2.1 onwards.

@echarlaix is it OK to have the requirement `diffusers>=0.17.0`?

### DIFF
--- a/optimum/exporters/onnx/utils.py
+++ b/optimum/exporters/onnx/utils.py
@@ -42,13 +42,15 @@ if is_diffusers_available():
             f"We found an older version of diffusers {_diffusers_version} but we require diffusers to be >= {DIFFUSERS_MINIMUM_VERSION}. "
             "Please update diffusers by running `pip install --upgrade diffusers`"
         )
-    TMP_DIFFUSERS_MAX_VERSION = "0.17.0"
-    if check_if_diffusers_greater(TMP_DIFFUSERS_MAX_VERSION):
-        logger.warning(
-            f"We found an newer version of diffusers {_diffusers_version} but we require diffusers to be < {TMP_DIFFUSERS_MAX_VERSION}."
-        )
-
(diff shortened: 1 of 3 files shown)


optimum/exporters/onnx/utils.py
```diff
@@ -190,13 +192,13 @@ def get_stable_diffusion_models_for_export(
...
     # PyTorch does not support the ONNX export of torch.nn.functional.scaled_dot_product_attention
-    pipeline.unet.set_attn_processor(CrossAttnProcessor())
+    pipeline.unet.set_attn_processor(AttnProcessor())
     models_for_export["unet"] = (pipeline.unet, unet_onnx_config)
 
     # VAE Encoder https://github.com/huggingface/diffusers/blob/v0.11.1/src/diffusers/models/vae.py#L565
     vae_encoder = copy.deepcopy(pipeline.vae)
-    if hasattr(vae_encoder.encoder.mid_block.attentions[0], "_use_2_0_attn"):
-        vae_encoder.encoder.mid_block.attentions[0]._use_2_0_attn = False
+    if not packaging.version.parse(torch.__version__) >= packaging.version.parse("2.1.0"):
```
echarlaix, line 200: It could make sense to add a [`is_torch_version`](https://github.com/huggingface/optimum-intel/blob/v1.9.1/optimum/intel/utils/import_utils.py#L176) in the future (or [`check_if_torch_greater`](https://github.com/huggingface/optimum/blob/a7670aae20a22ae77487db5b1fce57c07de29eb5/optimum/utils/import_utils.py#L120))

echarlaix (approved):
LGTM, thanks a lot for the fix @fxmarty !!

Issue #351, opened 2022-08-16
Add all available ONNX models to ORTConfigManager


ChainYo:
This issue is linked to the [ONNXConfig for all](https://huggingface.co/OWG) working group created for implementing an ONNXConfig for all available models. Let's extend our work and try to add all models with a fully functional ONNXConfig implemented to ORTConfigManager.

Adding models to ORTConfigManager will allow 🤗Optimum users to boost even more their model with ONNX optimization capacity!

Feel free to join us in this adventure! Join the org by clicking [here](https://huggingface.co/organizations/OWG/share/TskjfGaGjGnMXXssbPPXrQWEIbosGqZshZ)

Here is a non-exhaustive list of models that have one ONNXConfig and could be added to ORTConfigManager:

(This includes only models with ONNXConfig implemented, if your target model doesn't have an ONNXConfig, please open an issue/or implement it (even cooler) in the 🤗Transformers repository. Check [this issue](https://github.com/huggingface/transformers/issues/16308) to know how to do)

* [x] Albert
* [x] BART
* [ ] BeiT
* [x] BERT
* [x] BigBird
* [ ] BigBirdPegasus
* [ ] Blenderbot
* [ ] BlenderbotSmall
* [ ] BLOOM
* [x] CamemBERT
* [ ] CLIP
* [x] CodeGen
* [ ] ConvNext
* [ ] ConvBert
* [ ] Data2VecText
* [ ] Data2VecVision
* [x] Deberta
* [x] Deberta-v2
* [ ] DeiT
* [ ] DETR
* [x] Distilbert
* [x] ELECTRA
* [ ] Flaubert
* [x] GPT2
* [ ] GPTJ
* [x] GPT-NEO
* [ ] I-BERT
* [ ] LayoutLM
* [ ] LayoutLMv2
* [ ] LayoutLMv3
* [ ] LeViT
* [ ] LongT5
* [ ] M2M100
* [ ] mBART
* [x] MT5
* [ ] MarianMT
* [ ] MobileBert
* [ ] MobileViT
* [ ] OpenAIGPT-2
* [ ] PLBart
* [ ] Perceiver
* [ ] ResNet
* [ ] RoFormer
* [x] RoBERTa
* [ ] SqueezeBERT
* [ ] T5
* [ ] ViT
* [ ] XLM
* [x] XLM-RoBERTa
* [ ] XLM-RoBERTa-XL
* [ ] YOLOS

If you want an example of implementation, I did one for `MT5` #341.

[...]

PR #1117, opened 2023-06-16
Add ViT to ORTConfigManager

baskrahmer:
# What does this PR do?

Part of #351

## Before submitting


fxmarty (approved):
LGTM, thank you for the addition! We should indeed integration with more of the optimizations from ORT team.

Issue #1056, opened 2023-05-16
Any support for GIT and BLIP-2 for BetterTransformer


Sayantad:
### Feature request

Support for GIT and BLIP-2 models in BetterTransformer API

### Motivation

Reduce inference timings

### Your contribution
NA

PR #1125, opened 2023-06-21
Add blip-2 to bettertransformer

baskrahmer:
# What does this PR do?
Add BLIP-2 to the BetterTransformer API.

Part of #1056 

## Before submitting

### DIFF
--- a/optimum/bettertransformer/models/__init__.py
+++ b/optimum/bettertransformer/models/__init__.py
@@ -181,6 +181,7 @@ class BetterTransformerManager:
         if model_type in [
             "blenderbot",
             "codegen",
+            "clip-2",
             "gpt2",
             "gptj",
             "gpt_neo",
--- a/optimum/bettertransformer/models/__init__.py
+++ b/optimum/bettertransformer/models/__init__.py
@@ -56,9 +56,7 @@ class BetterTransformerManager:
         "bert-generation": {"BertGenerationLayer": BertLayerBetterTransformer},
         "blenderbot": {"BlenderbotAttention": BlenderbotAttentionLayerBetterTransformer},
         "camembert": {"CamembertLayer": BertLayerBetterTransformer},
-        "blip-2": {
-            "T5Attention": T5AttentionLayerBetterTransformer,
-        },
+        "blip-2": {"T5Attention": T5AttentionLayerBetterTransformer},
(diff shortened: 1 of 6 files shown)


fxmarty (approved):
LGTM thank you! Could you just solve the conflict?

We should definitely have a nice table in the doc with the speedups.

baskrahmer:
@fxmarty yes I agree such a comparison would be nice! If there are already benchmarks then I would be interested to work on this.

I also thought about adding tests to assert speedups, but if the speeds fluctuate this could mess with the CI.

fxmarty:
There are some scripts that we used for blog posts, but we did not put results in the documentation itself: https://github.com/huggingface/optimum/tree/main/tests/benchmark

Encoder implementation may need to be revamped soon though, as currently we error out when they are used for training, while there's not really any reason to now.

kirills12345:
@baskrahmer thank you very much for contributing BLIP2 support!
I have a question - does it currently only support FlanT5 model? If so, I'm a bit confused of how does it support is since T5 model [can not be supported](https://github.com/huggingface/transformers/issues/20372#issuecomment-1324736962) due to nature of it's attention mechanism - shouldn't it be the same for FlanT5?

Also does it show any improvement in inference speed? Could not get any during my experiments, maybe I've done something wrong.
Thanks in advance!

baskrahmer:
Hey @kirillsemenov1314 :)

> I have a question - does it currently only support FlanT5 model? If so, I'm a bit confused of how does it support is since T5 model [can not be supported](https://github.com/huggingface/transformers/issues/20372#issuecomment-1324736962) due to nature of it's attention mechanism - shouldn't it be the same for FlanT5?

I believe this statement no longer holds. The BetterTransformer implementation of the T5 layer is found in [this file](https://github.com/huggingface/optimum/blob/main/optimum/bettertransformer/models/decoder_models.py), so I suggest going through it if you're interested in the implementation.

> Also does it show any improvement in inference speed on GPU T4? Could not get any during my experiments, maybe I've done something wrong. Thanks in advance!

This is an interesting topic. AFAIK active work is being done on [this tool](https://github.com/huggingface/optimum-benchmark) which can be used to also run benchmarks on BetterTransformer architectures. Inference speed is influenced by a variety of factors such as the model, dataset and hardware. It can thus very well be that there is no significant speedup using BetterTransformer in your case, and it does not necessarily apply you are doing something wrong.

PR #1129, opened 2023-06-23
add clip text model to better transformers

williamberman:
# What does this PR do?

This adds clip text model to better transformers

I'm not sure if this requires updating tests or docs. Up to optimum maintainers if so?

## Before submitting


fxmarty:
Hi @williamberman , yes it would be nice to add it to the tests, and possibly verify whether there is any speedup.

williamberman:
@fxmarty yes verified there's speed up. Could you point me towards what test to add?

PR #1131, opened 2023-06-23
Add CvTONNX Config

rishabbala:
# What does this PR do?
Added changes for CvT model

Fixes part of #555

## Before submitting

@michaelbenayoun @fxmarty

### DIFF
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -550,6 +550,11 @@ class ViTOnnxConfig(VisionOnnxConfig):
         return {"pixel_values": {0: "batch_size", 1: "num_channels", 2: "height", 3: "width"}}
 
 
+class CvTOnnxConfig(ViTOnnxConfig):
+    DEFAULT_ONNX_OPSET = 13
+    ATOL_FOR_VALIDATION = 1e-2
+
+
 class LevitOnnxConfig(ViTOnnxConfig):
     pass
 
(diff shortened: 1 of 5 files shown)


fxmarty:
Hi, thank you for the addition! Tests indeed pass, except the style one. Could you run `make style`?

optimum/exporters/onnx/model_configs.py
```diff
@@ -550,6 +550,11 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
         return {"pixel_values": {0: "batch_size", 1: "num_channels", 2: "height", 3: "width"}}
 
 
+class CvTOnnxConfig(ViTOnnxConfig):
+    DEFAULT_ONNX_OPSET = 13
+    ATOL_FOR_VALIDATION = 1e-2
```
michaelbenayoun, line 555: This seems pretty high tbh, do you have any idea which op makes it fail?

rishabbala:
@fxmarty  Done

optimum/exporters/onnx/model_configs.py
michaelbenayoun: I am talking about the tolerance, we might have a hidden bug if we need `1e-2`, or we might not, but just wanted to raise the potential issue here.

fxmarty (approved):
LGTM, thank you for the addition!

@rishabbala Maybe before merging, could you try to run the export on a larger CvT model and see what the diff is versus pytorch in the validation?

rishabbala:
The diff for cvt-13 is 3.5e-3, for cvt-13-384-22k it is 6e-3 and for cvt-21 it is 2.7e-3. Should I reduce the tolerance?

fxmarty:
Sounds fine to me, thanks!

PR #1132, opened 2023-06-23
Refactor BetterTransformerManager requirement validation methods

baskrahmer:
# What does this PR do?

Refactoring of methods to make them less ugly ;)

### DIFF
--- a/optimum/bettertransformer/models/__init__.py
+++ b/optimum/bettertransformer/models/__init__.py
@@ -119,16 +119,16 @@ class BetterTransformerManager:
     }
 
     REQUIRES_NESTED_TENSOR = {
-            "blenderbot",
-            "codegen",
-            "gpt2",
-            "gptj",
-            "gpt_neo",
-            "gpt_neox",
-            "llama",
-            "opt",
-            "pegasus",
-            "t5",
+        "blenderbot",
+        "codegen",
+        "gpt2",
+        "gptj",
+        "gpt_neo",
+        "gpt_neox",
+        "llama",
+        "opt",
+        "pegasus",
+        "t5",
     }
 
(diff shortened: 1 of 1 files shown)


optimum/bettertransformer/models/__init__.py
```diff
@@ -118,6 +118,49 @@ class BetterTransformerManager:
         "glpn": "GLPN has a convolutional layer present in the FFN network, which is not supported in PyTorch's BetterTransformer.",
     }
 
+    REQUIRES_NESTED_TENSOR = {
```
ydshieh, line 121: The naming and the logic below is actually reversed.

```python
def requires_nested_tensor
    return model_type not in BetterTransformerManager.REQUIRES_NESTED_TENSOR
```

baskrahmer: I don't follow. Can you give a corrected example?

ydshieh: The name is `REQUIRES_NESTED_TENSOR`. But for a model_type in `REQUIRES_NESTED_TENSOR`, the method return `False`, but the method name is `requires_nested_tensor`. The name of the method and what it actually does is completely different thing, if I don't make mistake.

ydshieh: One quick way is to rename `REQUIRES_NESTED_TENSOR` to `NOT_REQUIRES_NESTED_TENSOR`.

baskrahmer: Ah gotcha! Good point, thanks for the feedback I will update the PR.

fxmarty (approved):
Thanks!

PR #1137, opened 2023-06-26
update the default block size

rui-ren:
# What does this PR do?

Set the default `block_size = model_max_length`.

Fixes # (issue)

## Before submitting

### DIFF
--- a/examples/onnxruntime/training/language-modeling/run_clm.py
+++ b/examples/onnxruntime/training/language-modeling/run_clm.py
@@ -435,12 +435,10 @@ def main():
 
     if data_args.block_size is None:
         block_size = tokenizer.model_max_length
-        if block_size > 1024:
-            logger.warning(
-                f"The tokenizer picked seems to have a very large `model_max_length` ({tokenizer.model_max_length}). "
-                "Picking 1024 instead. You can change that default value by passing --block_size xxx."
-            )
-            block_size = 1024
+        logger.warning(
+            f"The tokenizer picked seems to have a very large `model_max_length` ({tokenizer.model_max_length}). "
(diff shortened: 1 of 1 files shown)


JingyaHuang (approved):
LGTM, thanks for contributing.

rui-ren:
@JingyaHuang Hi, in [Transformer PyTorch Example ](https://github.com/rui-ren/transformers/blob/c817bc44e2915593a292df9e7d4e2c0dfefb6620/examples/pytorch/language-modeling/run_clm.py#L474-L480) ,  The block_size is default to `1027 if block_size > 1024:`. Do we need to update it accordingly? Otherwise the benchmarking for PyTorch vs Onnxruntime might be different cases. Thanks.

rui-ren:
FYI, @JingyaHuang  https://github.com/huggingface/transformers/pull/24583

JingyaHuang:
Hi @rui-ren, just to confirm the reason that you would not limit the block size, would it boost the performance of ORT, is the change ORT specific?

JingyaHuang:
Yeah, so we suggest to keep the example aligned with what's in transformers unless the change benefits ORT's performance.

rui-ren:
@JingyaHuang, 

I used this `run_clm.py` to run benchmark tests for models that `block_size` is larger than 2k, if we use the default of `block_size=1024`, and we might fail to test memory usage from the model itself.  That's the reason for this this PR.

However, it makes sense that I can add `--block-size xxx` to override the default 1024. I will revert this PR.

Thanks!

rui-ren:
@JingyaHuang please check here https://github.com/huggingface/optimum/pull/1162 Thanks!

Issue #1150, opened 2023-06-30
ORTModelForCausalLM - ONNX - Merge model fails to run


claeyzre:
### System Info

```shell
System: Windows 11
Optimum Version: 1.9 (tried in 1.8.x as well and didn't work)
Onnxruntime Version: 1.15.1 (tried in 1.13.x as well and didn't work)
Onnx Version: 1.13.1
Pytorch Version:2.0.0 (tried in 1.X as well and didn't work)
```

### Who can help?

@fxmarty as I saw you working a lot of these kind of models in the previous issues

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction

I have an onnx error while trying to execute this small script
```
text = "I want my money and"
model_id = "hf-internal-testing/tiny-random-BloomForCausalLM"
tokenizer = AutoTokenizer.from_pretrained(model_id)

model = ORTModelForCausalLM.from_pretrained(model_id, from_transformers=True, use_merged=True)
input_ids = tokenizer(text, return_tensors="pt").input_ids
results_hf = model.generate(inputs=input_ids,
                            num_beams=2,
                            do_sample=False,
```

I have the following onnx error
```
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running If node. Name:'optimum::if' Status Message: Non-zero status code returned while running Concat node. Name:'/transformer/h.0/self_attention/Concat_4' Status Message: concat.cc:159 onnxruntime::ConcatBase::PrepareForCompute Non concat axis dimensions must match: Axis 0 has mismatched dimensions of 8 and 2
```

Note that this error arises with num_beams = 2 but not with num_beams=1 so this occurs only when the decoder takes a "batched" input.

### Expected behavior

The generation runs smoothly

fxmarty:
Thank you! Interestingly I can reproduce the issue on bloom but not gpt2 / gpt-neo archs. But indeed the CI tests only `num_beams=1`, I'll fix for bloom and extend our tests.

PR #1152, opened 2023-06-30
Fix bloom KV cache usage in ORTForCausalLM

fxmarty:
Fixes https://github.com/huggingface/optimum/issues/1150

@michaelbenayoun @echarlaix Basically not all models share the exact same `_reorder_cache` and `prepare_inputs_for_generation`, in particular here bloom. What do you think of this solution? @echarlaix I would guess there is the same bug in optimum-intel.

I think my solution is very ugly (now `ORTBloomForCausalLM` and `ORTModelForCausalLM` need to be in the same file forever). An other approach is to move all shared methods to `ORTModelDecoder` (effectively making it a mixin class) and having `ORTBloomForCausalLM` not inherit from `ORTModelForCausalLM`, but it does not solve the issue of "all classes in one file", and more importantly I believe that changing the inheritance is too breaking of a change (i.e. `isinstance(ort_bloom_model, ORTModelForCausalLM)` not working anymore).

Other solution: have a single `prepare_inputs_for_generation`, `_reorder_cache`, and dispatch to the relevant function from a dictionary. This adds dynamism, which I think is better to avoid.

Note: should add tests for `num_beams>1` in this PR as well


regisss (approved):
This solution looks good enough to me :+1:

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -695,3 +696,54 @@ def _reorder_cache(past: Tuple[Tuple[torch.Tensor]], beam_idx: torch.Tensor) ->
...
+    def _from_pretrained(
+        cls,
+        model_id: Union[str, Path],
+        config: "PretrainedConfig",
+        **kwargs,
+    ):
+        if config.model_type == "bloom":
+            return super()._from_pretrained(model_id, config, init_cls=ORTBloomForCausalLM, **kwargs)
+        else:
+            return super()._from_pretrained(model_id, config, init_cls=ORTModelForCausalLM, **kwargs)
```
michaelbenayoun, line 710:
```suggestion
        return super()._from_pretrained(model_id, config, init_cls=ORTModelForCausalLM, **kwargs)
```

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -695,3 +696,54 @@ def _reorder_cache(past: Tuple[Tuple[torch.Tensor]], beam_idx: torch.Tensor) ->
...
+
+
+class ORTBloomForCausalLM(ORTModelForCausalLM):
+    # Adapted from transformers.models.bloom.modeling_bloom.BloomForCausalLM._reorder_cache
```
michaelbenayoun, line 714:
```suggestion
    # Adapted from transformers.models.bloom.modeling_bloom.BloomForCausalLM.prepare_inputs_for_generation
```

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -315,6 +315,7 @@ def _from_pretrained(
         cls,
         model_id: Union[str, Path],
         config: "PretrainedConfig",
+        init_cls,
```
michaelbenayoun, line 318:
```suggestion
        init_cls: Type[ORTDecoderModel],
```

michaelbenayoun (approved):
Looks good for now. If this conditioning on the model type grows maybe we can find a nicer way of doing it, but right now it seems acceptable to me.

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -695,3 +696,54 @@ def _reorder_cache(past: Tuple[Tuple[torch.Tensor]], beam_idx: torch.Tensor) ->
...
+    # Adapted from transformers.models.bloom.modeling_bloom.BloomForCausalLM._reorder_cache
+    def prepare_inputs_for_generation(self, input_ids, past_key_values=None, **kwargs):
+        attention_mask = kwargs.get("attention_mask", None)
+        use_cache = kwargs.get("use_cache", None)
+
+        # only last token for input_ids if past is not None
+        if past_key_values:
+            # the cache may be in the stardard format (e.g. in contrastive search), convert to bloom's format if needed
+            if past_key_values[0][0].shape[0] == input_ids.shape[0]:
+                past_key_values = BloomForCausalLM._convert_to_bloom_cache(past_key_values)
```
echarlaix, line 723: Not a super satisfying approach but here is what we've done for OpenVINO : https://github.com/huggingface/optimum-intel/blob/v1.9.3/optimum/intel/openvino/modeling_decoder.py#L322 in order to rely as little as possible on `transformers` (which will be a dependency until we've our own `.generate()` so doesn't change much for now)

fxmarty: Thanks! I see the bug has been fixed already in optimum-intel. Feel free to ping me / open an issue in optimum in such case, as it's likely the bug is there as well. Trying to do as well!

echarlaix: > Thanks! I see the bug has been fixed already in optimum-intel. Feel free to ping me / open an issue in optimum in such case, as it's likely the bug is there as well. Trying to do as well!

This was integrated in https://github.com/huggingface/optimum-intel/pull/274 when enabling past key values re-use, so never saw that as a bug per say that could also happen in optimum, otherwise always do !> I'm personally not super fan of if/else, I think in this case subclassing makes sense (though here it is still ugly). But I'll remove the import indeed!

echarlaix: > Thanks! I see the bug has been fixed already in optimum-intel. Feel free to ping me / open an issue in optimum in such case, as it's likely the bug is there as well. Trying to do as well!

This was integrated in https://github.com/huggingface/optimum-intel/pull/274 when enabling past key values re-use for openvino models, so never saw that as a bug to be fixed in optimum as well, otherwise always do!

PR #1154, opened 2023-06-30
[ONNX export] Fixed incomplete model memory release issue

sharpbai:
# What does this PR do?

This PR resolves incomplete model memory release issue. The original code only delete the reference of model and onnx_model but not to trigger python gc and to release video memory. I found this issue when I tried to export llama based 13B model using fp16 on a A100 80G card. After showing that `Saving external data to one file...`, the the video memory usage rose up from 52G to 80G, which made the process crashed.

The export command is down below

```bash
optimum-cli export onnx --model TheBloke/Wizard-Vicuna-13B-Uncensored-HF \
--task causal-lm-with-past --fp16 --device cuda {save_dir}
```

There is no existing opening issue about this.

Now the export process will not use more than double the size of model weights memory for correct releasing models, which makes exporting 13B models on 80G video card possible.

## Before submitting

### DIFF
--- a/optimum/exporters/onnx/convert.py
+++ b/optimum/exporters/onnx/convert.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 """ONNX model check and export functions."""
 
+import gc
 import multiprocessing as mp
 import os
 import traceback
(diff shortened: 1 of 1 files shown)


fxmarty (approved):
LGTM thank you for fixing it!

Dont know if you have see the latest release notes @sharpbai , but there's also currently bugs in pytorch 2.0.1 that result in duplicate memory during the onnx export: https://github.com/huggingface/optimum/releases/tag/v1.9.0 It may be a good idea to use Pytorch nightly for now to export large models.

PR #1159, opened 2023-07-05
Andreyan/exporters model configs

andreyanufr:
# What does this PR do?
1) Added configuration for onnx conversion of MPTForCausalLM  architecture (https://huggingface.co/mosaicml/mpt-7b-instruct/blob/main/config.json).
2) Changed default seq_len for correct conversion of MPTForCausalLM  with use_cache=True flag (past key values)


fxmarty:
Hi @andreyanufr , thank you for your contribution! MPT is currently not supported in transformers (but will be), will have a look once it is in a release.

Meanwhile, I would recommend to use custom ONNX configs following https://github.com/huggingface/optimum/pull/1166

regisss:
@MaximProshin You can follow the advancement here: https://github.com/huggingface/transformers/pull/24629
But it's hard to say when this PR will be merged.

PR #1160, opened 2023-07-05
Add ONNX export for MGP-STR

NielsRogge:
# What does this PR do?

This PR adds support for the [MGP-STR](https://huggingface.co/docs/transformers/model_doc/mgp-str) model which can perform text scene recognition in images.

However this is not an official task yet, so I was wondering what the best way is to support this. For the moment I've only included "feature-extraction" but would like to also include the head model.

### DIFF
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -571,6 +571,10 @@ class ConvNextOnnxConfig(ViTOnnxConfig):
     pass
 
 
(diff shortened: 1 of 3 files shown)


fxmarty:
Hi @NielsRogge , would https://github.com/huggingface/optimum/pull/1166 fit?

NielsRogge:
Will close this PR as it worked with PyTorch's native ONNX export.

PR #1168, opened 2023-07-06
Enable SD XL ONNX export and ONNX Runtime inference

echarlaix:
Add Stable Diffusion XL ONNX export and pipelines to enable ONNX Runtime inference for text-to-image and image-to-image 

Export using the CLI :
```
optimum-cli export onnx --model stabilityai/stable-diffusion-xl-base-0.9 --task stable-diffusion-xl ./sd_xl_onnx
```
 
For inference :
```python
from optimum.onnxruntime import ORTStableDiffusionXLPipeline

model_id = "stabilityai/stable-diffusion-xl-base-0.9"
pipeline = ORTStableDiffusionXLPipeline.from_pretrained(model_id, export=True)
```

PR #1173, opened 2023-07-07
fix stable diffusion ONNX export

echarlaix:


echarlaix:
added in https://github.com/huggingface/optimum/pull/1168

michaelbenayoun (approved):
LGTM

PR #1188, opened 2023-07-14
Fix OwlViT exporter config

### LINKED ISSUE #1183: Cannot convert owlvit-base-patch32 model to ONNX and run inference
### System Info

```shell
Optimum version: 1.9.1
Python version: 3.11.3
OS: MacOS
```

### Who can help?

@mich

### Information

- [X] My own modified scripts

### Tasks

- [X] My own task or dataset (give details below)

### Reproduction

When using the CLI command
`optimum-cli export onnx --model google/owlvit-base-patch32 --task zero-shot-object-detection object_detection/owlvit_onnx` 
I'm able to get a converted ONNX format. Then, when using the following code to perform inference with the converted model:
`checkpoint = "google/owlvit-base-patch32"`
`processor = AutoProcessor.from_pretrained(checkpoint)`

`image = skimage.data.astronaut()`
`image = Image.fromarray(np.uint8(image)).convert("RGB")`
`text_queries = ["human face", "rocket", "nasa badge", "star-spangled banner", "woman", "smile", "hair", 'human head', 'human eye']`

`np_inputs = processor(text=text_queries, images=image, return_tensors="np")`
`session = ort.InferenceSession("object_detection/owlvit_onnx/model.onnx")`

`out =session.run(['logits', 'pred_boxes', 'text_embeds', 'image_embeds'], np_inputs)`

I get the following error: 
`RuntimeException: [ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Non-zero status code returned while running Reshape node. Name:'/Reshape_3' Status Message: /Users/runner/work/1/s/onnxruntime/core/providers/cpu/tensor/reshape_helper.h:41 onnxruntime::ReshapeHelper::ReshapeHelper(const onnxruntime::TensorShape &, onnxruntime::TensorShapeVector &, bool) gsl::narrow_cast(input_shape.Size()) == size was false. The input tensor cannot be reshaped to the requested shape. Input shape:{9,16}, requested shape:{2,4,16}`

Now it seems to be related to some input being wrong, but I cannot get what is wrong. The pre-processing step is the same as for the HF model, the only difference being instead of returning "pt" tensors I'm returning "np" so it can work with ONNX. Here are my input shapes:

input_ids: (9, 16)
attention_mask: (9, 16)
pixel_values: (1, 3, 768, 768)

Thanks in advance!

### Expected behavior

Inference to run successfully and outputs to be very similar to that of the original torch model.

regisss:
# What does this PR do?

Fixes #1183.
Non-max suppression is used in OwlViT to filter box predictions here: https://github.com/huggingface/transformers/blob/91d7df58b6537d385e90578dac40204cb550f706/src/transformers/models/owlvit/modeling_owlvit.py#L1498
This implies looping over the batch size, which basically hard-codes this batch size in this for-loop when exporting to ONNX.

Thus, the model must be exported with the same batch size as the one that will be used at inference.

## Before submitting


optimum/exporters/onnx/model_configs.py
```diff
@@ -803,10 +803,10 @@ def outputs(self) -> Dict[str, Dict[int, str]]:
             outputs["logits_per_image"] = {0: "image_batch_size", 1: "text_batch_size"}
             outputs["logits_per_text"] = {0: "text_batch_size", 1: "image_batch_size"}
         elif self.task == "zero-shot-object-detection":
-            outputs["logits"] = {0: "batch_size", 1: "num_queries"}
-            outputs["pred_boxes"] = {0: "batch_size", 1: "num_queries"}
+            outputs["logits"] = {0: "image_batch_size", 2: "num_queries"}
```
fxmarty, line 806: Is the fix changing 1 to 2 here?

Shouldn't it rather be

```
        elif self.task == "zero-shot-object-detection":
            outputs["logits"] = {1: "num_queries"}
            outputs["pred_boxes"] = {1: "num_queries"}
```

if the batch size can't be dynamic?

regisss: I just kept the batch size for "consistency" with other ONNX configs, but yes it makes sense to remove it. Will do :+1:

regisss: The change from 1 to 2 is needed because the second dimension of `logits` is not the number of queries but is the number of patches in the images: https://github.com/huggingface/transformers/blob/5bb4430edc7df9f9950d412d98bbe505cc4d328b/src/transformers/models/owlvit/modeling_owlvit.py#L1699

regisss: Actually ONNX doesn't like when I remove `"image_batch_size"`, it seems to get confused between text and image batch sizes. So let's keep it that way.

regisss:
@fxmarty

fxmarty:
@regisss  The cause was numpy usage, not https://github.com/huggingface/transformers/blob/91d7df58b6537d385e90578dac40204cb550f706/src/transformers/models/owlvit/modeling_owlvit.py#L1498 (that is never used in the model forward)

https://github.com/huggingface/optimum/pull/1188

fxmarty:
https://github.com/huggingface/transformers/pull/29326

PR #1189, opened 2023-07-15
Bt benchmarks

baskrahmer:
# What does this PR do?

Refactors the benchmark scripts; add CPU support.
Next step is to make a script to run these benchmarks on all BetterTransformer architectures and generate the speedup tables :)

Still need to test on GPU

https://github.com/huggingface/optimum/pull/1125#pullrequestreview-1499893051

## Before submitting

### DIFF
--- a/optimum/onnxruntime/runs/__init__.py
+++ b/optimum/onnxruntime/runs/__init__.py
@@ -1,4 +1,3 @@
-import copy
 import os
 from pathlib import Path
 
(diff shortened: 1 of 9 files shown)


regisss:
@IlyasMoutawwakil Do we already support BetterTransformer in Optimum Benchmark?

IlyasMoutawwakil:
@regisss Yep we do,

baskrahmer:
@regisss @IlyasMoutawwakil ah I did not know of this repo but it seems useful! That leaves me questioning the purpose of this PR and generally the need for benchmark scripts in the Optimum library.

regisss:
> @regisss @IlyasMoutawwakil ah I did not know of this repo but it seems useful! That leaves me questioning the purpose of this PR and generally the need for benchmark scripts in the Optimum library.

Not sure exactly if these scripts are still maintained. Maybe @fxmarty knows more about this?

PR #1198, opened 2023-07-18
Add SD XL documentation

echarlaix:
should be merged just before the next release


echarlaix:
> LGTM @echarlaix!! We'll need to rebase to have the PR doc but it looks good to me.

Will do it right now, thanks @regisss !

PR #1199, opened 2023-07-18
Add bark into bettertransformer

ylacombe:
# What does this PR do?

This PR aims at integrating Bark, a TTS model, to `bettertransformer`.

The [original bark implementation ](https://github.com/suno-ai/bark) uses SDPA to get significant speed improvement. An internal benchmark has also shown that transformers's Bark could greatly benefit from an integration into `bettertransformer`.

The two types of `SelfAttention` of the original Bark [(causal self-attention](https://github.com/suno-ai/bark/blob/599fed040e52c89e0b3580e02e2684b2c9100701/bark/model.py#L23-L92) and [full self-attention](https://github.com/suno-ai/bark/blob/599fed040e52c89e0b3580e02e2684b2c9100701/bark/model_fine.py#L15-L61)) have been integrated into a [single `BarkSelfAttention` class](https://github.com/huggingface/transformers/blob/9c875839c0800c55c23f651ec4a1821d0c20ca04/src/transformers/models/bark/modeling_bark.py#L56-L178).

`Bark` was designed and trained by [Suno-AI team](https://github.com/suno-ai/bark) and is made of 4 main components:

- A `semantic model ` (also named `text model`), i.e a causal autoregressive transformer (GPT2-like), which takes into input a tokenized text
- A `coarse acoustics model` (also named `coarse model`), also a causal autoregressive transformer, taking into input the results of the last model. It aims at regressing the first two audio codebooks necessary to `encodec`.
- A `fine acoustics model` (`fine model`), this time a non-causal autoencoder transformer, which iteratively predicts the 6 last codebooks based on the sum of the previous codebooks embeddings.
- having predicted 8 codebooks channels of `encodec`, Bark uses `encodec` to generate the output audio array.

## TODO
- [ ] Make sure that the current test list is relevant and that tests pass.
- [ ] Make sure `bark_wrapped_scaled_dot_product` and `BarkAttentionLayerBetterTransformer` are well implemented.


ylacombe (commented):
Hi @younesbelkada and @fxmarty, I'm currently adding SDPA support to `Bark` a TTS model!

Could you have a look when you have time?

Many thanks!

BTW, I'm not sure that it's the right version of `transformers` since it was just added in the last release! Should I wait a bit?

optimum/bettertransformer/models/attention.py
```diff
@@ -89,6 +89,35 @@ def gpt2_wrapped_scaled_dot_product(
...
+    # to work around this we set is_causal=False.
+    is_causal = self.is_causal and query.shape[2] != 1
+
+    if is_causal:
+        sdpa_result = torch.nn.functional.scaled_dot_product_attention(
+            query, key, value, attn_mask=None, dropout_p=self.dropout, is_causal=is_causal
+        )
+    else:
+        sdpa_result = torch.nn.functional.scaled_dot_product_attention(
+            query, key, value, attn_mask=attention_mask, dropout_p=self.dropout, is_causal=False
```
fxmarty, line 115: The point of the batch size / training & eval check in the other functions is that SDPA does not dispatch to mem efficient attention / flash attention in case a custom attention mask is passed.

So here with this current implem, if somebody fine-tunes bark (which typically does not involve any padding, thus no need of custom attention mask) with batch size > 1, there would be close to no speedup.

Is there anything special that makes passing an attention mask required?

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -167,6 +169,46 @@ def forward(self, *args, **kwargs):
         return super().forward(*args, **kwargs)
 
 
+class BarkAttentionLayerBetterTransformer(BetterTransformerBaseLayer, BarkSelfAttention, nn.Module):
+    _attn = bark_wrapped_scaled_dot_product
+
+    def __init__(self, layer: "nn.Module", config: "PretrainedConfig", is_causal: bool = False):
+        super().__init__(config)
+
+        is_causal = layer.is_causal  # True # TODO
```
fxmarty, line 178:
```suggestion
        is_causal = layer.is_causal
```

maybe

tests/bettertransformer/test_audio.py
```diff
@@ -15,23 +15,184 @@
...
+
+            torch.manual_seed(0)
+            output_bt = converted_model.generate(
+                **inputs, fine_temperature=None, do_sample=False, semantic_max_new_tokens=length
+            )
+
+            self.assertTrue(
+                torch.allclose(output_hf, output_bt),
+                f"Maxdiff: {(output_hf - output_bt).abs().max()}",
+            )
```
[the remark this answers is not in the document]
fxmarty: How big is it?

I had lots of issues due to padding/batched case when integrating SDPA, maybe you could print intermediary results (& compare with https://www.diffchecker.com/) without/with SDPA and see if all values match.

optimum/bettertransformer/models/__init__.py
```diff
@@ -114,6 +116,8 @@ class BetterTransformerManager:
         "clip": ["text_model"],
         # blip-2's Q-former and vision model should not be identified as the last layers of the model
         "blip-2": ["qformer.encoder.layer", "vision_model.encoder.layers"],
+        # encodec should not be identified as the last layer
+        "bark": ["codec_model.encoder.layers"],
```
fxmarty, line 120: I forgot who added blip-2, but what's the link between this dict and not being identified as last layer?

fxmarty (commented):
Looks great, thanks a lot!

optimum/bettertransformer/models/__init__.py
ylacombe: The comment is not the good one, I should have said (and I will correct):
`bark.codec_model.encoder is not supported in BetterTransformer`

Without it, the transform doesn't work.

ylacombe:
```suggestion
        # bark.codec_model.encoder is not supported in BetterTransformer
        "bark": ["codec_model.encoder.layers"],
```

optimum/bettertransformer/models/attention.py
ylacombe: I didn't know about the use of the training & eval check. What about the point of the `batch_size == 1` check?

At the moment, Bark is only used for inference, but I could probably change the code with regards to the training & eval check, just in case training is supported in the future.

The attention mask is not really used in `BarkModel` yet but the underlying sub-models might require it for finetuning in the future I guess.

tests/bettertransformer/test_audio.py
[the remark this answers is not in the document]
ylacombe: It's around 0.25 atm, I might check intermediary results with your recommendation!

ylacombe (commented):
Hi @fxmarty , thanks for the review! 

I will investigate the `fp16` issue but I'm open to any suggestions with regards to what to be careful about since you had issues with padding/batching!

I would love to have your feedback on the self-attention method as well!

ylacombe:
Hi @fxmarty , I hope you are doing well!

I've updated the `bark_wrapped_scaled_dot_product` method and some of my code/comments, according to your comments!

I still get small numerical differences when testing `fp16`, though. 
I investigated a bit, and found that it was due to small numerical differences between the original self-attention and flash attention which accumulate when passing through Bark's `fine` submodel ([see here for an overview of Bark!](https://huggingface.co/docs/transformers/v4.31.0/en/model_doc/bark#overview)) which, during the generation pass, adds its outputs again and again, amplifying the numerical differences.

I still have to check if it produces huge qualitative differences between the generated waveform, but I don't think it's possible to pass this test. WDYT?

tests/bettertransformer/test_audio.py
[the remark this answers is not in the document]
fxmarty: @ylacombe is 0.25 actually small relatively? Maybe you could rather check something like the mean relative diff:

```
((hf_result - bt_result).abs() / (hf_result + eps).abs()).mean() < rtol
```

fxmarty (approved):
LGTM, I left a comment concerning the fp16 issue. I think we can merge as is and look into the fp16 later.

Do you confirm you have no issue in fp32?

Maybe you could use `with torch.backends.cuda.sdp_kernel(enable_math=True, enable_flash=False, enable_mem_efficient=False):` context manager to see if it is better? But it's probably just tiny numerical diff.

There's no hacky dtype conversion in the attention of bark, e.g. the softmax in fp32 in all cases?

optimum/bettertransformer/models/attention.py
```diff
@@ -89,6 +89,30 @@ def gpt2_wrapped_scaled_dot_product(
...
+    raise_on_head_mask(head_mask)
+
+    # When `past_kv` is provided, we're doing incremental decoding and `q.shape[2] == 1`: q only contains
+    # the query for the last token. scaled_dot_product_attention interprets this as the first token in the
+    # sequence, so if is_causal=True it will mask out all attention from it. This is not what we want, so
+    # to work around this we set is_causal=False.
+    is_causal = self.is_causal and query.shape[2] != 1
+
+    sdpa_result = torch.nn.functional.scaled_dot_product_attention(
+        query, key, value, attn_mask=None, dropout_p=self.dropout, is_causal=is_causal
```
fxmarty, line 110:
```suggestion
        query, key, value, attn_mask=None, dropout_p=self.dropout if self.training else 0., is_causal=is_causal
```

ylacombe:
Hey @fxmarty ,
Indeed I have no issues with `fp32`, the tests are running well!

I tried using the context manager, with no amelioration tbh. Also, the outputs are waveform, with low value (`output_bt.abs().max()=0.26`). Thus using the mean relative difference is giving higher results (around 0.7). Using mean difference it's around 0.015! 
For now I've set the tests `fp16` with a high tolerance on mean difference, they pass, but I can reconsider it if you want.

ylacombe:
Hi @fxmarty,
I've corrected some of the code with respect to optimum recent changes!
Let me know if I can do anything else!

fxmarty:
It's good thanks! Sorry for merging my PR first!

Issue #1192, opened 2023-07-17
Converted Vision Encoder Decoder onnx model seems like doesn't use past key values


kyle-bong:
### System Info

optimum = 1.9.1
python = 3.10
device = cpu
OS = linux

I made a vision encoder decoder model using huggingface transformers library. And I converted the model to onnx using this code:

```python
from optimum.onnxruntime import ORTModelForVision2SeqORTQuantizer
from transformers import PreTrainedTokenizerFast
model_checkpoint = "../checkpoint"
save_directory = "../onnx/"
# Load a model from transformers and export it to ONNX
ort_model = ORTModelForVision2Seq.from_pretrained(model_checkpoint, 
                                                  export=True, 
                                                  from_transformers=True,
                                                  use_cache=True)
tokenizer = PreTrainedTokenizerFast.from_pretrained(model_checkpoint)
# Save the onnx model and tokenizer
ort_model.save_pretrained(save_directory)
tokenizer.save_pretrained(save_directory)
```
```
The argument `from_transformers` is deprecated, and will be removed in optimum 2.0.  Use `export` instead
Framework not specified. Using pt to export to ONNX.
Using framework PyTorch: 2.0.1+cu117
Overriding 1 configuration item(s)
	- use_cache -> False
/home/bs/anaconda3/envs/text_rec/lib/python3.8/site-packages/transformers/models/deit/modeling_deit.py:126: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if num_channels != self.num_channels:
/home/bs/anaconda3/envs/text_rec/lib/python3.8/site-packages/transformers/models/deit/modeling_deit.py:130: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if height != self.image_size[0] or width != self.image_size[1]:
============= Diagnostic Run torch.onnx.export version 2.0.1+cu117 =============
verbose: False, log level: Level.ERROR
======================= 0 NONE 0 NOTE 0 WARNING 0 ERROR ========================

Using framework PyTorch: 2.0.1+cu117
Overriding 1 configuration item(s)
	- use_cache -> True
Using framework PyTorch: 2.0.1+cu117
Overriding 1 configuration item(s)
	- use_cache -> True
[...]

PR #1210, opened 2023-07-19
Fix vision encoder decoder that may not cache cross-attention

fxmarty:
As per title

Fixes https://github.com/huggingface/optimum/issues/1192


michaelbenayoun (approved):
LGTM!

PR #1216, opened 2023-07-20
Add GPTQ Quantization 

SunMarc:
# What does this PR do?
This PR adds the possibility to perform GTPQ quantization. I tried to be as generic as possible to support any kind of models ( whether it is Transformer models or not). The backend relies on [auto_gptq](https://github.com/PanQiWei/AutoGPTQ) library where we use `GTPQ` class and `QuantLinear` class. With this API you can do conversion, saving and loading quantized weights. We have a dependence on `accelerate` to save the weights and load the quantized weights efficiently without allocating more memory than needed. 

## Quantize model: 
```py
model_name = "bigscience/bloom-1b7"
from transformers import AutoModelForCausalLM, AutoTokenizer
from optimum.gptq import GPTQQuantizer, load_quantized_model

tokenizer = AutoTokenizer.from_pretrained(model_name, use_fast=True)
model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype = torch.float16, device_map = "auto")
quantizer = GPTQQuantizer(bits=4, dataset="c4", group_size=128, desc_act=False)
quantized_model = quantizer.quantize_model(quantized_model, tokenizer)
```
## Save model 
```py
save_folder = 'bloom-1b7-quantized_optimum'
quantizer.save(model, save_folder)
```

## Convert model and Load quantized weights
```py
from accelerate import init_empty_weights
with init_empty_weights():
    empty_model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
empty_model.tie_weights()
model_from_saved = load_quantized_model(empty_model, save_folder=save_folder, device_map="auto")
```

 PS: In a follow-up [PR](https://github.com/huggingface/transformers/pull/25062), I will use this API to integrate GTPQ quantization into transformers,  so that we take into account GPTQ quantized model, just like what we did for `bitsandbytes`

TODO: 
- [x] Doc 
- [x] CI
- [x] cpu offload (for now, if some modules are offloaded to the cpu (using device_map), the quantization will not work )
- [x] exllama integration
- [ ] script to calculate the perplexity, maybe in a new PR ...


younesbelkada (commented):
Thanks a lot for this work and integration! Left some initial comments before the review of @fxmarty with respect to potential transformers integration later on

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+            if not isinstance(getattr(self, boolean), bool):
+                raise ValueError(f"{boolean} must be a float")
+        if self.model_seqlen is not None and not isinstance(self.model_seqlen, int):
+            raise ValueError("model_seqlen must be an int")
+        if self.block_name_to_quantize is not None and not isinstance(self.block_name_to_quantize, str):
+            raise ValueError("block_name_to_quantize must be a string")
+        if self.module_name_preceding_first_block is not None and not isinstance(
+            self.module_name_preceding_first_block, list
+        ):
+            raise ValueError("block_name_to_quantize must be a list of string")
```
fxmarty, line 124: same, not needed

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+
+    def to_dict(self):
+        """
+        Return the args in dict format
```
fxmarty, line 128:
```suggestion
        Returns the args in dict format.
```

Let's stick to descriptive style within optimum (as in transformers)
https://github.com/huggingface/doc-builder#writing-documentation-for-hugging-face-libraries
https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+                quantization config
+
+        Returns:
+            `GPTQQuantizer`:  The quantizer object instantiated from those parameters.
+        """
+        return cls(**config_dict)
+
+    def _replace_by_quant_layers(self, module: nn.Module, names: List[str], name: str = ""):
+        """
+        Replace linear layers in `module` by `QuantLinear`
```
fxmarty, line 148:
```suggestion
        Replaces linear layers in `module` by `QuantLinear`
```

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+
+        Args:
+            module (`nn.Module`):
+                Module to quantize
+            names (`List[str]`):
+                List of names of the module to quantize
+            name (`str`, *optional*, defaults to `""`):
```
fxmarty, line 155:
```suggestion
            name (`str`, defaults to `""`):
```

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+    def quantize_model(
+        self,
+        model: nn.Module,
+        tokenizer: Any,
+        dataset: Union[List[str], str],
+        batch_size: int = 1,
+        pad_token_id: Optional[int] = None,
+    ):
+        """
+        Quantize the model using the dataset
```
fxmarty, line 199:
```suggestion
        Quantizes the model using the dataset
```

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+
+        Args:
+            model (`nn.Module`):
+                The model to quantize
+            dataset (`Union[List[str],str]`):
+                The dataset used for quantization. You can provide your own dataset in a list of string or just use the original datasets used
+                in the paper ['wikitext2','c4'].
```
fxmarty, line 206:
```suggestion
                in GPTQ paper ['wikitext2','c4'].
```

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+                new_layer.device = device
+                setattr(module, attr, new_layer.to(device))
+        for name1, child in module.named_children():
+            self._replace_by_quant_layers(child, names, name + "." + name1 if name != "" else name1)
+
+    @torch.no_grad()
+    def quantize_model(
+        self,
+        model: nn.Module,
+        tokenizer: Any,
```
fxmarty, line 193: Why Any?

tests/gptq/test_quantization.py
```diff
@@ -0,0 +1,174 @@
...
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from optimum.gptq import GPTQQuantizer, load_quantized_model
+from optimum.gptq.data import get_dataset
+from optimum.utils.testing_utils import require_accelerate, require_autogptq, require_torch_gpu
+
+
+@require_autogptq
+@require_torch_gpu
+class GTPQTest(unittest.TestCase):
```
fxmarty, line 30: can you make these tests slow? We probably don't want to run them in the PR/main CI given that they use large models. You would need to add a yml file in .github/workflows as well for gptq.

tests/gptq/test_quantization.py
```diff
@@ -0,0 +1,174 @@
...
+    model_name = "bigscience/bloom-560m"
+
+    input_text = "Hello my name is"
+    EXPECTED_OUTPUTS = set()
+    EXPECTED_OUTPUTS.add("Hello my name is John and I am a professional photographer. I")
+    EXPECTED_OUTPUTS.add("Hello my name is John.\nI am a friend of your father.\n")
```
fxmarty, line 36: I'm not sure to get why there's two? In my experience the small numerical difference between runs (due e.g. to unpredictable ordering) should not impact the generation

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+                The batch size of the dataset
+            pad_token_id (`Optional[int]`, *optional*, defaults to `None`):
+                The pad token id. Needed to prepare the dataset when `batch_size` > 1.
+
+        Returns:
+            `nn.Module`: The quantized model
+        """
+
+        if not torch.cuda.is_available():
+            raise RuntimeError("No GPU found. A GPU is needed to quantize model.")
```
mfuntowicz, line 219: Do we actually need a GPU to quantize the model? I don't have the details right out of my head now, but is it a strong requirement?

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+        model.eval()
+
+        # For Transformer model
+        has_config = False
+        if hasattr(model, "config"):
+            has_config = True
+        if hasattr(model, "hf_device_map"):
+            # If the model has a device_map, we don't move to model. We have already dispatch the hook that will do the work
+            has_device_map = True
+        device = get_device(model)
```
fxmarty, line 230: This will break in case all the weights of the model are not on the same device as the first weight.

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+
+        if not has_device_map:
+            # put modules from module_name_preceding_first_block on cuda
+            for module_name in self.module_name_preceding_first_block:
+                module = get_module_by_name_prefix(model, module_name)
+                if module is None:
+                    raise ValueError(f"Module {module_name} was not found in model")
+                module = module.to(0)
+            # get inputs by running self.module_name_preceding_first_block + first block on gpu
+            blocks[0] = blocks[0].to(0)
```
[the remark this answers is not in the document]
younesbelkada: I see you have added a check above, feel free to ignore this comment

optimum/gptq/quantizer.py
mfuntowicz: (Not considering the actual inference process which I understand requires a GPU, maybe quantization/calibration can be done on CPU too?)

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+        quantizers: Dict[str, Tuple],
+    ):
+        """
+        Pack the model by replacing the layers by quantized layers
+
+        Args:
+            model (`nn.Module`):
+                _description_
+            quantizers (`Dict[str,Tuple]`):
+                _description_
```
fxmarty, line 423: to do

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+        elif isinstance(dataset, list):
+            dataset = [tokenizer(data, return_tensors="pt") for data in dataset]
+        dataset = prepare_dataset(dataset, pad_token_id=pad_token_id, batch_size=batch_size)
+
+        # Step 2: get the input of the 1st block
+        layer_inputs = []
+        layer_outputs = []
+        layer_input_kwargs = []
+
+        class Catcher(nn.Module):
```
fxmarty, line 251: I don't know if you copied this bit from elsewhere, but I think you could use forward pre hooks?

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+                    )
+                    # if we pack the model at the end
+                    if not self.pack_sequentially:
+                        pass
+                        quantizers[f"{self.block_name_to_quantize}.{i}.{name}"] = (
+                            gptq[name].quantizer,
+                            scale,
+                            zero,
+                            g_idx,
+                        )
```
fxmarty, line 368: What happens here (with the pass)?

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+            desc_act (`bool`, *optional*, defaults to `True`):
+                Whether to quantize columns in order of decreasing activation size.
+                Setting it to False can significantly speed up inference but the perplexity may become slightly worse.
+                Also known as act-order.
+            sym (`bool`, *optional*, defaults to `True`):
+                Whether to use symetric quantization.
+            true_sequential (`bool`, *optional*, defaults to `True`):
+                Whether to performing sequential quantization even within a single Transformer block.
+            pack_sequentially (`bool`, *optional*, defaults to `True`):
+                Whether to pack the layer just after it is quantized. If False, we will pack the model at the end.
```
fxmarty, line 79: Pack meaning bitpack?

optimum/gptq/utils.py
```diff
@@ -0,0 +1,94 @@
...
+                    return {name: module}
+            else:
+                return {name: module}
+    res = {}
+    for name1, child in module.named_children():
+        res.update(get_layers(child, layers=layers, prefix=prefix, name=name + "." + name1 if name != "" else name1))
+    return res
+
+
+def get_block_name(model: nn.Module):
```
fxmarty, line 48: Can you add a docstring? I think the function name should be more explicit.

optimum/gptq/utils.py
```diff
@@ -0,0 +1,94 @@
...
+logger = getLogger(__name__)
+
+
+def get_module_by_name_prefix(model: nn.Module, module_name: str):
+    for name, module in model.named_modules():
+        if name.startswith(module_name):
+            return module
+
+
+def get_layers(module: nn.Module, layers=[Conv1D, nn.Conv2d, nn.Linear], prefix: Optional[str] = None, name: str = ""):
```
fxmarty, line 34: Can you add a docstring?

IMO vs recursive function it is also more readable to iterate over `name_modules()` and use recursive getattr if needed: https://github.com/huggingface/optimum/blob/e72901cc3937d6bb647e8a63648368da9b560759/optimum/utils/modeling_utils.py#L18

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+                    if k not in ["hidden_states"]:
+                        other_kwargs[k] = v
+                layer_input_kwargs.append(other_kwargs)
+                raise ValueError
+
+        # get block_name
+        if self.block_name_to_quantize is None:
+            self.block_name_to_quantize = get_block_name(model)
+
+        # get modules_name that are preceding the first block
```
fxmarty, line 278: `self.module_name_preceding_first_block = get_preceding_modules(...)` is probably explicit enough, but it could be nice to explain why (just run part of the model up to the given block, correct?)

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+        if self.module_name_preceding_first_block is None:
+            self.module_name_preceding_first_block = get_preceding_modules(model, self.block_name_to_quantize)
+
+        # get block
```
fxmarty, line 282:
```suggestion
```

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+        blocks = get_module_by_name_prefix(model, self.block_name_to_quantize)
+
+        if not has_device_map:
+            # put modules from module_name_preceding_first_block on cuda
```
fxmarty, line 286: This is already explicit from `module = module.to(0)`, but you could explain why do that rather than a `model = model.to("cuda")`. My guess is that if the user does not use accelerate, we try to avoid OOM by putting only part of the model on cuda device.

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+            for module_name in self.module_name_preceding_first_block:
+                module = get_module_by_name_prefix(model, module_name)
+                if module is None:
+                    raise ValueError(f"Module {module_name} was not found in model")
+
+        torch.cuda.empty_cache()
+
+        # Step 3: Quantize the blocks
+        quantizers = {}
+        # start quantizing the blocks
```
fxmarty, line 316:
```suggestion
```

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+        for i, block in enumerate(blocks):
+            logger.info(f"Start quantizing block {self.block_name_to_quantize} {i + 1}/{len(blocks)}")
+            # move block to cuda if needed
```
fxmarty, line 319: same

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+            damp_percent (`float`, *optional*, defaults to `0.01`):
+                The percent of the average Hessian diagonal to use for dampening, recommended value is 0.01.
+            desc_act (`bool`, *optional*, defaults to `True`):
+                Whether to quantize columns in order of decreasing activation size.
+                Setting it to False can significantly speed up inference but the perplexity may become slightly worse.
+                Also known as act-order.
+            sym (`bool`, *optional*, defaults to `True`):
+                Whether to use symetric quantization.
+            true_sequential (`bool`, *optional*, defaults to `True`):
+                Whether to performing sequential quantization even within a single Transformer block.
```
fxmarty, line 77: Can you explain in the docstring what sequential is? It is not mentioned in GPTQ paper and the original repo is not explicit about it.
```suggestion
                Whether to perform sequential quantization even within a single Transformer block.
```

fxmarty (commented):
It is in good shape, thank you for working on it!

I still think that using exllama kernel would be very interesting (for ref https://github.com/fxmarty/q4f16-gemm-gemv-benchmark & https://github.com/huggingface/text-generation-inference/pull/553#issuecomment-1642568803, second one is with Triton, that I see is not used here), but integrating with AutoGPTQ is already a nice step.

Is there a hard requirement on accelerate? If so, why?

I think it would be good to add a documentation as well in this PR, and CI.

It is also not clear to me what should go in accelerate and optimum, if this PR puts a hard requirement on accelerate. Why not put it in accelerate directly? Why bitsandbytes in accelerate & this in optimum? I think it can be quite confusing to users.

optimum/gptq/constants.py
```diff
@@ -0,0 +1,23 @@
...
+
+SEQLEN_KEYS_TRANFORMERS = ["max_position_embeddings", "seq_length", "n_positions"]
+BLOCK_PATTERNS = [
+    "transformer.h",
+    "model.decoder.layers",
+    "gpt_neox.layers",
+    "model.layers",
+]
+
+GPTQ_CONFIG = "quantization_config.json"
```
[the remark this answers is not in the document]
SunMarc: We need to keep those constants. For the `GTPQ_CONFIG`, we won't use that in `transformers`. We will save the quantization config as discussed.

optimum/gptq/data.py
```diff
@@ -0,0 +1,235 @@
...
+            Split of the dataset. Can be either "train" or "validation"
+    Returns:
+        `List[Dict[str,torch.LongTensor]]`: The tokenized dataset.
+    """
+    random.seed(seed)
+    np.random.seed(seed)
+    torch.random.manual_seed(seed)
+    if split not in ["train", "test"]:
+        raise ValueError(f"The split need to be 'train' or 'validation' but found {split}")
+    if dataset_name == "wikitext2":
```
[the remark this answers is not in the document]
SunMarc: Thx for the design !

optimum/gptq/data.py
```diff
@@ -0,0 +1,235 @@
...
+        i = random.randint(0, enc.input_ids.shape[1] - seqlen - 1)
+        j = i + seqlen
+        inp = enc.input_ids[:, i:j]
+        attention_mask = torch.ones_like(inp)
+        dataset.append({"input_ids": inp, "attention_mask": attention_mask})
+
+    return dataset
+
+
+def get_ptb_new(tokenizer, seqlen, nsamples, split="train"):
```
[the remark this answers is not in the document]
SunMarc: done

optimum/gptq/data.py
```diff
@@ -0,0 +1,235 @@
...
+        i = random.randint(0, enc.input_ids.shape[1] - seqlen - 1)
+        j = i + seqlen
+        inp = enc.input_ids[:, i:j]
+        attention_mask = torch.ones_like(inp)
+        dataset.append({"input_ids": inp, "attention_mask": attention_mask})
+
+    return dataset
+
+
+def get_ptb(tokenizer, seqlen, nsamples, split="train"):
```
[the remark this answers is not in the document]
SunMarc: done

optimum/gptq/data.py
```diff
@@ -0,0 +1,235 @@
...
+        i = random.randint(0, enc.input_ids.shape[1] - seqlen - 1)
+        j = i + seqlen
+        inp = enc.input_ids[:, i:j]
+        attention_mask = torch.ones_like(inp)
+        dataset.append({"input_ids": inp, "attention_mask": attention_mask})
+
+    return dataset
+
+
+def get_c4_new(tokenizer, seqlen, nsamples, split="train"):
```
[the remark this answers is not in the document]
SunMarc: done

optimum/gptq/data.py
```diff
@@ -0,0 +1,235 @@
...
+        "input_ids": torch.cat(input_ids_blocks, dim=0).long(),
+        "attention_mask": torch.cat(attention_mask_blocks, dim=0).long(),
+    }
+    if contain_labels:
+        data["labels"] = torch.cat(label_blocks, dim=0).long()
+
+    return data
+
+
+def get_wikitext2(tokenizer, seqlen, nsamples, split="train"):
```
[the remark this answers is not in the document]
SunMarc: done

optimum/gptq/data.py
```diff
@@ -0,0 +1,235 @@
...
+            "You need to pass a `pad_token_id` in `quantize_model` if you want to have examples with batch size > 1"
+        )
+    new_examples = [
+        collate_data(new_examples[start : start + batch_size], pad_token_id)
+        for start in range(0, len(new_examples), batch_size)
+    ]
+    return new_examples
+
+
+def collate_data(
```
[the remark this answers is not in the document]
SunMarc: done

optimum/gptq/data.py
```diff
@@ -0,0 +1,235 @@
...
+
+import random
+from typing import Any, Dict, List, Union
+
+import numpy as np
+import torch
+from datasets import load_dataset
+
+
+def prepare_dataset(
```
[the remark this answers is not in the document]
SunMarc: done

optimum/gptq/data.py
```diff
@@ -0,0 +1,235 @@
+#  Copyright 2023 The HuggingFace Team. All rights reserved.
```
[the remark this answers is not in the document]
SunMarc: done

optimum/gptq/utils.py
SunMarc: added and changed the function name by `get_block_name_with_pattern`. Let me know if it is better

optimum/gptq/utils.py
SunMarc: Replaced and added the docstring

tests/gptq/test_quantization.py
SunMarc: Done for the slow tests. I will update the workflows later

tests/gptq/test_quantization.py
SunMarc: Oh this is something that @younesbelkada did for `bitsandbytes` when we do 4bit quantization. So I just replicated but indeed I've only seen so far the first output with my test. I will only keep the first as you confirmed that it is pretty stable. 
cc @younesbelkada

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+        block_name_to_quantize: Optional[str] = None,
+        module_name_preceding_first_block: Optional[List[str]] = None,
+        *args,
+        **kwargs,
+    ):
+        """
+        Args:
+            bits (`int`):
+                The number of bits to quantize to, supported numbers are (2, 3, 4, 8).
+            group_size (int, *optional*, defaults to -1):
```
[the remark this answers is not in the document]
SunMarc: Oh thx for the clarification. I will modify that for every docstring

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+                The group size to use for quantization. Recommended value is 128 and -1 uses full row.
```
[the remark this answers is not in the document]
SunMarc: my bad, yes the per-column formulation is better. Changed it

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+            sym (`bool`, *optional*, defaults to `True`):
+                Whether to use symetric quantization.
+            true_sequential (`bool`, *optional*, defaults to `True`):
+                Whether to performing sequential quantization even within a single Transformer block.
+            pack_sequentially (`bool`, *optional*, defaults to `True`):
+                Whether to pack the layer just after it is quantized. If False, we will pack the model at the end.
+            use_cuda_fp16 (`bool`, *optional*, defaults to `True`):
+                Whether or not to use optmized cuda kernel for fp16 model. Need to have model in fp16.
+            model_seqlen (`int`, *optional*, defaults to `None`):
+                The model sequence length
```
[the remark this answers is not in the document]
SunMarc: The maximum sequence length that the model can take. For example, Bloom can take sequence length of 2048 token.

optimum/gptq/quantizer.py
SunMarc: Basically right now, we call the `pack_model` function at the end of the quantization. I wanted to see if packing the layers just after the `.fasterquant` method would help us to save memory without degradation. I will do some tests and it is useless, I will remove that.

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+                Whether to quantize columns in order of decreasing activation size.
+                Setting it to False can significantly speed up inference but the perplexity may become slightly worse.
+                Also known as act-order.
+            sym (`bool`, *optional*, defaults to `True`):
+                Whether to use symetric quantization.
+            true_sequential (`bool`, *optional*, defaults to `True`):
+                Whether to performing sequential quantization even within a single Transformer block.
+            pack_sequentially (`bool`, *optional*, defaults to `True`):
+                Whether to pack the layer just after it is quantized. If False, we will pack the model at the end.
+            use_cuda_fp16 (`bool`, *optional*, defaults to `True`):
```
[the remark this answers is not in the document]
SunMarc: If you prefer, we can set it to False and if the model is in fp16, we decide to set it to True?  For the transformers integration, i think we will make sure that the model is loaded in fp16. @younesbelkada

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+            raise ValueError("damp_percent must between 0 and 1.")
+        for boolean in [
+            "desc_act",
+            "sym",
+            "true_sequential",
+            "pack_sequentially",
+            "use_cuda_fp16",
+        ]:
+            if not isinstance(getattr(self, boolean), bool):
+                raise ValueError(f"{boolean} must be a float")
```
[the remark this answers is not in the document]
SunMarc: sounds good

optimum/gptq/quantizer.py
SunMarc: I didn't want to put `AutoTokenizer` to be as generic as possible. Any idea for a better solution ?

optimum/gptq/quantizer.py
SunMarc: Ofc, the reason is to make sure that the module preceding the first block are on the same device as the first bloc. (We move them to `cuda`) then we run the model. It will stop at the first bloc as we modified the block to raise an Exception.

optimum/gptq/quantizer.py
SunMarc: Yes that's the reason.
 In the case we don't use accelerate, there is two cases: 
- model is on cpu : we only put part of the model on cuda device for inference
- model is on gpu : we do nothing except at the end where we need to put the mode on cpu for packing 
If we use accelerate, we don't need to move the module by manually except at the end where we need to put the mode on cpu for packing.

optimum/gptq/quantizer.py
SunMarc: forgot to remove that ;)

optimum/gptq/quantizer.py
SunMarc: oups

optimum/gptq/quantizer.py
SunMarc: If the user don't use accelerate, can we make the assumption that the model is only on one device ?

optimum/gptq/quantizer.py
SunMarc: Yes ofc, i will change that

optimum/gptq/quantizer.py
SunMarc: I don't think it is a strong requirement. I need to test but you can have a look at `fasterquant` method from `GPTQ` class. Looks like it uses `cuda` to speed up some operations (`cholesky`, `cholesky_inverse`, `matmul`) and I can see that he calls `torch.cuda.synchronize()`.

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+                The number of bits to quantize to, supported numbers are (2, 3, 4, 8).
+            group_size (int, *optional*, defaults to -1):
+                The group size to use for quantization. Recommended value is 128 and -1 uses full row.
+            damp_percent (`float`, *optional*, defaults to `0.01`):
+                The percent of the average Hessian diagonal to use for dampening, recommended value is 0.01.
+            desc_act (`bool`, *optional*, defaults to `True`):
+                Whether to quantize columns in order of decreasing activation size.
+                Setting it to False can significantly speed up inference but the perplexity may become slightly worse.
+                Also known as act-order.
+            sym (`bool`, *optional*, defaults to `True`):
```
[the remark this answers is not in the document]
SunMarc: I wanted to change too but it is better to stay with the same parameters as auto_gptq because we want to make them easier to be compatible to the existing config on some [model](https://huggingface.co/TheBloke/Llama-2-13B-GPTQ/blob/main/quantize_config.json) on the hub> cc @younesbelkada

optimum/gptq/quantizer.py
SunMarc: Yes, I will do that. If you want more information you can check this [link](https://github.com/IST-DASLab/gptq#new-features) from the original repo of gtpq. They say that it improves the performance of the model.

SunMarc:
> I still think that using exllama kernel would be very interesting (for ref https://github.com/fxmarty/q4f16-gemm-gemv-benchmark & [huggingface/text-generation-inference#553 (comment)](https://github.com/huggingface/text-generation-inference/pull/553#issuecomment-1642568803), second one is with Triton, that I see is not used here), but integrating with AutoGPTQ is already a nice step (making sure that `cuda-old` is used when possible).
 
Yes, adding exllama kernel would be awesome. The best would to be add it on auto_gtpq library.

> Is there a hard requirement on accelerate? If so, why?

For the quantization, no hard requirement on accelerate. However, there is a hard requirement for saving and loading the quantized weights. I did that to mimic the behavior that we will have in transformers integration (empty model + loading weights + dispatch according to the device_map). To be clear, the saving and loading function in `GPTQQuantizer` are not mean to be used for HF transformers model. It is just a way for non Transformer model to save and load their model. 

> It is also not clear to me what should go in accelerate and optimum, if this PR puts a hard requirement on accelerate. Why not put it in accelerate directly? Why bitsandbytes in accelerate & this in optimum? I think it can be quite confusing to users.

To me, the hard requirement on accelerate is not a big problem as `transformers` has `accelerate` as core depedency and `optimum` is dependent on `transformers`. However, it might be confusing to users since bnb is on `accelerate`. 
wdyt @younesbelkada ?

gante:
Checking in from a `transformers` API perspective: looks good to me, assuming the `from_pretrained` and `save_pretrained` utilities are going to be implemented as described in the plan 🙌 

BTW, on `script to calculate the perplexity` -> this will be VERY helpful to the community, we should definitely publicize it when it is ready!

optimum/gptq/quantizer.py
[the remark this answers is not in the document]
fxmarty: I'm not sure, it's just that pytorch defaults to fp32 so it seems risky to assume a model is using fp16.

optimum/gptq/quantizer.py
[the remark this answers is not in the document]
fxmarty: Okay, it could be worth adding to the docstring.

optimum/gptq/quantizer.py
fxmarty: No sounds good!

optimum/gptq/quantizer.py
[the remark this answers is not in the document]
younesbelkada: Does this behaviour can be determined on the fly?

tests/gptq/test_quantization.py
younesbelkada: SGTM !

optimum/gptq/quantizer.py
fxmarty: What I meant (and in my next comment) is that this comment is not informative duplicate of `self.module_name_preceding_first_block = get_preceding_modules(...)`), and it can be more interesting to read the "why" we do that.

It's just taste obviously!

optimum/gptq/quantizer.py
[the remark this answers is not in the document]
SunMarc: Yes !

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,571 @@
...
+        os.makedirs(save_dir, exist_ok=True)
+        if not model._is_quantized_gptq:
+            raise EnvironmentError("can only save quantized model, please execute .quantize first.")
+        model = model.to("cpu")
+        # save model and config
+        accelerator = Accelerator()
+        accelerator.save_model(model, save_dir, max_shard_size=max_shard_size, safe_serialization=safe_serialization)
+        with open(os.path.join(save_dir, GPTQ_CONFIG), "w", encoding="utf-8") as f:
+            json.dump(self.to_dict(), f, indent=2)
+        # TODO: remove that when the integration with transformers is done
```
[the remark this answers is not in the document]
SunMarc: removed

optimum/gptq/quantizer.py
SunMarc: Thx for the suggestion ! I've removed the comment and added a comment on why we do that

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,78 @@
+# Quantization
+
+## `AutoGPTQ` Integration
+
+🤗 Optimum collaborated with `AutoGPTQ` library to provides a simple API that perform GPTQ quantization on language models.  With GPTQ quantization, you can quantize your favorite lanmguage model to 8,6,4 or even 2 bits. This comes without a big drop of performance and with faster inference speed. This is supported by most GPU hardwares.
```
younesbelkada, line 5:
```suggestion
🤗 Optimum collaborated with `AutoGPTQ` library to provide a simple API that apply GPTQ quantization on language models.  With GPTQ quantization, you can quantize your favorite lanmguage model to 8,6,4 or even 2 bits. This comes without a big drop of performance and with faster inference speed. This is supported by most GPU hardwares.
```

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+If you want to quantize 🤗 Transformers models with GPTQ, follow this [documentation](https://huggingface.co/docs/transformers/main_classes/quantization). 
+
+To learn more about the the quantization model, check out: 
+- the [GPTQ](https://arxiv.org/pdf/2210.17323.pdf) paper
+- the [`AutoGPTQ`](https://github.com/PanQiWei/AutoGPTQ) library used as the backend
+
```
younesbelkada, line 12:
```suggestion
For more advanced usage (triton backend and fused attention), please check out the `auto-gptq` library in more details.

```

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,78 @@
...
+`pip install auto-gptq`
+
+- Install latest `optimum` from source 
+`pip install --upgrade optimum`
+
+- Install latest `transformers` from source 
+`pip install --upgrade transformers`
+
+- Install latest `accelerate` from source 
+`pip install --upgrade accelerate`
```
younesbelkada, line 27:
```suggestion
- Install latest `optimum` from source 
`pip install --upgrade git+https://github.com/huggingface/optimum.git`

- Install latest `transformers` from source 
`pip install --upgrade git+https://github.com/huggingface/transformers.git`

- Install latest `accelerate` library 
`pip install --upgrade accelerate`
```

.github/workflows/test_gptq.yml
```diff
@@ -0,0 +1,90 @@
...
+    needs: start-runner # required to start the main job when the runner is ready
+    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    env:
+      AWS_REGION: us-east-1
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build image
+        run: |
+          docker build -f tests/onnxruntime/docker/Dockerfile_quantization_gpu -t gptq-gpu .
```
younesbelkada, line 63:
```suggestion
          docker build -f tests/gptq/docker/Dockerfile_quantization_gpu -t gptq-gpu .
```

younesbelkada (approved):
Amazing work @SunMarc ! All this looks great! Left few nits but the overall structure looks really nice. Thanks a lot for working on this!
Let's wait for @fxmarty's review for merging this PR :D

.github/workflows/test_gptq.yml
SunMarc: oups ;)

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,629 @@
...
+            tokenizer (`Any`):
+                The tokenizer to use in order to prepare the dataset. You can pass either:
+                    - A custom tokenizer object.
+                    - A string, the *model id* of a predefined tokenizer hosted inside a model repo on huggingface.co.
+                      Valid model ids can be located at the root-level, like `bert-base-uncased`, or namespaced under a
+                      user or organization name, like `dbmdz/bert-base-german-cased`.
+                    - A path to a *directory* containing vocabulary files required by the tokenizer, for instance saved
+                      using the [`~PreTrainedTokenizer.save_pretrained`] method, e.g., `./my_model_directory/`.
+            verbose (`bool`):
+                Whether to show the tqdm bar or not
```
fxmarty, line 248: Why not show it by default and remove this? If it takes several hours I think it can be useful

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,629 @@
...
+                # put the data on gpu, we won't put them back to cpu
+                data[k] = v.to(0)
+            try:
+                model(**data)
+            except ValueError:
+                pass
+
+        handle.remove()
+        if not has_device_map:
+            blocks[0].to(device)
```
fxmarty, line 365: Here `device` may be cpu, is it correct?

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,629 @@
...
+        ):
+            logger.info(f"Start quantizing block {self.block_name_to_quantize} {i + 1}/{len(blocks)}")
+            # move block to cuda if needed
+            # in case we have offload modules, we need to put them on cuda because of GPTQ object
+            if not has_device_map or get_device(block) == torch.device("cpu"):
+                block = block.to(0)
+            layers = get_layers(block)
+            if self.true_sequential:
+                # lazy sequential but works well
+                layers_name_list = [[key] for key in layers.keys()]
```
fxmarty, line 386: This is true sequential?

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,629 @@
...
+
+                        return tmp
+
+                    # because it adding a hook will replace the old one.
+                    handles.append(subset_layers[name].register_forward_hook(add_batch(name)))
+                # update Hessian for each layer in subset_layers thanks to the hook
+                for j in range(len(dataset)):
+                    # the args are already on the gpu
+                    # don't need to store the output
+                    block(layer_inputs[j], **layer_input_kwargs[j])
```
fxmarty, line 413: There's something I don't get here. The `layer_inputs` are the inputs to the first layer in `blocks`. For example, for neox, the inputs to the first `GPTNeoXLayer` from `self.layers =  nn.ModuleList([GPTNeoXLayer(config) for _ in range(config.num_hidden_layers)])`.

Here, we use the inputs of the first layer for the n-th layer?

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,629 @@
...
+                        device_layer = get_device(quantized_layer)
+                        quantized_layer = quantized_layer.to("cpu")
+                        quantized_layer.pack(subset_layers[name], scale, zero, g_idx)
+                        quantized_layer = quantized_layer.to(device_layer)
+                    gptq[name].free()
+                del subset_layers
+            # we get the new output from the partial quantized block
+            for j in range(len(dataset)):
+                layer_output = block(layer_inputs[j], **layer_input_kwargs[j])[0]
+                layer_outputs.append(layer_output)
```
fxmarty, line 450: This does not seem to be used

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,629 @@
...
+        # if we pack the model at the end
+        if not self.pack_sequentially:
+            self.pack_model(model=model, quantizers=quantizers)
+
+        model._is_gptq_quantized = True
+        if has_config:
+            model.config.use_cache = use_cache
+            model.config.quantization_config = self.to_dict()
+
+        # Step 5: Compatibility with peft
```
fxmarty, line 470: ?

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,79 @@
...
+You need to have the following requirements installed to run the code below: 
+
+- Install latest `AutoGPTQ` library
+`pip install auto-gptq`
+
+- Install latest `optimum` from source 
+`pip install --upgrade git+https://github.com/huggingface/optimum.git`
+
+- Install latest `transformers` from source 
+`pip install --upgrade git+https://github.com/huggingface/transformers.git`
```
fxmarty, line 25: Same, not sure we want that. Also, why would 4.31 not work?

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+- Install latest `accelerate` library 
+`pip install --upgrade accelerate`
+
+### Load and quantize model
```
fxmarty, line 30:
```suggestion
### Load and quantize a model
```

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+```py
+from transformers import AutoModelForCausalLM, AutoTokenizer
+from optimum.gptq import GPTQQuantizer, load_quantized_model
+model_name = "facebook/opt-125m"
+tokenizer = AutoTokenizer.from_pretrained(model_name)
+model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
+
+quantizer = GPTQQuantizer(bits=4, dataset="c4", block_name_to_quantize = "model.decoder.layers", model_seqlen = 2048)
+quantized_model = quantizer.quantize_model(quantized_model, tokenizer)
```
fxmarty, line 48: `quantized_model` is not defined here

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,79 @@
...
+quantizer.save(model,save_folder)
+```
+
+### Load quantized weights
+
+You can load your quantized weights by using the [`~optimum.gptq.load_quantized_model`] function.
+```python
+from optimum.gptq import load_quantized_model
+model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
+model_from_saved = load_quantized_model(model, save_folder=save_folder, device_map = "auto")
```
fxmarty, line 69: would
```suggestion
quantized_model = load_quantized_model(model, save_folder=save_folder, device_map = "auto")
```
make sense? Does it duplicate memory between model + quantized_model?

fxmarty, line 69: Would
```suggestion
with torch.device("meta"):
    model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
quantized_model = load_quantized_model(model, save_folder=save_folder, device_map = "auto")
```

work?

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,79 @@
...
+save_folder = "/path/to/save_folder/"
+quantizer.save(model,save_folder)
+```
+
+### Load quantized weights
+
+You can load your quantized weights by using the [`~optimum.gptq.load_quantized_model`] function.
+```python
+from optimum.gptq import load_quantized_model
+model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
```
fxmarty, line 68: `model_name` is not defined, it would help if example codes are copy pastable and runable

fxmarty (commented):
Looks great! I left a few questions/style comments

docs/source/optimization_toolbox/usage_guides/quantization.mdx
SunMarc: indeed, any version of transformers should work.

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+### Requirements
+
+You need to have the following requirements installed to run the code below: 
+
+- Install latest `AutoGPTQ` library
+`pip install auto-gptq`
+
+- Install latest `optimum` from source 
+`pip install --upgrade git+https://github.com/huggingface/optimum.git`
```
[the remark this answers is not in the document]
SunMarc: sounds good

docs/source/optimization_toolbox/package_reference/quantization.mdx
```diff
@@ -0,0 +1,17 @@
+<!--Copyright 2023 The HuggingFace Team. All rights reserved.
```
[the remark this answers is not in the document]
SunMarc: where should i put the file ?

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -0,0 +1,79 @@
...
+The [`~optimum.gptq.GPTQQuantizer`] class is used to quantize your model. In order to quantize your model, you need to provide a few arguemnts:
+- the number of bits: `bits`
+- the dataset used to calibrate the quantization: `dataset`
+- the model sequence length used to process the dataset: `model_seqlen`
+- the block name to quantize: `block_name_to_quantize`
+
+With 🤗 Transformers integration, you don't need to pass the `block_name_to_quantize` and `model_seqlen` as we can retrieve them. However, for custom model, you need to specify them. Also, make sure that your model is converted to `torch.float16` before quantization.
+
+```py
+from transformers import AutoModelForCausalLM, AutoTokenizer
```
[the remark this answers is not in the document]
SunMarc: fixed

docs/source/optimization_toolbox/usage_guides/quantization.mdx
SunMarc: ofc

docs/source/optimization_toolbox/usage_guides/quantization.mdx
SunMarc: It will indeed work. `init_empty_weights` from accelerate do that but it can also take care of buffer if specified. I forgot that load_quantized_model relies on accelerate. I think it will be better to merge the two examples.

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,629 @@
...
+        module_name_preceding_first_block: Optional[List[str]] = None,
+        batch_size: int = 1,
+        pad_token_id: Optional[int] = None,
+        *args,
+        **kwargs,
+    ):
+        """
+        Args:
+            bits (`int`):
+                The number of bits to quantize to, supported numbers are (2, 3, 4, 8).
```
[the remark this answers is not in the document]
SunMarc: This is a constrain from the backend. See [doc](https://github.com/PanQiWei/AutoGPTQ/blob/a7167b108c438f570938f0ced46a52fe515f4a59/auto_gptq/nn_modules/qlinear/qlinear_cuda_old.py#L35C8-L36C74)

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,629 @@
...
+            dataset (`Union[List[str],str]`, defaults to None):
```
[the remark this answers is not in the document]
SunMarc: We don't need the dataset when loading quantized weights. I decided to put `None`  because the repo from `TheBloke` don't have a dataset field. This way, we don't need to pass a hardcoded dataset to load these models.

optimum/gptq/quantizer.py
```diff
@@ -0,0 +1,629 @@
...
+            group_size (int, defaults to -1):
+                The group size to use for quantization. Recommended value is 128 and -1 uses per-column quantization.
+            damp_percent (`float`, defaults to `0.01`):
+                The percent of the average Hessian diagonal to use for dampening, recommended value is 0.01.
+            desc_act (`bool`, defaults to `True`):
+                Whether to quantize columns in order of decreasing activation size.
+                Setting it to False can significantly speed up inference but the perplexity may become slightly worse.
+                Also known as act-order.
+            sym (`bool`, defaults to `True`):
+                Whether to use symetric quantization.
```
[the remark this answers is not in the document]
SunMarc: This was the default value in autogptq. should i changed it to False ?

optimum/gptq/quantizer.py
SunMarc: works for me !

optimum/gptq/quantizer.py
SunMarc: yes !

optimum/gptq/quantizer.py
SunMarc: Yes. we update one layer at a time.

optimum/gptq/quantizer.py
SunMarc: it is used here : `layer_inputs, layer_outputs = layer_outputs, []`. See comment above.

optimum/gptq/quantizer.py
SunMarc: No, we use the output of the n-1 layer as the inputs of the n layers. This is done by getting the outputs and with: 
`layer_inputs, layer_outputs = layer_outputs, []`

optimum/gptq/quantizer.py
SunMarc: removed

docs/source/optimization_toolbox/package_reference/quantization.mdx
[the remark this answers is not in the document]
fxmarty: Maybe the easiest would be to move

```
[[autodoc]] gtpq.GPTQQuantizer
    - all

[[autodoc]] gtpq.load_quantized_model
    - all
```

to quantization.mdx?

optimum/gptq/quantizer.py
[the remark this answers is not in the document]
fxmarty: Thanks

optimum/gptq/quantizer.py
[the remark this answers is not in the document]
fxmarty: Thank you. In my opinion this implementation is not the most elegant one, there is in my opinion two distinct tasks:
* Quantizing a model that is in fp16, for which you need to know all of the `damp_percent`, `dataset`, `batch_size`, `true_sequential`, etc. and saving it to disk.
* Loading a model *that is already quantized*, for which you only need to have the knowledge of bits, g_idx, qweights, qzeros, scales, group_size and possibly sym. Having the knowledge of which layers were replaced is known from the state_dict keys.

Here it feels like the `GPTQQuantizer` mixes up the two tasks, by hacking the `__init__` signature with defaults that should not really be there.

It is fine to keep as is, but I think a simpler design dissociating the two could have been possible. Let me know your thoughts!

optimum/gptq/quantizer.py
fxmarty: Thank you, I missed it!

docs/source/optimization_toolbox/usage_guides/quantization.mdx
fxmarty: Adding the `QuantizationMethod.GPTQ`, I think this won't work anymore with 4.31.

fxmarty (approved):
LGTM, thank you for adding it!

Just left taste comments / answers to the threads above. I believe this breaks with transformers 4.31 now, but maybe that's fine

fxmarty:
Can you run `make style`?

optimum/gptq/quantizer.py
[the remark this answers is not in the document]
SunMarc: Totally agree with you. At first, I also wanted to separe these two tasks. However, it was not compatible with Transformers integration if we wanted to quantize a model or load an already quantized model (must have a `quantization_config`) with only the `from_pretrained()`method. I had to put all the arguments inside the `GPTQQuantizer`.

docs/source/optimization_toolbox/package_reference/quantization.mdx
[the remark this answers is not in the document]
SunMarc: sounds good

docs/source/optimization_toolbox/usage_guides/quantization.mdx
SunMarc: nice catch

docs/source/_toctree.yml
```diff
@@ -125,6 +125,11 @@
     isExpanded: false
   title: BetterTransformer
   isExpanded: false
+- sections:
+  - local: optimization_toolbox/usage_guides/quantization
+    title: GPTQ quantization
+  title: Optimization toolbox
```
regisss, line 131: Not sure about this title as many things in Optimum are about optimization :sweat_smile: 
I would just name it "GPTQ" for now, or maybe better "LLM quantization". WDYT?

regisss (commented):
Very nice PR @SunMarc :fire: :rocket:  
Just commenting the title of the new doc section.
Also, could you add a bullet point for GPTQ in [this section](https://github.com/huggingface/optimum/blob/94bf76698638e8a5ef48392e5951375d09e72183/docs/source/concept_guides/quantization.mdx#L175) please?

docs/source/_toctree.yml
SunMarc: Yes ! LLM quantization sounds good =)

docs/source/optimization_toolbox/usage_guides/quantization.mdx
```diff
@@ -73,6 +73,23 @@ empty_model.tie_weights()
 quantized_model = load_quantized_model(empty_model, save_folder=save_folder, device_map="auto")
 ```
 
+### Exllama kernels for faster inference
+
+For 4-bit model, you can use the exllama kernels in order to a faster inference speed. You just need to pass `disable_exllama = False` in [`~optimum.gptq.load_quantized_model`]. In order to use these kernels, you need to have the entire model on gpus.
```
fxmarty: Isn't False the default?

SunMarc: Yes, the default behavior is [False](url). I will clarify the message.

PR #1223, opened 2023-07-24
Set default value to unet config sample size

echarlaix:
Set default value for model without any config such as [tiny-random-OnnxStableDiffusionPipeline](https://huggingface.co/hf-internal-testing/tiny-random-OnnxStableDiffusionPipeline). If no configuration provided, the default `height` and `width` value will be [512](https://github.com/huggingface/diffusers/blob/v0.8.0/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py#L392)

### DIFF
--- a/optimum/pipelines/diffusers/pipeline_stable_diffusion.py
+++ b/optimum/pipelines/diffusers/pipeline_stable_diffusion.py
@@ -286,8 +286,8 @@ class StableDiffusionPipelineMixin(DiffusionPipelineMixin):
             list of `bool`s denoting whether the corresponding generated image likely represents "not-safe-for-work"
             (nsfw) content, according to the `safety_checker`.
         """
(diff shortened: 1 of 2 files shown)


fxmarty:
This seem different compared to diffusers that uses `height = height or self.unet.config.sample_size * self.vae_scale_factor`? Should we just add the relevant config to the tiny model?

fxmarty:
Shouldn't this fix be in diffusers?

echarlaix:
> Shouldn't this fix be in diffusers?

I'm talking about models that have already been exported such as https://huggingface.co/axodoxian/stable_diffusion_onnx or https://huggingface.co/sayakpaul/onnx-stable-diffusion-v1-5

fxmarty (approved):
LGTM

echarlaix:
Merging as failing tests are unrelated

PR #1228, opened 2023-07-25
add upgrade strategy

echarlaix:
As discussed with @regisss, adding `--upgrade-strategy eager` so that the latest subpackages are installed.

With `pip install --upgrade optimum[intel]`  the latest optimum version will be downloaded but not the latest version of optimum-intel


fxmarty (approved):
LGTM

regisss (approved):
Thanks @echarlaix :smiley: 

If you have time, you can add entries for Furiosa in both tables, otherwise we'll do it in another PR :slightly_smiling_face:

echarlaix:
> If you have time, you can add entries for Furiosa in both tables, otherwise we'll do it in another PR 🙂

Good point,  will do it right now !

PR #1232, opened 2023-07-27
fix SD loading when safetensors weights only

echarlaix:
Fix ONNX export of SD model with safetensors weights such as [`stable-diffusion-xl-base-1.0`](https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0)


fxmarty (approved):
LGTM

PR #1234, opened 2023-07-27
fix optimum-intel min version

echarlaix:
Make sure latest `optimum-intel` version is installed for the `openvino` subpackage to be compatible with latest `optimum`  (keeping compatiblity even for `pip install --upgrade optimum[intel]`)


fxmarty:
Great, maybe we can remove the `--upgrade-strategy eager` from the README then (or at least when all subpackages align) and use back `--upgrade`?

echarlaix:
> Great, maybe we can remove the `--upgrade-strategy eager` from the README then (or at least when all subpackages align) and use back `--upgrade`?

Yes fine with me ! Let's wait for all subpackages to be aligned cc @regisss @JingyaHuang @michaelbenayoun @mht-sharma

PR #1235, opened 2023-07-27
Support Llama 2 in BetterTransformer.

noamwies:
# This PR update the `llama_forward` function in order for BetterTransformer to support the grouped self-attention in Llama-2.

### DIFF
--- a/optimum/bettertransformer/models/attention.py
+++ b/optimum/bettertransformer/models/attention.py
@@ -17,8 +17,7 @@ from typing import Optional, Tuple
 import torch
 from transformers.models.llama.modeling_llama import _expand_mask as _llama_expand_mask
 from transformers.models.llama.modeling_llama import _make_causal_mask as _llama_make_causal_mask
-from transformers.models.llama.modeling_llama import apply_rotary_pos_emb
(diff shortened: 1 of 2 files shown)


fxmarty:
Thank you for the update @noamwies , can you add a test for it? For example with https://huggingface.co/BlackSamorez/llama-2-tiny-testing (that has no tokenizer, so you could fork it and add a tokenizer)

I think you would need to edit https://github.com/huggingface/optimum/blob/608188c4ddecc28401af289c7b5f9ca7eef4e0d0/tests/bettertransformer/testing_utils.py#L52,

and replace

```
model_id = MODELS_DICT[model_type]
```
by
```python
model_ids = [MODELS_DICT[model_type]] if isinstance(MODELS_DICT[model_type], str) else MODELS_DICT[model_type]
for model_id in model_ids:
    ....
```

in the tests here https://github.com/huggingface/optimum/blob/main/tests/bettertransformer/test_decoder.py

noamwies:
Thanks, per your suggestion, I added a test for my changes. I saw that in other models such as the CLIP model, you handle the multiple variant issue by having a dedicated key for each variant. Hence, I added a dedicated Llama 2 variant with GQA.

@fxmarty

Issue #1243, opened 2023-08-01
transformers.convert_graph_to_onnx.quantize equivalent with optimum?


jobergum:
Historically, I've used the following to quantize a model after training:

```python
import sys
from pathlib import Path 
from transformers.convert_graph_to_onnx import quantize

input_file = sys.argv[1]
print("Performing quantization of model '{}'".format(input_file))
quantized_model_path =  quantize(Path(input_file))
print("Rename quantized model '{}' to '{}'".format(quantized_model_path.name, input_file))
quantized_model_path.replace(input_file)
```

Is there a way to accomplish the same type of quantization using`optimum-cli? The quantize method from above (that is deprecated) produces a much smaller model than optimum-cli. 

```
Original model 448M multilingual-e5-small-onnx/model.onnx
Model after above 112M  multilingual-e5-small-onnx/model.onnx
```

I've tried the following export/quantize commands, but the model file size is still above 400MB

```
$ optimum-cli export onnx --task sentence-similarity -m intfloat/multilingual-e5-small --optimize O3 multilingual-e5-small-onnx
$ optimum-cli onnxruntime quantize --onnx_model multilingual-e5-small-onnx --avx2 --output test
```

```
403M Aug  1 09:38 test/model_quantized.onnx
```
Thank you!

fxmarty:
Hi, thank you for the report! The differences lies in the default operators quantized in the previous transformers script vs Optimum. By default, for some reason, ORT optimization replaces most `MatMul` operations by `Attention` ops, which is for some reason not in the defaults ops to quantize in Optimum (`MatMul`, `Add`). https://github.com/huggingface/optimum/pull/1246 fixes it.

Using this fix, the model size goes 450 MiB to 112.7 MiB. Of course it is then better to check whether the prediction quality is still good after quantization.

Unfortunately, the CLI currently does not allow to specify which ops to quantize.

Note that for some reason, passing `--optimize O3` or `--optimize O2` during the export seem to lower quite a bit the precision of the output - I don't think it is an issue on our end but I'll double check.

jobergum:
Thank you for the swift resolution of this!

>Using this fix, the model size goes 450 MiB to 112.7 MiB. Of course it is then better to check whether the prediction quality is still good after quantization.

Yes, this is a dense retriever model, and on the dataset, I'm evaluating on there is only a 2% drop in retrieval accuracy (NDCG@10) with the quantized model.

PR #1246, opened 2023-08-01
ORT quantizes by default all ops

fxmarty:
As per title

Partially fixes https://github.com/huggingface/optimum/issues/1243 and align with the previous transformers script


michaelbenayoun (commented):
LGTM

PR #1250, opened 2023-08-03
Drop mask for training in all cases for BetterTransformer & precise documentation

fxmarty:
cc @younesbelkada

### DIFF
--- a/docs/source/bettertransformer/overview.mdx
+++ b/docs/source/bettertransformer/overview.mdx
@@ -12,17 +12,24 @@ specific language governing permissions and limitations under the License.
 
 # Overview
 
-🤗 Optimum provides an integration with `Better Transformer`, a fast path of standard PyTorch Transformer APIs to benefit from interesting speedups on CPU & GPU through sparsity and fused kernels. For now, it supports Transformer encoders, basically fast path of [`nn.TransformerEncoderLayer`](https://pytorch.org/blog/a-better-transformer-for-fast-transformer-encoder-inference/),
-support for decoders and training path is coming soon.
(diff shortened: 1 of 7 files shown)


younesbelkada (approved):
Thanks for this and for clarifying the docs! 🚀

michaelfeil:
@fxmarty With torch2.2+, is the kernel still not able to dispatch an `attn_mask`?

fxmarty:
@michaelfeil It appears not to (`torch==2.2.0`):

```
--- no mask
--- with mask
/home/felix/test_sdpa.py:16: UserWarning: Memory efficient kernel not used because: (Triggered internally at ../aten/src/ATen/native/transformers/cuda/sdp_utils.cpp:415.)
  res = torch.nn.functional.scaled_dot_product_attention(query, key, value, attn_mask=attn_mask)
/home/felix/test_sdpa.py:16: UserWarning: Memory Efficient attention has been runtime disabled. (Triggered internally at ../aten/src/ATen/native/transformers/sdp_utils_cpp.h:456.)
  res = torch.nn.functional.scaled_dot_product_attention(query, key, value, attn_mask=attn_mask)
/home/felix/test_sdpa.py:16: UserWarning: Flash attention kernel not used because: (Triggered internally at ../aten/src/ATen/native/transformers/cuda/sdp_utils.cpp:417.)
  res = torch.nn.functional.scaled_dot_product_attention(query, key, value, attn_mask=attn_mask)
/home/felix/test_sdpa.py:16: UserWarning: Both fused kernels do not support non-null attn_mask. (Triggered internally at ../aten/src/ATen/native/transformers/sdp_utils_cpp.h:272.)
  res = torch.nn.functional.scaled_dot_product_attention(query, key, value, attn_mask=attn_mask)
  File "/home/felix/test_sdpa.py", line 16, in <module>
    res = torch.nn.functional.scaled_dot_product_attention(query, key, value, attn_mask=attn_mask)
RuntimeError: No available kernel. Aborting execution.
```

We could however update BT so that memory-efficient attention is used, that supports a custom attn_mask. This is however not at the top of my priority, as we are rather adding native SDPA support in Transformers: https://huggingface.co/docs/transformers/perf_infer_gpu_one#pytorch-scaled-dot-product-attention

PR #1251, opened 2023-08-03
Fix doc build

JingyaHuang:


JingyaHuang:
@fxmarty As the Trainer now largely use accelerate, I will need `accelerate` as a dependency for building the doc (otherwise error like [this](https://github.com/huggingface/optimum/actions/runs/5751329591/job/15589861592)), WDYT, would it makes sense to have it in the base requirements?

fxmarty:
What do you think @michaelbenayoun ? Maybe we could make it a soft dependency, having a check when using ORTTrainer whether `accelerate` is available, and erroring out if it is not? And just `pip install accelerate` in the Dockerfile / workflow?

fxmarty:
Okay so accelerate is now a hard requirement of transformers[torch] following https://github.com/huggingface/transformers/pull/22752. It is not a hard requirement of transformers though: https://github.com/huggingface/transformers/blob/641adca55832ed9c5648f54dcd8926d67d3511db/setup.py#L415

Personally I don't think we should put accelerate as a hard requirement given that it will impact all subpackages, that may not require it.

Let me know if https://github.com/huggingface/optimum/pull/1253 is good @JingyaHuang

JingyaHuang:
Closing this one for #1253

PR #1253, opened 2023-08-04
Accelerate as a soft dependency

fxmarty:
As per title

### DIFF
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -33,13 +33,28 @@ from transformers.integrations import (
     is_fairscale_available,
 )
 
+from transformers.utils import is_accelerate_available
+from packaging import version
+
+if is_accelerate_available():
+    from accelerate import __version__ as accelerate_version
+    from accelerate.utils import DistributedDataParallelKwargs
+
+    if version.parse(accelerate_version) >= version.parse("0.16"):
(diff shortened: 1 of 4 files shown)


optimum/onnxruntime/trainer.py
```diff
@@ -33,13 +33,28 @@
     is_fairscale_available,
 )
 
+from transformers.utils import is_accelerate_available
+from packaging import version
+
+if is_accelerate_available():
```
JingyaHuang, line 39: From what I remember, this won't help solving the doc build dependency issue.

fxmarty: @JingyaHuang I edited the dockerfile for docbuild accordingly.

fxmarty:
@JingyaHuang WDYT?

JingyaHuang:
@fxmarty I am for setting accelerate as soft dependency as far as the doc build works.

optimum/onnxruntime/trainer.py
JingyaHuang: Why not, let's just keep it.

fxmarty:
It worked! I'll do a patch release for the documentation.

JingyaHuang:
Do you want to update that in the v1.11.0 branch as well, I think the doc is missing for the release.

PR #1257, opened 2023-08-04
Enable the export of only one decoder 

echarlaix:
In this PR we refactorize the `ORTModelForCausalLM` class so that the decoder doesn't need to be exported twice (and optionally merged) when `use_cache=True`. To enable this, we create empty `past_key_values` when the ONNX model expects it and no `past_key_values` have been computed yet (during the model first generation).
All previously exported models are still compatible and will be loaded by default with : 
* `use_merged=True` -> `"decoder_model_merged.onnx"`
* `use_cache=True` -> `"decoder_model_with_past.onnx"`
* `use_cache=False` -> `"decoder_model.onnx"`

For [theses](https://github.com/huggingface/optimum/blob/refactorization-decoder-ort/optimum/onnxruntime/modeling_decoder.py#L977) architectures, when exporting the model with a sequence length of 1, the attention masks will be generated incorrectly for sequence length greater than 1 https://github.com/huggingface/transformers/blob/0ee45906845c8d58b9bd2df5acd90e09b00047ff/src/transformers/models/bloom/modeling_bloom.py#L654 for this reason we have introduced specific model patcher for these architectures. This results in the previously exported model to generate incorrect decoder attention masks and ultimately to generate outputs which differs from the expected output, in this specific case we raise an error asking to re-export the model : it would make sense to include it in a v2.0 release alongside with the new `position_ids` inputs fix : https://github.com/huggingface/optimum/pull/1381


optimum/commands/export/onnx.py
```diff
@@ -217,6 +217,11 @@ def parse_args_onnx(parser):
         default=DEFAULT_DUMMY_SHAPES["nb_points_per_image"],
         help="For Segment Anything. It corresponds to the number of points per segmentation masks.",
     )
+    optional_group.add_argument(
+        "--legacy",
```
fxmarty, line 221: We could merge the `no_position_ids` and `legacy` as they correspond to the previous export behavior and `no_position_ids` is not in a release yet. WDYT?

optimum/exporters/onnx/config.py
```diff
@@ -164,10 +164,7 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
         # generating wrong position_ids in the model itself:
         # https://github.com/huggingface/transformers/blob/v4.33.1/src/transformers/models/gpt2/modeling_gpt2.py#L802
         if not self.no_position_ids and self.task == "text-generation":
-            if self.use_past_in_inputs:
-                common_inputs["position_ids"] = {0: "batch_size"}
-            else:
-                common_inputs["position_ids"] = {0: "batch_size", 1: "sequence_length"}
+            common_inputs["position_ids"] = {0: "batch_size", 1: "sequence_length"}
```
fxmarty, line 167: Why?

optimum/exporters/onnx/model_configs.py
```diff
@@ -400,7 +428,7 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
         return int_tensor
 
 
-class BartOnnxConfig(TextSeq2SeqOnnxConfig):
+class M2M100OnnxConfig(TextSeq2SeqOnnxConfig):
```
fxmarty, line 431: I'm not sure to understand the inheritance change here for M2M100/Bart/Mbart.

optimum/exporters/onnx/model_patcher.py
```diff
@@ -342,3 +344,114 @@ def patched_forward(
...
+class BlenderbotSmallModelPatcher(BartModelPatcher):
+    pass
+
+
+class BlenderbotModelPatcher(BartModelPatcher):
+    pass
+
+
+class PegasusModelPatcher(BartModelPatcher):
+    pass
```
fxmarty, line 457: This is just taste but for these archs I would just use the parent model patcher for the child models.

optimum/exporters/onnx/utils.py
```diff
@@ -158,15 +159,16 @@ def _get_submodels_for_export_stable_diffusion(
...
 ) -> Dict[str, Union["PreTrainedModel", "TFPreTrainedModel"]]:
     """
     Returns the decoder part of the model.
     """
-    models_for_export = {}
+    models_for_export = {ONNX_DECODER_NAME if legacy else "model": model}
 
-    models_for_export[ONNX_DECODER_NAME] = model
-    if use_past:
+    if legacy and use_past:
```
fxmarty, line 171: How about encoder-decoders when legacy=False?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -648,39 +703,440 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
     )
     def forward(
         self,
-        input_ids: torch.LongTensor = None,
+        input_ids: torch.LongTensor,
         attention_mask: Optional[torch.FloatTensor] = None,
-        past_key_values: Optional[Tuple[Tuple[torch.Tensor]]] = None,
         position_ids: Optional[torch.LongTensor] = None,
+        past_key_values: Optional[Tuple[Tuple[torch.Tensor]]] = None,
```
fxmarty, line 709: Why this reordering?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -648,39 +703,440 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
...
+        )
+        new_model_save_dir = model_cache_path.parent
+
+        # model_save_dir can be provided in kwargs as a TemporaryDirectory instance, in which case we want to keep it
+        # instead of the path only.
+        if model_save_dir is None:
+            model_save_dir = new_model_save_dir
+
+        # Since v1.7.0 decoder with past models have fixed sequence length of 1
+        # To keep these models compatible we set this dimension to dynamic
```
fxmarty, line 1029: I don't understand what happens here. Why is it needed to patch the ONNX models? Why is the output logits sequence length = 1 for with-past ONNX models an issue with the decoder_model.onnx/decoder_with_past_model.onnx?

tests/onnxruntime/test_modeling.py
```diff
@@ -1971,26 +1968,39 @@ class ORTModelForCausalLMIntegrationTest(ORTModelTestMixin):
...
+            outputs = model.generate(**tokens, num_beams=1, do_sample=False, min_new_tokens=30, max_new_tokens=30)
+            self.assertTrue(torch.allclose(outputs_onnx, outputs))
 
     def test_load_model_from_hub_onnx(self):
         model = ORTModelForCausalLM.from_pretrained("fxmarty/onnx-tiny-random-gpt2-without-merge")
 
         self.assertFalse(model.use_merged)
         self.assertTrue(model.use_cache)
-        self.assertTrue(model.decoder_with_past is not None)
+        self.assertIsInstance(model.model, onnxruntime.capi.onnxruntime_inference_collection.InferenceSession)
```
fxmarty, line 1995:
```suggestion
        self.assertIsInstance(model.model, onnxruntime.InferenceSession)
```

tests/onnxruntime/test_modeling.py
```diff
@@ -2025,12 +2036,12 @@ def test_merge_from_onnx_and_save(self, model_arch):
...
 
         with tempfile.TemporaryDirectory() as tmpdir:
-            main_export(model_id, tmpdir, task=task)
+            main_export(model_id, tmpdir, task=task, legacy=True)
 
             model = ORTModelForCausalLM.from_pretrained(tmpdir)
 
             self.assertTrue(model.use_merged)
-            self.assertTrue(model.decoder_with_past is None)
+            self.assertIsInstance(model.model, onnxruntime.capi.onnxruntime_inference_collection.InferenceSession)
```
fxmarty, line 2044:
```suggestion
            self.assertIsInstance(model.model, onnxruntime.InferenceSession)
```

tests/onnxruntime/test_modeling.py
```diff
@@ -2009,12 +2019,13 @@ def test_merge_from_transformers_and_save(self, model_arch):
...
             model.save_pretrained(tmpdir)
-            save_path = os.path.join(tmpdir, ONNX_DECODER_MERGED_NAME)
-            self.assertTrue(has_onnx_input(save_path, "use_cache_branch"))
+            save_path = os.path.join(tmpdir, ONNX_WEIGHTS_NAME)
+            self.assertFalse(has_onnx_input(save_path, "use_cache_branch"))
 
             folder_contents = os.listdir(tmpdir)
             self.assertTrue(ONNX_DECODER_NAME not in folder_contents)
             self.assertTrue(ONNX_DECODER_WITH_PAST_NAME not in folder_contents)
+            self.assertTrue(ONNX_DECODER_MERGED_NAME not in folder_contents)
```
fxmarty, line 2028: I'm not sure to understand. `use_merged=True`, `save_pretrained` and the merged decoder not in the saved folder?

tests/onnxruntime/test_modeling.py
```diff
@@ -2063,21 +2075,11 @@ def test_compare_to_transformers(self, test_name: str, model_arch: str, use_cach
...
-        self.assertIsInstance(onnx_model.decoder, ORTDecoder)
-        if onnx_model.use_cache is True and onnx_model.use_merged is False:
-            self.assertIsInstance(onnx_model.decoder_with_past, ORTDecoder)
-        if onnx_model.use_cache is True and onnx_model.use_merged is True:
-            self.assertTrue(onnx_model.decoder_with_past is None)
 
+        model_path = Path(self.onnx_model_dirs[test_name], ONNX_WEIGHTS_NAME)
+        self.assertFalse(has_onnx_input(model_path, "use_cache_branch"))
+        self.assertFalse(onnx_model.use_merged)
+        self.assertIsInstance(onnx_model.model, onnxruntime.capi.onnxruntime_inference_collection.InferenceSession)
```
fxmarty, line 2082:
```suggestion
        self.assertIsInstance(onnx_model.model, onnxruntime.InferenceSession)
```

tests/onnxruntime/test_modeling.py
```diff
@@ -2312,15 +2314,14 @@ def test_compare_merged_and_not_merged_models_outputs(self, test_name: str, mode
...
+        self.assertFalse(model_not_merged.use_merged)
 
         model_merged = ORTModelForCausalLM.from_pretrained(model_merged_dir)
-        merged_onnx_path = Path(model_merged_dir, ONNX_DECODER_MERGED_NAME)
-        self.assertTrue(has_onnx_input(merged_onnx_path, "use_cache_branch"))
-        self.assertEqual(model_merged.decoder_with_past, None)
-        self.assertEqual(model_merged.use_merged, True)
+        merged_onnx_path = Path(model_merged_dir, ONNX_WEIGHTS_NAME)
+        self.assertFalse(has_onnx_input(merged_onnx_path, "use_cache_branch"))
+        self.assertFalse(model_merged.use_merged)
```
fxmarty, line 2324: ?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -648,39 +703,440 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
...
+                "blenderbot-small",
+                "blenderbot",
+                "opt",
+                "pegasus",
+                "bart",
+            }
+            if file_name == ONNX_DECODER_WITH_PAST_NAME and config.model_type in MODEL_TO_PATCH_FOR_PAST:
+                raise ValueError(
+                    f"{ONNX_DECODER_WITH_PAST_NAME} not supported for the following architecture : {', '.join(MODEL_TO_PATCH_FOR_PAST)}. Please re-export your model or set use_cache=False."
+                )
```
fxmarty, line 990: I am not sure to understand what is happening here.

fxmarty (commented):
Great work @echarlaix!

optimum/commands/export/onnx.py
```diff
@@ -217,6 +217,11 @@ def parse_args_onnx(parser):
...
+        action="store_true",
+        help=("Export decoder only models in two (without + with past) model as a single ONNX file."),
```
michaelbenayoun, line 223:
```suggestion
        help="Export decoder only models in two (without + with past) model as a single ONNX file.",
```
QQ: the `--legacy` flag should export two ONNX files no?

optimum/exporters/onnx/model_configs.py
```diff
@@ -261,6 +284,11 @@ def add_past_key_values(self, inputs_or_outputs: Dict[str, Dict[int, str]], dire
                 1: decoder_sequence_name,
             }
 
+    def patch_model_for_export(
+        self, model: Union["PreTrainedModel", "TFPreTrainedModel"], model_kwargs: Optional[Dict[str, Any]] = None
+    ) -> "ModelPatcher":
+        return BloomModelPatcher(self, model, model_kwargs=model_kwargs)
```
michaelbenayoun, line 290: Since this pattern is repeated multiple times, should we add a class attribute called `MODEL_PATCHER_CLASS`, and have one method implemented in the base class making use of it?

optimum/exporters/onnx/model_patcher.py
```diff
@@ -342,3 +344,114 @@ def patched_forward(
...
+        super().__init__(config, model, model_kwargs)
+
+        self.patch = self.real_config.task == "text-generation" and self.real_config.use_past
+        if self.patch:
+            self.orig_prepare_attn_mask = getattr(self._model.model, "_prepare_decoder_attention_mask")
+
+    def __enter__(self):
+        super().__enter__()
+        if self.patch:
+            setattr(self._model.model, "_prepare_decoder_attention_mask", _prepare_decoder_attention_mask)
```
michaelbenayoun, line 389: That's a nit but if you are patching methods, I would bind them as well:
```suggestion
            self._model.model._prepare_decoder_attention_mask =  _prepare_decoder_attention_mask.__get__(self._model.model)
```

optimum/exporters/onnx/model_patcher.py
michaelbenayoun: I agree with you, but it would also differ from the design we've been following. But yes it makes sense.

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -630,14 +634,65 @@ def to(self, device: Union[torch.device, str, int]):
...
+        preprocessors: Optional[List] = None,
+        generation_config: Optional[GenerationConfig] = None,
+        use_cache: Optional[bool] = None,
+        **kwargs,
+    ):
+        if use_io_binding is None:
+            if model.get_providers()[0] in ["CPUExecutionProvider", "CUDAExecutionProvider"]:
+                use_io_binding = True
+            else:
+                use_io_binding = False
```
michaelbenayoun, line 660:
```suggestion
            use_io_binding = model.get_providers()[0] in ["CPUExecutionProvider", "CUDAExecutionProvider"]
```

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -630,14 +634,65 @@ def to(self, device: Union[torch.device, str, int]):
...
+        self.onnx_paths = [self.model_path]
+        self.use_merged = "use_cache_branch" in self.inputs_names
+
+        self.use_fp16 = False
+        for inp in model.get_inputs():
+            if inp.name == "past_key_values" and inp.type == "tensor(float16)":
+                self.use_fp16 = True
+                break
+
+        if use_cache ^ self.use_cache:
```
michaelbenayoun, line 682: I personally find this clearer but both are correct.
```suggestion
        if use_cache != self.use_cache:
```

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -648,39 +703,440 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
...
     def forward(
         self,
-        input_ids: torch.LongTensor = None,
+        input_ids: torch.LongTensor,
         attention_mask: Optional[torch.FloatTensor] = None,
-        past_key_values: Optional[Tuple[Tuple[torch.Tensor]]] = None,
         position_ids: Optional[torch.LongTensor] = None,
+        past_key_values: Optional[Tuple[Tuple[torch.Tensor]]] = None,
         labels: Optional[torch.LongTensor] = None,
+        use_cache_branch: None = None,
```
michaelbenayoun, line 711: Missing type?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -648,39 +703,440 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
...
+            if use_merged is True:
+                raise ValueError(
+                    "The parameters combination use_cache=False, use_merged=True is not supported."
+                    " To use a merged decoder, past key values must be used."
+                )
+            use_merged = False
+
+        # TODO : deprecate
+        decoder_file_name = kwargs.pop("decoder_file_name", None)
+        decoder_with_past_file_name = kwargs.pop("decoder_with_past_file_name", None)
```
michaelbenayoun, line 936: Should we deprecate?

optimum/exporters/onnx/model_configs.py
echarlaix: we want to override `patch_model_for_export` for OnnxConfig of some architectures but not for `M2M100OnnxConfig`  see https://github.com/huggingface/optimum/blob/refactorization-decoder-ort/optimum/exporters/onnx/model_configs.py#L555-L570

optimum/exporters/onnx/config.py
echarlaix: The sequence length of the `position_ids` can vary and is not necessarily 1 anymore (first generation)

optimum/commands/export/onnx.py
echarlaix: Yes good idea, will merge both

optimum/exporters/onnx/model_patcher.py
echarlaix: yes and this means that modifying `BloomModelPatcher` will have an impact on other architectures which is a bit misleading IMO, no strong opinion though so will modify it

optimum/exporters/onnx/model_configs.py
echarlaix: yes let's add this in a following PR

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -630,14 +634,65 @@ def to(self, device: Union[torch.device, str, int]):
 
 
 @add_end_docstrings(ONNX_MODEL_END_DOCSTRING)
-class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
+class ORTModelForCausalLM(ORTModel, GenerationMixin):
```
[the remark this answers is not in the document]
echarlaix: sure will remove it!

optimum/onnxruntime/modeling_decoder.py
echarlaix: added a warning

optimum/exporters/onnx/utils.py
echarlaix: this will impact only decoder models as this modification is done in `_get_submodels_for_export_decoder`, encoder-decoder models will not be impacted (not sure if that answers your question)

optimum/onnxruntime/modeling_decoder.py
echarlaix: When loading a model previously exported, we will only load the decoder_with_past_model if `use_cache=True` and `use_merged=False`. Since v1.7.0 decoder_with_past_model have a fixed sequence length of 1, to keep these models compatible we set this dimension to dynamic, this is needed for the first generation

optimum/commands/export/onnx.py
echarlaix: It actually export 3 ONNX files (without past, with past, merged decoder), I updated the description so that it's clearer

optimum/onnxruntime/modeling_decoder.py
echarlaix: For [theses](https://github.com/huggingface/optimum/blob/refactorization-decoder-ort/optimum/onnxruntime/modeling_decoder.py#L977) architectures, when exporting the model with a sequence length of 1, the attention masks will be generated incorrectly for sequence length greater than 1 https://github.com/huggingface/transformers/blob/0ee45906845c8d58b9bd2df5acd90e09b00047ff/src/transformers/models/bloom/modeling_bloom.py#L654 for this reason we have introduced specific model patcher for these architectures. This results in the previously exported decoder_model_with_past to generate incorrect decoder attention masks and ultimately to generate outputs which can differ from what's expected in this specific case we raise an error asking to re-export the model

optimum/onnxruntime/modeling_decoder.py
echarlaix: Follows https://github.com/huggingface/optimum/blob/8c296d36f1f9c795bf2a330c6e62381751ca0ed1/optimum/onnxruntime/base.py#L329-L330, needed for IO Binding https://github.com/huggingface/optimum/blob/8c296d36f1f9c795bf2a330c6e62381751ca0ed1/optimum/onnxruntime/base.py#L389 as the input names are ordered given the forward signature https://github.com/huggingface/optimum/blob/8c296d36f1f9c795bf2a330c6e62381751ca0ed1/optimum/onnxruntime/utils.py#L300

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -648,39 +703,440 @@ class ORTModelForCausalLM(ORTModelDecoder, GenerationMixin):
...
+
+        save_dir = TemporaryDirectory()
+        save_dir_path = Path(save_dir.name)
+
+        main_export(
+            model_name_or_path=model_id,
+            output=save_dir_path,
+            task=task,
+            do_validation=False,
+            no_post_process=True,
```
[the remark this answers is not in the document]
echarlaix: Good point, will add it

optimum/onnxruntime/optimization.py
```diff
@@ -97,17 +97,11 @@ def from_pretrained(
...
-                if model_or_path.use_merged is True:
-                    raise NotImplementedError(
-                        "ORTOptimizer does not support ORTModelForCausalLM models that use a single ONNX for both the without/with past cases."
-                        " Please pass an ORTModelForCausalLM that uses a separate ONNX for each without/with past cases. This can be done"
-                        " by using `ORTModelForCausalLM.from_pretrained(..., export=True, use_merged=False)`, or by"
-                        " using the option `--no-post-process` in the optimum-cli ONNX export tool."
-                    )
-                onnx_model_path.append(model_or_path.decoder_model_path)
-                if model_or_path.use_cache:
-                    onnx_model_path.append(model_or_path.decoder_with_past_model_path)
```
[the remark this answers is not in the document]
echarlaix: Not sure to understand, we only export the decoder once and also load only one model (even for the previously exported models), what modifications are you thinking about ?

optimum/onnxruntime/quantization.py
```diff
@@ -136,13 +135,6 @@ def from_pretrained(
         path = None
         if isinstance(model_or_path, ORTModelForConditionalGeneration):
             raise NotImplementedError(ort_quantizer_error_message)
-        elif isinstance(model_or_path, ORTModelForCausalLM):
-            if model_or_path.use_cache is False:
-                path = Path(model_or_path.decoder_model_path)
-            elif model_or_path.use_cache is True and model_or_path.use_merged is False:
-                raise NotImplementedError(ort_quantizer_error_message)
-            else:
-                path = Path(model_or_path.decoder_model_path)
```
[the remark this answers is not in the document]
echarlaix: There is only one decoder loaded so an instance of `ORTModelForCausalLM` should be treated as the other instances of `ORTModel`

optimum/utils/modeling_utils.py
```diff
@@ -39,3 +42,77 @@ def recurse_setattr(module, name, value):
     else:
         name, rest = name.split(".", 1)
         recurse_setattr(getattr(module, name), rest, value)
+
+
+# Modified from transformers.models.bloom.modeling_bloom._make_causal_mask
```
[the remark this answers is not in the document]
echarlaix: added an explanation

tests/onnxruntime/test_modeling.py
echarlaix: `use_merged=True` was deprecated for export via `ORTModel`, but still available via the CLI (with `--legacy`), here we just verify that the model was exported as expected and not merged, which is not super useful, might make sense to remove it!

tests/onnxruntime/test_modeling.py
echarlaix: same !

echarlaix: changed the logic in https://github.com/huggingface/optimum/pull/1257/commits/4a32f7a1c8ac7e0f6d7612eee503949277ce2b0f

tests/onnxruntime/test_modeling.py
```diff
@@ -846,12 +846,9 @@ def test_save_load_decoder_model_with_external_data(self, use_cache: bool):
...
             # verify external data is exported
             folder_contents = os.listdir(tmpdirname)
-            self.assertTrue(ONNX_DECODER_NAME in folder_contents)
-            self.assertTrue(ONNX_DECODER_NAME + "_data" in folder_contents)
-
-            if use_cache:
-                self.assertTrue(ONNX_DECODER_WITH_PAST_NAME in folder_contents)
-                self.assertTrue(ONNX_DECODER_WITH_PAST_NAME + "_data" in folder_contents)
+            self.assertTrue(ONNX_WEIGHTS_NAME in folder_contents)
+            # self.assertTrue(ONNX_WEIGHTS_NAME + "_data" in folder_contents)
```
echarlaix, line 850: will remove this test entirely if we keep post processing default for decoder models wtih  [`no_post_process=False`](https://github.com/huggingface/optimum/blob/refactorization-decoder-ort/optimum/onnxruntime/modeling_decoder.py#L593)
https://github.com/huggingface/optimum/blob/c98cb87e3b7151627ddc562341ffa46e39b1def0/optimum/exporters/onnx/base.py#L513

echarlaix: added a slow tets instead in https://github.com/huggingface/optimum/pull/1257/commits/5406f95b52756a831e2ef24d609c9a42628164e8 after dicussion with @fxmarty

optimum/exporters/onnx/config.py
fxmarty: Isn't this "breaking" the legacy export? Breaking in the sense that previously the with past decoder was hard-coding `1` shape and now would not. It is not a very minor issue anyway given that the merged decoder grab the shape from the `decoder_model.onnx`.

optimum/exporters/onnx/model_configs.py
fxmarty: @echarlaix can you send a static link to refer to?

optimum/exporters/onnx/utils.py
fxmarty: Yes makes sense thank you!

optimum/onnxruntime/modeling_decoder.py
fxmarty: > When loading a model previously exported, we will only load the decoder_with_past_model

Is the motivation code simplification? How about merged model? Is the inference dropped as well?

> Since v1.7.0 decoder_with_past_model have a fixed sequence length of 1, to keep these models compatible we set this dimension to dynamic, this is needed for the first generation

I would rather raise an error as unsupported (if we stop supporting) and suggesting to re-export the models.

optimum/onnxruntime/optimization.py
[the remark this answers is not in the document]
fxmarty: We previously did support `decoder_model.onnx` + `decoder_with_past_model.onnx`.

optimum/onnxruntime/quantization.py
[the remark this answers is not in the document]
fxmarty: Okay so it is the same question as above - I'm not sure to understand why we always load only one decoder (namely `decoder_with_past_model.onnx`), even for old models.

As discussed offline, some archs as llama (and bloom?) behave differently for the attention mask generation during prefill & decode, so I am not sure why we are doing this change for legacy models.

IMO we should either error out or keep the old support.

optimum/onnxruntime/modeling_decoder.py
fxmarty: IMO we should raise an error for all archs, or instead keep the legacy support (decoder_model, decoder_with_past_model).

optimum/exporters/onnx/config.py
echarlaix: discussed internally and resolved

optimum/exporters/onnx/model_configs.py
echarlaix: discussed internally and resolved

optimum/onnxruntime/optimization.py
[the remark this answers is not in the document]
echarlaix: discussed internally and resolved

tests/onnxruntime/test_modeling.py
claeyzre: If i may ask this question:
> use_merged=True was deprecated for export via ORTModel, but still available via the CLI (with --legacy)

Does this mean that merging model has been proven to be non effective and not supported in a near future (as the legacy word seems to indicate ?)

Thanks,

echarlaix: for decoder only models, the decoder will now be exported only once by default, previously we had to export the model twice (with and without kv cache) before we could create a "merged" model which was quite memory intensive. do you have a case where you need to use a merged model / cannot use the now default export ?

PR #1269, opened 2023-08-11
Fix build PR doc workflow

fxmarty:
The Furiosa docbuilding requires secrets.

PRs from forks need to be labeled `build-pr-doc` to run the workflow.


fxmarty:
Maybe I need not fork for this

PR #1270, opened 2023-08-11
Fix build PR doc workflow

fxmarty:
The Furiosa docbuilding requires secrets.

PRs from forks need to be labeled `build-pr-doc` to run the workflow.

### DIFF
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -6,6 +6,10 @@ on:
   pull_request_target:
     branches: [ main ]
     types: [ opened, synchronize, reopened, labeled ]
+    paths:
+      - "optimum/**.py"
+      - "docs/**.mdx"
+      - "docs/**.yml"
 
 concurrency:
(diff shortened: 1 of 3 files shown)


fxmarty:
@regisss I don't get why the build PR doc workflow is not picked now. Any idea?

regisss:
> @regisss I don't get why the build PR doc workflow is not picked now on this PR. I don't expect we need to merge first for it to be triggered. Any idea?
> 
> edit: let's just try to merge

I think this is because of
```
 paths:
      - "optimum/**.py"
      - "docs/**.mdx"
      - "docs/**.yml"
```
Since you just modified GH workflows but nothing in `optimum/` or `docs/`, the PR doc build workflow was not triggered.

fxmarty:
I did in https://github.com/huggingface/optimum/pull/1270/commits/7a4a333863d46129d5ea51155a0a63f99cb5c6ba

fxmarty:
Not sure why but looks to work now.

PR #1284, opened 2023-08-14
Fix gptq params

SunMarc:
# What does this PR do 
This PR fixes a few nits on the GPTQ integration such as default args values and the doc.


fxmarty (approved):
LGTM

PR #1292, opened 2023-08-17
Add AutoGPTQ benchmark

fxmarty:
As per title, still need to add results.

### DIFF
--- a/tests/benchmark/README.md
+++ b/tests/benchmark/README.md
@@ -1,4 +1,8 @@
-## GPTQ benchmark
+# BetterTransformer benchmark
+
+Please refer to https://medium.com/pytorch/bettertransformer-out-of-the-box-performance-for-huggingface-transformers-3fbe27d50ab2 & https://pytorch.org/blog/out-of-the-box-acceleration/ for reproduction.
+
+# GPTQ benchmark
 
 Run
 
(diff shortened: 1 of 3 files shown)


tests/benchmark/README.md
```diff
@@ -0,0 +1,20 @@
...
+CUDA_VISIBLE_DEVICES=0 python benchmark_gptq.py --model daryl149/llama-2-13b-chat-hf --sweep --num-batches 4 --task text-generation
+```
+
+and
+
+```
+git clone --branch main https://huggingface.co/TheBloke/Llama-2-13B-chat-GPTQ
+cd Llama-2-13B-chat-GPTQ
+mv gptq_model-4bit-128g.safetensors model.safetensors
+mv quantize_config.json quantization_config.json
```
SunMarc, line 15: This will not work as there is an issue with the metadata with TheBloke GPTQ model. However it is solved with the most recent model that he uploaded as a fix was introduced in AutoGPTQ

tests/benchmark/benchmark_gptq.py
```diff
@@ -0,0 +1,232 @@
...
+    is_decoder = False
+
+if args.gptq:
+    if not args.gptq_model:
+        raise ValueError("The argument --gptq-model needs to be provided when benchmarking GPTQ.")
+    
+    with init_empty_weights():
+        empty_model = AutoModelForCausalLM.from_pretrained(args.model, torch_dtype=torch.float16)
+    empty_model.tie_weights()
+    model = load_quantized_model(empty_model, save_folder=args.gptq_model, state_dict_name="model.safetensors", device_map="auto")
```
SunMarc, line 161: In this function `disable_exllama` is set to False by default, maybe good to have the choice to disable it ?

SunMarc (approved):
LGTM ! Let's get some results before merging.

tests/benchmark/benchmark_gptq.py
fxmarty: Yes I'll add both! Also no act-order / act-order.

Issue #1288, opened 2023-08-16
onnxruntime support for pix2struct using ORT class


krathul:
### Feature request

It will be great to have an ORT class for pix2struct. The export of the model using optimum was supported in #937 but no onnxruntime support have been implemented.

### Motivation

Currently there is no onnxruntime support for pix2struct in optimum. An ORTModel class will help in applying optimizations and exporting.

### Your contribution

I can help with the ORT class, I currently have a working implementation of an ORT class for supporting pix2struct. I used the same decoder used for all classes in [modeling_seq2seq.py](https://github.com/huggingface/optimum/blob/main/optimum/onnxruntime/modeling_seq2seq.py) with an additional decoder_attention_mask argument and an encoder and a main class written by referencing the `ORTModelForSpeechSeq2Seq` and `transformers.Pix2StructForConditionalGeneration`.

fxmarty:
Will do this week

fxmarty:
Fixed by https://github.com/huggingface/optimum/pull/1296

PR #1296, opened 2023-08-18
Pix2Struct onnxruntime support

krathul:
This PR implements onnxruntime support for Pix2struct using ORT class (#1288)


optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -1362,3 +1455,217 @@ def _reorder_cache(past, beam_idx) -> Tuple[Tuple[torch.FloatTensor]]:
...
+        # Encode if needed : first prediction pass
+        # Encode if needed (training, first prediction pass)
+        if encoder_outputs is None:
+            encoder_outputs = self.encoder(
+                flattened_patches=flattened_patches,
+                attention_mask=attention_mask,
+            )
+        # work around for Unexpected input data type (tensor(float)) for attention_mask
+        # Need to be looked into
+        attention_mask = attention_mask.to(torch.int64)
```
fxmarty, line 1520: Did you find the issue?

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -1362,3 +1455,217 @@ def _reorder_cache(past, beam_idx) -> Tuple[Tuple[torch.FloatTensor]]:
...
+    """
+    Pix2struct model with a language modeling head for ONNX Runtime inference.
+    """
+
+    # pix2struct cannot be loaded using AutoModel
+    auto_model_class = Pix2StructForConditionalGeneration
+    main_input_name = "flattened_patches"
+
+    # Copied from transformers.models.t5.modeling_t5.T5PreTrainedModel._shift_right with T5->Pix2Struct
+    def _shift_right(self, input_ids):
```
fxmarty, line 1470: @JingyaHuang In transformers, this seem to be used for training. I never really looked at the ONNX config with loss, but does this include a `label` input? i.e., in ORTTrainer, is the export going through e.g. https://github.com/huggingface/transformers/blob/e42587f596181396e1c4b63660abf0c736b10dae/src/transformers/models/t5/modeling_t5.py#L1701-L1703 ?

@krathul I think for now it is reasonable to remove this method. We can have a look at training later.

fxmarty (commented):
Thank you, it looks in good shape! I'll add tests in this PR if that is fine to you.

optimum/onnxruntime/modeling_seq2seq.py
krathul: Ok

krathul:
@fxmarty you can add tests.

optimum/onnxruntime/modeling_seq2seq.py
krathul: Yes, the attention_mask returned by pix2structProcessor is of type torch.float32, it is converted for use in `ORTEncoderForPix2Struct` but is directly used in `ORTDecoderForSeq2Seq` for the decoder, which requires a long tensor. As a solution, I guess the conversion can be done in forward method since the decoder method is generic.

fxmarty: Yes, we should fix the dtype of the attention mask in the ONNX export. As a dirty workaround I kept yours, but it is not ideal.

optimum/exporters/onnx/model_configs.py
```diff
@@ -1275,12 +1284,46 @@ def inputs(self):
...
         return common_inputs
 
+    @property
+    def outputs(self) -> Dict[str, Dict[int, str]]:
+        if self._behavior is ConfigBehavior.ENCODER:
+            common_outputs = {
+                "last_hidden_state": {0: "batch_size"}
+            }  # The last hidden state dim=1 is constant, no need for it to be dynamic.
+        else:
+            common_outputs = super(OnnxConfigWithPast, self).outputs
```
fxmarty, line 1301: We redefine because of this comment.

fxmarty:
The PR is in good shape. On my CPU I am getting a 17% speedup over pytorch.

fxmarty (approved):
@krathul Thanks you for the clean implementation!

krathul:
> @krathul Thank you for the clean implementation!

@fxmarty Thanks for helping with the merge

PR #1326, opened 2023-08-31
Fix PyTorch tied weights being duplicated in the exported ONNX models

fxmarty:
This is the case for example when the embedding weight and language modeling head share the same weight.

Although PyTorch has a pass to deduplicate identical initializers (https://github.com/pytorch/pytorch/blob/main/torch/csrc/jit/passes/onnx/deduplicate_initializers.cpp), one issue arises when torch.onnx.export does constant folding, that may e.g. transpose a weight (in the nn.Linear case), that later messes up with the initializer deduplication. See details: https://github.com/pytorch/pytorch/issues/108342

For small models, the duplication can result in ONNX weights >30% larger vs pytorch weight. The issue is less severe for large models. Given that doing constant folding is generally good, just passing `do_constant_folding=False` is not really a satisfying solution hence this PR.

e.g. bloom-560m goes from 3.0 GiB to 2.1 GiB.

Fixes https://github.com/pytorch/pytorch/issues/108342 as a post-processing step for nn.Embedding & nn.Linear.

### DIFF
--- a/optimum/exporters/onnx/base.py
+++ b/optimum/exporters/onnx/base.py
@@ -19,6 +19,7 @@ import enum
 import gc
 import inspect
 import itertools
+import os
 import re
 from abc import ABC, abstractmethod
 from collections import OrderedDict
@@ -26,7 +27,14 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Tuple, Union
 
 import numpy as np
-from transformers.utils import is_torch_available
+import onnx
+from transformers.utils import is_accelerate_available, is_torch_available
+
+from ...onnx import remove_duplicate_weights_from_tied_info
+
+
+if is_torch_available():
+    import torch.nn as nn
 
 from ...onnx import merge_decoders
 from ...utils import (
@@ -42,9 +50,12 @@ from ...utils import TRANSFORMERS_MINIMUM_VERSION as GLOBAL_MIN_TRANSFORMERS_VER
 from ...utils.doc import add_dynamic_docstring
 from ...utils.import_utils import check_if_transformers_greater, is_onnx_available, is_onnxruntime_available
 from ..base import ExportConfig
-from .constants import ONNX_DECODER_MERGED_NAME, ONNX_DECODER_NAME, ONNX_DECODER_WITH_PAST_NAME, ONNX_ENCODER_NAME
+from .constants import ONNX_DECODER_MERGED_NAME, ONNX_DECODER_NAME, ONNX_DECODER_WITH_PAST_NAME
 from .model_patcher import ModelPatcher, Seq2SeqModelPatcher
 
+
+if is_accelerate_available():
+    from accelerate.utils import find_tied_parameters
 
 if TYPE_CHECKING:
     from transformers import PretrainedConfig, PreTrainedModel, TFPreTrainedModel
@@ -505,9 +516,27 @@ class OnnxConfig(ExportConfig, ABC):
             models_and_onnx_configs (`Dict[str, Tuple[Union["PreTrainedModel", "TFPreTrainedModel", "ModelMixin"], "OnnxConfig"]]`):
                 A dictionnary containing the models t apply post-processing on, and their corresponding ONNX configuration.
             onnx_files_subpaths (`List[str]`):
-            The relative paths from the export directory to the ONNX files to do post-processing on. The order must be the same as*
-            the order of submodels in the ordered dict `models_and_onnx_configs`.
+                The relative paths from the export directory to the ONNX files to do post-processing on. The order must be the same as
+                the order of submodels in the ordered dict `models_and_onnx_configs`.
         """
+        first_key = next(iter(models_and_onnx_configs))
+        if is_torch_available() and isinstance(models_and_onnx_configs[first_key][0], nn.Module):
+            if is_accelerate_available():
+                print("Deduplicating shared (tied) weights...")
+                keys = list(models_and_onnx_configs.keys())
+                for i, subpath in enumerate(onnx_files_subpaths):
+                    onnx_model = onnx.load(os.path.join(path, subpath))
+
+                    torch_model = models_and_onnx_configs[keys[i]][0]
+                    tied_params = find_tied_parameters(torch_model)
+                    remove_duplicate_weights_from_tied_info(
+                        onnx_model, torch_model, tied_params, save_path=os.path.join(path, subpath)
+                    )
+            else:
+                logger.warning(
+                    "Weight deduplication check in the ONNX export requires accelerate. Please install accelerate to run it."
+                )
+
         return models_and_onnx_configs, onnx_files_subpaths
 
 
@@ -918,14 +947,14 @@ class OnnxSeq2SeqConfigWithPast(OnnxConfigWithPast):
         ],
         onnx_files_subpaths: List[str],
     ):
+        models_and_onnx_configs, onnx_files_subpaths = super().post_process_exported_models(
+            path, models_and_onnx_configs, onnx_files_subpaths
+        )
+
         # Attempt to merge only if the decoder was exported without/with past
         if self.use_past is True and len(models_and_onnx_configs) == 3:
-            if onnx_files_subpaths is not None:
-                decoder_path = Path(path, onnx_files_subpaths[1])
-                decoder_with_past_path = Path(path, onnx_files_subpaths[2])
-            else:
-                decoder_path = Path(path, ONNX_DECODER_NAME + ".onnx")
-                decoder_with_past_path = Path(path, ONNX_DECODER_WITH_PAST_NAME + ".onnx")
+            decoder_path = Path(path, onnx_files_subpaths[1])
+            decoder_with_past_path = Path(path, onnx_files_subpaths[2])
             decoder_merged_path = Path(path, ONNX_DECODER_MERGED_NAME + ".onnx")
             try:
                 # The decoder with past does not output the cross attention past key values as they are constant,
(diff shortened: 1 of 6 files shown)


xenova:
Awesome! 🚀 
I've tested and bloom-560m does go down to 2.1 GiB! However, after quantizing (with [this script](https://github.com/xenova/transformers.js/blob/main/scripts/convert.py#L139-L197)), the quantized model is around 1.5GiB 😅

xenova:
```sh
optimum-cli onnxruntime quantize --avx512 --onnx_model bloom_onnx --output ./quantized_bloom_onnx
```

I also tried using `optimum-cli` directly, but same results (1.48GiB in size)

fxmarty:
@xenova This should be fixed. Could you give a second try? I'm now getting a quantized bloom-560m that is 536 MiB.

optimum/exporters/onnx/base.py
```diff
@@ -505,9 +516,27 @@ def post_process_exported_models(
...
             onnx_files_subpaths (`List[str]`):
-            The relative paths from the export directory to the ONNX files to do post-processing on. The order must be the same as*
-            the order of submodels in the ordered dict `models_and_onnx_configs`.
+                The relative paths from the export directory to the ONNX files to do post-processing on. The order must be the same as
+                the order of submodels in the ordered dict `models_and_onnx_configs`.
         """
+        first_key = next(iter(models_and_onnx_configs))
+        if is_torch_available() and isinstance(models_and_onnx_configs[first_key][0], nn.Module):
+            if is_accelerate_available():
+                print("Deduplicating shared (tied) weights...")
```
michaelbenayoun, line 525: nit: use a logger instead of printing.

optimum/exporters/onnx/base.py
```diff
@@ -505,9 +516,27 @@ def post_process_exported_models(
...
                 A dictionnary containing the models t apply post-processing on, and their corresponding ONNX configuration.
             onnx_files_subpaths (`List[str]`):
-            The relative paths from the export directory to the ONNX files to do post-processing on. The order must be the same as*
-            the order of submodels in the ordered dict `models_and_onnx_configs`.
+                The relative paths from the export directory to the ONNX files to do post-processing on. The order must be the same as
+                the order of submodels in the ordered dict `models_and_onnx_configs`.
         """
+        first_key = next(iter(models_and_onnx_configs))
+        if is_torch_available() and isinstance(models_and_onnx_configs[first_key][0], nn.Module):
+            if is_accelerate_available():
```
michaelbenayoun, line 524: nit:import torch.nn here?

optimum/onnx/graph_transformations.py
```diff
@@ -59,6 +68,33 @@ def remove_duplicate_weights(model: ModelProto, inplace: bool = False) -> ModelP
...
 
+def remove_duplicate_weights_from_tied_info(
+    onnx_model: ModelProto, torch_model: "nn.Module", tied_params: List[List[str]], save_path: str
+):
+    """
+    Tries to remove potential duplicate ONNX initializers from the tied information in tied_params.
+
+    Args:
+        onnx_model (`onnx.ModelProto`):
+            The ONNX model for which to tie potentially duplicate initializers.
```
michaelbenayoun, line 79: You forgot `torch_model` docstring.

optimum/onnx/graph_transformations.py
```diff
@@ -89,6 +125,54 @@ def replace_atenops_to_gather(model: ModelProto) -> ModelProto:
...
+                pass
+            else:
+                raise e
+        if save_path:
+            # Overwrite.
+            save_path = str(save_path)
+            if save_path.endswith(".onnx") and os.path.isfile(save_path):
+                os.remove(save_path)
+            save_path = Path(save_path).as_posix()
+            onnx.save(model, save_path)
```
michaelbenayoun, line 146: So no saving is done if the model is small enough and we dont provide a `save_path`, only the check?

optimum/onnx/graph_transformations.py
```diff
@@ -89,6 +125,54 @@ def replace_atenops_to_gather(model: ModelProto) -> ModelProto:
...
+            all_tensors_to_one_file=True,
+            location=external_file_name,
+        )
+        try:
+            onnx.checker.check_model(save_path)
+        except Exception as e:
+            if "No Op registered for" in str(e):
+                pass
+            else:
+                raise e
```
michaelbenayoun, line 171: In this case we save the model first, meaning that it is possible that an ONNX file is saved with the check failing afterwards? 
It's the opposite of the first case so just checking.

optimum/onnx/transformations_utils.py
```diff
@@ -312,3 +316,223 @@ def cast_int64_tensorproto_to_int32(initializer: onnx.TensorProto, cast: bool =
 
     initializer.CopyFrom(tensor)
     initializer.name = original_name
+
+
+def _get_weights_to_tie(tied_params, torch_model: "nn.Module") -> Tuple[List[List[str]]]:
+    """
+    Find tied weights in the PyTorch model `model`, and separate them in tied weights
+    for which an untying strategy exists and do not exist.
```
michaelbenayoun, line 324: nit:
```suggestion
    for which an untying strategy exists and does not exist.
```

michaelbenayoun: Also I did not really understand the docstring.

optimum/onnx/transformations_utils.py
```diff
@@ -312,3 +316,223 @@ def cast_int64_tensorproto_to_int32(initializer: onnx.TensorProto, cast: bool =
...
+    for params in tied_params:
+        skip_group = False
+        for param_name in params:
+            module_name = ".".join(param_name.split(".")[:-1])
+
+            module = recurse_getattr(torch_model, module_name)
+            if module.__class__.__name__ not in SUPPORTED_DEDUPLICATION_OPS:
+                skip_group = True
+            if len(params) != 2:
+                skip_group = 2
```
michaelbenayoun, line 338: `skip_group` is a `bool` or an `int`?

optimum/onnx/transformations_utils.py
```diff
@@ -312,3 +316,223 @@ def cast_int64_tensorproto_to_int32(initializer: onnx.TensorProto, cast: bool =
...
+    Find tied weights in the PyTorch model `model`, and separate them in tied weights
+    for which an untying strategy exists and do not exist.
+    """
+    SUPPORTED_DEDUPLICATION_OPS = ("Embedding", "Linear")
+    tied_groups_to_tie = []
+    tied_groups_ignored = []
+    for params in tied_params:
+        skip_group = False
+        for param_name in params:
+            module_name = ".".join(param_name.split(".")[:-1])
```
michaelbenayoun, line 332:
```suggestion
            module_name = param_name.rsplit(".", maxsplit=1)[0]
```
Not sure it is more clear though

michaelbenayoun (approved):
LGTM, thanks a lot @fxmarty !

optimum/exporters/onnx/base.py
fxmarty: I put it rather as the top of the file under `if is_torch_available()`.

optimum/onnx/graph_transformations.py
fxmarty: Yes

optimum/onnx/graph_transformations.py
fxmarty: Yes but that is intended. onnx checker can either take a ModelProto or a path, and in case of a model larger than 2 GB, a path needs to be used.

optimum/onnx/transformations_utils.py
fxmarty: The implementation can only re-tie Gather and MatMul weights on the ONNX side, while there may be other tied weights on the torch side. We use the pytorch model to split the tied weights groups in one for which we have an available implementation to tie on the ONNX side, and an other for which we don't have (we will detect whether weights are duplicated, but won't fix).

optimum/exporters/onnx/base.py
```diff
@@ -505,9 +516,27 @@ def post_process_exported_models(
...
+                logger.info("Deduplicating shared (tied) weights...")
+                keys = list(models_and_onnx_configs.keys())
+                for i, subpath in enumerate(onnx_files_subpaths):
+                    onnx_model = onnx.load(os.path.join(path, subpath))
+
+                    torch_model = models_and_onnx_configs[keys[i]][0]
+                    tied_params = find_tied_parameters(torch_model)
+                    remove_duplicate_weights_from_tied_info(
+                        onnx_model, torch_model, tied_params, save_path=os.path.join(path, subpath)
+                    )
```
xenova, line 534:
```suggestion
                for subpath, key in zip(onnx_files_subpaths, models_and_onnx_configs):
                    onnx_model = onnx.load(os.path.join(path, subpath))

                    torch_model = models_and_onnx_configs[key][0]
                    tied_params = find_tied_parameters(torch_model)
                    remove_duplicate_weights_from_tied_info(
                        onnx_model, torch_model, tied_params, save_path=os.path.join(path, subpath)
                    )
```

optimum/onnx/graph_transformations.py
```diff
@@ -14,25 +14,32 @@
...
+    _get_weights_to_tie,
     _remove_redundant_initializers,
     _replace_input_names,
     _unify_onnx_outputs,
     cast_int64_tensorproto_to_int32,
 )
 
 
+if TYPE_CHECKING:
+    import torch.nn as nn
```
xenova, line 40: Should this also be under a `is_torch_available()`, or can we assume it's available here?

optimum/onnx/transformations_utils.py
```diff
@@ -312,3 +316,224 @@ def cast_int64_tensorproto_to_int32(initializer: onnx.TensorProto, cast: bool =
...
+                        "initializer_name": {param_name},
+                        "nodes_containing_initializer": nodes_containing_initializer,
+                    }
+                )
+                identical_initializer = True
+
+            # If not found (e.g. "lm_head.weight"), we greedily search for all initializers from potentially matching node names (e.g. "lm_head").
+            # This greedy approach may found more initializers than wanted.
+            if not identical_initializer:
+                module_name = ".".join(param_name.split(".")[:-1])
```
xenova, line 394: see https://github.com/huggingface/optimum/pull/1326/files#r1312812437

xenova (approved):
Can confirm that https://github.com/huggingface/optimum/pull/1326/commits/1a5dbfc907e1c06268b533a440e7ae6c052d5bb7 fixed it and the quantized model is now of the correct size! 🥳 

Just some minor suggestions in the review:

optimum/onnx/graph_transformations.py
fxmarty: We can probably assume that a dev has all the dependencies installed so probably not needed.

PR #1329, opened 2023-08-31
Version bump + add max_input_length to gptq 

SunMarc:
# What does this PR do ? 
This PR bumps the minimum required version of `auto_gptq` and adds the possibility to change the max input length when using exllama + act_order with gptq. 
We also do a couple of fixes: 
- disable exllama kernel when doing quantization with act order because the user may save the model with the weights in the wrong order. exllama + act_order should only be used for inference and the user should not save the model. This is a temporary measure until we are able to reorder the weights.
- `self.quant_method = QuantizationMethod.GPTQ `for people who wants to quantize with optimum api and load it with transformers like this [tutorial](https://www.philschmid.de/gptq-llama) .

### DIFF
--- a/tests/gptq/test_quantization.py
+++ b/tests/gptq/test_quantization.py
@@ -18,7 +18,7 @@ import unittest
 
 import torch
 from parameterized import parameterized
-from transformers import AutoModelForCausalLM, AutoTokenizer
+from transformers import AutoModelForCausalLM, AutoTokenizer, AutoConfig
 from transformers.testing_utils import slow
 
 from optimum.gptq import GPTQQuantizer, load_quantized_model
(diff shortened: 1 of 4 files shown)


tests/gptq/test_quantization.py
```diff
@@ -126,14 +128,89 @@ def test_serialization(self):
...
+            quantized_model_from_saved = load_quantized_model(
+                empty_model, save_folder=tmpdirname, device_map={"": 0}, disable_exllama=False, max_input_length=4028
+            )
+
+            prompt = "I am in Paris and" * 1000
+            inp = self.tokenizer(prompt, return_tensors="pt").to(0)
+            self.assertTrue(inp["input_ids"].shape[1] > 4028)
+            with self.assertRaises(RuntimeError) as cm:
+                quantized_model_from_saved.generate(**inp, num_beams=1, min_new_tokens=3, max_new_tokens=3)
+                self.assertTrue("temp_state buffer is too small" in str(cm.exception))
```
fxmarty, line 208: The error shown from AutoGPTQ is irrelevant if the model is loaded through transformers right? Should I do a fix on AutoGPTQ side to have the exllama_set_max_input_length method behave well with transformers?

fxmarty (approved):
LGTM thank you!

younesbelkada (approved):
Thanks a lot @SunMarc !

tests/gptq/test_quantization.py
fxmarty: Calling `exllama_set_max_input_length` directly on a transformers model will fail, see: https://github.com/PanQiWei/AutoGPTQ/issues/315

fxmarty: Why is this fine? The error shown by AutoGPTQ suggests to use `exllama_set_max_input_length` which does not work with transformers.

SunMarc: What is the fix that you want to do ? I think i'm missing something 😅

SunMarc: That would be great indeed ! The workaround that is used for now in optimum/transformers is: 
```py 
class StoreAttr(object):
    pass

model.quantize_config = StoreAttr()
model.quantize_config.desc_act = self.desc_act
```

PR #1331, opened 2023-09-01
Fix opset custom onnx export

mht-sharma:
# What does this PR do?

Fix opset for onnx export

Current tests fail with this error 

```bash
torch.onnx.errors.UnsupportedOperatorError: Exporting the operator 'aten::tril' to ONNX opset version 13 is not supported. Support for this operator was added in version 14, try exporting with this version.
```


fxmarty (approved):
Thank you!

PR #1332, opened 2023-09-01
Replace mpt to ernie custom export

mht-sharma:
The MPT model is now supported for export, so `test_custom_export_trust_remote_error` is failing. This PR proposes to replace the MPT model with the Ernie model.


fxmarty (approved):
LGTM

PR #851, opened 2023-03-03
Add encoder decoder model

mht-sharma:
# What does this PR do?

Support encoder-decoder export and inference in ORT

Fixes #367 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

PR #1338, opened 2023-09-03
Add text2text-generation-with-past test for encoder-decoder model

mht-sharma:
# What does this PR do?

Creating a test for an `text2text-generation-with-past` tasks for  an encoder-decoder model  [https://github.com/huggingface/optimum/pull/851](https://github.com/huggingface/optimum/pull/851)

- Fixes a bug to use `--with-past` with the model

## Before submitting


tests/onnxruntime/test_modeling.py
```diff
@@ -3099,84 +3114,100 @@ def test_load_vanilla_transformers_which_is_not_supported(self):
...
-        model_id = MODEL_NAMES[model_arch]
-        model = ORTModelForSeq2SeqLM.from_pretrained(self.onnx_model_dirs[test_name], use_cache=use_cache)
-        tokenizer = get_preprocessor(model_id)
-        text = "This is a sample output"
-        tokens = tokenizer(text, return_tensors="pt")
+        model_ids = self._get_model_ids(model_arch)
+        for model_id in model_ids:
+            if (
+                model_arch == "encoder-decoder"
+                and use_cache is True
```
michaelbenayoun, line 3124:
```suggestion
                and use_cache
```

tests/onnxruntime/test_modeling.py
```diff
@@ -3082,6 +3082,21 @@ class ORTModelForSeq2SeqLMIntegrationTest(ORTModelTestMixin):
     GENERATION_LENGTH = 100
     SPEEDUP_CACHE = 1.1
 
+    def _get_model_ids(self, model_arch):
+        model_ids = MODEL_NAMES[model_arch]
+        if isinstance(model_ids, dict):
+            model_ids = list(model_ids.keys())
+        else:
+            model_ids = [model_ids]
+        return model_ids
```
michaelbenayoun, line 3091: So what you call `model_arch` is a "model type" like `bert`, `t5` and so on?
This function returns a list of model ids from the Hub for a given architecture?

tests/onnxruntime/test_modeling.py
```diff
@@ -3189,61 +3220,70 @@ def test_compare_to_transformers(self, test_name: str, model_arch: str, use_cach
...
-        tokenizer = get_preprocessor(model_id)
-        tokens = tokenizer("This is a sample output", return_tensors="pt")
-        decoder_start_token_id = transformers_model.config.decoder_start_token_id if model_arch != "mbart" else 2
-        if model_arch == "encoder-decoder":
-            decoder_start_token_id = tokenizer.cls_token_id
+        model_ids = self._get_model_ids(model_arch)
+        for model_id in model_ids:
+            if (
+                model_arch == "encoder-decoder"
+                and use_cache is True
```
michaelbenayoun, line 3227: nit
```suggestion
                and use_cache
```

tests/onnxruntime/test_modeling.py
```diff
@@ -3256,34 +3296,45 @@ def test_pipeline_text_generation(self, test_name: str, model_arch: str, use_cac
...
         self._setup(model_args)
 
-        model_id = MODEL_NAMES[model_arch]
-        onnx_model = ORTModelForSeq2SeqLM.from_pretrained(self.onnx_model_dirs[test_name], use_cache=use_cache)
-        tokenizer = get_preprocessor(model_id)
+        model_ids = self._get_model_ids(model_arch)
+        for model_id in model_ids:
+            if (
+                model_arch == "encoder-decoder"
+                and use_cache is True
```
michaelbenayoun, line 3303: Same

michaelbenayoun (approved):
Left a few nits.
I do not understand everyting, but looks good to me from what I could read.
Thanks @mht-sharma !

tests/onnxruntime/test_modeling.py
mht-sharma: I guess this is the common terminology used in the repo, to explicitly check the value?

tests/onnxruntime/test_modeling.py
mht-sharma: It's a helper function for the test to get the list of `model_ids` for a given `model_arch(model_type)` from `MODEL_NAMES`. Earlier there was one `model_id` for a `model_arch`. But with this PR there can be multiple (Format is same as used in onnx export).

tests/onnxruntime/test_modeling.py
fxmarty: It's me who advocated to use that :D But more and more I agree with Michael, we should not explicitly check `is True` when the type is `bool`, we should only for `Optional[bool]`.

fxmarty (approved):
LGTM thanks for adding the tests! It is fine to me as long as they pass.

tests/onnxruntime/test_modeling.py
mht-sharma: I guess this explicit check is something which is also followed in the transformers, I remember Lewis mentioning this, but yeah I agree

PR #1343, opened 2023-09-04
[`BetterTransformer`] Add falcon to `BetterTransformer`

younesbelkada:
# What does this PR do?

As per title, adds Falcon to BT ! 
We probably need to drop the call to torch.scaled_dot_product_attention in the modeling code of falcon to avoid code de-duplication

https://github.com/huggingface/transformers/blob/main/src/transformers/models/falcon/modeling_falcon.py#L434

cc @fxmarty

### DIFF
--- a/optimum/bettertransformer/models/attention.py
+++ b/optimum/bettertransformer/models/attention.py
@@ -916,6 +916,9 @@ def falcon_forward(
     # 3 x [batch_size, seq_length, num_heads, head_dim]
     (query_layer, key_layer, value_layer) = self._split_heads(fused_qkv)
 
+    if output_attentions is True:
+        raise ValueError("output_attentions=True can not be supported with BetterTransformer.")
+
     batch_size, query_length, _, _ = query_layer.shape
 
     query_layer = query_layer.transpose(1, 2).reshape(batch_size * self.num_heads, query_length, self.head_dim)
@@ -949,9 +952,19 @@ def falcon_forward(
     value_layer_ = value_layer.reshape(batch_size, num_kv_heads, -1, self.head_dim)
 
     if alibi is None:
-        attn_output = torch.nn.functional.scaled_dot_product_attention(
-            query_layer_, key_layer_, value_layer_, attention_mask_float, 0.0, is_causal=False
-        )
+        if batch_size == 1 or self.training:
+            if query_length > 1:
+                attn_output = torch.nn.functional.scaled_dot_product_attention(
+                    query_layer_, key_layer_, value_layer_, attn_mask=None, dropout_p=0.0, is_causal=True
+                )
+            else:
+                attn_output = torch.nn.functional.scaled_dot_product_attention(
+                    query_layer_, key_layer_, value_layer_, attn_mask=None, dropout_p=0.0, is_causal=False
+                )
+        else:
+            attn_output = torch.nn.functional.scaled_dot_product_attention(
+                query_layer_, key_layer_, value_layer_, attention_mask_float, 0.0, is_causal=False
+            )
 
         attn_output = attn_output.view(batch_size, self.num_heads, query_length, self.head_dim)
         attn_output = attn_output.permute(0, 2, 1, 3)
--- a/optimum/bettertransformer/models/attention.py
+++ b/optimum/bettertransformer/models/attention.py
@@ -899,3 +899,85 @@ def bloom_forward(
         present = None
 
     return (output_tensor, present)
+
+
+def falcon_forward(
+    self,
+    hidden_states: torch.Tensor,
+    alibi: Optional[torch.Tensor],
+    attention_mask: torch.Tensor,
+    layer_past: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
+    head_mask: Optional[torch.Tensor] = None,
+    use_cache: bool = False,
+    output_attentions: bool = False,
+):
+    fused_qkv = self.query_key_value(hidden_states)  # [batch_size, seq_length, 3 x hidden_size]
+    num_kv_heads = self.num_heads if self.new_decoder_architecture else self.num_kv_heads
+    # 3 x [batch_size, seq_length, num_heads, head_dim]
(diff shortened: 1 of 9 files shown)


tests/bettertransformer/test_decoder.py
```diff
@@ -25,14 +25,26 @@
...
+        "bloom",
+        "codegen",
+        "gpt2",
+        "gpt_bigcode",
+        "gptj",
+        "gpt_neo",
+        "gpt_neox",
+        "llama",
+        "opt",
+        "falcon",
```
fxmarty, line 46: Can you keep it ordered?

optimum/bettertransformer/models/decoder_models.py
```diff
@@ -19,6 +19,7 @@
 from transformers.models.blenderbot.modeling_blenderbot import BlenderbotAttention
 from transformers.models.bloom.modeling_bloom import BloomAttention
 from transformers.models.codegen.modeling_codegen import CodeGenAttention
+from transformers.models.falcon.modeling_falcon import FalconAttention
```
fxmarty, line 22: Can you add it under a guard with the relevant transformers version? So as to support older transformers versions.

See below:
```python
# TODO: remove once we are much higher than 4.31
if check_if_transformers_greater("4.31"):
    from transformers.models.llama.modeling_llama import LlamaAttention
else:
    from ...utils.dummy_bettertransformer_objects import LlamaAttention
```

optimum/utils/input_generators.py
```diff
@@ -858,3 +858,29 @@ def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int
             )
             for _ in range(self.num_layers)
         ]
+
+
+class FalconDummyPastKeyValuesGenerator(DummyPastKeyValuesGenerator):
+    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+        self.num_kv_heads = 1
```
younesbelkada, line 865: Tiny falcon uses `self.new_decoder_architecture=True` and `self.multi_query=True` and in that case the number of kv heads is set to 1

```python
self.num_kv_heads = config.num_kv_heads if (self.new_decoder_architecture or not self.multi_query) else 1
```

https://github.com/huggingface/transformers/blob/main/src/transformers/models/falcon/modeling_falcon.py#L287C9-L287C114

optimum/bettertransformer/models/attention.py
```diff
@@ -899,3 +899,84 @@ def bloom_forward(
...
+    attention_mask_float = (attention_mask * 1.0).masked_fill(attention_mask, float("-1e9")).to(query_layer.dtype)
+
+    query_layer_ = query_layer.reshape(batch_size, self.num_heads, -1, self.head_dim)
+    key_layer_ = key_layer.reshape(batch_size, num_kv_heads, -1, self.head_dim)
+    value_layer_ = value_layer.reshape(batch_size, num_kv_heads, -1, self.head_dim)
+
+    if alibi is None:
+        attn_output = torch.nn.functional.scaled_dot_product_attention(
+            query_layer_, key_layer_, value_layer_, attention_mask_float, 0.0, is_causal=False
+        )
```
fxmarty, line 954: It would probably be best to drop the attention mask altogether for batch size = 1 and training, as done for the other architectures. Otherwise you are never dispatching to flash/mem efficient.

optimum/bettertransformer/models/attention.py
```diff
@@ -899,3 +899,84 @@ def bloom_forward(
...
+        alibi = alibi.reshape(batch_size, -1, *alibi.shape[1:])
+        alibi = torch.masked_fill(alibi, attention_mask, torch.finfo(alibi.dtype).min)
+
+        context_layer = torch.nn.functional.scaled_dot_product_attention(
+            query_layer,
+            key_layer,
+            value_layer,
+            attn_mask=alibi,
+            dropout_p=self.attention_dropout if self.training else 0.0,
+        )
```
fxmarty, line 974: Here we have no choice but to keep the mask. In torch 2.1 it will dispatch to mem-efficient attention.

optimum/bettertransformer/models/attention.py
```diff
@@ -899,3 +899,84 @@ def bloom_forward(
...
+
+def falcon_forward(
+    self,
+    hidden_states: torch.Tensor,
+    alibi: Optional[torch.Tensor],
+    attention_mask: torch.Tensor,
+    layer_past: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
+    head_mask: Optional[torch.Tensor] = None,
+    use_cache: bool = False,
+    output_attentions: bool = False,
```
fxmarty, line 912: Can you raise an error if this is True? Like in the other archs

optimum/bettertransformer/models/decoder_models.py
fxmarty: You will also need to modify the `BACKENDS_MAPPING` in utils/import_utils.py, and add the relevant dummy objects

tests/bettertransformer/testing_utils.py
```diff
@@ -34,6 +34,7 @@
     "blenderbot": "hf-internal-testing/tiny-random-BlenderbotModel",
     "blip-2": "hf-internal-testing/tiny-random-Blip2Model",
     "bloom": "hf-internal-testing/tiny-random-BloomModel",
+    "falcon": "Rocketknight1/tiny-random-falcon-7b",
```
fxmarty, line 37: Can you keep it ordered?

younesbelkada:
Thanks @fxmarty for the review, I have updated the PR according to your review, my only question is about https://github.com/huggingface/optimum/pull/1343#discussion_r1314723241
As discussed offline, I am unsure I did it the correct way, also I am unsure about from which transformers version Falcon support has been added, as per my understanding that arch is only available on main currently

fxmarty (approved):
LGTM

fxmarty:
@younesbelkada It appears `from transformers.models.falcon.modeling_falcon import FalconAttention` goes fine with 4.32 so we're good!

Issue #1337, opened 2023-09-03
ORTForModelVision2Seq is running very slow on onnxruntime than the pytorch version


aftnix:
### System Info

```shell
- `optimum` version: 1.12.0
- `transformers` version: 4.32.1
- Platform: Linux-4.14.318-167.530.amzn1.x86_64-x86_64-with-glibc2.10
- Python version: 3.8.13
- Huggingface_hub version: 0.16.4
- PyTorch version (GPU?): 1.13.0a0+d321be6 (cuda available: True)
```

### Who can help?

@JingyaHuang @echarlaix 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction (minimal, reproducible, runnable)

```
from transformers import AutoTokenizer
import torch
import re
from transformers import DonutProcessor, VisionEncoderDecoderModel
from datasets import load_dataset
import huggingface_hub
import time
from optimum.onnxruntime import ORTModelForVision2Seq
from PIL import Image
import onnxruntime as onnxrt

#model_info = huggingface_hub.model_info("naver-clova-ix/donut-base-finetuned-cord-v2")

hf_model = VisionEncoderDecoderModel.from_pretrained(
    "eng_donut"
)

hf_model.to('cuda')
onnx_dir = "eng_donut_onnx"
tokenizer = AutoTokenizer.from_pretrained(onnx_dir)
session_options = onnxrt.SessionOptions()
session_options.log_severity_level = 1
session_options.enable_profiling = True
session_options.graph_optimization_level = onnxrt.GraphOptimizationLevel.ORT_ENABLE_EXTENDED
onnx_model = ORTModelForVision2Seq.from_pretrained(onnx_dir,
        session_options,
        use_io_binding=False,
        from_transformers=False,
        provider="CUDAExecutionProvider", 
        provider_options={"cudnn_conv_algo_search": "DEFAULT"})
onnx_model.to('cuda')
onnx_model.name_or_path = onnx_dir
processor = DonutProcessor.from_pretrained(onnx_dir)


dataset = load_dataset("hf-internal-testing/example-documents", split="test")
#image = dataset[2]["image"]
image = Image.open('0114.jpg')
task_prompt = "<s_cord-v2>"
decoder_input_ids = processor.tokenizer(
[...]

PR #1345, opened 2023-09-04
Add name_or_path for donut generation

fxmarty:
Fix following https://github.com/huggingface/transformers/pull/22955

cc @mht-sharma @IlyasMoutawwakil

Currently if `decoder_input_ids` is passed (which is typically the case for donut, see https://huggingface.co/docs/transformers/v4.32.1/en/model_doc/donut#inference), inference with ORT for donut will fail:

```
  File "/home/fxmarty/hf_internship/transformers/src/transformers/generation/utils.py", line 1499, in generate
    input_ids, model_kwargs = self._prepare_decoder_input_ids_for_generation(
  File "/home/fxmarty/hf_internship/transformers/src/transformers/generation/utils.py", line 695, in _prepare_decoder_input_ids_for_generation
    elif self.config.model_type == "vision-encoder-decoder" and "donut" in self.name_or_path.lower():
AttributeError: 'ORTModelForVision2Seq' object has no attribute 'name_or_path'
```

due to https://github.com/huggingface/transformers/blob/bfb1895e3346cb8a2bf2560c75d45e70edf46a47/src/transformers/generation/utils.py#L694

Found out about the issue thanks to https://github.com/huggingface/optimum/issues/1337

### DIFF
--- a/optimum/onnxruntime/modeling_seq2seq.py
+++ b/optimum/onnxruntime/modeling_seq2seq.py
@@ -607,6 +607,7 @@ class ORTModelForConditionalGeneration(ORTModel, ABC):
             preprocessors=preprocessors,
         )
         self.config = config
+        self.name_or_path = config.name_or_path
 
         self.onnx_paths = onnx_paths
         self.use_cache = use_cache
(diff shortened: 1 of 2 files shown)


mht-sharma (approved):
LGTM, thanks @fxmarty

Issue #1334, opened 2023-09-01
Enable CLI export of decoder-only models without present outputs


mgoin:
### Feature request

Currently `optimum-cli export onnx` only supports exporting text-generation models with present outputs (`--task text-generation`) or with past+present outputs (``--task text-generation-with-past`). It would be useful to be able to export a variant without any caching structures if they will not be used.

Example of how `--task text-generation` is not sufficient for this usecase:
<details>

```
optimum-cli export onnx --model facebook/opt-125m --task text-generation TEST
...
Validating ONNX model TEST/decoder_model.onnx...
        -[✓] ONNX model output names match reference model (present.7.key, present.2.key, present.3.key, present.2.value, present.3.value, present.10.value, logits, present.8.key, present.0.value, present.10.key, present.1.key, present.1.value, present.11.key, present.9.value, present.6.value, present.4.value, present.7.value, present.5.value, present.5.key, present.8.value, present.9.key, present.4.key, present.6.key, present.0.key, present.11.value)
        - Validating ONNX Model output "logits":
                -[✓] (2, 16, 50272) matches (2, 16, 50272)
                -[x] values not close enough, max diff: 3.719329833984375e-05 (atol: 1e-05)
        - Validating ONNX Model output "present.0.key":
                -[✓] (2, 12, 16, 64) matches (2, 12, 16, 64)
                -[✓] all values close (atol: 1e-05)
        - Validating ONNX Model output "present.0.value":
                -[✓] (2, 12, 16, 64) matches (2, 12, 16, 64)
                -[✓] all values close (atol: 1e-05)
        - Validating ONNX Model output "present.1.key":
                -[✓] (2, 12, 16, 64) matches (2, 12, 16, 64)
                -[✓] all values close (atol: 1e-05)
        - Validating ONNX Model output "present.1.value":
                -[✓] (2, 12, 16, 64) matches (2, 12, 16, 64)
                -[✓] all values close (atol: 1e-05)
        - Validating ONNX Model output "present.2.key":
                -[✓] (2, 12, 16, 64) matches (2, 12, 16, 64)
                -[✓] all values close (atol: 1e-05)
        - Validating ONNX Model output "present.2.value":
                -[✓] (2, 12, 16, 64) matches (2, 12, 16, 64)
                -[✓] all values close (atol: 1e-05)
        - Validating ONNX Model output "present.3.key":
                -[✓] (2, 12, 16, 64) matches (2, 12, 16, 64)
[...]

PR #1358, opened 2023-09-06
Do not output KV cache when not using `with-past` in the ONNX export

fxmarty:
As per title, likely fixes a longstanding bug. Should anyway refactor all that to use `variant` at some point.

Fixes https://github.com/huggingface/optimum/issues/1334


optimum/exporters/onnx/base.py
```diff
@@ -821,24 +821,24 @@ def __init__(
...
+            if use_present_in_outputs is None:
+                self.use_present_in_outputs = True
 
     def with_behavior(
         self,
         behavior: Union[str, ConfigBehavior],
         int_dtype: str = "int64",
         float_dtype: str = "fp32",
         use_past: bool = False,
+        use_present_in_outputs: Optional[bool] = None,
```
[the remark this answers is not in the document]
fxmarty: I was just unwilling to change the signature of the `__init__` that is for some reason `use_present_in_outputs: Optional[bool] = None` and to have to fix potential things breaking due to that.

fxmarty:
Actually it happens that `use_present_in_outputs` is likely useless thanks to @echarlaix notice.

optimum/exporters/onnx/base.py
```diff
@@ -821,24 +821,24 @@ def __init__(
...
+        # Override this to specify custom attribute change for a given behavior.
         if self._behavior is ConfigBehavior.ENCODER:
             self.task = "feature-extraction"
             self.use_past_in_inputs = False
             self.use_present_in_outputs = False
         if self._behavior is ConfigBehavior.DECODER:
             self.use_past_in_inputs = self.use_past
-            self.use_present_in_outputs = True
+            if use_present_in_outputs is None:
+                self.use_present_in_outputs = True
```
echarlaix, line 833: Could this be removed as well (to keep [`use_past`](https://github.com/huggingface/optimum/blob/v1.12.0/optimum/exporters/onnx/base.py#L503) default), also not sure to understand the point of `use_present_in_outputs`, when should it be different from `use_past` ? cc @michaelbenayoun worked on https://github.com/huggingface/optimum/commit/2a4b790be4ac1640e01f3f54b1381e4b594d1695

echarlaix (approved):

optimum/exporters/onnx/base.py
fxmarty: As discussed offline with Michael we can remove it.

mht-sharma (approved):

fxmarty:
In the end this PR removes the `use_present_in_outputs`, and changes the behavior of `use_past` which is now true for the export of both `decoder_model.onnx` and `decoder_with_past_model.onnx` when we export with `use_cache=True`. WDYT?

mgoin (approved):
> In the end this PR removes the use_present_in_outputs, and changes the behavior of use_past which is now true for the export of both decoder_model.onnx and decoder_with_past_model.onnx when we export with use_cache=True. WDYT?

This satisfies my use case and I've tested it works to replace the previous workaround. Appreciate it!

PR #1359, opened 2023-09-06
[Better Transformer] Add MusicGen

### LINKED ISSUE #1357: [RFC] MusicGen `.to_bettertransformer()` integration
### Feature request

Add support for MusicGen Better Transformer integration. MusicGen is composed of three sub-models:

1. Text encoder: maps the text inputs to a sequence of hidden-state representations. The pre-trained MusicGen models use a frozen text encoder from either T5 or Flan-T5
2. MusicGen decoder: a language model (LM) that auto-regressively generates audio tokens (or codes) conditional on the encoder hidden-state representations. The pre-trained MusicGen models use the BART decoder structure
3. Audio codec: used to encode an audio prompt to use as prompt tokens, and recover the audio waveform from the audio tokens predicted by the decoder. The pre-trained MusicGen models use the [EnCodec model](https://huggingface.co/docs/transformers/main/model_doc/encodec)

=> the text encoder uses the T5 attention module, and the MusicGen decoder uses the BART attention module. Thus, there are no extra attention layers we need to add to optimum. The audio codec is not transformer based, so we don't need to export it to better transformer.

The question is simply how to get the integration working with the sub-model structure. The config file for MusicGen is nested in the same way as the model structure, containing sub-configs for each of the three components: https://huggingface.co/docs/transformers/main/model_doc/musicgen#transformers.MusicgenConfig 

=> this means that the text encoder config is accessed as `config.text_encoder`, and the text encoder model as `model.text_encoder`. Likewise, the MusicGen decoder config is accessed as `config.decoder`, and the text encoder model as `model.decoder`. We need to export the pairs of {models, configs} to their better transformer counterparts, e.g. {`model.text_encoder`, `config.text_encoder`} -> `better_transformer_text_encoder`, and {`model.decoder`, `config.decoder`} -> `better_transformer_decoder`.

Ideally, we'd like to be able to export the entire model to better transformer in one go:
```python
from transformers import  MusicgenForConditionalGeneration

model = MusicgenForConditionalGeneration.from_pretrained("facebook/musicgen-small")
model = model.to_bettertransformer()
```

However. we can't simply export {`model`, `config`} like this, since the top-level config does not contain the config attributes for the sub-models. It's just a place-holder for the sub-model configs.

A simple workaround is to export the text encoder and decoder separately:
```python
from transformers import  MusicgenForConditionalGeneration

model = MusicgenForConditionalGeneration.from_pretrained("facebook/musicgen-small")
model.text_encoder = model.text_encoder.to_bettertransformer()
model.decoder = model.decoder.to_bettertransformer()
```
=> but this diverges from the better transformer API

### Motivation

~9M MusicGen [downloads](https://huggingface.co/models?search=facebook/musicgen) per month -> huge interest in running the model!

### Your contribution

Happy to help with the integration!

sanchit-gandhi:
# What does this PR do?

Fixes #1357


optimum/bettertransformer/models/__init__.py
```diff
@@ -97,6 +98,10 @@ class BetterTransformerManager:
         },
         "markuplm": {"MarkupLMLayer": BertLayerBetterTransformer},
         "mbart": {"MBartEncoderLayer": MBartEncoderLayerBetterTransformer},
+        "musicgen": {
+            "T5Attention": T5AttentionLayerBetterTransformer,
+            "MusicgenAttention": MusicGenAttentionLayerBetterTransformer,
```
sanchit-gandhi, line 103: Is it sufficient just to transform the attention layer @fxmarty, or do we need to convert the whole transformer block, as per this guide? https://huggingface.co/docs/optimum/bettertransformer/tutorials/contribute#step-2-building-the-xxxlayerbettertransformer-module

tests/bettertransformer/test_audio.py
```diff
@@ -172,6 +176,318 @@ def test_save_load_invertible(self, test_name: str, model_type: str, keep_origin
...
+            hf_random_model = automodel_class.from_pretrained(model_id, torch_dtype=torch.float16).to(0)
+            converted_model = BetterTransformer.transform(hf_random_model, keep_original_model=False)
+
+            # MusicGen does not support deepcopy, hence we make an external copy here
+            hf_random_model = automodel_class.from_pretrained(model_id, torch_dtype=torch.float16).to(0)
+        else:
+            hf_random_model = automodel_class.from_pretrained(model_id).to(0)
+            converted_model = BetterTransformer.transform(hf_random_model, keep_original_model=False)
+
+            # MusicGen does not support deepcopy, hence we make an external copy here
```
fxmarty, line 218: Here keep_original_model=False so deepcopy should not be called

tests/bettertransformer/test_audio.py
```diff
@@ -172,6 +176,318 @@ def test_save_load_invertible(self, test_name: str, model_type: str, keep_origin
...
+        """
+        # The first row of the attention mask needs to be all ones -> check: https://github.com/pytorch/pytorch/blob/19171a21ee8a9cc1a811ac46d3abd975f0b6fc3b/test/test_nn.py#L5283
+        inputs = self.prepare_inputs_for_class(model_id=model_id, model_type=model_type, **preprocessor_kwargs).to(0)
+
+        torch.manual_seed(0)
+        if not use_to_operator:
+            hf_random_model = automodel_class.from_pretrained(model_id, torch_dtype=torch.float16).to(0)
+            converted_model = BetterTransformer.transform(hf_random_model, keep_original_model=False)
+
+            # MusicGen does not support deepcopy, hence we make an external copy here
```
fxmarty, line 212: same

fxmarty (approved):
LGTM, thank you for the addition! Did you make sure on a real-sized model that the output matches? Also, it could be interesting to run a script similar to `tests/benchmark/benchmark_bettertransformer.py` (and maybe training) to see what speedup there is.

For the tests, it seems that many methods are reimplemented, was it necessary for all of them?

Note to self: there is no test for audio models to make sure that gradient matches (contrary to NLP models), this should maybe be added sometime.

fxmarty:
@sanchit-gandhi Can you run `make style`?

fxmarty:
Could you also run tests on GPU and check whether they pass: `pytest tests/bettertransformer -s -k "BetterTransformersMusicgenTest" -vvvvv`

fxmarty:
@sanchit-gandhi Reading the failing tests, you should probably add musicgen to https://github.com/huggingface/optimum/blob/f7292664c38ff18004e934de061a6123b899d8e6/optimum/bettertransformer/transformation.py#L338

optimum/bettertransformer/models/__init__.py
fxmarty: No it is good as is!

sanchit-gandhi:
Hey @fxmarty - thanks for the review and comments! Much appreciated :)

1. I ran a logits test on the full-sized model: they match for the baseline model and better transformers one ✅ The logits test was more or less the same as we have in `transformers`: https://github.com/huggingface/transformers/blob/0188739a74dca8a9cf3f646a9a417af7f136f1aa/tests/models/musicgen/test_modeling_musicgen.py#L1164
I can add this as a slow test in this PR to confirm
2. I'm struggling to get any real speed-up on my toy benchmark. If I run the model in full-precision (fp32), I get inf errors:
```python
from transformers import MusicgenConfig, MusicgenForConditionalGeneration, MusicgenProcessor, set_seed

config = MusicgenConfig.from_pretrained('facebook/musicgen-small')
model = MusicgenForConditionalGeneration.from_pretrained('facebook/musicgen-small')
processor = MusicgenProcessor.from_pretrained('facebook/musicgen-small')

model = model.to("cuda:0")
model = model.to_bettertransformer()

set_seed(1)
audio_better = model.generate(**inputs.to("cuda:0"), max_new_tokens=256)
```

**Traceback:**
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
File <timed exec>:1

File ~/hf/lib/python3.10/site-packages/torch/utils/_contextlib.py:115, in context_decorator.<locals>.decorate_context(*args, **kwargs)
    112 @functools.wraps(func)
    113 def decorate_context(*args, **kwargs):
    114     with ctx_factory():
--> 115         return func(*args, **kwargs)

File ~/transformers/src/transformers/models/musicgen/modeling_musicgen.py:2456, in MusicgenForConditionalGeneration.generate(self, inputs, generation_config, logits_processor, stopping_criteria, synced_gpus, streamer, **kwargs)
   2448     input_ids, model_kwargs = self._expand_inputs_for_generation(
   2449         input_ids=input_ids,
   2450         expand_size=generation_config.num_return_sequences,
   2451         is_encoder_decoder=self.config.is_encoder_decoder,
   2452         **model_kwargs,
   2453     )
   2455     # 12. run sample
-> 2456     outputs = self.sample(
   2457         input_ids,
   2458         logits_processor=logits_processor,
   2459         logits_warper=logits_warper,
   2460         stopping_criteria=stopping_criteria,
   2461         pad_token_id=generation_config.pad_token_id,
   2462         eos_token_id=generation_config.eos_token_id,
   2463         output_scores=generation_config.output_scores,
   2464         return_dict_in_generate=generation_config.return_dict_in_generate,
   2465         synced_gpus=synced_gpus,
   2466         streamer=streamer,
   2467         **model_kwargs,
   2468     )
   2470 else:
   2471     raise ValueError(
   2472         "Got incompatible mode for generation, should be one of greedy or sampling."
   2473         "Ensure that beam search is de-activated by setting `num_beams=1` and `num_beam_groups=1`."
   2474     )

File ~/transformers/src/transformers/generation/utils.py:2766, in GenerationMixin.sample(self, input_ids, logits_processor, stopping_criteria, logits_warper, max_length, pad_token_id, eos_token_id, output_attentions, output_hidden_states, output_scores, return_dict_in_generate, synced_gpus, streamer, **model_kwargs)
   2764 # sample
   2765 probs = nn.functional.softmax(next_token_scores, dim=-1)
-> 2766 next_tokens = torch.multinomial(probs, num_samples=1).squeeze(1)
   2768 # finished sentences should have their next token be a padding token
   2769 if eos_token_id is not None:

RuntimeError: probability tensor contains either `inf`, `nan` or element < 0
```

If I run it in float16, the generation works, but no faster than the un-optimised counterpart:
```python
from transformers import MusicgenConfig, MusicgenForConditionalGeneration, MusicgenProcessor, set_seed
import time

config = MusicgenConfig.from_pretrained('facebook/musicgen-small')
model = MusicgenForConditionalGeneration.from_pretrained('facebook/musicgen-small')
processor = MusicgenProcessor.from_pretrained('facebook/musicgen-small')

model = model.to("cuda:0").half()

# warm-up step
_ = model.generate(**inputs.to("cuda:0"), max_new_tokens=256)

start = time.time()
set_seed(1)
audio_base = model.generate(**inputs.to("cuda:0"), max_new_tokens=256)
time_base = time.time() - start

model = model.to_bettertransformer()

start = time.time()
set_seed(1)
audio_better = model.generate(**inputs.to("cuda:0"), max_new_tokens=256)
time_better = time.time() - start

print(time_base)
print(time_better)
```

On an NVIDIA Titan RTX (`vorace`) this gives:
```
3.161407470703125
2.879577398300171
```
=> so an 8.8% improvement with better transformer. Does this seem about right? I'm running PT 2.2 nightly, so should get a good speed-up on an RTX machine? The flash attention 2 claims are closer to 2x speed-up right?

Either way, I can do a deeper benchmark with the script provided!
3. The reason we override all of the tests is because MusicGen has two models in `transformers`: the decoder-only LM, and the full encoder-decoder. I (maybe mistakenly) added the decoder-only model to the AutoModel mapping, and not the full encoder-decoder, so when we call `AutoModel.from_pretrained` in the better transformer tests, we only load the decoder-only model. Thus, I've override the tests to call the MusicGen encoder-decoder class

fxmarty:
@sanchit-gandhi I'll try to have a look this week! Flash v2 is not supported in PyTorch SDPA (and won't unfortunately neither be in the 2.1 release). Also, the x2 speedup (or whatever) is only theoretical. It is quite possible that MusicGen is bottlenecked by other components that are not attention. Using PyTorch profiler is useful for that.

Overall though, I would not trust the benchmark code that you use, there are several possible flaws there.

In general, forward + backward speedup is more significant: https://pytorch.org/blog/out-of-the-box-acceleration/

Finally, it can be good to try and use the context manager https://pytorch.org/docs/master/backends.html#torch.backends.cuda.sdp_kernel to check on which kernel you are actually dispatching.

sanchit-gandhi:
Going to close this since sdpa is now supported in `transformers` and we'll probably just integrate this feature directly there!

PR #1364, opened 2023-09-07
add SDXL panorama pipeline and tiled VAE for ORT

ssube:
# What does this PR do?

Adds a new SDXL pipeline based on the multi-diffusion paper (aka panorama) and a tiled VAE for ORT.

The pipeline is based on https://multidiffusion.github.io/ and https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion_panorama.py, and supports both txt2img and img2img.

Used like:

```python
from optimum.onnxruntime import ORTStableDiffusionXLPanoramaPipeline

model_id = "stabilityai/stable-diffusion-xl-base-1.0"
pipeline = ORTStableDiffusionXLPipeline.from_pretrained(model_id)

# patch pipeline with tiled VAE
pipeline.vae_decoder = ORTModelTiledVaeWrapper(pipeline.vae_decoder, True, 64, 0.5)
pipeline.vae_encoder = ORTModelTiledVaeWrapper(pipeline.vae_encoder, False, 64, 0.5)

# run txt2img
prompt = "sailing ship in storm by Leonardo da Vinci"
image = pipeline(prompt).images[0]

# you can use the same cached pipeline for img2img
image = pipeline(prompt, image=image, strength=0.5).images[0]
```

I have a few questions about the style:

- Should `ORTModelTiledVaeWrapper` be integrated with the other ORT model VAE classes?
- Should the VAE patch be a method on the pipeline?
- How should the window/stride size be set for the pipeline? for the VAE?
- Is the dual-mode `__call__` with `if "image" in kwargs` acceptable?
  - Keeping the pipeline loaded and reusing the cached models can provide a pretty significant performance increase when using the "hires fix" or other high resolution modes that use txt2img followed by many img2img runs
  - If not, should the pipeline be split up into two?
- What is the correct return type for a encoder/decoder?

I still need to clean up `check_inputs` to handle the txt2img parameters correctly, but wanted to posted this to get some feedback.

## Before submitting

There are tests and they are now passing. ~but they are not passing:~ 

```
================================================================================================================== short test summary info ===================================================================================================================
FAILED tests/onnxruntime/test_stable_diffusion_pipeline.py::ORTStableDiffusionXLPanoramaPipelineTest::test_compare_to_diffusers_0_stable_diffusion_xl - onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Non-zero status code returned while running Conv node. Name:'/conv_in/Conv_output_0_nchwc' Status Message: Invalid input shape: {32,0}
FAILED tests/onnxruntime/test_stable_diffusion_pipeline.py::ORTStableDiffusionXLPanoramaPipelineTest::test_image_reproducibility_0_stable_diffusion_xl - onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Non-zero status code returned while running Conv node. Name:'/conv_in/Conv_output_0_nchwc' Status Message: Invalid input shape: {32,0}
====================================================================================================== 2 failed, 5951 deselected, 30 warnings in 53.20s ======================================================================================================
```

~The pipeline does work for me elsewhere, so I'm guessing I messed something up while moving it into optimum or in the tests. I am looking into that.~

The `get_views` method does not work correctly when the pipeline's height/width are less than the window size. Reducing the window size for the tests fixed them.

Some example outputs:

image

image

The seams are a little bit more visible with SDXL than with SD v1.5, but upscaling and/or applying img2img repeatedly fixes that:

sdxl-panorama-highres

(cropped to remove some upscaling errors)

### DIFF
--- a/docs/source/onnxruntime/usage_guides/models.mdx
+++ b/docs/source/onnxruntime/usage_guides/models.mdx
@@ -248,3 +248,26 @@ image = refiner(prompt=prompt, image=image[None, :]).images[0]
 image.save("sailing_ship.png")
 ```
 
+
+### Panorama
+
+Here is an example of how you can use the SDXL panorama pipeline with the tiled VAE to render images that would
+otherwise be too large to fit into VRAM:
+
+```python
+from optimum.onnxruntime import ORTStableDiffusionXLPanoramaPipeline
+
+model_id = "stabilityai/stable-diffusion-xl-base-1.0"
+pipeline = ORTStableDiffusionXLPipeline.from_pretrained(model_id)
+
+# patch pipeline with tiled VAE
+pipeline.vae_decoder = ORTModelTiledVaeWrapper(pipeline.vae_decoder, True, 64, 0.5)
+pipeline.vae_encoder = ORTModelTiledVaeWrapper(pipeline.vae_encoder, False, 64, 0.5)
+
+# run txt2img
+prompt = "sailing ship in storm by Leonardo da Vinci"
+image = pipeline(prompt).images[0]
+
+# you can use the same cached pipeline for img2img
+image = pipeline(prompt, image=image, strength=0.5).images[0]
+```
--- a/optimum/onnxruntime/modeling_diffusion.py
+++ b/optimum/onnxruntime/modeling_diffusion.py
@@ -30,6 +30,9 @@ from diffusers import (
     StableDiffusionPipeline,
     StableDiffusionXLImg2ImgPipeline,
 )
+from diffusers.models.autoencoder_kl import AutoencoderKLOutput
+from diffusers.models.vae import DecoderOutput
+from diffusers.pipelines.onnx_utils import ORT_TO_NP_TYPE
 from diffusers.schedulers.scheduling_utils import SCHEDULER_CONFIG_NAME
 from diffusers.utils import CONFIG_NAME, is_invisible_watermark_available
 from huggingface_hub import snapshot_download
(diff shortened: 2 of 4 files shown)


ssube:
@echarlaix totally fair. I ran into that while writing tests, there is no reference pipeline for the comparison. :sweat_smile: 

I forgot to mention in the initial post that this is all derived from a panorama pipeline that I have for SD v1/v2, so I can open a PR for that instead, if you would prefer. Eventually I would like to upstream both of them, but the code is very very similar.

> focusing our effort on pipelines already supported by diffusers and which are more in demand (such as controlnet or upscaler pipelines)

Is that an XL upscaler or are you referring to implementing the pipeline for `stabilityai/stable-diffusion-x4-upscaler` in optimum? I wrote the ONNX upscaler pipeline in diffusers (https://github.com/huggingface/diffusers/pull/2158) and may be able to help port that.

echarlaix:
> I forgot to mention in the initial post that this is all derived from an ORT panorama pipeline that I have for SD v1/v2, so I can open a PR for that instead, if you would prefer. Eventually I would like to upstream both of them, but the code is very very similar.

Yes I think it makes a lot of sense to first support the panorama pipeline for SD models supported in [diffusers](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion_panorama.py) that you shared. For SDXL pipelines I'm not against the integration as long as we keep it "experimental" (warning message + stated in the documentation) until integrated to diffusers as we won't be able to compare it with the diffusers pipeline, wdyt @ssube ?


> Is that an XL upscaler or are you referring to implementing the pipeline for stabilityai/stable-diffusion-x4-upscaler in optimum? I wrote the ONNX upscaler pipeline in diffusers (https://github.com/huggingface/diffusers/pull/2158) and may be able to help port that.

That's awesome! Yes I was talking about the upscaler [pipeline](https://github.com/huggingface/diffusers/blob/v0.21.1/src/diffusers/pipelines/stable_diffusion/pipeline_onnx_stable_diffusion_upscale.py#L56) to which you contributed! We are planning to add ORT support for the pipelines the most used by the community :  in priority we would like to support the pipelines with no ORT integration from diffusers (like [`StableDiffusionControlNetPipeline`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/controlnet/pipeline_controlnet_sd_xl.py#L102) or [`KandinskyV22Pipeline`](https://github.com/huggingface/diffusers/blob/v0.21.0/src/diffusers/pipelines/kandinsky2_2/pipeline_kandinsky2_2.py#L67)) and in a second time the ones which are already integrated (like the ORT upscaler pipeline mentionned above) so that they can later be deprecated in diffusers. We don't have a lot of bandwidth to do so at the moment, but more than happy to help you on any integrations, don't hesitate to ping me!

ssube:
I opened an MR for the panorama pipeline for v1 and v2: https://github.com/huggingface/optimum/pull/1390. It has a few test failures, which I am working on. That also includes the tiled VAE, so if/when one of these is merged, I will rebase the other and remove the VAE.

> For SDXL pipelines I'm not against the integration as long as we keep it "experimental" (warning message + stated in the documentation) until integrated to diffusers as we won't be able to compare it with the diffusers pipeline, wdyt @ssube ?

That seems fine to me. How/where should I add that warning in the pipeline? 

> We are planning to add ORT support for the pipelines the most used by the community : in priority we would like to support the pipelines with no ORT integration from diffusers (like [StableDiffusionControlNetPipeline](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/controlnet/pipeline_controlnet_sd_xl.py#L102) or [KandinskyV22Pipeline](https://github.com/huggingface/diffusers/blob/v0.21.0/src/diffusers/pipelines/kandinsky2_2/pipeline_kandinsky2_2.py#L67)) and in a second time the ones which are already integrated (like the ORT upscaler pipeline mentionned above) so that they can later be deprecated in diffusers.

I'm not very familiar with the Kandinsky pipelines, but there is an ONNX pipeline for ControlNet. The only problem is that it was developed under the GPL license, so I don't think you can include it as-is: https://github.com/uchuusen/onnx_stable_diffusion_controlnet

The biggest difference is in the Unet inputs. afaik, ORT does not support passing tuples to the model, so those have to be expanded into a fixed number of inputs. I think that limits the number of controlnets you can use, and I'm not sure if that will be a problem for XL, but it seems to work pretty well for v1. My tool and at least one other ONNX GUI tool export two copies of the Unet with and without controlnet inputs, which takes up more disk but we only load one or the other, so memory is the same.

PR #1371, opened 2023-09-08
upgrade min compatible optimum-intel version

echarlaix:


fxmarty (approved):
thanks for removing my prints :)

fxmarty:
will add to the patch

echarlaix:
> will add to the patch

Thanks! 

Also penend a PR to fix doc building https://github.com/huggingface/optimum-furiosa/pull/12

PR #1289, opened 2023-08-16
Add optional position ids to forward pass

kunal-vaishnavi:
# What does this PR do?

This PR adds support for providing `position_ids` in the forward pass. Per the [Hugging Face glossary](https://huggingface.co/docs/transformers/glossary#position-ids), absolute positional embeddings are used by default if the `position_ids` input is not provided. Because LLaMA uses rotary positional embeddings, the `position_ids` input is needed. Currently, the exported ONNX models for LLaMA do not have this input because it is missing in the forward pass.

The PR changes are needed to export and run LLaMA with Optimum + ONNX Runtime in this [ORT PR](https://github.com/microsoft/onnxruntime/pull/17020).

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #1381, opened 2023-09-13
Add position ids in ONNX export and ORT

fxmarty:
Supersede https://github.com/huggingface/optimum/pull/1289 & https://github.com/huggingface/optimum/pull/984

Fix a critical issue where the exported ONNX models could not produce meaningful results for batch size > 1 due to a very strong assumption in transformers: that all sequences are of the same length if `position_ids` is not passed. See: https://github.com/huggingface/transformers/blob/a796f7eea6c86b54671a6f522cebbe41f630bb62/src/transformers/models/gpt2/modeling_gpt2.py#L802

Left to do:
- [x] Add an option to export without position_ids for backward compatibility (disabled by default)
- [ ] Fix models on the Hub [will only do for those under Optimum namespace - but not until we have a release]
- [x] Add ORT tests that would have catched the issue [the added batched generation test indeed fails on `main`)

cc @xenova

### DIFF
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -155,6 +155,23 @@ class TextDecoderOnnxConfig(OnnxConfigWithPast):
         return models_and_onnx_configs, onnx_files_subpaths
 
 
+class TextDecoderWithPositionIdsOnnxConfig(TextDecoderOnnxConfig):
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        common_inputs = super().inputs
+
+        # Decoders based on GPT2 require a position_ids input to avoid
+        # generating wrong position_ids in the model itself:
+        # https://github.com/huggingface/transformers/blob/v4.33.1/src/transformers/models/gpt2/modeling_gpt2.py#L802
+        if not self.no_position_ids and self.task == "text-generation":
+            if self.use_past_in_inputs:
+                common_inputs["position_ids"] = {0: "batch_size"}
+            else:
+                common_inputs["position_ids"] = {0: "batch_size", 1: "sequence_length"}
+
+        return common_inputs
+
+
 class TextSeq2SeqOnnxConfig(OnnxSeq2SeqConfigWithPast):
     """
     Handles encoder-decoder-based text architectures.
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -66,6 +66,29 @@ class TextDecoderOnnxConfig(OnnxConfigWithPast):
     DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, DummyPastKeyValuesGenerator)
     DUMMY_PKV_GENERATOR_CLASS = DummyPastKeyValuesGenerator
 
+    def __init__(
+        self,
+        config: "PretrainedConfig",
+        task: str = "feature-extraction",
+        int_dtype: str = "int64",
+        float_dtype: str = "fp32",
+        use_past: bool = False,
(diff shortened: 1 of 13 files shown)


xenova:
Thanks for this fix @fxmarty! I'll be testing this with transformers.js hopefully soon.

optimum/exporters/onnx/config.py
```diff
@@ -66,6 +66,29 @@ class TextDecoderOnnxConfig(OnnxConfigWithPast):
...
+            config=config,
+            task=task,
+            int_dtype=int_dtype,
+            float_dtype=float_dtype,
+            use_past=use_past,
+            use_past_in_inputs=use_past_in_inputs,
+            preprocessors=preprocessors,
+        )
+        # TODO: remove no_position_ids once optimum is sufficiently above 1.13
+        self.no_position_ids = no_position_ids
```
michaelbenayoun, line 90: Should we make helper function like `deprecated(version_number)` that does nothing when the version is lower than that and fails after?
This way we dont forget to remove deprecated code.

optimum/exporters/onnx/model_configs.py
```diff
@@ -218,15 +258,29 @@ class LlamaOnnxConfig(TextDecoderOnnxConfig):
...
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        common_inputs = super().inputs
+
+        # Refer to GPT2OnnxConfig inputs comment.
+        if not self.no_position_ids and self.task == "text-generation":
+            if self.use_past_in_inputs:
+                common_inputs["position_ids"] = {0: "batch_size"}
+            else:
+                common_inputs["position_ids"] = {0: "batch_size", 1: "sequence_length"}
+        return common_inputs
```
michaelbenayoun, line 271: Should we make a `TextDecoderWithPositionIdsOnnxConfig` since it seems to be repeated accross many models?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -748,3 +774,33 @@ def _reorder_cache(past: Tuple[Tuple[torch.Tensor]], beam_idx: torch.Tensor) ->
...
+            "use_cache": use_cache,
+            "position_ids": None,
+            "attention_mask": attention_mask,
+        }
+
+
+class ORTMPTForCausalLM(ORTModelForCausalLM):
+    # Adapted from transformers.models.gpt2.modeling_gpt2.GPT2LMHeadModel.prepare_inputs_for_generation
+    def prepare_inputs_for_generation(self, input_ids, past_key_values=None, **kwargs):
+        attention_mask = kwargs.get("attention_mask", None)  # input_ids.new_ones(input_ids.shape)
```
michaelbenayoun, line 797: Remove comment?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -748,3 +774,33 @@ def _reorder_cache(past: Tuple[Tuple[torch.Tensor]], beam_idx: torch.Tensor) ->
             for layer_past in standardized_past
         )
         return bloom_convert_to_bloom_cache(reordered_past)
+
+
+class ORTOPTForCausalLM(ORTModelForCausalLM):
+    # Adapted from transformers.models.gpt2.modeling_gpt2.GPT2LMHeadModel.prepare_inputs_for_generation
+    def prepare_inputs_for_generation(self, input_ids, past_key_values=None, **kwargs):
+        attention_mask = kwargs.get("attention_mask", None)  # input_ids.new_ones(input_ids.shape)
```
michaelbenayoun, line 782: Remove comment?

michaelbenayoun (approved):
Thanks for handling this @fxmarty

Issue #1172, opened 2023-07-07
Add support for Falcon model to export to ONNX


mindest:
### Feature request

Add support so that we can export Falcon model to ONNX format using `optimum-cli`, like

    optimum-cli export onnx -m tiiuae/falcon-7b falcon_7b --trust-remote-code

### Motivation

Wanted to run Falcon model using ONNXRuntime. Currently error will be raised when exporting Falcon model to ONNX, using the above command:
`KeyError: "refinedwebmodel is not supported yet. Only {'electra', 'mt5', ...} are supported. If you want to support refinedwebmodel please propose a PR or open up an issue."`

### Your contribution

Support may be dependent on updates in transformers, like huggingface/transformers#24523.

Johnno1011:
Yo, just curious, why do we have to wait for a proper release (IE: pip etc) before this can be resolved? I was able to build from source and it works just fine.
Thanks.

fxmarty:
Our CI relies on transformers latest release.

Pradipve2011:
> 

Hey John,

I am looking for solution on this. Can you please share how do you able to convert it?

WilliamTambellini:
up

fxmarty:
Hi, the support is added in https://github.com/huggingface/optimum/pull/1391

PR #1391, opened 2023-09-16
Add ONNX and ORT support for Falcon

fxmarty:
This one was more painful than it should have been because:
* Exporting without `position_ids` input and then using `generate` is bugged because the generated position_ids https://github.com/huggingface/transformers/blob/0a55d9f7376f72ad3ff296d4249840021b03bcc4/src/transformers/models/falcon/modeling_falcon.py#L932 have a different shape than the `position_ids` generated in `generate`. I believe this is a bug in many Transformers models.
* Transormers implementation reformats the KV cache directly in the model, which is new: https://github.com/huggingface/transformers/pull/26199
* Optimum ORT has a rather bad support of MQA. Especially, the shape computation for IO Binding is very ugly with a lot of controlflow, expensive calls to `normalized_config` - this should be refactored with inheritance to avoid any controlflow at all.
* ORT does not support bool input to `Trilu` op https://github.com/microsoft/onnxruntime/issues/16189

Remaining issue: I think `repeat_interleave` ONNX export inserts `Loop` in the ONNX, which we may want to avoid. EDIT: fixed in pytorch 2.1


fxmarty:
Fixes https://github.com/huggingface/optimum/issues/1172

optimum/exporters/onnx/model_patcher.py
```diff
@@ -223,6 +226,202 @@ def patched_forward(*args, **kwargs):
...
+        self,
+        config: "OnnxConfig",
+        model: Union["PreTrainedModel", "TFPreTrainedModel"],
+        model_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        super().__init__(config, model, model_kwargs)
+
+        # This is kind of ugly and bug prone if other FalconModel are instantiated.
+        if config.task == "text-generation":
+            model.transformer.__class__.forward = falcon_model_forward_without_kv_reformatting
```
michaelbenayoun, line 384: You can change it only for this specific instance:

```suggestion
        # This is kind of ugly and bug prone if other FalconModel are instantiated.
        if config.task == "text-generation":
            model.transformer.forward = falcon_model_forward_without_kv_reformatting.__get__(model.transformer)
```

optimum/exporters/onnx/model_patcher.py
```diff
@@ -223,6 +226,202 @@ def patched_forward(*args, **kwargs):
...
+
+        self.original_make_causal = transformers.models.falcon.modeling_falcon._make_causal_mask
+        transformers.models.falcon.modeling_falcon._make_causal_mask = _make_causal_mask_falcon_patched
```
michaelbenayoun, line 387: We can make this in `__enter__` no?

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -707,8 +707,100 @@ def _from_pretrained(
...
+            onnx_paths=onnx_paths,
+            decoder_with_past_session=decoder_with_past_session,
+            use_cache=use_cache,
+            use_io_binding=use_io_binding,
+            model_save_dir=model_save_dir,
+            preprocessors=preprocessors,
+            generation_config=generation_config,
+            **kwargs,
+        )
+        # self.num_kv_heads = config.num_kv_heads if (config.new_decoder_architecture or not config.multi_query) else 1
```
michaelbenayoun, line 745: To remove?

optimum/utils/normalized_config.py
```diff
@@ -211,7 +211,7 @@ class NormalizedConfigManager:
         "blenderbot": BartLikeNormalizedTextConfig,
         "blenderbot_small": BartLikeNormalizedTextConfig,
         "bloom": NormalizedTextConfig.with_args(num_layers="n_layer"),
-        "falcon": NormalizedTextConfig.with_args(num_layers="num_hidden_layers", num_attention_heads="num_kv_heads"),
+        "falcon": NormalizedTextConfig,
```
michaelbenayoun, line 214: Question: does `NormalizedConfig` have a `NUM_KV_HEADS` attribute to normalize it or not?

fxmarty: No

optimum/onnxruntime/modeling_decoder.py
fxmarty: Let's keep it for now

optimum/exporters/onnx/model_patcher.py
fxmarty: Thank you, did it

PR #1394, opened 2023-09-19
add normalizedconfig for baichuan and mixformer-sequential

changwangss:
# What does this PR do?

Fixes # (issue)
we want to use the models https://huggingface.co/baichuan-inc/Baichuan2-13B-Base and https://huggingface.co/microsoft/phi-1, but the normalizedconfig doesn't contain.

## Before submitting


michaelbenayoun:
I think the issu is that we can end-up with multiple names potentially pointing to different `NormalizedConfig`s. 

I think what we can do is allow the user to provide a `NormalizedConfig` on its own: 

1. When exporting the model via the `export` function it is trivial to do so
2. When exporting the model via the CLI, it is less trivial: either the desired `NormalizedConfig` already exists and in this case it can simply be the name of the `NormalizedConfig` to use, or it does not exist yet, in which case I think we can point the user to use the `export` function?

fxmarty:
couldn't we just greedy and have a heuristic of most commonly used key names, and try to use those?

michaelbenayoun:
Yes, we could do that.

PenghuiCheng (approved):

changwangss:
> I think the issu is that we can end-up with multiple names potentially pointing to different `NormalizedConfig`s.
> 
> I think what we can do is allow the user to provide a `NormalizedConfig` on its own:
> 
> 1. When exporting the model via the `export` function it is trivial to do so
> 2. When exporting the model via the CLI, it is less trivial: either the desired `NormalizedConfig` already exists and in this case it can simply be the name of the `NormalizedConfig` to use, or it does not exist yet, in which case I think we can point the user to use the `export` function?

Could you have the plan continue to do this feature?  @fxmarty @michaelbenayoun 
I close the PR first and hope the feature can be completed soon

PR #1408, opened 2023-09-22
Fix normalized config key for models architecture

echarlaix:
Replace key of `_conf` with underscore to dash so that we have the same format as the `TasksManager._SUPPORTED_MODEL_TYPE` https://github.com/huggingface/optimum/blob/6d1ae0eb9d3660b46f1b2e705ff2f61b38baacd6/optimum/exporters/tasks.py#L281 Currently there is a mix between the two which makes this model https://huggingface.co/hf-internal-testing/tiny-random-BlenderbotSmallForCausalLM/blob/main/config.json#L27 not compatible for example

### DIFF
--- a/optimum/onnxruntime/utils.py
+++ b/optimum/onnxruntime/utils.py
@@ -101,8 +101,8 @@ class ORTConfigManager:
         "albert": "bert",
         "bart": "bart",
         "bert": "bert",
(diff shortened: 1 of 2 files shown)


fxmarty (approved):
Sorry I missed it. LGTM!

PR #1419, opened 2023-09-27
add_exllamav2

SunMarc:
# What does this PR do ? 
This PR adds the possibility to choose exllamav2 kernels for GPTQ model This PR follows the [integration](https://github.com/PanQiWei/AutoGPTQ/pull/349) of the kernels in auto-gptq. I've also added a test to check that we are able to load and infer using exllamav2 kernel. I will update the benchmark in a follow-up PR. 

- [ ] Merge after the release of auto-gptq ?

### DIFF
--- a/docs/source/llm_quantization/usage_guides/quantization.mdx
+++ b/docs/source/llm_quantization/usage_guides/quantization.mdx
@@ -76,7 +76,7 @@ quantized_model = load_quantized_model(empty_model, save_folder=save_folder, dev
 
 ### Exllama kernels for faster inference
 
-For 4-bit model, you can use the exllama kernels in order to a faster inference speed. If you want to change its value, you just need to pass `disable_exllama` in [`~optimum.gptq.load_quantized_model`]. In order to use these kernels, you need to have the entire model on gpus.
+For 4-bit model, you can use the exllama kernels in order to have a faster inference speed. If you want to change its value, you just need to pass `disable_exllama` in [`~optimum.gptq.load_quantized_model`]. In order to use these kernels, you need to have the entire model on gpus.
 
 ```py
 from optimum.gptq import GPTQQuantizer, load_quantized_model
--- a/docs/source/llm_quantization/usage_guides/quantization.mdx
+++ b/docs/source/llm_quantization/usage_guides/quantization.mdx
@@ -86,7 +86,7 @@ from accelerate import init_empty_weights
 with init_empty_weights():
     empty_model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
 empty_model.tie_weights()
-quantized_model = load_quantized_model(empty_model, save_folder=save_folder, device_map="auto", disable_exllama=False)
+quantized_model = load_quantized_model(empty_model, save_folder=save_folder, device_map="auto", disable_exllama=False, disable_exllamav2=True)
 ```
 
 With the release of the exllamav2 kernel, you can get faster inference speed compared to the exllama kernels. You just need to 
@@ -104,6 +104,9 @@ quantized_model = load_quantized_model(empty_model, save_folder=save_folder, dev
 ```
 
 Note that only 4-bit models are supported with exllama/exllamav2 kernels for now. Furthermore, it is recommended to disable the exllama kernel when you are finetuning your model with peft.
+
+You can find the benchmark of these kernels [here](https://github.com/huggingface/optimum/tree/main/tests/benchmark#gptq-benchmark)
+
 #### Fine-tune a quantized model
 
 With the official support of adapters in the Hugging Face ecosystem, you can fine-tune models that have been quantized with GPTQ.
--- a/docs/source/llm_quantization/usage_guides/quantization.mdx
+++ b/docs/source/llm_quantization/usage_guides/quantization.mdx
@@ -76,20 +76,7 @@ quantized_model = load_quantized_model(empty_model, save_folder=save_folder, dev
 
 ### Exllama kernels for faster inference
 
-For 4-bit model, you can use the exllama kernels in order to have a faster inference speed. If you want to change its value, you just need to pass `disable_exllama` in [`~optimum.gptq.load_quantized_model`]. In order to use these kernels, you need to have the entire model on gpus.
-
-```py
-from optimum.gptq import GPTQQuantizer, load_quantized_model
-import torch
-
-from accelerate import init_empty_weights
-with init_empty_weights():
-    empty_model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
-empty_model.tie_weights()
-quantized_model = load_quantized_model(empty_model, save_folder=save_folder, device_map="auto", disable_exllama=False, disable_exllamav2=True)
-```
-
-With the release of the exllamav2 kernel, you can get faster inference speed compared to the exllama kernels. It is activated by default: `disable_exllamav2=False` in [`~optimum.gptq.load_quantized_model`]. 
+With the release of the exllamav2 kernel, you can get faster inference speed compared to the exllama kernels for 4-bit model. It is activated by default: `disable_exllamav2=False` in [`~optimum.gptq.load_quantized_model`]. In order to use these kernels, you need to have the entire model on gpus.
 
 ```py
 from optimum.gptq import GPTQQuantizer, load_quantized_model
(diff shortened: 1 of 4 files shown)


docs/source/llm_quantization/usage_guides/quantization.mdx
```diff
@@ -86,10 +86,26 @@ from accelerate import init_empty_weights
 with init_empty_weights():
     empty_model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
 empty_model.tie_weights()
-quantized_model = load_quantized_model(empty_model, save_folder=save_folder, device_map="auto", disable_exllama=False)
+quantized_model = load_quantized_model(empty_model, save_folder=save_folder, device_map="auto", disable_exllama=False, disable_exllamav2=True)
 ```
 
-Note that only 4-bit models are supported with exllama kernels for now. Furthermore, it is recommended to disable the exllama kernel when you are finetuning your model with peft.
+With the release of the exllamav2 kernel, you can get faster inference speed compared to the exllama kernels. You just need to
```
fxmarty, line 92: The default is already `disable_exllamav2=False` so I am not sure to understand. Are these new kernels selected by default?

docs/source/llm_quantization/usage_guides/quantization.mdx
```diff
@@ -86,10 +86,26 @@ from accelerate import init_empty_weights
 with init_empty_weights():
     empty_model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
 empty_model.tie_weights()
-quantized_model = load_quantized_model(empty_model, save_folder=save_folder, device_map="auto", disable_exllama=False)
+quantized_model = load_quantized_model(empty_model, save_folder=save_folder, device_map="auto", disable_exllama=False, disable_exllamav2=True)
```
fxmarty, line 89: same comment

optimum/gptq/quantizer.py
```diff
@@ -137,6 +141,11 @@ def __init__(
             raise ValueError("group_size must be greater than 0 or equal to -1")
         if not (0 < self.damp_percent < 1):
             raise ValueError("damp_percent must between 0 and 1.")
+        if not self.disable_exllamav2 and not self.disable_exllama:
+            logger.warning(
+                "You have activated exllama and exllamav2 backend. Setting `disable_exllama=True` and keeping `disable_exllamav2=False`"
+            )
+            self.disable_exllama = True
```
fxmarty, line 148: I think you should raise an error here, but it is personal taste.

optimum/gptq/quantizer.py
```diff
@@ -440,13 +450,21 @@ def tmp(_, input, output):
...
+                        "Found modules on cpu/disk. Using Exllama backend requires all the modules to be on GPU. Setting `disable_exllama=True`"
+                    )
+                    self.disable_exllama = True
+                if not self.disable_exllamav2:
+                    logger.warning(
+                        "Found modules on cpu/disk. Using Exllamav2 backend requires all the modules to be on GPU. Setting `disable_exllamav2=True`"
+                    )
+                    self.disable_exllamav2 = True
+            # act order and exllama
+            elif self.desc_act and not self.disable_exllama:
```
fxmarty, line 467: Is it not the same thing with act-order + exllamav2? Or exllamav2 dropped support for act-order?

tests/gptq/test_quantization.py
```diff
@@ -178,7 +187,7 @@ def test_exllama_serialization(self):
                 )
             empty_model.tie_weights()
             quantized_model_from_saved = load_quantized_model(
-                empty_model, save_folder=tmpdirname, device_map={"": 0}, disable_exllama=False
+                empty_model, save_folder=tmpdirname, device_map={"": 0}, disable_exllama=False, disable_exllamav2=True
```
fxmarty, line 190: I would add a `@parameterized.expand` here to test both exllama and exllama v2. Edit: but the approach with two classes works (though it duplicates code).

tests/gptq/test_quantization.py
```diff
@@ -197,7 +206,12 @@ def test_exllama_max_input_length(self):
                 )
             empty_model.tie_weights()
             quantized_model_from_saved = load_quantized_model(
-                empty_model, save_folder=tmpdirname, device_map={"": 0}, disable_exllama=False, max_input_length=4028
+                empty_model,
```
fxmarty, line 209: same

tests/gptq/test_quantization.py
```diff
@@ -213,6 +227,42 @@ def test_exllama_max_input_length(self):
...
+    disable_exllama = True
+    disable_exllamav2 = True
+
+    def test_generate_quality(self):
+        # don't need to test
+        pass
+
+    def test_serialization(self):
+        # don't need to test
+        pass
```
fxmarty, line 241: why not?

docs/source/llm_quantization/usage_guides/quantization.mdx
SunMarc: I clarified the doc, thanks.

docs/source/llm_quantization/usage_guides/quantization.mdx
SunMarc: `disable_exllamav2` is set  by default to `False`

optimum/gptq/quantizer.py
SunMarc: Yeah, good point. The user will have to be more careful.

optimum/gptq/quantizer.py
SunMarc: When i tested it, exllamav2 modifies the qweight with or without act-order. So in the end, it is not feasible save a model using exllamav2 and reload it with the same weights. This makes the quantization process complicated with exllamav2. Basically, the user should only use exllamav2 after it has been quantized and saved.

tests/gptq/test_quantization.py
SunMarc: I prefer to keep the duplicated code. The two classes are not exactly the same for exllama and exllamav2. For exllama, we need to differentiate between act-order and no act-order whereas it is not the case for exllamav2.

tests/gptq/test_quantization.py
SunMarc: Here, we quantize the model using the cuda-old kernel and save the model to later load it with exllamav2 for the `test_exllama_serialization` test. Since these tests will use the cuda-old kernel, we don't need to test them as we already do so in a previous test.

docs/source/llm_quantization/usage_guides/quantization.mdx
SunMarc: Yeah, those are selected by default. I will reformulate the doc so that it is clearer.

docs/source/llm_quantization/usage_guides/quantization.mdx
```diff
@@ -76,7 +76,7 @@ quantized_model = load_quantized_model(empty_model, save_folder=save_folder, dev
 
 ### Exllama kernels for faster inference
 
-For 4-bit model, you can use the exllama kernels in order to a faster inference speed. It is activated by default. If you want to change its value, you just need to pass `disable_exllama` in [`~optimum.gptq.load_quantized_model`]. In order to use these kernels, you need to have the entire model on gpus.
+For 4-bit model, you can use the exllama kernels in order to a faster inference speed. If you want to change its value, you just need to pass `disable_exllama` in [`~optimum.gptq.load_quantized_model`]. In order to use these kernels, you need to have the entire model on gpus.
```
fxmarty, line 79:
```suggestion
For 4-bit model, you can use the exllama kernels in order to have a faster inference speed. If you want to change its value, you just need to pass `disable_exllama` in [`~optimum.gptq.load_quantized_model`]. In order to use these kernels, you need to have the entire model on gpus.
```

tests/gptq/test_quantization.py
fxmarty: And how about generate_quality?

docs/source/llm_quantization/usage_guides/quantization.mdx
fxmarty, line 79: Sorry being picky: it could make sense to first give the example with exllamav2 (aka no need to pass any arg), and then have a small section explaining how to use the previous kernels from exllama. Here it is a bit weird having the least important part first.

fxmarty (approved):
LGTM, that's great! As commented on transformers PR by somebody else, that's true that the inflation of disable args is not very scalable, that was probably a bad idea on my end to use that in AutoGPTQ.

tests/gptq/test_quantization.py
SunMarc: This is also tested in the `GPTQTest` class. The wording is confusing but `test_exllama_serialization` in `GPTQTestExllamav2` does two things: test the loading quantized weights with exllamav2 kernels + test the inference correctness.

docs/source/llm_quantization/usage_guides/quantization.mdx
SunMarc: Makes sense !

achew010:
Hi @SunMarc,

I was running some GPTQ-PEFT experiments through optimum's GPTQ interface and noticed the loss curves for the exllama kernel diverges,

drawing

Im wondering if you guys observed a similar behaviour in your tests previously?

PR #1420, opened 2023-09-27
gptq_benchmark_update

SunMarc:
# What does this PR do ? 
This PR updates the gptq script and the results with the new exllamav2 kernel that are being integrated into [optimum](https://github.com/huggingface/optimum/pull/1419) and [transformers](https://github.com/huggingface/transformers/pull/26437). I also add perplexity measure for comparison.

- [x] merge optimum and transformers PRs.
- [x] update readme with new results for generation with A100
- [x] edit benchmark requirement version


tests/benchmark/benchmark_gptq.py
```diff
@@ -288,46 +302,22 @@ def benchmark_memory(
...
-        save_folder=args.gptq_model,
-        state_dict_name="model.safetensors",
+    quantization_config = GPTQConfig(
+        bits=4, disable_exllama=args.disable_exllama, disable_exllamav2=args.disable_exllamav2
+    )
+    model = autoclass.from_pretrained(
+        args.model,
+        revision=args.revision,
+        quantization_config=quantization_config,
+        torch_dtype=torch.float16,
```
fxmarty, line 314: No strong feeling about it, but this depends on Transformers, while it did not before. I don't think it matters a lot though.

tests/benchmark/benchmark_gptq.py
```diff
@@ -364,82 +375,100 @@ def benchmark_memory(
...
-output_file = open(file_name, "w")
-header = "quantization, act_order, bits, group_size, kernel, num_batches, batch_size, prompt_length, new_tokens, Load time (s), Per-token latency (ms), Throughput (tok/s), Max memory (MB)\n"
-output_file.write(header)
-
-latencies = {}
-throughputs = {}
-all_max_mem = {}
-print(
-    "WARNING: The reported peak memory is only a rough estimate, and can NOT be precisely relied upon to estimate an OOM limit."
-)
```
fxmarty, line 377: Can you leave this warning?

fxmarty (approved):
LGTM thanks a lot!

tests/benchmark/benchmark_gptq.py
SunMarc: I decided to go with that as we don't need to clone any model

PR #1421, opened 2023-09-28
Fix arg in bettertransformer llama attention 

SunMarc:
# What does this PR do ? 
This PR fixes the integration of llama in bettertransformers. Since the [PR](https://github.com/huggingface/transformers/pull/25598) about flashattention2 in llama models was merged, a new argument `padding_mask` was introduced silently in the forward of the attention module. This breaks the forward of the `llama_forward` in bettertransformer. 

@younesbelkada We may have to do it for every model that supports flashattention2. LMK if I should just do one PR for all supported + plan to support models ?

### DIFF
--- a/optimum/bettertransformer/models/attention.py
+++ b/optimum/bettertransformer/models/attention.py
@@ -242,6 +242,7 @@ def opt_forward(
     attention_mask: Optional[torch.Tensor] = None,
     layer_head_mask: Optional[torch.Tensor] = None,
     output_attentions: bool = False,
+    **kwargs,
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Thanks ! I think that you can just pass dummy `**kwargs` all over the places, what do you think?

SunMarc:
Yeah, we can do that too since FA2 is experimental at this stage and will be removed in the future from transformers. LMK what you think @fxmarty.

fxmarty:
LGTM, thank you for fixing! Yes, feel free to add kwargs everywhere in this PR.

Issue #1399, opened 2023-09-20
Export in ONNX/FP16 of PY007/TinyLlama-1.1B-Chat-v0.2 fails 


claeyzre:
### System Info

```shell
Python: v3.9.13

Optimum: v1.13.1
Transformers: v4.33.2
Onnxruntime: v1.15.1
CUDA: v11.7

System: Windows 11
```

### Who can help?

@fxmarty  ?

### Reproduction (minimal, reproducible, runnable)

While running this command
```
optimum-cli export onnx --model PY007/TinyLlama-1.1B-Chat-v0.2  --task causal-lm-with-past --fp16 --for-ort --device cuda tiny-llamav0.2-onnx
```
This model is a "Tiny" Llama-2.

I got this stack, I assume it's during the post-processing step:
```
(venv) PS D:\john.doe\optimum-test> optimum-cli export onnx --model PY007/TinyLlama-1.1B-Chat-v0.2  --task causal-lm-with-past --fp16 --for-ort --device cuda tiny-llamav0.2-onnx
The option --for-ort was passed, but its behavior is now the default in the ONNX exporter and passing it is not required anymore.
Framework not specified. Using pt to export to ONNX.
Downloading (…)lve/main/config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 652/652 [00:00<?, ?B/s]
D:\john.doe\optimum-test\venv\lib\site-packages\huggingface_hub\file_download.py:137: UserWarning: `huggingface_hub` cache-system uses symlinks by default to efficiently store duplicated files but your machine does not support them in D:\cache\huggingface\hub. Caching files will still work but in a degraded version that might require more space on your disk. This warning can be disabled by setting the `HF_HUB_DISABLE_SYMLINKS_WARNING` environment variable. For more details, see https://huggingface.co/docs/huggingface_hub/how-to-cache#limitations.
To support symlinks on Windows, you either need to activate Developer Mode or to run Python as an administrator. In order to see activate developer mode, see this article: https://docs.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development
  warnings.warn(message)
Downloading model.safetensors: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4.40G/4.40G [01:50<00:00, 39.8MB/s]
[...]

PR #975, opened 2023-04-17
Add llama onnx export & onnxruntime support

fxmarty:
As per title

PR #1425, opened 2023-10-02
Add Mistral models ONNX export support

echarlaix:

PR #1432, opened 2023-10-05
Fix llama ONNX export

fxmarty:
LLama uses an optional configuration key `num_key_value_heads` for the number of heads, and uses `num_attention_heads` to compute the head dimension. This was unfortunately not implemented in https://github.com/huggingface/optimum/pull/975 (apart from llama2 70b, the llama and llama2 series do not make use of this num_key_value_heads key), as the key probably did not exist at the time.

Fixes https://github.com/huggingface/optimum/issues/1399

### DIFF
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -218,7 +218,7 @@ class OPTOnnxConfig(TextDecoderOnnxConfig):
 
 class LlamaOnnxConfig(TextDecoderWithPositionIdsOnnxConfig):
     DEFAULT_ONNX_OPSET = 13
-    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig
+    NORMALIZED_CONFIG_CLASS = NormalizedTextConfig.with_args(num_attention_heads="num_key_value_heads")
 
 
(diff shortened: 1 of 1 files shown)


optimum/exporters/onnx/model_configs.py
```diff
@@ -216,7 +216,48 @@ class OPTOnnxConfig(TextDecoderOnnxConfig):
...
+        super().__init__(
+            task=task,
+            normalized_config=normalized_config,
+            batch_size=batch_size,
+            sequence_length=sequence_length,
+            random_batch_size_range=random_batch_size_range,
+            random_sequence_length_range=random_sequence_length_range,
+            **kwargs,
+        )
+        self.num_key_value_heads = normalized_config.num_key_value_heads
```
echarlaix, line 239: should we also add a fix in [`prepare_inputs_for_merged`](https://github.com/huggingface/optimum/blob/mistral/optimum/onnxruntime/base.py#L221)

(as done in https://github.com/huggingface/optimum/pull/1425)

echarlaix (commented):
thanks for the fix

optimum/exporters/onnx/model_configs.py
fxmarty: yes thank you for taking care of it!

PR #1449, opened 2023-10-14
Fix perceiver tests and dummy inputs for ONNX

baskrahmer:


fxmarty (approved):
Great work, thank you for the fix!

Issue #1414, opened 2023-09-24
Converting OwlViT model results in `ValueError: Unrecognized configuration class`


xenova:
### System Info

```shell
optimum version: 1.13.2
platform: linux
python version: 3.10.12
```

### Who can help?

@fxmarty @michaelbenayoun 

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction (minimal, reproducible, runnable)

Running
```
optimum-cli export onnx --model google/owlvit-base-patch32 out2
```

Results in the following error:
```py
Framework not specified. Using pt to export to ONNX.
Traceback (most recent call last):
  File "/usr/local/bin/optimum-cli", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/dist-packages/optimum/commands/optimum_cli.py", line 163, in main
    service.run()
  File "/usr/local/lib/python3.10/dist-packages/optimum/commands/export/onnx.py", line 232, in run
    main_export(
  File "/usr/local/lib/python3.10/dist-packages/optimum/exporters/onnx/__main__.py", line 323, in main_export
    model = TasksManager.get_model_from_task(
  File "/usr/local/lib/python3.10/dist-packages/optimum/exporters/tasks.py", line 1708, in get_model_from_task
    model = model_class.from_pretrained(model_name_or_path, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/transformers/models/auto/auto_factory.py", line 566, in from_pretrained
    raise ValueError(
ValueError: Unrecognized configuration class <class 'transformers.models.owlvit.configuration_owlvit.OwlViTConfig'> for this kind of AutoModel: AutoModelForObjectDetection.
Model type should be one of ConditionalDetrConfig, DeformableDetrConfig, DetaConfig, DetrConfig, TableTransformerConfig, YolosConfig.
```

### Expected behavior

The model should export correctly.

PR #1453, opened 2023-10-16
Fix owlvit task detection

fxmarty:
Fixes https://github.com/huggingface/optimum/issues/1414

### DIFF
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -1401,7 +1401,7 @@ class TasksManager:
                 pipeline_tag = getattr(model_info, "pipeline_tag", None)
                 # The Hub task "conversational" is not a supported task per se, just an alias that may map to
                 # text-generaton or text2text-generation.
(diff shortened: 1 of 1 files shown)


optimum/exporters/tasks.py
```diff
@@ -1399,9 +1399,11 @@ def _infer_task_from_model_name_or_path(
                 inferred_task_name = "image-classification"
             else:
                 pipeline_tag = getattr(model_info, "pipeline_tag", None)
-                # conversational is not a supported task per se, just an alias that may map to
-                # text-generaton or text2text-generation
-                if pipeline_tag is not None and pipeline_tag != "conversational":
+                # The Hub task "conversational" is not a supported task per se, just an alias that may map to
+                # text-generaton or text2text-generation.
+                # The Hub task "object-detection" is not a supported task per so, as in Transformers this may map to either
```
fxmarty, line 1404:
```suggestion
                # The Hub task "object-detection" is not a supported task per se, as in Transformers this may map to either
```

PR #1457, opened 2023-10-16
Disable bart onnx export for text-classification and question-answering

fxmarty:
PyTorch 2.1 introduced a breaking change in torchscript graph capture that make the export of bart for those tasks fail.

Given that there is probably low usage anyway, let's simply remove the support for it and let's try to get the CI green.

See details: https://github.com/pytorch/pytorch/issues/110597


claeyzre:
Now that this has been fixed in [pytorch](https://github.com/pytorch/pytorch/issues/110597), would it be possible to revert this to get back Bart support ?

PR #1458, opened 2023-10-16
Remove ARCH_MODEL_MAP from tests

baskrahmer:
# What does this PR do?
Followup of https://github.com/huggingface/optimum/pull/1449
The variable is no longer used and it is cleaner to just remove it.


fxmarty (approved):
LGTM thank you for the cleanup!

PR #1460, opened 2023-10-17
ONNX exporter library_name infer fixes

baskrahmer:
# What does this PR do?
Small changes to `library_name` inference


fxmarty (approved):
LGTM thank you!

Just for my knowledge, do you have an example of model on the Hub where `model_info` does not have the attribute `library_name`?

Issue #1450, opened 2023-10-14
open_clip model onnx export not supported 


alachyan-ml:
### System Info

```shell
Python 3.9.16
optimum 1.13.2
Ubuntu 20.04.6 LTS
```

### Who can help?

@michaelbenayoun 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction (minimal, reproducible, runnable)

 ```% optimum-cli export onnx -m laion/CLIP-ViT-B-32-256x256-DataComp-s34B-b86K --framework pt ./clip --task zero-shot-image-classification
Traceback (most recent call last):
  File "/raid/lingo/alachyan/anaconda3/envs/fan_trt/bin/optimum-cli", line 8, in <module>
    sys.exit(main())
  File "/raid/lingo/alachyan/anaconda3/envs/fan_trt/lib/python3.9/site-packages/optimum/commands/optimum_cli.py", line 163, in main
    service.run()
  File "/raid/lingo/alachyan/anaconda3/envs/fan_trt/lib/python3.9/site-packages/optimum/commands/export/onnx.py", line 232, in run
    main_export(
  File "/raid/lingo/alachyan/anaconda3/envs/fan_trt/lib/python3.9/site-packages/optimum/exporters/onnx/__main__.py", line 323, in main_export
    model = TasksManager.get_model_from_task(
  File "/raid/lingo/alachyan/anaconda3/envs/fan_trt/lib/python3.9/site-packages/optimum/exporters/tasks.py", line 1683, in get_model_from_task
    model_class = TasksManager.get_model_class_for_task(
  File "/raid/lingo/alachyan/anaconda3/envs/fan_trt/lib/python3.9/site-packages/optimum/exporters/tasks.py", line 1148, in get_model_class_for_task
    tasks_to_model_loader = TasksManager._LIBRARY_TO_TASKS_TO_MODEL_LOADER_MAP[library]
KeyError: 'open_clip' 

### Expected behavior

Getting key_error for exporting onnx model based on the open_clip library. Seems that the `TasksManager._LIBRARY_TO_TASKS_TO_MODEL_LOADER_MAP` map does not have open_clip integrated. 

Open AI CLIP exporting working as expected without any problems. 

[...]

PR #1466, opened 2023-10-18
[WIP] Support open-clip onnx export

isaac-chung:
# What does this PR do?

Fixes # (issue) https://github.com/huggingface/optimum/issues/1450

## Before submitting


isaac-chung:
@fxmarty here's the WIP. Currently when I run the export command, I get
```
(v1) isaacchung@Isaacs-MBP optimum % optimum-cli export onnx -m laion/CLIP-ViT-B-32-laion2B-s34B-b79K --framework pt clip_onnx
Automatic task detection to zero-shot-image-classification.
`text_config_dict` is provided which will be used to initialize `CLIPTextConfig`. The value `text_config["id2label"]` will be overriden.
`text_config_dict` is provided which will be used to initialize `CLIPTextConfig`. The value `text_config["bos_token_id"]` will be overriden.
`text_config_dict` is provided which will be used to initialize `CLIPTextConfig`. The value `text_config["eos_token_id"]` will be overriden.
/Users/isaacchung/work/transformers/src/transformers/models/clip/feature_extraction_clip.py:28: FutureWarning: The class CLIPFeatureExtractor is deprecated and will be removed in version 5 of Transformers. Please use CLIPImageProcessor instead.
  warnings.warn(
Using the export variant default. Available variants are:
        - default: The default ONNX variant.
`text_config_dict` is provided which will be used to initialize `CLIPTextConfig`. The value `text_config["id2label"]` will be overriden.
`text_config_dict` is provided which will be used to initialize `CLIPTextConfig`. The value `text_config["bos_token_id"]` will be overriden.
`text_config_dict` is provided which will be used to initialize `CLIPTextConfig`. The value `text_config["eos_token_id"]` will be overriden.
Using framework PyTorch: 2.0.0
  File "/Users/isaacchung/virtualenv/v1/bin/optimum-cli", line 8, in <module>
    sys.exit(main())
  File "/Users/isaacchung/work/optimum/optimum/commands/optimum_cli.py", line 163, in main
    service.run()
  File "/Users/isaacchung/work/optimum/optimum/commands/export/onnx.py", line 239, in run
    main_export(
  File "/Users/isaacchung/work/optimum/optimum/exporters/onnx/__main__.py", line 505, in main_export
    _, onnx_outputs = export_models(
  File "/Users/isaacchung/work/optimum/optimum/exporters/onnx/convert.py", line 752, in export_models
    export(
  File "/Users/isaacchung/work/optimum/optimum/exporters/onnx/convert.py", line 855, in export
    export_output = export_pytorch(
  File "/Users/isaacchung/work/optimum/optimum/exporters/onnx/convert.py", line 542, in export_pytorch
    dummy_inputs = config.generate_dummy_inputs(framework="pt", **input_shapes)
  File "/Users/isaacchung/work/optimum/optimum/exporters/onnx/base.py", line 454, in generate_dummy_inputs
    dummy_inputs_generators = self._create_dummy_input_generator_classes(**kwargs)
  File "/Users/isaacchung/work/optimum/optimum/exporters/onnx/base.py", line 222, in _create_dummy_input_generator_classes
    first_inputs_gen = self.DUMMY_INPUT_GENERATOR_CLASSES[0](self.task, self._normalized_config, **kwargs)
  File "/Users/isaacchung/work/optimum/optimum/utils/input_generators.py", line 343, in __init__
    self.vocab_size = normalized_config.vocab_size
  File "/Users/isaacchung/work/optimum/optimum/utils/normalized_config.py", line 109, in __getattr__
    return super().__getattr__(attr_name)
  File "/Users/isaacchung/work/optimum/optimum/utils/normalized_config.py", line 69, in __getattr__
    raise AttributeError(f'Could not find the attribute named "{leaf_attr_name}" in the normalized config.')
AttributeError: Could not find the attribute named "vocab_size" in the normalized config.
```
I'm stepping away for a bit this evening but I will continue afterwards.

fxmarty:
Yes, it appears there are no small open clip models on the Hub. Probably the most reasonable option (if you don't want to go into the hassle of addind a tiny testing model) is to just add a nightly test for open clip with a somewhat "small" model.

fxmarty:
@isaac-chung Thank you. It appears that the `tokenize` function from open_clip and the `open_clip.tokenizer.HFTokenizer` do pad by default to `tokenizer.tokenizer.model_max_length`, while Transformers CLIP do not. It is unclear to me why there is this difference.

Something you could try in the OpenCLIPOnnxConfig is first to remove the dynamic axis for the sequence length by redefining the `inputs` property:

```python
    @property
    def inputs(self) -> Dict[str, Dict[int, str]]:
        return {
            "input_ids": {0: "text_batch_size"},
            "pixel_values": {0: "image_batch_size", 1: "num_channels", 2: "height", 3: "width"},
            "attention_mask": {0: "text_batch_size"},
        }
```

As for the shape itself used during the export, it should be held in this kwargs: https://github.com/huggingface/optimum/blob/15b8d1eed4d83c5004d3b60f6b6f13744b358f01/optimum/exporters/onnx/base.py#L451

What I may suggest is to override the shape from the openclip onnx config:
```python
    def generate_dummy_inputs(self, framework: str = "pt", **kwargs):
        # override sequence_length shape here
        return super().generate_dummy_inputs(framework, **kwargs)
```

You will need to access the tokenizer though, that hopefully you can access under the `self._preprocessors` attribute.

isaac-chung:
Thanks for the tips so far, @fxmarty ! I am wondering if we could draw some inspiration from `pix2StructOnnxConfig`, esp. `def overwrite_shape_and_generate_input(...)`, where they also require padding inputs to the max length. 
```
# pix2struct takes inputs whose so-called sequence length is **static** to max_patches, so we do NOT use
# the passed sequence_length that behaves as a dynamic shape.
```
Probably needs to be massaged to fit into our `OpenCLIPOnnxConfig`. Will investigate.

fxmarty:
@isaac-chung Yes that is a good option as well! It is just that `overwrite_shape_and_generate_input` is not defined at the top level `OnnxConfig`, so you would need to define it there (and basically just generate inputs, no override of shapes at the tope level), and redefine it in OpenClipOnnxConfig to override the relevant shapes.

isaac-chung:
I see what you mean now, thanks again! Will continue with the `generate_dummy_inputs` approach. Seems more straightforward 💪

isaac-chung:
Seeing a different error, so I'll count this as progress :D It seems like [this issue](https://github.com/pytorch/pytorch/issues/98190) has been resolved, but specifying `--opset 18` and installing torch nightly yielded the following
```
 File "/Users/isaacchung/work/optimum/optimum/exporters/onnx/convert.py", line 883, in export
    config.fix_dynamic_axes(output, device=device, input_shapes=input_shapes, dtype=dtype)
  File "/Users/isaacchung/work/optimum/optimum/exporters/onnx/base.py", line 302, in fix_dynamic_axes
    session = InferenceSession(model_path.as_posix(), providers=providers, sess_options=session_options)
  File "/Users/isaacchung/virtualenv/v1/lib/python3.9/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 419, in __init__
    self._create_inference_session(providers, provider_options, disabled_optimizers)
  File "/Users/isaacchung/virtualenv/v1/lib/python3.9/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 463, in _create_inference_session
    sess.initialize_session(providers, provider_options, disabled_optimizers)
onnxruntime.capi.onnxruntime_pybind11_state.NotImplemented: [ONNXRuntimeError] : 9 : NOT_IMPLEMENTED : Could not find an implementation for ArgMax(13) node with name '/ArgMax'
```

This looks similar to [this issue](https://github.com/microsoft/onnxruntime/issues/10068) but doesn't seem to apply to the types here (read that we only deal with fp16).

fxmarty:
Hi @isaac-chung , could you check where is the argmax in the modeling code? What do you mean by `we only deal with fp16`?

isaac-chung:
The argmax is within `model.encode_text`. In this [line](https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/model.py#L279), `self.text_pool_type` is `arg_max`. 

> What do you mean by we only deal with fp16?

I misread things, please disregard.

fxmarty:
Thank you! Can you check the dtype of `text` here https://github.com/mlfoundations/open_clip/blob/7b8dd2cbaf9cca13ca5b1defa6a321a145eb166c/src/open_clip/transformer.py#L558?

A solution would be to use the ModelPatcher (see e.g. https://github.com/huggingface/optimum/blob/15b8d1eed4d83c5004d3b60f6b6f13744b358f01/optimum/exporters/onnx/model_patcher.py#L638-L651 & https://github.com/huggingface/optimum/blob/15b8d1eed4d83c5004d3b60f6b6f13744b358f01/optimum/exporters/onnx/model_configs.py#L1415-L1418) to cast the arg `text` in [`text_global_pool`](https://github.com/mlfoundations/open_clip/blob/7b8dd2cbaf9cca13ca5b1defa6a321a145eb166c/src/open_clip/transformer.py#L550) to a dtype accepted by ONNX Runtime: https://github.com/microsoft/onnxruntime/blob/main/docs/OperatorKernels.md

Something like: 

```python
if is_open_clip_available():
    import open_clip

class OpenClipModelPatcher(...):
    def __enter__(...):
        ...
        open_clip.transformer.text_global_pool = text_global_pool_patched

    def __exit__(...):
```

isaac-chung:
`text` is of type `torch.Tensor`. 

And great, thank you! Let me take a look and report back.

fxmarty:
I meant dtype!

isaac-chung:
Whoops, let me see. I got int64. Maybe we should cast it to something like int8?
```
In [5]: text.dtype
Out[5]: torch.int64
```
It is odd that the [main branch](https://github.com/mlfoundations/open_clip/blob/main/src/open_clip/transformer.py#L672) does not have the same method.

fxmarty:
@isaac-chung According to https://github.com/microsoft/onnxruntime/blob/main/docs/OperatorKernels.md you could try to add a cast to `torch.int32` before the operation.

optimum/exporters/onnx/model_patcher.py
```diff
@@ -448,3 +449,45 @@ def __init__(
...
+def _text_global_pool_patched(x, text: Optional[torch.Tensor] = None, pool_type: str = 'argmax'):
+    text.to(dtype=torch.int32)
+    if pool_type == 'first':
+        pooled, tokens = x[:, 0], x[:, 1:]
+    elif pool_type == 'last':
+        pooled, tokens = x[:, -1], x[:, :-1]
+    elif pool_type == 'argmax':
+        # take features from the eot embedding (eot_token is the highest number in each sequence)
+        assert text is not None
+        pooled, tokens = x[torch.arange(x.shape[0]), text.argmax(dim=-1)], x
```
fxmarty, line 469:
```suggestion
    if pool_type == 'first':
        pooled, tokens = x[:, 0], x[:, 1:]
    elif pool_type == 'last':
        pooled, tokens = x[:, -1], x[:, :-1]
    elif pool_type == 'argmax':
        text = text.to(dtype=torch.int32)  # ONNX Runtime is unable to run argmax with int64 input, hence this cast.
        # take features from the eot embedding (eot_token is the highest number in each sequence)
        assert text is not None
        pooled, tokens = x[torch.arange(x.shape[0]), text.argmax(dim=-1)], x
```

optimum/exporters/onnx/model_patcher.py
```diff
@@ -448,3 +449,45 @@ def __init__(
             self._patch_func = _prepare_decoder_attention_mask
             self._orig_func_name = "_prepare_decoder_attention_mask"
             self._orig_func = self._model_to_patch._prepare_decoder_attention_mask
+
+
+def is_open_clip_available():
+    return importlib.util.find_spec("open_clip") is not None
```
fxmarty, line 455: Could you move this in `optimum/utils/import_utils.py`?

isaac-chung: Done.

optimum/exporters/onnx/model_patcher.py
isaac-chung: Nice catch, makes sense.

Still getting the same error. I wonder if there's something missing with how the model patcher is setup.

fxmarty: That's weird. I'll give it a try later.

optimum/exporters/onnx/model_patcher.py
```diff
@@ -448,3 +451,42 @@ def __init__(
...
+        config: "OnnxConfig",
+        model: Union["PreTrainedModel", "TFPreTrainedModel"],
+        model_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        super().__init__(config, model, model_kwargs)
+        self.original_text_global_pool = open_clip.transformer.text_global_pool
+
+    def __enter__(self):
+        self.patch_ops()
+        open_clip.transformer.text_global_pool = _text_global_pool_patched
```
fxmarty, line 487: Can you try:
```suggestion
        open_clip.transformer.text_global_pool.__code__ = _text_global_pool_patched.__code__
```

I believe that given that the model is already instanciated, the old `text_global_pool` is still used despite patching. This can be checked by printing stuff in `_text_global_pool_patched`.

isaac-chung: Thanks! Yep with the addition of `__code__`, print stmt works within `_text_global_pool_patched`. The `torch.export` seems to have run without issues, and the onnx input names are now `['images','text']`. Later this error is raised:
```
  File "/Users/isaacchung/work/optimum/optimum/exporters/onnx/base.py", line 331, in fix_dynamic_axes
    outputs = session.run(None, onnx_inputs)
  File "/Users/isaacchung/virtualenv/v1/lib/python3.9/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 216, in run
    self._validate_input(list(input_feed.keys()))
  File "/Users/isaacchung/virtualenv/v1/lib/python3.9/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 198, in _validate_input
    raise ValueError(
ValueError: Required inputs (['image', 'text']) are missing from input feed (['input_ids', 'pixel_values', 'attention_mask']).
```
Interestingly, in `fix_dynamic_axes`, setting `to_fix = []` allows the ONNX export to succeed, but the following warning is raised:
```
The ONNX export succeeded with the warning: The exported ONNX model does not have the exact same outputs as what is provided in OpenCLIPOnnxConfig. Difference: 4624.
```

fxmarty: Yes, the input name should be fixed as well there. You can check how the inputs are generated and whether the input remap is called & where/why/why not.

isaac-chung: Made some good progress in this new commit. I feel that we are very close. The only diff now is in the values are mismatching:
```
Validating ONNX model clip_onnx/model.onnx...
        -[✓] ONNX model output names match reference model (logit_scale, image_features, text_features)
        - Validating ONNX Model output "text_features":
                -[✓] (2, 512) matches (2, 512)
                -[x] values not close enough, max diff: 0.3609406352043152 (atol: 1e-05)
        - Validating ONNX Model output "image_features":
                -[✓] (2, 512) matches (2, 512)
                -[x] values not close enough, max diff: 0.36094051599502563 (atol: 1e-05)
        - Validating ONNX Model output "logit_scale":
                -[✓] () matches ()
                -[✓] all values close (atol: 1e-05)
The ONNX export succeeded with the warning: The maximum absolute difference between the output of the reference model and the ONNX exported model is not within the set tolerance 1e-05:
- text_features: max diff = 0.3609406352043152
- image_features: max diff = 0.36094051599502563.
 The exported model was saved at: clip_onnx
```

fxmarty: Awesome, great work! I wonder if 0.36 is actually small compared to the output distribution or not. It could be worth checking the mean absolute difference. Maybe there's a dropout with `training=True` somewhere (like here https://github.com/huggingface/transformers/blob/0baa9246cb1ddac355db1df7824a521426599eb7/src/transformers/models/speecht5/modeling_speecht5.py#L720, though I doubt it)

isaac-chung: Thanks! The mean absolute difference is around 0.043, so maybe 0.36 is quite significant.
```
(Pdb) np.mean(np.abs(ref_value - ort_value))
0.042761594
```
The max diff value changes slightly at every run, but I think that's likely due to the random dummy inputs. Dropout is set to be 0 in the configs too. I initially ran these on a CPU, so trying now on a GPU.

isaac-chung: Same error - so it does not seem to be a device issue.

mertalev: I'm actually working on exporting open_clip models and ran into an issue when calling `torch.jit.trace`. Not sure if it's the same issue, but I had to explicitly do

```
for param in model.parameters():
   param._requires_grad(False)
```

`.eval()` and `with torch.no_grad()` didn't help; it would get stuck while exporting unless I added the above snippet.

isaac-chung: Where did you add the snippet btw?

mertalev: Just before calling `torch.jit.trace`. [Here's the link](https://github.com/immich-app/immich/commit/b5badc07acfbdc934cef13d2689f16af1acf0f38#diff-3951c3cdd96468dc58826b8555b35b0216a5e61d2fa224298a1b97fe77999326R57)

isaac-chung: Thanks. Just tested it out by add that within `TasksManager.get_model_from_task`, and setting `jit=False`, still the same error. I don't think the issue here is jit related.

isaac-chung:
Just a heads up - I anticipate a busier week this week, so hoping to continue the investigation afterwards. If any new ideas pop to mind, feel free to continue the discussion and I can catch back up.

isaac-chung:
It is possible that the discrepancy is due to the task being auto-detected as "zero-shot-classification" instead of "feature-extraction". Currently only the image/text embeds are in the output, where as for ZS-classification we need the logits ([I was reading this](https://gist.github.com/rwightman/c79fd0241ed3c860e898114931c07990#file-convert_open_clip_to_hf-py-L244)). The scale more-or-less matches the magnitude of logit_scale. 

I haven't been able to confirm this yet - will start looking into adding the feature-extraction task as well and how to add the logic for generating those logits from features in the model config. It seems like this might require the model patcher when the task is ZS-classification?

isaac-chung:
Sorry for the long pause. 
Printing out the validation step ref outputs vs. onnx outputs, at a glance the values look very close. 
```
Validating ONNX model clip_onnx/model.onnx...
ref_outputs_dict={'image_features': tensor([[ 0.0082, -0.0139, -0.0986,  ...,  0.0043,  0.0170, -0.0243],
        [ 0.0125, -0.0246, -0.0875,  ...,  0.0078,  0.0198, -0.0267]]), 'text_features': tensor([[-0.0114, -0.0058, -0.0060,  ..., -0.0212, -0.0546, -0.0145],
        [-0.0216, -0.0326, -0.0414,  ...,  0.0304,  0.0093,  0.0012]]), 'logit_scale': tensor(100.0000)}
onnx_outputs=[array([[ 0.00823098, -0.01392082, -0.09864607, ...,  0.00433102,
         0.01700564, -0.02425753],
       [ 0.01247843, -0.02459271, -0.08752848, ...,  0.00782712,
         0.01979438, -0.02666864]], dtype=float32), array([[-0.01138339, -0.00584862, -0.0059775 , ..., -0.02118963,
        -0.05462093, -0.01450986],
       [-0.02160461, -0.03263872, -0.04141301, ...,  0.03040083,
         0.00926695,  0.00119737]], dtype=float32), array(100.00001, dtype=float32)]
```

Validation would only pass if `ATOL_FOR_VALIDATION = 1`.

kechan:
Thanks all very much, looks like lot of work done to support CLIP indeed. I am looking to try export ONNX.
I am using the L14 and here's the cli output. Will utilize and report back if I see any issues.

optimum-cli export onnx -m laion/CLIP-ViT-L-14-laion2B-s32B-b82K --framework pt clip_onnx
.../python310_env/lib/python3.10/site-packages/huggingface_hub/file_download.py:1132: FutureWarning: `resume_download` is deprecated and will be removed in version 1.0.0. Downloads always resume when possible. If you want to force a new download, use `force_download=True`.
  warnings.warn(
Automatic task detection to zero-shot-image-classification.
.../python310_env/lib/python3.10/site-packages/transformers/models/clip/feature_extraction_clip.py:28: FutureWarning: The class CLIPFeatureExtractor is deprecated and will be removed in version 5 of Transformers. Please use CLIPImageProcessor instead.
  warnings.warn(
Using the export variant default. Available variants are:
    - default: The default ONNX variant.

***** Exporting submodel 1/1: CLIPModel *****
Using framework PyTorch: 2.2.1
.../python310_env/lib/python3.10/site-packages/transformers/models/clip/modeling_clip.py:279: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if attn_weights.size() != (bsz * self.num_heads, tgt_len, src_len):
.../python310_env/lib/python3.10/site-packages/transformers/models/clip/modeling_clip.py:319: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if attn_output.size() != (bsz * self.num_heads, tgt_len, self.head_dim):
.../python310_env/lib/python3.10/site-packages/transformers/modeling_attn_mask_utils.py:86: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if input_shape[-1] > 1 or self.sliding_window is not None:
.../python310_env/lib/python3.10/site-packages/transformers/modeling_attn_mask_utils.py:162: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if past_key_values_length > 0:
.../python310_env/lib/python3.10/site-packages/transformers/models/clip/modeling_clip.py:287: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if causal_attention_mask.size() != (bsz, 1, tgt_len, src_len):
.../python310_env/lib/python3.10/site-packages/transformers/models/clip/modeling_clip.py:296: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if attention_mask.size() != (bsz, 1, tgt_len, src_len):
.../python310_env/lib/python3.10/site-packages/torch/onnx/symbolic_opset9.py:5859: UserWarning: Exporting aten::index operator of advanced indexing in opset 14 is achieved by combination of multiple ONNX operators, including Reshape, Transpose, Concat, and Gather. If indices include negative values, the exported graph will produce incorrect results.
  warnings.warn(
Post-processing the exported models...
Deduplicating shared (tied) weights...

Validating ONNX model clip_onnx/model.onnx...
	-[✓] ONNX model output names match reference model (image_embeds, text_embeds, logits_per_text, logits_per_image)
	- Validating ONNX Model output "logits_per_image":
		-[✓] (2, 2) matches (2, 2)
		-[✓] all values close (atol: 1e-05)
	- Validating ONNX Model output "logits_per_text":
		-[✓] (2, 2) matches (2, 2)
		-[✓] all values close (atol: 1e-05)
	- Validating ONNX Model output "text_embeds":
		-[✓] (2, 768) matches (2, 768)
		-[✓] all values close (atol: 1e-05)
	- Validating ONNX Model output "image_embeds":
		-[✓] (2, 768) matches (2, 768)
		-[✓] all values close (atol: 1e-05)
The ONNX export succeeded and the exported model was saved at: clip_onnx

sushilkhadkaanon:
Hi @isaac-chung @fxmarty , I've clone your repo and installed optimum from isaac-chung:support-open-clip-onnx-export, 
But still it's not working for me, 
`orch.onnx.errors.UnsupportedOperatorError: Exporting the operator 'aten::_native_multi_head_attention' to ONNX opset version 18 is not supported. Please feel free to request support or submit a pull request on PyTorch GitHub: https://github.com/pytorch/pytorch/issues.`

I think this issue might be due to newer version of dependencies, Could you guys please give me info about the environment so that I can reproduce. Thanks!

PR #1469, opened 2023-10-19
Enable Latent Consistency models ONNX export

echarlaix:
Add Latent Consistency models (added in `diffusers` in https://github.com/huggingface/diffusers/pull/5438 and https://github.com/huggingface/diffusers/pull/5448 ) ONNX export and pipeline to enable ONNX Runtime inference for text-to-image 

```python
from optimum.onnxruntime import ORTLatentConsistencyModelPipeline

pipe = ORTLatentConsistencyModelPipeline.from_pretrained("SimianLuo/LCM_Dreamshaper_v7", export=True)
prompt = "sailing ship in storm by Leonardo da Vinci"
images = pipe(prompt=prompt, num_inference_steps=4, guidance_scale=8.0).images
```

OpenVINO integration : https://github.com/huggingface/optimum-intel/pull/463


optimum/exporters/onnx/base.py
```diff
@@ -435,7 +435,7 @@ def ordered_inputs(self, model: Union["PreTrainedModel", "TFPreTrainedModel"]) -
             sig = inspect.signature(model.call)
 
         for param in sig.parameters:
-            param_regex = re.compile(rf"{param}(\.\d*)?")
+            param_regex = re.compile(rf"{param}(\..*)?$")
```
echarlaix, line 438: this modification comes from `timestep` matching both `timestep` and `timestep_cond` previously (behavior that we don't want), we still want `past_key_value` to match `past_key_values.0.key` though

tests/onnxruntime/test_stable_diffusion_pipeline.py
```diff
@@ -483,3 +488,55 @@ def test_vae_image_processor_pil(self):
...
+    @require_diffusers
+    def test_compare_to_diffusers(self, model_arch: str):
+        ort_pipeline = self.ORTMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch], export=True)
+        self.assertIsInstance(ort_pipeline.text_encoder, ORTModelTextEncoder)
+        self.assertIsInstance(ort_pipeline.vae_decoder, ORTModelVaeDecoder)
+        self.assertIsInstance(ort_pipeline.vae_encoder, ORTModelVaeEncoder)
+        self.assertIsInstance(ort_pipeline.unet, ORTModelUnet)
+        self.assertIsInstance(ort_pipeline.config, Dict)
+
+        from diffusers import LatentConsistencyModelPipeline
```
fxmarty, line 511: This is taste, but maybe imports at the top with a `if_diffusers_is_available()` (or maybe not even needed given that the test is all about diffusers) is preferable.

https://peps.python.org/pep-0008/#imports

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -562,6 +565,15 @@ class ORTStableDiffusionInpaintPipeline(ORTStableDiffusionPipelineBase, StableDi
     __call__ = StableDiffusionInpaintPipelineMixin.__call__
 
 
+@add_end_docstrings(ONNX_MODEL_END_DOCSTRING)
+class ORTLatentConsistencyModelPipeline(ORTStableDiffusionPipelineBase, LatentConsistencyPipelineMixin):
```
fxmarty, line 569: Could you add this one in the documentation?

optimum/utils/dummy_diffusers_objects.py
```diff
@@ -68,3 +68,14 @@ def __init__(self, *args, **kwargs):
     @classmethod
     def from_pretrained(cls, *args, **kwargs):
         requires_backends(cls, ["diffusers"])
+
+
+class ORTLatentConsistencyModelPipeline(metaclass=DummyObject):
```
fxmarty, line 73: Is this essentially putting a requirement on diffusers==v0.22 (next release)? Could it make sense to https://github.com/huggingface/optimum/blob/f223f5e330b678767c76489b3c30154d6959c80c/optimum/utils/dummy_bettertransformer_objects.py#L19 ? Just wondering

.github/workflows/test_onnxruntime.yml
```diff
@@ -29,6 +29,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
+        pip install git+https://github.com/huggingface/diffusers
```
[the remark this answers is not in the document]
echarlaix: yes just added it to trigger the test for latent consistency models and plan to remove it before merging :)

optimum/utils/dummy_diffusers_objects.py
echarlaix: No it doesn't, however we'll need diffusers>=v0.22.0 for the tests (so that we can compare the ORT pipeline with the diffusers `LatentConsistencyModelPipeline` pipeline)

optimum/onnxruntime/modeling_diffusion.py
echarlaix: added it in https://github.com/huggingface/optimum/pull/1469/commits/824fc5702532eed226c139fdeba897be9cb7be16

tests/onnxruntime/test_stable_diffusion_pipeline.py
fxmarty: I see. Then what I would do is a try/except import, and add a decorator (or condition + skiptest) on latent consistency tests.

echarlaix: we already use the skip decorator, concerning your try/except import suggestion I'm not sure to get why it would be cleaner in this case, could you elaborate ? (I'd rather add a diffusers version check or keep as is)

michaelbenayoun (approved):
LGTM, thanks @echarlaix !

tests/onnxruntime/test_stable_diffusion_pipeline.py
fxmarty: It would be cleaner because it would respect PIP 8.

tests/onnxruntime/test_stable_diffusion_pipeline.py
```diff
@@ -483,3 +488,55 @@ def test_vae_image_processor_pil(self):
...
+class ORTLatentConsistencyModelPipelineTest(ORTModelTestMixin):
+    SUPPORTED_ARCHITECTURES = [
+        "latent-consistency",
+    ]
+    ORTMODEL_CLASS = ORTLatentConsistencyModelPipeline
+    TASK = "text-to-image"
+
+    @parameterized.expand(SUPPORTED_ARCHITECTURES)
+    @require_diffusers
+    @unittest.skipIf(parse(_diffusers_version) <= Version("0.21.4"), "not supported with this diffusers version")
```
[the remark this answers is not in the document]
echarlaix: good point, will add it !

tests/onnxruntime/test_stable_diffusion_pipeline.py
fxmarty: Oh, I was referring to something like:

```python
try:
    from diffusers import LatentConsistencyModelPipeline
except Exception:
    LatentConsistencyModelPipeline = None
```

PR #1479, opened 2023-10-24
Added cache_block_outputs parameter to handle models with non-regular structure such as ChatGLM

AlexKoff88:
`cache_block_outputs` enables the collection of the block output to speed up GPTQ process. However, it does not work for some models such as ChatGLM where the LayerNorm is the first layer in the block.
Just compare:
- OPT structure:
model.decoder.layers.0.self_attn
model.decoder.layers.0.self_attn.k_proj
model.decoder.layers.0.self_attn.v_proj
model.decoder.layers.0.self_attn.q_proj
model.decoder.layers.0.self_attn.out_proj
model.decoder.layers.0.activation_fn
model.decoder.layers.0.self_attn_layer_norm
model.decoder.layers.0.fc1
model.decoder.layers.0.fc2
model.decoder.layers.0.final_layer_norm

- ChatGLM structure:
transformer.encoder.layers.0
**transformer.encoder.layers.0.input_layernorm**
transformer.encoder.layers.0.self_attention
transformer.encoder.layers.0.self_attention.query_key_value
transformer.encoder.layers.0.self_attention.core_attention
transformer.encoder.layers.0.self_attention.core_attention.attention_dropout
transformer.encoder.layers.0.self_attention.dense
transformer.encoder.layers.0.post_attention_layernorm
transformer.encoder.layers.0.mlp
transformer.encoder.layers.0.mlp.dense_h_to_4h
transformer.encoder.layers.0.mlp.dense_4h_to_h

The solution is to disable SA block output caching and collect the quantizing block inputs starting from the beginning of the model. It slows down the optimization a bit but works more stable.

Related PR to Transformers: https://github.com/huggingface/transformers/pull/27032


optimum/gptq/quantizer.py
```diff
@@ -112,6 +112,8 @@ def __init__(
             max_input_length (`Optional[int]`, defaults to `None`):
                 The maximum input length. This is needed to initialize a buffer that depends on the maximum expected input length.
                 It is specific to the exllama backend with act-order.
+            cache_block_outputs (`bool`, defaults to `False`):
```
SunMarc, line 115: You should add the arg in the `init`

optimum/gptq/quantizer.py
```diff
@@ -130,6 +132,7 @@ def __init__(
         self.disable_exllama = disable_exllama
         self.max_input_length = max_input_length
         self.quant_method = QuantizationMethod.GPTQ
+        self.cache_block_outputs = False
```
SunMarc, line 135: use the value from the init

optimum/gptq/quantizer.py
```diff
@@ -340,10 +343,10 @@ def quantize_model(self, model: nn.Module, tokenizer: Any):
 
         def store_input_hook(_, input, *args):
             kwargs = args[0]
-            input = input[0]
+            input = (input[0],) if len(input) < 3 else input
```
SunMarc, line 346: Let's not have the condition and just use the whole input. 
```suggestion

```

optimum/gptq/quantizer.py
```diff
@@ -428,16 +445,22 @@ def tmp(_, input, output):
...
-            # put back to device
-            if not has_device_map:
-                blocks[i] = block.to(device)
-            del layers
-            del layer_inputs
-            layer_inputs, layer_outputs = layer_outputs, []
+            if self.cache_block_outputs:
+                for j in range(len(dataset)):
+                    layer_output = block(*layer_inputs[j], **layer_input_kwargs[j])
+                    layer_output = layer_output
```
SunMarc, line 451: to remove

SunMarc (commented):
Thanks for adding more flexibility to GPTQ  @AlexKoff88 ! This PR looks good. I have some questions regarding `cache_block_outputs` : 
- Did you run some tests to see if we really use less memory ? Because it seems like we are recomputing the inputs from scratch to get the output of the previous block that we get here (L450):  `layer_output = block(*layer_inputs[j], **layer_input_kwargs[j])`. To me, we get the same input in both cases. 
- The issue with the chatglm is that the entire input was required for the forward of the attention block. This was solved by passing *layer_inputs[j] and keeping in the whole input previously. I was able to run chatgml model with `cache_block_outputs=True` and `cache_block_outputs = False `. Here the snippet: 
```py
from transformers import GPTQConfig, AutoTokenizer, AutoModelForCausalLM

model_id = "THUDM/chatglm2-6b"
dataset = ["auto-gptq is an easy-to-use model quantization library with user-friendly apis, based on GPTQ algorithm."]
tokenizer = AutoTokenizer.from_pretrained("THUDM/chatglm2-6b", trust_remote_code=True)
gptq_config = GPTQConfig(bits=4, dataset = dataset, tokenizer=tokenizer, block_name_to_quantize = "transformer.encoder.layers", cache_block_outputs=False)
model = AutoModelForCausalLM.from_pretrained(model_id, quantization_config = gptq_config)
```

optimum/gptq/quantizer.py
AlexKoff88: Agree, thanks

optimum/gptq/quantizer.py
AlexKoff88: fixed, thanks

optimum/gptq/quantizer.py
AlexKoff88: fixed, thanks

AlexKoff88:
> Thanks for adding more flexibility to GPTQ @AlexKoff88 ! This PR looks good. I have some questions regarding `cache_block_outputs` :
> 
> * Did you run some tests to see if we really use less memory ? Because it seems like we are recomputing the inputs from scratch to get the output of the previous block that we get here (L450):  `layer_output = block(*layer_inputs[j], **layer_input_kwargs[j])`. To me, we get the same input in both cases.
> * The issue with the chatglm is that the entire input was required for the forward of the attention block. This was solved by passing *layer_inputs[j] and keeping in the whole input previously. I was able to run chatgml model with `cache_block_outputs=True` and `cache_block_outputs = False `. Here the snippet:
> 
> ```python
> from transformers import GPTQConfig, AutoTokenizer, AutoModelForCausalLM
> 
> model_id = "THUDM/chatglm2-6b"
> dataset = ["auto-gptq is an easy-to-use model quantization library with user-friendly apis, based on GPTQ algorithm."]
> tokenizer = AutoTokenizer.from_pretrained("THUDM/chatglm2-6b", trust_remote_code=True)
> gptq_config = GPTQConfig(bits=4, dataset = dataset, tokenizer=tokenizer, block_name_to_quantize = "transformer.encoder.layers", cache_block_outputs=False)
> model = AutoModelForCausalLM.from_pretrained(model_id, quantization_config = gptq_config)
> ```

Thanks @SunMarc for the prompt review. I get the following error with your code if I don't set `cache_block_outputs=True`:
`TypeError: forward() missing 1 required positional argument: 'rotary_pos_emb'`
This happens when the optimization process goes to the second block. The problem is that the block input is not just an output of the preceding block.

optimum/gptq/quantizer.py
```diff
@@ -112,6 +113,8 @@ def __init__(
             max_input_length (`Optional[int]`, defaults to `None`):
                 The maximum input length. This is needed to initialize a buffer that depends on the maximum expected input length.
                 It is specific to the exllama backend with act-order.
+            cache_block_outputs (`bool`, defaults to `False`):
+                Whether to cache block outputs to reuse as inputs for the succeeding block.
```
SunMarc, line 117: Maybe add a comment saying that it might not work for some models such as chatgml.

optimum/gptq/quantizer.py
```diff
@@ -71,6 +71,7 @@ def __init__(
         pad_token_id: Optional[int] = None,
         disable_exllama: bool = False,
         max_input_length: Optional[int] = None,
+        cache_block_outputs: Optional[bool] = False,
```
SunMarc, line 74: I would prefer to leave it True to not break BC.

optimum/gptq/quantizer.py
```diff
@@ -112,6 +113,8 @@ def __init__(
             max_input_length (`Optional[int]`, defaults to `None`):
                 The maximum input length. This is needed to initialize a buffer that depends on the maximum expected input length.
                 It is specific to the exllama backend with act-order.
+            cache_block_outputs (`bool`, defaults to `False`):
```
SunMarc, line 116: same

optimum/gptq/quantizer.py
SunMarc, line 117: Maybe add a comment saying that it might not work for some models such as chatgml.

optimum/gptq/quantizer.py
AlexKoff88: Fixed

optimum/gptq/quantizer.py
AlexKoff88: Fixed

optimum/gptq/quantizer.py
AlexKoff88: Fixed

AlexKoff88:
> `cache_block_outputs`
@SunMarc, I've added a test

SunMarc:
Awesome @AlexKoff88 ! This should be good to be merged once the conflits are resolved =)

AlexKoff88:
Done

SunMarc (approved):
LGTM ! Thanks again @AlexKoff88 for iterating !

fxmarty (approved):
LGTM, thank you @AlexKoff88

PR #1482, opened 2023-10-24
disable_exllamav2_for_quantization

SunMarc:
# What does this do ? 
This PR removes the possibility to quantize a model with exllamav2 as we don't be save to save the weights as they are somehow reordered.

### DIFF
--- a/optimum/gptq/quantizer.py
+++ b/optimum/gptq/quantizer.py
@@ -457,11 +457,6 @@ class GPTQQuantizer(object):
                         "Found modules on cpu/disk. Using Exllama backend requires all the modules to be on GPU. Setting `disable_exllama=True`"
                     )
                     self.disable_exllama = True
-                if not self.disable_exllamav2:
-                    logger.warning(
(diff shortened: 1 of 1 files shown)


optimum/gptq/quantizer.py
```diff
@@ -457,18 +457,19 @@ def tmp(_, input, output):
...
                     "Using Exllama backend with act_order will reorder the weights offline, thus you will not be able to save the model with the right weights."
                     "Setting `disable_exllama=True`. You should only use Exllama backend with act_order for inference. "
                 )
                 self.disable_exllama = True
+            elif not self.disable_exllamav2:
+                logger.warning(
+                    "Using Exllamav2 backend will reorder the weights offline, thus you will not be able to save the model with the right weights."
+                    "Setting `disable_exllamav2=True`. You should only use Exllamav2 backend for inference. "
+                )
+                self.disable_exllamav2 = True
```
fxmarty, line 472: User tend to ignore warnings so I like to raise errors. But it is just taste, LGTM

fxmarty (approved):
LGTM

PR #1487, opened 2023-10-26
Handling block_size value for models with no sequence length limit.

AdamLouly:
# What does this PR do?
Models with no sequence length limit will fail to create a dataset because the block_size will be -1.
and it will error this:
`ValueError: num_samples should be a positive integer value, but got num_samples=0`

this PR handles the case where max_position_embeddings  is -1, we set block size to a default value.

### DIFF
--- a/examples/onnxruntime/training/language-modeling/run_clm.py
+++ b/examples/onnxruntime/training/language-modeling/run_clm.py
@@ -495,6 +495,8 @@ def main():
 
     if hasattr(config, "max_position_embeddings"):
         max_pos_embeddings = config.max_position_embeddings
+        if max_pos_embeddings < 0:
+            max_pos_embeddings = 1024
     else:
         # Define a default value if the attribute is missing in the config.
         max_pos_embeddings = 1024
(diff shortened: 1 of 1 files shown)


AdamLouly:
@JingyaHuang

examples/onnxruntime/training/language-modeling/run_clm.py
```diff
@@ -506,7 +506,10 @@ def tokenize_function(examples):
                 f"The tokenizer picked seems to have a very large `model_max_length` ({tokenizer.model_max_length}). "
                 f"Using block_size={min(1024, max_pos_embeddings)} instead. You can change that default value by passing --block_size xxx."
             )
+        if max_pos_embeddings > 0:
```
JingyaHuang, line 509: What about checking if `max_pos_embeddings>0` while fetching its value:
https://github.com/huggingface/optimum/blob/6c80244465cc92fd46c0381a12a287ddc8456e3f/examples/onnxruntime/training/language-modeling/run_clm.py#L496-L497
something like
```python
max_pos_embeddings = getattr(config, "max_position_embeddings", 1024)
if max_pos_embeddings < 0:
    max_pos_embeddings = 1024

```

AdamLouly: That also works, I changed it

JingyaHuang (approved):
Looks good to me, thanks again for patching the example for a wider range of models! @AdamLouly

AdamLouly:
Thanks for the review.

PR #1489, opened 2023-10-26
Fix minimum auto-gptq version logic

baskrahmer:
Basically fixes this case:
```
ImportError: Found an incompatible version of auto-gptq. Found version 0.4.2, but only version above 0.4.2 are supported
```

Which I assume to be a bug, but please correct me if I'm wrong!

cc @SunMarc

### DIFF
--- a/optimum/utils/import_utils.py
+++ b/optimum/utils/import_utils.py
@@ -110,7 +110,7 @@ def is_timm_available():
 def is_auto_gptq_available():
     if _auto_gptq_available:
         version_autogptq = packaging.version.parse(importlib_metadata.version("auto_gptq"))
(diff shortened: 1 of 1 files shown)


SunMarc:
Hi @baskrahmer, we merged a PR in optimum that modified the minimum version of auto-gptq (we need 0.5.0dev). We will do a release shortly in autogptq so that you don't have to install from source. There are two ways to solve your issue, either download autogptq from source or you can work on a branch with the[commit](https://github.com/huggingface/optimum/pull/1419) reverted. Sorry that it impacted your work.

baskrahmer:
@SunMarc ah I see. All good, thanks for the context!

PR #1503, opened 2023-11-01
Fix whisper export

mht-sharma:
# What does this PR do?

Fixes whisper export for [https://huggingface.co/sanchit-gandhi/medium-24-2-tpu-timestamped-prob-0.2/](https://huggingface.co/sanchit-gandhi/medium-24-2-tpu-timestamped-prob-0.2/)

## Before submitting


xenova (approved):
Can confirm it worked! I also just tested `openai/whisper-tiny.en`, and that worked too ;)

> (and maybe we can use the whisper-distil in the CI instead of whisper-tiny, if the former is smaller)

whisper-tiny is still smaller :)

PR #1505, opened 2023-11-01
switch to a config_dict instead of disable_exllamav2

SunMarc:
# What does this PR do ?
This PR removes the `disable_exllamav2` arg in favor of `exllama_config` which is more flexible. See related [PR](https://github.com/huggingface/transformers/pull/27111) in Transformers. I will have to merge this PR first before Transformers one. 

TODO: tests to check

### DIFF
--- a/optimum/gptq/quantizer.py
+++ b/optimum/gptq/quantizer.py
@@ -15,6 +15,7 @@
 import copy
 import json
 import os
+from enum import Enum
 from logging import getLogger
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -49,6 +50,11 @@ if is_auto_gptq_available():
 logger = getLogger(__name__)
 
 
+class ExllamaVersion(int, Enum):
+    ONE = 1
+    TWO = 2
+
+
 class GPTQQuantizer(object):
     r"""
     A simple API for GPTQ Quantization
@@ -69,8 +75,8 @@ class GPTQQuantizer(object):
         module_name_preceding_first_block: Optional[List[str]] = None,
         batch_size: int = 1,
         pad_token_id: Optional[int] = None,
-        disable_exllama: Optional[bool] = None,
-        disable_exllamav2: bool = False,
+        disable_exllama: bool = False,
+        exllama_config: Dict[str, Any] = None,
         max_input_length: Optional[int] = None,
         cache_block_outputs: Optional[bool] = True,
         *args,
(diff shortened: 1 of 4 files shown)


optimum/gptq/quantizer.py
```diff
@@ -69,8 +75,8 @@ def __init__(
         module_name_preceding_first_block: Optional[List[str]] = None,
         batch_size: int = 1,
         pad_token_id: Optional[int] = None,
-        disable_exllama: Optional[bool] = None,
-        disable_exllamav2: bool = False,
+        disable_exllama: bool = False,
+        exllama_config: Dict[str, Any] = None,
```
fxmarty, line 79: It is unclear in the docstring which of the two take priority

optimum/gptq/quantizer.py
```diff
@@ -146,16 +152,18 @@ def __init__(
...
+            self.exllama_config = {"version": ExllamaVersion.TWO}
+        else:
+            if "version" not in self.exllama_config:
+                raise ValueError("`exllama_config` needs to have a `version` key")
+            elif self.exllama_config["version"] not in [ExllamaVersion.ONE, ExllamaVersion.TWO]:
+                version = self.exllama_config["version"]
+                raise ValueError(
+                    f"Only supported versions are in [ExllamaVersion.ONE, ExllamaVersion.TWO] - not recognized version {version}"
+                )
+        self.exllama_version = self.exllama_config["version"]
```
fxmarty, line 166: what if I want to use the cuda/cuda-old kernel?

optimum/gptq/quantizer.py
```diff
@@ -627,8 +647,8 @@ def load_quantized_model(
     offload_folder: Optional[str] = None,
     offload_buffers: Optional[str] = None,
     offload_state_dict: bool = False,
-    disable_exllama: Optional[bool] = None,
-    disable_exllamav2: bool = False,
+    disable_exllama: bool = False,
+    exllama_config: Dict[str, Any] = None,
```
fxmarty, line 651:
```suggestion
    exllama_config: Optional[Dict[str, Any]] = None,
```

maybe elsewhere

optimum/gptq/quantizer.py
```diff
@@ -664,8 +684,8 @@ def load_quantized_model(
             picked contains `"disk"` values.
         disable_exllama (`Optional[bool]`, defaults to `None`):
             Whether to use exllama backend. Only works with `bits` = 4.
-        disable_exllama (`bool`, defaults to `False`):
-            Whether to use exllamav2 backend. Only works with `bits` = 4.
+        exllama_config (`Dict[str, Any]`, *optional*):
```
fxmarty, line 687:
```suggestion
        exllama_config (`Optional[Dict[str, Any]]`, defaults to `None`):
```

optimum/gptq/quantizer.py
```diff
@@ -664,8 +684,8 @@ def load_quantized_model(
...
+            The exllama config. You can specify the version of the exllama kernel through the `version` key. Defaults to `{"version": 2}` if unset.
```
fxmarty, line 688: How do we select cuda/cuda-old kernel?

SunMarc: You set `disable_exllama` = True.

optimum/gptq/quantizer.py
SunMarc: exllama_config is just the config that is used when `disable_exllama = False`. I will add more info in the docstring.

optimum/gptq/quantizer.py
fxmarty: Oh I see, so the `disable_exllama` is disable exllama altogether (v1/v2?)?

SunMarc: Yes !

fxmarty: Makes sense!

fxmarty (approved):
LGTM, feel free to merge once the `disable_exllama` docstring & type annotations are fixed!

PR #1511, opened 2023-11-02
Deprecate ORTDecoder

echarlaix:


optimum/onnxruntime/base.py
```diff
@@ -145,363 +146,13 @@ def __init__(
...
-                and "sequence_length" in self.expected_key_symbolic_shape[-1]
-            ):
-                self.key_sequence_length_idx = -1
-
-            self.value_sequence_length_idx = -2
-            if (
-                isinstance(self.expected_value_symbolic_shape[-1], str)
-                and "sequence_length" in self.expected_value_symbolic_shape[-1]
-            ):
-                self.value_sequence_length_idx = -1
```
fxmarty, line 197: this is not useful for seq2seq models?

fxmarty (approved):
LGTM - ideally we should make sure tests pass with IO Binding as well but I don't see why they wouldn't

optimum/onnxruntime/base.py
echarlaix: it's only used for computing the pkv shapes for decoder only models : https://github.com/huggingface/optimum/blob/v1.13.2/optimum/onnxruntime/base.py#L299 (not used by encoder/decoder models https://github.com/huggingface/optimum/blob/v1.13.2/optimum/onnxruntime/base.py#L504)

PR #1515, opened 2023-11-03
fix generation input for decoder

echarlaix:


fxmarty (approved):
LGTM - why did we change to the more verbose version? Is it used in transformers?

PR #1539, opened 2023-11-15
Refactor NormalizedConfigs for GQA

michaelbenayoun:
# What does this PR do?

It creates a `NormalizedTextConfigWithGQA` class that inherints `NormalizedTextConfig` to handle models that use GQA instead of vanilla MHA.


michaelbenayoun:
No, not really. But the changes for `Mistral` are also useful to `Llama` because it can use GQA as well.

echarlaix (approved):

PR #1585, opened 2023-12-11
add `modules_in_block_to_quantize` arg for gptq

SunMarc:
# What does this PR do?
This PR adds the `inside_layer_modules` arg for GPTQ quantization. This will enable the user to quantize specific modules inside a block. This is needed for quantizing model and run quantized model such as https://huggingface.co/TheBloke/Mixtral-8x7B-v0.1-GPTQ

### DIFF
--- a/optimum/gptq/quantizer.py
+++ b/optimum/gptq/quantizer.py
@@ -221,7 +221,7 @@ class GPTQQuantizer(object):
             layers_to_keep = sum(self.modules_in_block_to_quantize, [])
             for name in list(layers_to_be_replaced.keys()):
                 if not any(name.endswith(layer) for layer in layers_to_keep):
-                    logger.info(f"Quantization disabled for {name} (only modules_in_block_to_quantize={modules_in_block_to_quantize} are quantized)")
+                    logger.info(f"Quantization disabled for {name} (only modules_in_block_to_quantize={self.modules_in_block_to_quantize} are quantized)")
                     del layers_to_be_replaced[name]
         self._replace_by_quant_layers(model, layers_to_be_replaced)
         return model
--- a/optimum/gptq/quantizer.py
+++ b/optimum/gptq/quantizer.py
@@ -77,6 +77,7 @@ class GPTQQuantizer(object):
         exllama_config: Dict[str, Any] = None,
         max_input_length: Optional[int] = None,
         cache_block_outputs: Optional[bool] = True,
+        inside_layer_modules: Optional[list] = None,
         *args,
         **kwargs,
     ):
@@ -123,6 +124,8 @@ class GPTQQuantizer(object):
             cache_block_outputs (`bool`, defaults to `True`):
                 Whether to cache block outputs to reuse as inputs for the succeeding block. It allows optimization of non-standard models
                 (e.g. ChatGLM) but can require more time.
+            inside_layer_modules (`List`, *optional*, defaults to `None`):
+                List of module names to quantize inside block_name_to_quantize. If not set, we will quantize all the linear layers.
         """
 
         self.bits = bits
@@ -143,6 +146,7 @@ class GPTQQuantizer(object):
         self.max_input_length = max_input_length
         self.quant_method = QuantizationMethod.GPTQ
         self.cache_block_outputs = cache_block_outputs
+        self.inside_layer_modules = inside_layer_modules
 
         self.serialization_keys = [
             "bits",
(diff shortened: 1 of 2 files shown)


optimum/gptq/quantizer.py
```diff
@@ -123,6 +124,8 @@ def __init__(
             cache_block_outputs (`bool`, defaults to `True`):
                 Whether to cache block outputs to reuse as inputs for the succeeding block. It allows optimization of non-standard models
                 (e.g. ChatGLM) but can require more time.
+            inside_layer_modules (`List[List[str]]`, *optional*, defaults to `None`):
```
fxmarty, line 127:
```suggestion
            inside_layer_modules (`Optional[List[List[str]]]`, defaults to `None`):
```

optimum/gptq/quantizer.py
```diff
@@ -123,6 +124,8 @@ def __init__(
...
+                List of module names to quantize inside block_name_to_quantize. If not set, we will quantize all the linear layers.
```
fxmarty, line 128:
```suggestion
                List of module names to quantize inside block_name_to_quantize. This argument is useful to exclude certain linear modules from being quantized. If `None`, we will quantize all the linear layers. Example: `inside_layer_modules=[["self_attention.query_key_value"], ["mlp.dense_h_to_4h"]]`
```

fxmarty: Why can't we just use `block_name_to_quantize`?

optimum/gptq/quantizer.py
```diff
@@ -77,6 +77,7 @@ def __init__(
         exllama_config: Dict[str, Any] = None,
         max_input_length: Optional[int] = None,
         cache_block_outputs: Optional[bool] = True,
+        inside_layer_modules: Optional[List[List[str]]] = None,
```
fxmarty, line 80: Can we have a more explicit name like `modules_in_block_to_quantize`? Why is it a `List[List[str]]` and not list `List[str]`?

tests/gptq/test_quantization.py
```diff
@@ -300,6 +303,16 @@ class GPTQTestNoBlockCaching(GPTQTest):
     EXPECTED_OUTPUTS.add("Hello my name is John, I am a student in the University of")
 
 
+class GPTQTestInsideLayerModules(GPTQTest):
+    # all layers are quantized apart from self_attention.dense
+    inside_layer_modules = [["self_attention.query_key_value"], ["mlp.dense_h_to_4h"], ["mlp.dense_4h_to_h"]]
```
fxmarty, line 308: Couldn't it be
```suggestion
    inside_layer_modules = ["self_attention.query_key_value", "mlp.dense_h_to_4h", "mlp.dense_4h_to_h"]
```

optimum/gptq/quantizer.py
```diff
@@ -210,8 +215,13 @@ def convert_model(self, model: nn.Module):
             self.block_name_to_quantize = get_block_name_with_pattern(model)
         block_name = self.block_name_to_quantize
         layers_to_be_replaced = get_layers(model, prefix=block_name)
+        if self.inside_layer_modules is not None:
+            layers_to_keep = sum(self.inside_layer_modules, [])
```
fxmarty, line 219: I don't understand this

optimum/gptq/quantizer.py
```diff
@@ -210,8 +215,13 @@ def convert_model(self, model: nn.Module):
...
+            for name in list(layers_to_be_replaced.keys()):
+                if not any(name.endswith(layer) for layer in layers_to_keep):
+                    logger.info(f"{name} has not been quantized. We don't convert it")
```
fxmarty, line 222:
```suggestion
                    logger.info(f"Quantization disabled for {name} (only modules_in_block_to_quantize={modules_in_block_to_quantize} are quantized)")
```

tests/gptq/test_quantization.py
SunMarc: no since we want to quantize sequentially the layers. First update `self_attention.query_key_value`, then `mlp.dense_h_to_4h` and finally `mlp.dense_4h_to_h`. Using a list is useful when we want for example to update the q,k,v layers at the same time since they are independent. However, for `mlp.dense_h_to_4h`, we want to have q,k,v quantized before.

optimum/gptq/quantizer.py
SunMarc: We put everything in a single list as we have a list of list.

optimum/gptq/quantizer.py
SunMarc: related to the quantization process. See comment below

optimum/gptq/quantizer.py
SunMarc: Yeah this will be a lot better compared to using the same one as AutoGPTQ

tests/gptq/test_quantization.py
```diff
@@ -142,6 +144,7 @@ def test_serialization(self):
         with tempfile.TemporaryDirectory() as tmpdirname:
             self.quantizer.save(self.quantized_model, tmpdirname)
             self.quantized_model.config.save_pretrained(tmpdirname)
+            print(self.quantized_model.config)
```
fxmarty, line 147: to remove

fxmarty: I happen to not have maintainer rights anymore so can't push myself unfortunately

SunMarc: mb thx !

fxmarty (approved):
LGTM thanks!

PR #1587, opened 2023-12-12
Add timm models export

mht-sharma:
# What does this PR do?

Add export for large number of Timm models 

* Introduced DefaultTimmConfig to facilitate Timm model exports.
     - The Timm configuration entails the architecture, synonymous with model_type in transformers. However, the architecture comprises multiple variations for input_size, hidden_layers, etc.
     - This results in numerous architecture types for a single model, making it impractical to incorporate all in the `TaskManager`.
     - The addition of `DefaultTimmConfig` streamlines support for all Timm models. Specific cases can still be added similarly to other exports.
     - This design works best for this library, and changes do not affect other exports
* This repository now accommodates over 1000 models available on the https://huggingface.co/timm page.

## Before submitting


mht-sharma:
Question: For TImm tests, should I add all the supported models for testing, which would be a large number.

optimum/exporters/onnx/convert.py
```diff
@@ -267,7 +267,6 @@ def _run_validation(
     if input_shapes is None:
         input_shapes = {}  # will use the defaults from DEFAULT_DUMMY_SHAPES
     reference_model_inputs = config.generate_dummy_inputs(framework=framework, **input_shapes)
-    reference_model_inputs = config.rename_ambiguous_inputs(reference_model_inputs)
```
fxmarty, line 270: Why is this moved?

optimum/exporters/tasks.py
```diff
@@ -280,6 +280,10 @@ class TasksManager:
         "visual-question-answering",
     )
 
+    _MODEL_TYPE_FOR_DEFAULT_CONFIG = {
+        "timm": "default-timm-config",
+    }
```
fxmarty, line 285: Could we rathe remove this and use `model.config.model_type = "timm-model"`?

optimum/exporters/tasks.py
```diff
@@ -815,12 +819,7 @@ class TasksManager:
         "resnet": supported_tasks_mapping(
             "feature-extraction", "image-classification", onnx="ResNetOnnxConfig", tflite="ResNetTFLiteConfig"
         ),
-        "resnext26ts": supported_tasks_mapping("image-classification", onnx="TimmResNextOnnxConfig"),
-        "resnext50-32x4d": supported_tasks_mapping("image-classification", onnx="TimmResNextOnnxConfig"),
-        "resnext50d-32x4d": supported_tasks_mapping("image-classification", onnx="TimmResNext50d_32x4dOnnxConfig"),
-        "resnext101-32x4d": supported_tasks_mapping("image-classification", onnx="TimmResNextOnnxConfig"),
-        "resnext101-32x8d": supported_tasks_mapping("image-classification", onnx="TimmResNextOnnxConfig"),
-        "resnext101-64x4d": supported_tasks_mapping("image-classification", onnx="TimmResNextOnnxConfig"),
+        "default-timm-config": supported_tasks_mapping("image-classification", onnx="TimmDefaultOnnxConfig"),
```
fxmarty, line 822:
```suggestion
        "timm-model": supported_tasks_mapping("image-classification", onnx="TimmDefaultOnnxConfig"),
```

tests/exporters/exporters_utils.py
```diff
@@ -286,6 +286,7 @@
 }
 
 PYTORCH_TIMM_MODEL = {
+    "default-timm-config": "timm/resnext26ts.ra2_in1k",
```
fxmarty, line 289: why not remove the `resnext` here?

optimum/exporters/tasks.py
```diff
@@ -1077,7 +1076,7 @@ def decorator(config_cls: Type) -> Type:
 
     @staticmethod
     def get_supported_tasks_for_model_type(
-        model_type: str, exporter: str, model_name: Optional[str] = None
+        model_type: str, exporter: str, model_name: Optional[str] = None, library_name: str = "transformers"
```
fxmarty, line 1079: Could you detail why we need to add this argument? Can't we make away with just modifying `model_type`?

mht-sharma: This I added just to ensure that in future if there are some edge cases then we can handle them by adding a their model_type. By fixing it to 'timm-model' we may not have that flexibility.

optimum/exporters/onnx/convert.py
mht-sharma: Only need to rename input here for timm model and not onnx model

optimum/exporters/tasks.py
fxmarty: Well you just use `default-timm-config` in the mapping anyway? I'd be in favor of not adding complexity if it is not required for now.

PR #1590, opened 2023-12-12
Compatibility with Transformers 4.36

fxmarty:
As per title.

* Support SDPA ONNX export.
* Fix Whisper that now uses in `generate` attributes as `self.model` that are not accessible to us.
* Fix Falcon following the KV cache refactor (now 4D instead of 3D before).

Failing XLM and sew_d tests are not due to us, see https://huggingface.slack.com/archives/C014N4749J9/p1702463890741009?thread_ts=1702373869.327359&cid=C014N4749J9 & https://github.com/huggingface/transformers/issues/28000


optimum/exporters/onnx/__main__.py
```diff
@@ -353,6 +357,10 @@ def main_export(
                 f"Asked to export a {model_type} model for the task {task}{autodetected_message}, but the Optimum ONNX exporter only supports the tasks {', '.join(model_tasks.keys())} for {model_type}. Please use a supported task. Please open an issue at https://github.com/huggingface/optimum/issues if you would like the task {task} to be supported in the ONNX export for {model_type}."
             )
 
+        # TODO: Fix in Transformers so that SdpaAttention class can be exported to ONNX. `attn_implementation` is introduced in Transformers 4.36.
+        if model_type in SDPA_ARCHS_ONNX_EXPORT_NOT_SUPPORTED and _transformers_version >= version.parse("4.35.99"):
+            loading_kwargs["attn_implementation"] = "eager"
```
fxmarty, line 362: Bart decoder and Whisper encoder + decoder should be fixable since the query length is static there (so no need to actually have a controlflow on it).

And for bart encoder, we are always non-causal no matter q_len so we could just rely on that.

PR #1611, opened 2023-12-22
Add ability to specify a LoRA to fuse with Stable Diffusion models before export

gfodor:
# What does this PR do?

Adds the --fuse_lora command which can optionally fuse a specified LoRA with the ONNX model for Stable Diffusion. This allows easily exporting fine tuned versions of the model.

### DIFF
--- a/optimum/commands/export/onnx.py
+++ b/optimum/commands/export/onnx.py
@@ -150,6 +150,12 @@ def parse_args_onnx(parser):
             "Also disable the use of position_ids for text-generation models that require it for batched generation. This argument is introduced for backward compatibility and will be removed in a future release of Optimum."
         ),
     )
+    optional_group.add_argument(
+        "--fuse_lora",
+        type=str,
+        default=None,
+        help=("For Stable Diffusion models only. LoRA to fuse before export."),
+    )
 
     input_group = parser.add_argument_group(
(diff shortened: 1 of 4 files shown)


optimum/commands/export/onnx.py
```diff
@@ -150,6 +150,12 @@ def parse_args_onnx(parser):
             "Also disable the use of position_ids for text-generation models that require it for batched generation. This argument is introduced for backward compatibility and will be removed in a future release of Optimum."
         ),
     )
+    optional_group.add_argument(
+        "--fuse_lora",
+        type=str,
+        default=None,
+        help=("For Stable Diffusion models only. LoRA to fuse before export."),
```
fxmarty, line 157: Could you precise what kind of lora checkpoints are supported, what needs to be passed here? A link to diffusers/whatever documentation would be nice if there exists one.

fxmarty (commented):
@gfodor thank you for the addition, sorry for the late reply! Looks good overall, could you add a test for this?

optimum/commands/export/onnx.py
shadowofsoul: I'd like to point that i have been able to use this patch to fuse LCM into SD and then exporting it to ONNX. tested it using `pipe LCMScheduler` loading the individual ONNX models and works. in that way, now i can rewrite LCMScheduler for ONNX and have 4 steps ineference on ONNX

PR #1624, opened 2024-01-04
Rework doc homepage

regisss:
# What does this PR do?

This PR introduces a new layout for the doc homepage. It splits subpackages into two sections (hardware partners and open-source integrations). This gives a clearer view of what the various subpackages provide.

## Before submitting


docs/source/index.mdx
```diff
@@ -19,11 +19,20 @@ As such, Optimum enables developers to efficiently use any of these platforms wi
 
 🤗 Optimum is distributed as a collection of packages - check out the links below for an in-depth look at each one.
 
+
+## Hardware partners
+
+The packages below enable you to get the best of the 🤗 Hugging Face ecosystem on various types of devices.
+
```
echarlaix, line 26: I like the idea of having an "hardware" section, the only thing that could be unclear would be for packages that also enable hardware agnostic optimization (like INC for example), but this can be specified in the description or even in a specific section, will think of something!

docs/source/index.mdx
```diff
@@ -33,25 +42,39 @@ As such, Optimum enables developers to efficiently use any of these platforms wi
...
       ><div class="w-full text-center bg-gradient-to-br from-yellow-400 to-yellow-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">BetterTransformer</div>
       <p class="text-gray-700">A one-liner integration to use <span class="underline" onclick="event.preventDefault(); window.open('https://pytorch.org/blog/a-better-transformer-for-fast-transformer-encoder-inference/', '_blank');">PyTorch's BetterTransformer</span> with Transformers models</p>
     </a>
+    <a class="!no-underline border dark:border-gray-700 p-5 rounded-lg shadow hover:shadow-lg" href="./torch_fx/overview"
+      ><div class="w-full text-center bg-gradient-to-br from-green-400 to-green-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">Torch FX</div>
+      <p class="text-gray-700">Create and compose custom graph transformations to optimize PyTorch Transformers models with <span class="underline" onclick="event.preventDefault(); window.open('https://pytorch.org/docs/stable/fx.html#', '_blank');">Torch FX</span></p>
+    </a>
   </div>
 </div>
```
echarlaix, line 80: Not related to this PR but the whole color layout is burning my eyes, will open a new PR to propose something (not that it'll be any better lol)

docs/source/index.mdx
```diff
@@ -33,25 +42,39 @@ As such, Optimum enables developers to efficiently use any of these platforms wi
...
     </a>
     <a class="!no-underline border dark:border-gray-700 p-5 rounded-lg shadow hover:shadow-lg" href="./furiosa/index"
       ><div class="w-full text-center bg-gradient-to-br from-green-400 to-green-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">FuriosaAI</div>
       <p class="text-gray-700">Fast and efficient inference on <span class="underline" onclick="event.preventDefault(); window.open('https://www.furiosa.ai/', '_blank');">FuriosaAI WARBOY</span></p>
     </a>
+  </div>
+</div>
+
+
+## Open-source integrations
```
echarlaix, line 57: Not sure about "Open-source" as it's also true for the previous section, but would go with something like "hardware-agnostic"

echarlaix (approved):
Looks good, thanks @regisss !

docs/source/index.mdx
regisss: Yes, this can be added to the small description in the tiles, or as a note below the tiles such as:

> Some packages provide hardware-agnostic features (e.g. INC interface in Optimum Intel)

docs/source/index.mdx
regisss: I meant that for instance Furiosa's or Habana's SDKs are not open-source but they are integrated into Optimum, unlike ORT, FX etc.
The idea I would like to convey here is that HPPs are not the only thing we do and that we also work on the integration of other popular libraries into Optimum. Maybe there is a better way to say it?

PR #1627, opened 2024-01-08
Remove "to ONNX" from info message when exporting model

helena-intel:
This message is also shown when exporting models to OpenVINO format (in optimum-intel), which does not use ONNX. Replaced "Using pt to export to ONNX" with "Using pt to export the model".

@echarlaix


echarlaix (approved):
LGTM, thanks @helena-intel

PR #1640, opened 2024-01-10
[ORT] Improve dummy mask & add tips for attention fusion in the doc

JingyaHuang:
# What does this PR do?

- [x] Improve the creation of dummy mask tensors (to avoid accuracy off issue)
- [x] Add tips for ORT attention fusion

More context here: https://github.com/microsoft/onnxruntime/issues/18945


optimum/utils/input_generators.py
```diff
@@ -364,13 +416,23 @@ def __init__(
...
             self.num_choices = num_choices
 
-    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+    def generate(
+        self,
+        input_name: str,
+        framework: str = "pt",
+        int_dtype: str = "int64",
+        float_dtype: str = "fp32",
+        padding_side: str = "right",
```
fxmarty, line 425: I'd rather have this `padding_side` in the init in order to keep the same signature for all the `generate` methods, but it does not matter too much

fxmarty (approved):
LGTM

PR #1641, opened 2024-01-11
Bump min torch version

echarlaix:
As done in transformers https://github.com/huggingface/transformers/pull/28207 (also fixes an import error as `is_torch_less_than_1_11` has been removed in #28207)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ except Exception as error:
 REQUIRED_PKGS = [
     "coloredlogs",
     "sympy",
-    "transformers @ git+https://github.com/huggingface/transformers.git",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -15,8 +15,9 @@
 REQUIRED_PKGS = [
     "coloredlogs",
     "sympy",
-    "transformers[sentencepiece]>=4.26.0",
-    "torch>=1.9",
+    "transformers[sentencepiece] @ git+https://github.com/huggingface/transformers.git",
+    # "transformers[sentencepiece]>=4.26.0",
```
echarlaix, line 19: (just temporary for the tests, will be removed)

michaelbenayoun (approved):

PR #1643, opened 2024-01-12
Enable export of model with fixed shape

mht-sharma:
# What does this PR do?

Enable export of the models with fixed shapes.

This is required to support export of model like "vovnet" which cannot be exported with dynamic shapes.

```bash
 optimum-cli export onnx --model timm/ese_vovnet39b.ra_in1k  out_vov --no-dynamic-axes
```

## Before submitting
- [X] Did you make sure to update the documentation with your changes?

### DIFF
--- a/optimum/exporters/onnx/__main__.py
+++ b/optimum/exporters/onnx/__main__.py
@@ -190,6 +190,7 @@ def main_export(
     _variant: str = "default",
     library_name: Optional[str] = None,
     legacy: bool = False,
+    no_dynamic_axes: bool = False,
     **kwargs_shapes,
 ):
     """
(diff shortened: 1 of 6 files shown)


optimum/exporters/onnx/__main__.py
```diff
@@ -444,6 +447,11 @@ def main_export(
             f" `--task {task_non_past} --monolith`, or `--task {task}` without the monolith argument."
         )
 
+    if library_name == "timm" and no_dynamic_axes is False and "vovnet" in model_type:
+        raise ValueError(
+            f"The export of {model_type} is not supported with dynamic axes. Please pass --no-dynamic-axes to export the model with fixed shapes"
+        )
```
fxmarty, line 453: I wonder rather avoid adding more model-specific code in here. Is it possible to move it elsewhere (like `model_configs.py`)?

optimum/exporters/onnx/__main__.py
```diff
@@ -444,6 +447,11 @@ def main_export(
             f" `--task {task_non_past} --monolith`, or `--task {task}` without the monolith argument."
         )
 
+    if library_name == "timm" and no_dynamic_axes is False and "vovnet" in model_type:
```
michaelbenayoun, line 450:
```suggestion
    if library_name == "timm" and not no_dynamic_axes and "vovnet" in model_type:
```

michaelbenayoun (commented):
Can you write a test to test that when specifying shapes for the dummy inputs with no dynamic axes requested, the proper shapes are used in the exported model?

optimum/exporters/onnx/__main__.py
mht-sharma: This could not be possible as the variable (`no_dynamic_axes`) is not under the scope of model_configs.py. 

For now I have removed this from here. In documentation, it is mentioned that the following architecture support is limited.

optimum/exporters/onnx/__main__.py
mht-sharma: Added a small test, testing only a few models from transformers and vovnet from timm.

fxmarty (approved):
LGTM

PR #1645, opened 2024-01-12
[Sentence Transformers] Complete the update of `model_type` 

JingyaHuang:
# What does this PR do?

Improve the update of `model_type`, with the previous update, the `model_type` remains as the previous one (eg. `bert`) when displaying the config (which is relatively ok):
image

But if we apply [`config.to_dict()`](https://github.com/huggingface/transformers/blob/c48787f347bd604f656c2cfff730e029c8f8c1fe/src/transformers/configuration_utils.py#L893), the `model_type` is not changed, so I would suggest to complete the change of `model_type` thoroughly.

### DIFF
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -1682,9 +1682,11 @@ class TasksManager:
             if "Transformer" in model[0].__class__.__name__:
                 model.config = model[0].auto_model.config
                 model.config.model_type = "sentence-transformers-transformer"
+                model.config.__class__.model_type = "sentence-transformers-transformer"
             elif "CLIP" in model[0].__class__.__name__:
                 model.config = model[0].model.config
                 model.config.model_type = "sentence-transformers-clip"
+                model.config.__class__.model_type = "sentence-transformers-clip"
             else:
                 raise ValueError(
                     f"The export of a sentence-transformers model with the first module being {model[0].__class__.__name__} is currently not supported in Optimum. Please open an issue or submit a PR to add the support."


optimum/exporters/tasks.py
```diff
@@ -1682,9 +1682,11 @@ def standardize_model_attributes(
             if "Transformer" in model[0].__class__.__name__:
                 model.config = model[0].auto_model.config
                 model.config.model_type = "sentence-transformers-transformer"
+                model.config.__class__.model_type = "sentence-transformers-transformer"
```
fxmarty, line 1685:
```suggestion
                model.config = PretrainedConfig.from_dict(model[0].auto_model.config.to_dict())
                model.config.model_type = "sentence-transformers-transformer"
```

optimum/exporters/tasks.py
```diff
@@ -1682,9 +1682,11 @@ def standardize_model_attributes(
...
             elif "CLIP" in model[0].__class__.__name__:
                 model.config = model[0].model.config
                 model.config.model_type = "sentence-transformers-clip"
+                model.config.__class__.model_type = "sentence-transformers-clip"
```
fxmarty, line 1689:
```suggestion
                model.config = PretrainedConfig.from_dict(model[0].model.config.to_dict())
                model.config.model_type = "sentence-transformers-clip"
```

fxmarty (commented):
Thank you for the notice, would this suggestion work?

optimum/exporters/tasks.py
JingyaHuang: nope:
image

fxmarty:
Right, the issue comes from https://github.com/huggingface/transformers/blob/881e966aced6f0f208f43d7b7e7e55bc680f8fa5/src/transformers/configuration_utils.py#L901-L902. This sounds like a bug in Transformers to me, `self.__class__` is for some reason taking priority over `self`.

In the meantime we rather fix with:

```python
class SentenceTransformersTransformerPretrainedConfig(PretrainedConfig):
    model_type = "sentence-transformers-transformer"

class SentenceTransformersClipPretrainedConfig(PretrainedConfig):
    model_type = "sentence-transformers-clip"
```

to avoid modifying class attributes?

PR #1652, opened 2024-01-18
Remove abstract decorator from `_export`

JingyaHuang:
It breaks subclasses which inherit the class without defining `_export`, I would suggest we remove it at least for now.

### DIFF
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -394,7 +394,7 @@ class OptimizedModel(PreTrainedModel):
         elif export and trust_remote_code is None:
             trust_remote_code = False
 
-        from_pretrained_method = cls._export if export else cls._from_pretrained
+        from_pretrained_method = cls._from_transformers if export else cls._from_pretrained
 
         return from_pretrained_method(
             model_id=model_id,
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -291,7 +291,6 @@ class OptimizedModel(PreTrainedModel):
         )
 
     @classmethod
-    @abstractmethod
     def _export(
         cls,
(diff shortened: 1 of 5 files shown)


mht-sharma (approved):
Thanks @JingyaHuang, LGTM

JingyaHuang:
I did another commit that I could probably revert, shall we migrate all existing `_from_transformers` to `_export`? @fxmarty @mht-sharma @echarlaix

optimum/modeling_base.py
```diff
@@ -395,7 +394,7 @@ def from_pretrained(
         elif export and trust_remote_code is None:
             trust_remote_code = False
 
-        from_pretrained_method = cls._from_transformers if export else cls._from_pretrained
+        from_pretrained_method = cls._export if export else cls._from_pretrained
```
fxmarty, line 397: This assumes that `_export` has been implemented

JingyaHuang: yeah, ifwe merge this, we might need to do a sync release for optimum and all sub-packages. Everyone changes.

fxmarty: okay sounds good. so at this rate we could leave the `@abstractmethod`?

JingyaHuang:
Never mind let's discuss that in another PR.

echarlaix (approved):

optimum/modeling_base.py
echarlaix: also the case for optimum-furiosa https://github.com/huggingface/optimum-furiosa/blob/bc77f8408cc5b7ceb66044bd6bc30c38fbf2c568/optimum/furiosa/modeling_base.py#L195 cc @mht-sharma

PR #1653, opened 2024-01-19
Switch back to `pull_request` event for PR doc build

regisss:
# What does this PR do?

This PR triggers the PR doc build for the `pull_request` event instead of `pull_request_target`. `pull_request_target` was used so that the secrets needed to build Furiosa's doc can be accessed from forks (after authorization). However, this event always triggers the associated workflow from main in PRs, which is not convenient when modifying this specific workflow because we cannot see the changes.
As a workaround, this PR switches back to `pull_request` and the doc of Furiosa is not built for PRs.
Needed for #1624.

## Before submitting

### DIFF
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -4,9 +4,8 @@ name: Build PR documentation
 
 on:
   workflow_dispatch:
-  pull_request_target:
+  pull_request:
     branches: [ main ]
-    types: [ opened, synchronize, reopened, labeled ]
     paths:
       - "optimum/**.py"
       - "docs/**.mdx"
@@ -17,14 +16,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  authorize:
-    if: (github.event.action == 'labeled' && github.event.label.name == 'build-pr-doc') || github.event_name != 'pull_request_target' || (! github.event.pull_request.head.repo.fork)
-    runs-on: ubuntu-latest
-    steps:
-      - run: true
-
   build_documentation:
-    needs: authorize
     runs-on: ubuntu-latest
     env:
       COMMIT_SHA: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build_pr_documentation.yml
```diff
@@ -42,7 +34,6 @@ jobs:
         with:
           repository: 'huggingface/optimum'
           path: optimum
-          ref: ${{ github.event.pull_request.merge_commit_sha }}
```
fxmarty, line 45: do you mean that this SHA is main latest commit SHA instead of a potential merge commit sha?

regisss: No, by default the ref will point to the commit SHA of the event that triggered the workflow

.github/workflows/build_pr_documentation.yml
```diff
@@ -4,9 +4,8 @@ name: Build PR documentation
 
 on:
   workflow_dispatch:
```
echarlaix, line 6: could be removed no ?
```suggestion
```

.github/workflows/build_pr_documentation.yml
```diff
@@ -4,9 +4,8 @@ name: Build PR documentation
```
echarlaix, line 4: would also remove the warning comment

echarlaix (approved):
Looks good, thanks @regisss

.github/workflows/build_pr_documentation.yml
regisss: I think it's good to be able to trigger the PR doc build on demand so I would rather keep it

.github/workflows/build_pr_documentation.yml
regisss: Done

.github/workflows/build_pr_documentation.yml
fxmarty: I don't understand. `github.event.pull_request.merge_commit_sha` should be the merge commit from a PR, is that not? Why is the documentation built against main?

regisss: I don't know the details, but for sure it doesn't take into account the current PR changes to the workflow. See this PR for many examples: https://github.com/huggingface/optimum/pull/1624

PR #1662, opened 2024-01-22
Fix gptq device_map = "cpu"

SunMarc:
# What does this do ? 
This PR fixes the case where the user passes device_map = "cpu". In that case, we still need to move each block to the gpu 0 on our own since we didn't add hooks. 
Fixes [28632](https://github.com/huggingface/transformers/issues/28632)

### DIFF
--- a/tests/gptq/test_quantization.py
+++ b/tests/gptq/test_quantization.py
@@ -54,7 +54,7 @@ class GPTQTest(unittest.TestCase):
     exllama_config = None
     cache_block_outputs = True
     modules_to_quantize_inside_block = None
-
+    device_map_for_quantization = {"": 0}
     dataset = [
         "auto-gptq is an easy-to-use model quantization library with user-friendly apis, based on GPTQ algorithm."
     ]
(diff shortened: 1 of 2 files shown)


tests/gptq/test_quantization.py
```diff
@@ -54,7 +54,7 @@ class GPTQTest(unittest.TestCase):
     exllama_config = None
     cache_block_outputs = True
     modules_to_quantize_inside_block = None
-
+    device_map_for_quantization = {"": 0}
```
fxmarty, line 57: Setting default dictionary is dangerous: https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

fxmarty (approved):
> In that case, we still need to move each block to the gpu 0 on our own since we didn't add hooks.

Do you mean that accelerate does not add hooks in case we have a single device in device_map?

What if the user does not have a GPU?

SunMarc:
> Do you mean that accelerate does not add hooks in case we have a single device in device_map?
Yes, that's right. We added initially them, but we removed them since it was confusing for most users. 

> What if the user does not have a GPU?
You need a GPU to quantize the model. It will trigger an error

tests/gptq/test_quantization.py
SunMarc: wow, thanks for the advice.

Issue #1485, opened 2023-10-25
Error while converting bart-large-mnli to ONNX format


remusao:
### System Info

```shell
Python 3.10.12
optimum[exporters]==1.13.2
Linux + Nvidia GPU with driver 525.85.12 (Although the issue seems independent from GPU)
```

### Who can help?

@michaelbenayoun (Tagging you since the issue template suggests it for exports to ONNX)

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction (minimal, reproducible, runnable)

Install optimum:
```sh
$ pip install optimum[exporters]==1.13.2
```

Then try to convert `facebook/bart-large-mnli` to ONNX format:
```sh
$ optimum-cli export onnx --model facebook/bart-large-mnli onnx_output
```

I also tried the following commands which all fail in a similar way:
```sh
$ python3 -m transformers.onnx --model=facebook/bart-large-mnli --feature=sequence-classification ./onnx_output
```

Or,
```sh
$ optimum-cli export onnx \
    --model=facebook/bart-large-mnli \
    --task=text-classification \
    --device cuda \
    --opset 17 \
    --framework pt \
    ./onnx_output
```

I also tried to use opset 12, and remove the safetensor model to force conversion from torch tensors (not sure if that should make any difference).

The output is then:
```
Framework not specified. Using pt to export to ONNX.
Downloading (…)lve/main/config.json: 100%|████████████████████████████████████████████████████████████████████████████████| 1.15k/1.15k [00:00<00:00, 10.3MB/s]
Downloading model.safetensors: 100%|███████████████████████████████████████████████████████████████████████████████████████| 1.63G/1.63G [00:06<00:00, 248MB/s]
Automatic task detection to text-classification (possible synonyms are: sequence-classification, zero-shot-classification).
Downloading (…)okenizer_config.json: 100%|███████████████████████████████████████████████████████████████████████████████████| 26.0/26.0 [00:00<00:00, 234kB/s]
Downloading (…)olve/main/vocab.json: 100%|██████████████████████████████████████████████████████████████████████████████████| 899k/899k [00:00<00:00, 30.6MB/s]
Downloading (…)olve/main/merges.txt: 100%|██████████████████████████████████████████████████████████████████████████████████| 456k/456k [00:00<00:00, 33.9MB/s]
[...]

PR #1666, opened 2024-01-23
Reactivate BART Onnx Export

claeyzre:
# What does this PR do?

This PR reactivate the Bart Onnx Export which was disabled because of a Pytorch 2.1 regression. 
Since Pytorch 2.1.2 it is now fixed.
This [issue](https://github.com/huggingface/optimum/issues/1485) was the original issue of the export failure.

Question: To prevent this issue, hard constraints might be put under torch to make sure nobody gets optimum installed with torch 2.0 trying to export those models ? It should be above 1.11 but under 2.0 and could be above 2.1 as well. 

Tagging @fxmarty as he disabled the tests.

### DIFF
--- a/tests/onnxruntime/test_modeling.py
+++ b/tests/onnxruntime/test_modeling.py
@@ -1227,7 +1227,7 @@ class ORTModelIntegrationTest(unittest.TestCase):
 class ORTModelForQuestionAnsweringIntegrationTest(ORTModelTestMixin):
     SUPPORTED_ARCHITECTURES = [
         "albert",
-        # "bart",  # see tasks.py
+        "bart",  # see tasks.py
         "bert",
         # "big_bird",
         # "bigbird_pegasus",
@@ -1592,7 +1592,7 @@ class ORTModelForMaskedLMIntegrationTest(ORTModelTestMixin):
 class ORTModelForSequenceClassificationIntegrationTest(ORTModelTestMixin):
     SUPPORTED_ARCHITECTURES = [
         "albert",
-        # "bart",  # see tasks.py
+        "bart",  # see tasks.py
         "bert",
         # "big_bird",
         # "bigbird_pegasus",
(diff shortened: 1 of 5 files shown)


tests/onnxruntime/test_modeling.py
```diff
@@ -1592,7 +1592,7 @@ def test_compare_to_io_binding(self, model_arch):
 class ORTModelForSequenceClassificationIntegrationTest(ORTModelTestMixin):
     SUPPORTED_ARCHITECTURES = [
         "albert",
-        # "bart",  # see tasks.py
+        "bart",  # see tasks.py
```
fxmarty, line 1595:
```suggestion
        "bart", 
```

tests/onnxruntime/test_modeling.py
```diff
@@ -1227,7 +1227,7 @@ def test_trust_remote_code(self):
 class ORTModelForQuestionAnsweringIntegrationTest(ORTModelTestMixin):
     SUPPORTED_ARCHITECTURES = [
         "albert",
-        # "bart",  # see tasks.py
+        "bart",  # see tasks.py
```
fxmarty, line 1230:
```suggestion
        "bart", 
```

fxmarty (approved):
Thank you for the fix! Yes, if you would like to add a check e.g. in the `__init__` of the BartOnnxConfig on PyTorch version, but I think it is also fine to set `MIN_TORCH_VERSION = version.parse("2.1.2")` in `model_configs.py` for BartOnnxConfig, up to you!

Also fine for me to merge as is.

claeyzre:
Thanks for the review. 
Went for the second option.

claeyzre:
Hi @fxmarty, It looks like the CI didn't pass because of 
`OSError: [Errno 28] No space left on device`
Could you relaunch the CI if the subsequent problem has been resolved ?
Also would this fix be included in a hotfix release soon ?
Thanks,

PR #1674, opened 2024-01-31
Fixed bug key error "last_hidden_state"

satishsilveri:
# What does this PR do?

Fixed Key Error "last_hidden_state" bug when generating the embeddings using optimized models.

Fixes # 1673


Whadup:
Nice, this indeed fixes the key error. thanks

fxmarty (approved):
LGTM, a bit hacky but sentence-transformers models can be used with ORTModelForFeatureExtraction.

PR #1683, opened 2024-02-05
CI: Avoid iterating over a mutated iterable

fxmarty:
As per title, fixes `RuntimeError: dictionary changed size during iteration`

### DIFF
--- a/optimum/exporters/onnx/model_patcher.py
+++ b/optimum/exporters/onnx/model_patcher.py
@@ -62,7 +62,9 @@ def patch_everywhere(attribute_name: str, patch: Any, module_name_prefix: Option
         module_name_prefix (`Optional[str]`, defaults to `None`):
             If set, only module names starting with this prefix will be considered for patching.
     """
-    for name, module in sys.modules.items():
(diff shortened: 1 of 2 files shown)


optimum/exporters/onnx/model_patcher.py
```diff
@@ -62,7 +62,9 @@ def patch_everywhere(attribute_name: str, patch: Any, module_name_prefix: Option
         module_name_prefix (`Optional[str]`, defaults to `None`):
             If set, only module names starting with this prefix will be considered for patching.
     """
-    for name, module in sys.modules.items():
+    # sys.modules may be updated while being iterated over, hence the list copy.
+    for name in list(sys.modules):
```
michaelbenayoun, line 66: But `sys.modules` is a dictionary no?
```suggestion
    for name, module in dict(sys.modules).items():
```

fxmarty: `list(sys.modules)` gives a static (copy) list of the keys.

michaelbenayoun (approved):

PR #1685, opened 2024-02-06
Move & rename `onnx_export`

fxmarty:
As per title. It should ideally not be in `__main__.py`, as this file is kept simply for backward compatibility of `pythonm -m optimum.exporters.onnx`.

Motivation: not seeing `from optimum.exporters.onnx.__main__ import onnx_export` in code snippets.

### DIFF
--- a/optimum/exporters/onnx/convert.py
+++ b/optimum/exporters/onnx/convert.py
@@ -1036,22 +1036,6 @@ def onnx_export_from_model(
             f"Trying to export a {model_type} model, that is a custom or unsupported architecture, but no custom onnx configuration was passed as `custom_onnx_configs`. Please refer to https://huggingface.co/docs/optimum/main/en/exporters/onnx/usage_guides/export_a_model#custom-export-of-transformers-models for an example on how to export custom models. Please open an issue at https://github.com/huggingface/optimum/issues if you would like the model type {model_type} to be supported natively in the ONNX export."
         )
 
-    if task is None:
-        # TODO : _infer_task_from_model_or_model_class should also infer task from timm model
-        if library_name == "timm":
-            task = "image-classification"
-        else:
-            task = TasksManager._infer_task_from_model_or_model_class(model)
-
-        if (
-            library_name != "diffusers"
-            and task + "-with-past"
-            in TasksManager.get_supported_tasks_for_model_type(model_type, "onnx", library_name=library_name)
-            and not monolith
-            and model.config.use_cache
-        ):
-            task += "-with-past"
-
     if task.startswith("text-generation") and model.config.is_encoder_decoder:
         raise ValueError(
             f"model.config.is_encoder_decoder is True and task is `{task}`, which are incompatible. If the task was auto-inferred, please fill a bug report"
--- a/optimum/exporters/onnx/convert.py
+++ b/optimum/exporters/onnx/convert.py
@@ -22,24 +22,36 @@ import traceback
 from inspect import signature
 from itertools import chain
 from pathlib import Path
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 import numpy as np
 import onnx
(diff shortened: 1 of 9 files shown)


optimum/exporters/onnx/convert.py
```diff
@@ -884,3 +896,348 @@ def export(
     if not disable_dynamic_axes_fix:
         config.fix_dynamic_axes(output, device=device, input_shapes=input_shapes, dtype=dtype)
     return export_output
+
+
+def onnx_export_from_model(
```
fxmarty, line 901: The function is also removed to `onnx_export_from_model` to explicit its role, and documented.

optimum/exporters/onnx/convert.py
```diff
@@ -884,3 +896,348 @@ def export(
...
+
+        if (
+            not custom_architecture
+            and library_name != "diffusers"
+            and task + "-with-past"
+            in TasksManager.get_supported_tasks_for_model_type(model_type, "onnx", library_name=library_name)
+            and not monolith
+        ):
+            # -with-past is the default.
+            task = task + "-with-past"
```
fxmarty, line 1018: This is added compared to main - the default should be the export with KV cache.

optimum/exporters/onnx/convert.py
```diff
@@ -884,3 +896,348 @@ def export(
...
+
+    >>> model = AutoModelForCausalLM.from_pretrained("gpt2")
+    >>> # At this point, we could override some submodules, forward methods, weights, etc. from the model.
+
+    >>> onnx_export_from_model(model, output="gpt2_onnx/")
+    ```
+    """
+    library_name = TasksManager._infer_library_from_model(model)
+
+    # TODO: call standardize_model_attributes here once its model_name_or_path argument is optional.
```
fxmarty, line 991: As `standardize_model_attributes` is not called here but only in `main_export`, calling this function with a timm/diffusers/sentence-transformers model will fail. Only Transformers is working, I believe.

optimum/exporters/tasks.py
```diff
@@ -1718,6 +1718,8 @@ def standardize_model_attributes(
             library_name (`Optional[str]`, *optional*)::
                 The library name of the model. Can be any of "transformers", "timm", "diffusers", "sentence_transformers".
         """
+        # TODO: make model_name_or_path an optional argument here.
```
fxmarty, line 1721: @mht-sharma Is this argument needed? For timm, the config indeed seem not to be attached to the `timm.models.xxx.Xxx` (e.g. `timm.models.resnet.ResNet`) instance, but is this an issue for the export? Do we need any info from the config.json?

For diffusers, sentence-transformers, we are fine.

mht-sharma: @fxmarty indeed its not required, its mapped to the `pretrained_cfg`. Created a PR with the changes in respective locations. https://github.com/huggingface/optimum/pull/1686. Feel free to merge after review

optimum/exporters/onnx/convert.py
```diff
@@ -884,3 +896,348 @@ def export(
...
+
+        if (
+            library_name != "diffusers"
+            and task + "-with-past"
+            in TasksManager.get_supported_tasks_for_model_type(model_type, "onnx", library_name=library_name)
+            and not monolith
+            and model.config.use_cache
+        ):
+            task += "-with-past"
+
```
echarlaix, line 1054: should be removed
```suggestion

```

echarlaix (approved):
Looks great, thanks @fxmarty

optimum/exporters/tasks.py
fxmarty: Thank you @mht-sharma, can you in https://github.com/huggingface/optimum/pull/1686 make it so that `standardize_model_attributes` is called here: https://github.com/huggingface/optimum/blob/32a51afb7f8255f36546012b3b6bf060110e59bb/optimum/exporters/onnx/convert.py#L991

mht-sharma: Added

PR #1690, opened 2024-02-13
Fix library detection

fxmarty:
cc @echarlaix, reference https://github.com/huggingface/optimum/commit/3b4f5ace6873c31ab2867fd56067eccd9c0a55a2#r138565902


tests/exporters/test_tasks_manager.py
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+from optimum.exporters import TasksManager
+
```
echarlaix, line 19: why not include it to [tests/exporters/common/test_tasks_manager.py](https://github.com/huggingface/optimum/blob/main/tests/exporters/common/test_tasks_manager.py) ?

echarlaix (approved):
thanks for the fix @fxmarty

tests/exporters/test_tasks_manager.py
fxmarty: Did not know this one existed

fxmarty:
Failing tests are unrelated.

PR #1694, opened 2024-02-15
Fix no space left on device in actions

fxmarty:
Github actions seem to have decreased the available space on its ubuntu runners, see https://github.com/actions/runner-images/discussions/9329 & https://github.com/actions/runner-images/pull/9251/files

Go from 19G to 54G free disk space in 3min.


fxmarty:
Merging as workflow passes

PR #1695, opened 2024-02-15
Add end-to-end Marlin benchmark

fxmarty:
As per title. Reran the whole benchmark with more recent torch / AutoGPTQ / transformers versions.


fxmarty:
Merging as very non-critical.

PR #1707, opened 2024-02-21
Pin pytest

fxmarty:
As per title, avoid the CI to fail due to a deprecated function that was removed in in pytest===8.0

see https://github.com/huggingface/transformers/pull/29154


fxmarty:
@JingyaHuang thank you

JingyaHuang (approved):
LGTM

PR #1719, opened 2024-02-25
Experimental StableLM support

kazssym:
# What does this PR do?

This pull request tries to add support for the StableLM model. It is still incomplete.

Fixes # (issue)

## Before submitting


optimum/onnxruntime/utils.py
```diff
@@ -127,6 +127,7 @@ class ORTConfigManager:
         "nystromformer": "bert",
         "pegasus": "bert",
         "roberta": "bert",
+        "stablelm": "gpt2",
```
kazssym, line 130: Not sure if it's gpt2-like.

fxmarty:
@kazssym let me know if you'd like a review.

optimum/onnxruntime/utils.py
D4ve-R: I think it is correct, [this comment](https://github.com/huggingface/transformers/blob/main/src/transformers/models/stablelm/modeling_stablelm.py#L4) mentions it's similar to gpt-neox with small mods.

xenova:
Can confirm that this PR works for transformers.js models 👍

@kazssym Just a reminder: you can add https://huggingface.co/hf-internal-testing/tiny-random-StableLmForCausalLM as a unit test.

optimum/onnxruntime/utils.py
kazssym: Thank you.

xenova:
> This pull request tries to add support for the StableLM model. It is still incomplete.

Hi @kazssym 👋 - it looks like the PR was very close to completion. Would you like to re-open and request a review?

kazssym:
I'm no longer interested in the StableLM models and lack the time to update the branch without DirectML support. I hope someone else can continue the work.

PR #1725, opened 2024-02-27
Make optimized ONNX external data files downloadable from Hugging Face Hub

kazssym:
# What does this PR do?

This PR adds `.data` to the external data file suffixes so that an optimized model can be downloaded from Hugging Face as the optimizer saves models with `.data` instead of `_data`. 

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -873,21 +873,23 @@ class ORTModel(OptimizedModel):
                 force_download=force_download,
                 local_files_only=local_files_only,
             )
-            # try download external data
-            try:
-                hf_hub_download(
-                    repo_id=model_path.as_posix(),
-                    subfolder=subfolder,
-                    filename=file_name + "_data",
-                    use_auth_token=use_auth_token,
-                    revision=revision,
-                    cache_dir=cache_dir,
-                    force_download=force_download,
-                    local_files_only=local_files_only,
-                )
-            except EntryNotFoundError:
-                # model doesn't use external data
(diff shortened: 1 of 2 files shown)


kazssym:
> Hi @kazssym, Optimum ONNX export uses `model.onnx_data` for the external data file, contrary to other tools that use `model.onnx.data`. Hence the check on the Hub. Has this caused problems to you?

When I stored my external data file as `.onnx_data` on the Hub, it could be downloaded by Optimum but ONNX Runtime failed to load the file as it assumes `.onnx.data` internally. Renaming to `.onnx.data` on the Hub makes Optimum ignore the file and ONNX Runtime cannot use it anyway. It is inconvenient if I must manually download model data from the Hub.

fxmarty:
Thank you @kazssym, I am not sure to understand. Could you open an issue about that with a reproduction of the issue?

For example,
```bash
optimum-cli export onnx --model saibo/llama-1B llama_onnx
```

followed by

```python

model = ORTModelForCausalLM.from_pretrained("/path/to/llama_onnx")
```

works for me.

fxmarty:
@kazssym Thanks, let me try.

kazssym:
I found it needs more changes for this problem and changed this PR to draft now.

kazssym:
#1736

kazssym:
Updated this PR to cope with this case:
https://github.com/huggingface/optimum/blob/fd47a73267c3a71ea4e3c02f92260ae61c5ae372/optimum/onnxruntime/optimization.py#L225

PR #1753, opened 2024-03-11
Add support for Ernie model

BrightXiaoHan:
Solve #555 
This submission enables optimum to support the Ernie family model and support onnxruntime export of related models. The onnx code related to this model has been verified.

https://github.com/huggingface/transformers/tree/main/src/transformers/models/ernie

Before submitting

### DIFF
--- a/docs/source/exporters/onnx/overview.mdx
+++ b/docs/source/exporters/onnx/overview.mdx
@@ -42,6 +42,7 @@ Supported architectures from [🤗 Transformers](https://huggingface.co/docs/tra
 - Donut-Swin
 - Electra
 - Encoder Decoder
+- Ernie
 - ESM
 - Falcon
 - Flaubert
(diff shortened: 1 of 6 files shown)


fxmarty (commented):
LGTM thank you, waiting for the tests to run

Two things:
* could you add Ernie to the list at https://huggingface.co/docs/optimum/main/en/exporters/onnx/overview (in the `docs/` folder)
* could you add ernie to https://github.com/huggingface/optimum/blob/23fe6f75974ee55cb8e9323698b78ef08db2d2c8/tests/onnxruntime/test_modeling.py#L1979 & [here](https://github.com/huggingface/optimum/blob/23fe6f75974ee55cb8e9323698b78ef08db2d2c8/tests/onnxruntime/utils_onnxruntime_tests.py#L98)?

fxmarty:
@BrightXiaoHan Could you add ernie as well to https://github.com/huggingface/optimum/blob/7e08a820b65a359a61444abe51df4eb96b26b2e3/tests/exporters/exporters_utils.py#L82 ? Also can you use https://huggingface.co/hf-internal-testing/tiny-random-ErnieModel instead of https://huggingface.co/hf-internal-testing/tiny-random-ErnieMModel? They appear to be two different model types in transformers and the test currently fail due to this.

PR #1759, opened 2024-03-15
Add DPT support for ORTModelForSemanticSegmentation

merveenoyan:
Hiya 👋 
`ORTModelForSemanticSegmentation` is already implemented, DPT model in ONNX is already implemented too, but semantic segmentation isn't a task that was supported in DPT and DPT to ORT semantic segmentation integration tests, so I added that. The tests passed on my local so I hope it's gonna work fine. 

I also wanted to add BeiT but I got `op_name = 'adaptive_avg_pool2d, input size not accessible`


echarlaix (approved):
LGTM, thanks for the addition @merveenoyan !

PR #1760, opened 2024-03-18
Fix ONNX tests tensorflow version

echarlaix:
Fix https://github.com/huggingface/optimum/actions/workflows/test_onnx.yml (see https://github.com/huggingface/optimum/actions/runs/8330008937)


fxmarty (approved):
LGTM

PR #1762, opened 2024-03-21
Add LCM to DIFFUSERS_TASKS_TO_MODEL_LOADERS

l-bat:
# What does this PR do?

Relates to https://github.com/huggingface/optimum-intel/pull/618
* Fix `stable-diffusion-xl` mapping
* Support `latent-consistency` task

## Before submitting

### DIFF
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -1563,7 +1563,12 @@ class TasksManager:
 
             if library_name == "diffusers":
                 class_name = model_info.config["diffusers"]["class_name"]
-                inferred_task_name = "stable-diffusion-xl" if "StableDiffusionXL" in class_name else "stable-diffusion"
+                if "LatentConsistency" in class_name:
+                    inferred_task_name = "latent-consistency"
+                elif "StableDiffusionXL" in class_name:
+                    inferred_task_name = f"stable-diffusion-xl{'-refiner' if 'Img2Img' in class_name else ''}"
+                else:
+                    inferred_task_name = "stable-diffusion"
             elif library_name == "timm":
                 inferred_task_name = "image-classification"
             else:
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -188,7 +188,9 @@ class TasksManager:
 
         _DIFFUSERS_TASKS_TO_MODEL_LOADERS = {
             "stable-diffusion": "StableDiffusionPipeline",
-            "stable-diffusion-xl": "StableDiffusionXLImg2ImgPipeline",
+            "stable-diffusion-xl": "StableDiffusionXLPipeline",
+            "stable-diffusion-xl-refiner": "StableDiffusionXLImg2ImgPipeline",
+            "latent-consistency": "LatentConsistencyModelPipeline",
         }
 
         _TIMM_TASKS_TO_MODEL_LOADERS = {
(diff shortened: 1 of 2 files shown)


AlexKoff88:
@echarlaix, can you please take a look? This is required for https://github.com/huggingface/optimum-intel/pull/618/files

optimum/exporters/tasks.py
```diff
@@ -189,7 +189,9 @@ class TasksManager:
 
         _DIFFUSERS_TASKS_TO_MODEL_LOADERS = {
             "stable-diffusion": "StableDiffusionPipeline",
-            "stable-diffusion-xl": "StableDiffusionXLImg2ImgPipeline",
+            "stable-diffusion-xl": "StableDiffusionXLPipeline",
```
echarlaix, line 192: `_optional_components` was introduced in diffusers v0.22.0 https://github.com/huggingface/diffusers/blob/v0.22.0/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L144, so we would need to upgrade [`DIFFUSERS_MINIMUM_VERSION`](https://github.com/huggingface/optimum/blob/dedb852eaea3d3d899920f4017b9fd899eb2756b/optimum/utils/import_utils.py#L53) as well to avoid any potential issue when loading a SDXL model (which don't have optional components)

optimum/exporters/tasks.py
```diff
@@ -189,7 +189,9 @@ class TasksManager:
...
+            "stable-diffusion-xl-refiner": "StableDiffusionXLImg2ImgPipeline",
+            "latent-consistency": "LatentConsistencyModelPipeline",
```
echarlaix, line 194: Could you extend on why this is needed ? To my knowledge this shouldn't be needed (both pipelines can be correctly loaded using the current `_DIFFUSERS_TASKS_TO_MODEL_LOADERS`

echarlaix: also at the moment inferred task for `diffusers` models will either be "stable-diffusion-xl" or "stable-diffusion" https://github.com/huggingface/optimum/blob/dedb852eaea3d3d899920f4017b9fd899eb2756b/optimum/exporters/tasks.py#L1564

l-bat: This is necessary to correctly determine the model class from task. Hybrid quantization supports the ` stable-diffusion-xl` task, but doesn't support the `stable-diffusion-xl-refiner`. Without these changes, it is not possible to apply hybrid quantization via CLI.

optimum/exporters/tasks.py
l-bat: [DIFFUSERS_MINIMUM_VERSION](https://github.com/huggingface/optimum/blob/dedb852eaea3d3d899920f4017b9fd899eb2756b/optimum/utils/import_utils.py#L53) was upgraded

optimum/exporters/tasks.py
echarlaix: Not sure that we need it in the TaskManager, from what I see it's only needed to get the class to load the model in the optimum-intel CLI for diffusers models, could we move it in optimum-intel directly instead ? We could get this information by combining the task (text-to-image, image-to-image,  inpaint, ..) and the architecture (SD, SDXL, LCMs, ..) using the model's config

l-bat: Are you suggesting to remove only the `latent-consistency` from `_DIFFUSERS_TASKS_TO_MODEL_LOADERS`?

As far as I understand, there are no `text-to-image, image-to-image, inpaint,..` tasks in optimum. Should I create these tasks only for optimum-intel instead of `stable-diffusion, stable-diffusion-xl, stable-diffusion-xl-refiner and latent-consistency`? In this case, we cann't use `TasksManager.infer_task_from_model` for diffusion models.

Issue #1774, opened 2024-03-25
Unable to export Table Transformer


aW3st:
### System Info

```shell
Output of optimum-cli env:

- `optimum` version: 1.18.0
- `transformers` version: 4.39.1
- Platform: Linux-5.4.247-162.350.amzn2.x86_64-x86_64-with-glibc2.35
- Python version: 3.10.6
- Huggingface_hub version: 0.22.0
- PyTorch version (GPU?): 2.2.1+cu121 (cuda availabe: True)
- Tensorflow version (GPU?): not installed (cuda availabe: NA)

Output of nvidia-smi:
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 470.182.03   Driver Version: 470.182.03   CUDA Version: 12.1     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla T4            Off  | 00000000:00:1E.0 Off |                    0 |
| N/A   29C    P0    24W /  70W |      0MiB / 15109MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+

Other relevant packages:
nvidia-cublas-cu12       12.1.3.1
nvidia-cuda-cupti-cu12   12.1.105
nvidia-cuda-nvrtc-cu12   12.1.105
nvidia-cuda-runtime-cu12 12.1.105
nvidia-cudnn-cu12        8.9.2.26
nvidia-cufft-cu12        11.0.2.54
nvidia-curand-cu12       10.3.2.106
nvidia-cusolver-cu12     11.4.5.107
nvidia-cusparse-cu12     12.1.0.106
nvidia-nccl-cu12         2.19.3
nvidia-nvjitlink-cu12    12.4.99
nvidia-nvtx-cu12         12.1.105
onnx                     1.16.0
onnxruntime-gpu          1.17.1 # installed for CUDA 12.x using method described here https://onnxruntime.ai/docs/install/#install-onnx-runtime-gpu-cuda-12x
torchvision              0.17.1
```

### Who can help?

@michaelbenayoun, @JingyaHuang

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

[...]

PR #1775, opened 2024-03-26
Fix table transformers cuda export

mht-sharma:
As per title!

Issue: #1774

### DIFF
--- a/optimum/exporters/onnx/convert.py
+++ b/optimum/exporters/onnx/convert.py
@@ -274,8 +274,7 @@ def _run_validation(
 
     # Create ONNX Runtime session
     session_options = SessionOptions()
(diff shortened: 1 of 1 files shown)


optimum/exporters/onnx/convert.py
```diff
@@ -274,8 +274,7 @@ def _run_validation(
 
     # Create ONNX Runtime session
     session_options = SessionOptions()
-    # We could well set ORT_DISABLE_ALL here, but it makes CUDA export with O4 of gpt_neo fail
```
mht-sharma, line 277: Was a little confused by comment, does the export of gpt_neo fails with DISABLE?

fxmarty:
@mht-sharma I'm not sure about this

Issue #1297, opened 2023-08-21
[ONNX export] Musicgen for text-to-audio


xenova:
### Feature request

Musicgen was recently added to 🤗 Transformers ([model doc](https://huggingface.co/docs/transformers/model_doc/musicgen)) and it would be great to be able to export those models to ONNX with Optimum.

### Motivation

This will allow me to support music generation models in Transformers.js

### Your contribution

I will integrate into transformers.js once available in optimum.

kanger45:
Hi, 
   I'm also interested in converting musicgen model to onnx format so I can try to deploy it to the device. May i know is it support on Optimum now?

MaiZhiHao:
It would be great if this feature is done.
Btw, how can I get the transformers.js ?

xenova:
> May i know is it support on Optimum now?

Not yet 😇 cc @fxmarty


> Btw, how can I get the transformers.js ?

You can check out the repo [here](https://github.com/xenova/transformers.js) or the documentation [here](https://huggingface.co/docs/transformers.js/index). Since musicgen is not yet available in Optimum, however, it won't be available in transformers.js until then.

kanger45:
hi @xenova,

   May i know if have a plan or schedule to support Optimum for convert it to ONNX model?

fxmarty:
Hi @kanger45 @MaiZhiHao @zeke-john https://github.com/huggingface/optimum/pull/1779 is merged, which exports Musicgen in several parts to generate audio samples conditioned on a text prompt (Reference: https://huggingface.co/docs/transformers/model_doc/musicgen#text-conditional-generation). This uses the decoder KV cache. The following subcomponents are exported:
* `text_encoder.onnx`: corresponds to the text encoder part in https://github.com/huggingface/transformers/blob/v4.39.1/src/transformers/models/musicgen/modeling_musicgen.py#L1457.
* `encodec_decode.onnx`: corresponds to the Encodec audio encoder part in https://github.com/huggingface/transformers/blob/v4.39.1/src/transformers/models/musicgen/modeling_musicgen.py#L2472-L2480.
* `decoder_model.onnx`: The Musicgen decoder, without past key values input, and computing cross attention. Not required at inference (use decoder_model_merged.onnx instead).
* `decoder_with_past_model.onnx`: The Musicgen decoder, with past_key_values input (KV cache filled), not computing cross attention. Not required at inference (use decoder_model_merged.onnx instead).
[...]

PR #1779, opened 2024-03-27
Musicgen ONNX export (text-conditional only)

fxmarty:
Exports Musicgen conditioned by a text prompt.

```
optimum-cli export onnx --model facebook/musicgen-small musicgen_onnx &> export.log
```

If we want to condition with audio, it is more tricky and we first need to be able to export `EncodecModel.encode` which requires a combination of jit.script/jit.trace as it has some unrollable loops, unfortuantely.

Only KV cache export is tested & supported.

The following subcomponents are exported:
* `text_encoder.onnx`: corresponds to the text encoder part in https://github.com/huggingface/transformers/blob/v4.39.1/src/transformers/models/musicgen/modeling_musicgen.py#L1457.
* `audio_encoder_decode.onnx`: corresponds to the Encodec audio encoder part in https://github.com/huggingface/transformers/blob/v4.39.1/src/transformers/models/musicgen/modeling_musicgen.py#L2472-L2480.
* `decoder_model.onnx`: The Musicgen decoder, without past key values input, and computing cross attention.
* `decoder_with_past_model.onnx`: The Musicgen decoder, with past_key_values input (KV cache filled), not computing cross attention.
* `decoder_model_merged.onnx`: The two previous models fused in one, to avoid duplicating weights. A boolean input `use_cache_branch` allows to select the branch to use. In the first forward pass where the KV cache is empty, dummy past key values inputs need to be passed and are ignored with use_cache_branch=False.
* `build_delay_pattern_mask.onnx`: corresponds to https://github.com/huggingface/transformers/blob/v4.39.1/src/transformers/models/musicgen/modeling_musicgen.py#L1054-L1125

Partially fixes https://github.com/huggingface/optimum/issues/1297

⚠️ Depends on https://github.com/huggingface/transformers/pull/29913, please check out on this branch for now


fxmarty:
Need https://github.com/huggingface/optimum/pull/1780 for the CI

ylacombe:
Will be great to do it with Musicgen Melody as well (where music conditioning is different, so could maybe work)

optimum/exporters/onnx/model_configs.py
```diff
@@ -1372,10 +1382,252 @@ def outputs(self) -> Dict[str, Dict[int, str]]:
...
+        return models_and_onnx_configs, onnx_files_subpaths_new
+
+
 class SpeechT5OnnxConfig(OnnxSeq2SeqConfigWithPast):
     # TODO: Transformers batched generation for Speecht5 is BROKEN (https://github.com/huggingface/transformers/pull/25943),
     # so we won't support for now.
-    NORMALIZED_CONFIG_CLASS = NormalizedConfig.with_args(decoder_num_layers="decoder_layers")
     NORMALIZED_CONFIG_CLASS = NormalizedSeq2SeqConfig.with_args(
         hidden_size="hidden_size",
         num_attention_heads="encoder_attention_heads",  # TODO: bugged in case encoder and decoder have different number of heads
```
xenova, line 1633: I believe this causes an issue with https://huggingface.co/facebook/musicgen-small, which has a text encoder with 12 heads, and a decoder with 16 heads.

fxmarty:
@ylacombe What are the main differences?

xenova:
I've been testing this out in transformers.js and will report back when the output matches!

fxmarty:
@xenova I need to add the attention_mask input in ONNX

xenova (approved):
Can confirm this works for transformers.js ([v3 branch](https://github.com/xenova/transformers.js/pull/545#issuecomment-2041205108))! 🚀

Example code:

```js
import { AutoTokenizer, MusicgenForConditionalGeneration } from '@xenova/transformers';

// Load tokenizer and model
const tokenizer = await AutoTokenizer.from_pretrained('Xenova/musicgen-small');
const model = await MusicgenForConditionalGeneration.from_pretrained(
  'Xenova/musicgen-small', { dtype: 'fp32' }
);

// Prepare text input
const prompt = '80s pop track with bassy drums and synth';
const inputs = tokenizer(prompt);

// Generate audio
const audio_values = await model.generate({
  ...inputs,
  max_new_tokens: 512,
  do_sample: true,
  guidance_scale: 3,
});

// (Optional) Write the output to a WAV file
import wavefile from 'wavefile';
import fs from 'fs';

const wav = new wavefile.WaveFile();
wav.fromScratch(1, model.config.audio_encoder.sampling_rate, '32f', audio_values.data);
fs.writeFileSync('musicgen_out.wav', wav.toBuffer());
```

Samples:

https://github.com/xenova/transformers.js/assets/26504141/558855d3-5f30-41fa-85c4-0c530bdabd98


https://github.com/xenova/transformers.js/assets/26504141/80e8785c-ffdf-4181-9fb0-010f69c72a06


https://github.com/xenova/transformers.js/assets/26504141/1b66801d-43c5-49cd-9ae4-fea2ca3f23d8

optimum/exporters/onnx/model_configs.py
```diff
@@ -1372,10 +1383,302 @@ def outputs(self) -> Dict[str, Dict[int, str]]:
...
+            raise ValueError(
+                f"model_part is {model_part} and behavior is {behavior}. This is not supported, please open an issue at https://github.com/huggingface/optimum/issues."
+            )
+
+        if behavior == ConfigBehavior.MONOLITH:
+            raise ValueError(
+                "Musicgen does not support behavior=ConfigBehavior.MONOLITH. Please open an issue at https://github.com/huggingface/optimum/issues."
+            )
+
+        if config.audio_encoder.model_type != "encodec":
```
JingyaHuang, line 1460: Isn't it a typo? (encoder?)

JingyaHuang: @fxmarty

fxmarty: What would be the typo?

PR #1793, opened 2024-04-03
Fix infer task fo stable diffusion

JingyaHuang:
# What does this PR do?

The current task inference is not working for stable diffusion checkpoints, eg. [`runwayml/stable-diffusion-v1-5`](https://huggingface.co/runwayml/stable-diffusion-v1-5/blob/1d0c4ebf6ff58a5caecab40fa1406526bca4b5b9/model_index.json#L2)

p.s. why did we use `class_name` I did not see any checkpoint using it instead of `_class_name`, if there is not we should remove it (keep it for now).

```python
from optimum.exporters import TasksManager

task = TasksManager.infer_task_from_model("runwayml/stable-diffusion-v1-5")

# KeyError: 'class_name'
```

## Before submitting

### DIFF
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -1560,7 +1560,9 @@ class TasksManager:
             library_name = TasksManager.infer_library_from_model(model_name_or_path, subfolder, revision)
 
             if library_name == "diffusers":
-                class_name = model_info.config["diffusers"]["class_name"]
+                class_name = model_info.config["diffusers"].get("class_name", None) or model_info.config[
+                    "diffusers"
+                ].get("_class_name", None)
                 inferred_task_name = "stable-diffusion-xl" if "StableDiffusionXL" in class_name else "stable-diffusion"
             elif library_name == "timm":
                 inferred_task_name = "image-classification"


optimum/exporters/tasks.py
```diff
@@ -1560,7 +1560,9 @@ def _infer_task_from_model_name_or_path(
             library_name = TasksManager.infer_library_from_model(model_name_or_path, subfolder, revision)
 
             if library_name == "diffusers":
-                class_name = model_info.config["diffusers"]["class_name"]
+                class_name = model_info.config["diffusers"].get("class_name", None) or model_info.config[
+                    "diffusers"
+                ].get("_class_name", None)
```
fxmarty, line 1565:
```suggestion
                if hasattr(model_info.config["diffusers"], "class_name"):
                    class_name = model_info.config["diffusers"]["class_name"]
                elif hasattr(model_info.config["diffusers"], "_class_name"):
                    class_name = model_info.config["diffusers"]["_class_name"]
                else:
                    raise ValueError(f"Could not automatically infer the task name for {model_name_or_path}. Please open an issue at https://github.com/huggingface/optimum/issues.")
```

JingyaHuang: @fxmarty `model_info.config["diffusers"]` is a dictionary and `class_name` or `_class_name` are keys instead of attributes.

fxmarty: feel free to edit accordingly

JingyaHuang: Done!

fxmarty: @JingyaHuang Is the suggestion fine as is in the end?

fxmarty (approved):
Thank you

optimum/exporters/tasks.py
JingyaHuang: It's good.

PR #1819, opened 2024-04-18
Apply deprecated `evaluation_strategy`

muellerzr:
# What does this PR do?

This PR applies the deprecated `evaluation_strategy` from https://github.com/huggingface/transformers/pull/30190.

TL;DR: `evaluation_strategy` -> `eval_strategy`

(Should go live in v4.41.0, but opening now so it's here when ready!)

Fixes # (issue)

## Before submitting

## Who can review?

- ONNX Runtime Training: @JingyaHuang

### DIFF
--- a/examples/onnxruntime/training/image-classification/README.md
+++ b/examples/onnxruntime/training/image-classification/README.md
@@ -39,7 +39,7 @@ torchrun --nproc_per_node=NUM_GPUS_YOU_HAVE run_image_classification.py \
     --per_device_eval_batch_size 32 \
     --logging_strategy steps \
     --logging_steps 10 \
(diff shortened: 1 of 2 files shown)


Rohan138:
@muellerzr @IlyasMoutawwakil could we merge this?

It currently breaks examples with transformers>=4.41 because `evaluation_strategy` is set to None (https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#1375, which errors at https://github.com/huggingface/optimum/blob/main/optimum/onnxruntime/training_args.py#L129 because EvaluationStrategy expects one of ['no', 'steps', 'epoch']

Issue #1761, opened 2024-03-20
ORTOptimizer for the model type table-transformer


zachmayer:
### Feature request

I'd like to add support for the new[ microsoft table transformers](https://huggingface.co/microsoft/table-transformer-detection) to ORTOptimizer

### Motivation

These table transformers are object detection models.  They have a backbone that's a CNN, so at the very least I think we could optimize the CNN part of these models?

### Your contribution

Here is my sample script:

```python
import os
from pathlib import Path

import torch
from optimum.onnxruntime import ORTModel, ORTOptimizer
from optimum.onnxruntime.configuration import OptimizationConfig
from transformers import AutoImageProcessor, TableTransformerForObjectDetection

# Define the input model and the output directory
model_name = "microsoft/table-transformer-detection"
onnx_path = Path("~/Downloads/onnx_model/").expanduser()
onnx_model_name = "table_transformer.onnx"

# Load the Processor and save its config
processor = AutoImageProcessor.from_pretrained(model_name)
processor.save_pretrained(onnx_path)

# Load the model and save its config
model = TableTransformerForObjectDetection.from_pretrained(
    "microsoft/table-transformer-detection"
)
model.eval()
model.config.save_pretrained(onnx_path)

# Convert the model to onnx
torch.onnx.export(
    model,
    torch.rand((1, 3, 800, 600)),
    os.path.join(onnx_path, onnx_model_name),
    export_params=True,
    do_constant_folding=True,
    opset_version=11,
    input_names=["pixel_values"],
    output_names=["max_probability"],
    dynamic_axes={
        "pixel_values": {0: "batch_size", 2: "height", 3: "width"},
        "max_probability": {0: "batch_size"},
    },
)

# Optimize the onnx model
onnx_model = ORTModel.from_pretrained(onnx_path, file_name=onnx_model_name)
optimizer = ORTOptimizer.from_pretrained(onnx_model)
optimizer.optimize(
    save_dir=onnx_path,
    optimization_config=OptimizationConfig(optimization_level=99),
)
```

Which raises `NotImplementedError: Tried to use ORTOptimizer for the model type table-transformer, but it is not available yet. Please open an issue or submit a PR at https://github.com/huggingface/optimum.`

I'm more than happy to make  PR, if someone could share some example PRs I could use as a template or give me some guidance as to how to start.

mht-sharma:
Hi @zachmayer  thanks for interest in adding the `table-transformer` model in `ORTOptimizer`.

[...]

PR #1820, opened 2024-04-18
ORTOptimizer for the model type Segformer

zachmayer:
# What does this PR do?
Adds the segformer model to ORTOptimizer.  Based on the advice I got in https://github.com/huggingface/optimum/issues/1761, but I decided to start with segformer.

Fixes # (issue)

## Before submitting

## Who can review?
@mht-sharma maybe?

### DIFF
--- a/optimum/utils/normalized_config.py
+++ b/optimum/utils/normalized_config.py
@@ -102,6 +102,30 @@ class NormalizedVisionConfig(NormalizedConfig):
     INPUT_SIZE = "input_size"
 
 
+class NormalizedSegformerConfig(NormalizedVisionConfig):
+    NUM_ATTENTION_HEADS = "num_attention_heads"
+    HIDDEN_SIZE = "decoder_hidden_size"
+
+    def __getattr__(self, attr_name):
+        if attr_name == "hidden_size":
+            attr_value = getattr(self.config, self.HIDDEN_SIZE, None)
+            if attr_value is None:
+                raise AttributeError(f"Attribute {self.HIDDEN_SIZE} not found in config")
+            if isinstance(attr_value, list):
+                return max(attr_value)
+            return attr_value
+
+        elif attr_name == "num_attention_heads":
+            attr_value = getattr(self.config, self.NUM_ATTENTION_HEADS, None)
+            if attr_value is None:
+                raise AttributeError(f"Attribute {self.NUM_ATTENTION_HEADS} not found in config")
+            if isinstance(attr_value, list):
+                return max(attr_value)
+            return attr_value
+
+        return super().__getattr__(attr_name)
+
+
 class NormalizedTextAndVisionConfig(NormalizedTextConfig, NormalizedVisionConfig):
     TEXT_CONFIG = None
     VISION_CONFIG = None
@@ -203,7 +227,6 @@ class NormalizedConfigManager:
         'owlvit',
         'perceiver',
         'roformer',
-        'segformer',
         'squeezebert',
         'table-transformer',
     """
@@ -256,6 +279,7 @@ class NormalizedConfigManager:
         "regnet": NormalizedVisionConfig,
         "resnet": NormalizedVisionConfig,
         "roberta": NormalizedTextConfig,
+        "segformer": NormalizedSegformerConfig,
         "speech-to-text": SpeechToTextLikeNormalizedTextConfig,
         "splinter": NormalizedTextConfig,
         "t5": T5LikeNormalizedTextConfig,
--- a/optimum/utils/normalized_config.py
+++ b/optimum/utils/normalized_config.py
@@ -111,8 +111,6 @@ class NormalizedSegformerConfig(NormalizedVisionConfig):
             attr_value = getattr(self.config, self.HIDDEN_SIZE, None)
             if attr_value is None:
                 raise AttributeError(f"Attribute {self.HIDDEN_SIZE} not found in config")
-            if isinstance(attr_value, list):
-                return max(attr_value)
             return attr_value
 
         elif attr_name == "num_attention_heads":
--- a/optimum/utils/normalized_config.py
+++ b/optimum/utils/normalized_config.py
@@ -106,10 +106,12 @@ class NormalizedSegformerConfig(NormalizedVisionConfig):
     NUM_ATTENTION_HEADS = "num_attention_heads"
     HIDDEN_SIZE = "hidden_sizes"
 
+    # If the attribute is a list, return 0
+    # 0 means let the optimizer infer the correct value based on the model graph
     def __getattr__(self, attr_name):
         attr_value = super().__getattr__(attr_name)
         if isinstance(attr_value, list):
-            attr_value = max(attr_value)
+            attr_value = 0
         return attr_value
 
 
--- a/optimum/utils/normalized_config.py
+++ b/optimum/utils/normalized_config.py
@@ -104,24 +104,13 @@ class NormalizedVisionConfig(NormalizedConfig):
 
 class NormalizedSegformerConfig(NormalizedVisionConfig):
     NUM_ATTENTION_HEADS = "num_attention_heads"
-    HIDDEN_SIZE = "decoder_hidden_size"
+    HIDDEN_SIZE = "hidden_sizes"
 
     def __getattr__(self, attr_name):
-        if attr_name == "hidden_size":
-            attr_value = getattr(self.config, self.HIDDEN_SIZE, None)
-            if attr_value is None:
-                raise AttributeError(f"Attribute {self.HIDDEN_SIZE} not found in config")
-            return attr_value
-
-        elif attr_name == "num_attention_heads":
-            attr_value = getattr(self.config, self.NUM_ATTENTION_HEADS, None)
-            if attr_value is None:
-                raise AttributeError(f"Attribute {self.NUM_ATTENTION_HEADS} not found in config")
-            if isinstance(attr_value, list):
-                return max(attr_value)
-            return attr_value
-
-        return super().__getattr__(attr_name)
+        attr_value = super().__getattr__(attr_name)
+        if isinstance(attr_value, list):
+            attr_value = sum(attr_value)
+        return attr_value
 
 
 class NormalizedTextAndVisionConfig(NormalizedTextConfig, NormalizedVisionConfig):
--- a/optimum/utils/normalized_config.py
+++ b/optimum/utils/normalized_config.py
@@ -109,7 +109,7 @@ class NormalizedSegformerConfig(NormalizedVisionConfig):
     def __getattr__(self, attr_name):
         attr_value = super().__getattr__(attr_name)
         if isinstance(attr_value, list):
-            attr_value = sum(attr_value)
+            attr_value = max(attr_value)
         return attr_value
 
 
--- a/tests/onnxruntime/test_optimization.py
+++ b/tests/onnxruntime/test_optimization.py
@@ -24,6 +24,12 @@ import numpy as np
 import onnx
 import pytest
 import torch
+from parameterized import parameterized
+from transformers import AutoTokenizer
+from transformers.onnx.utils import get_preprocessor
+from transformers.testing_utils import require_torch_gpu
+from utils_onnxruntime_tests import MODEL_NAMES
+
 from optimum.exporters import TasksManager
 from optimum.exporters.onnx import MODEL_TYPES_REQUIRING_POSITION_IDS
 from optimum.onnxruntime import (
@@ -37,11 +43,6 @@ from optimum.onnxruntime.configuration import OptimizationConfig
 from optimum.onnxruntime.modeling_decoder import ORTModelForCausalLM
 from optimum.onnxruntime.modeling_seq2seq import ORTModelForSeq2SeqLM, ORTModelForSpeechSeq2Seq
 from optimum.utils.testing_utils import grid_parameters
-from parameterized import parameterized
-from transformers import AutoTokenizer
-from transformers.onnx.utils import get_preprocessor
-from transformers.testing_utils import require_torch_gpu
-from utils_onnxruntime_tests import MODEL_NAMES
 
 
 class ORTOptimizerTestMixin(unittest.TestCase):
--- a/tests/onnxruntime/test_optimization.py
+++ b/tests/onnxruntime/test_optimization.py
@@ -24,12 +24,6 @@ import numpy as np
 import onnx
 import pytest
 import torch
-from parameterized import parameterized
-from transformers import AutoTokenizer
-from transformers.onnx.utils import get_preprocessor
-from transformers.testing_utils import require_torch_gpu
-from utils_onnxruntime_tests import MODEL_NAMES
-
 from optimum.exporters import TasksManager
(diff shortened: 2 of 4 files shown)


optimum/utils/normalized_config.py
```diff
@@ -102,6 +102,30 @@ class NormalizedVisionConfig(NormalizedConfig):
...
+    NUM_ATTENTION_HEADS = "num_attention_heads"
+    HIDDEN_SIZE = "decoder_hidden_size"
+
+    def __getattr__(self, attr_name):
+        if attr_name == "hidden_size":
+            attr_value = getattr(self.config, self.HIDDEN_SIZE, None)
+            if attr_value is None:
+                raise AttributeError(f"Attribute {self.HIDDEN_SIZE} not found in config")
+            if isinstance(attr_value, list):
+                return max(attr_value)
```
zachmayer, line 115: not sure if max is the way to go here, or if I need to add support for lists somewhere

tests/onnxruntime/test_optimization.py
```diff
@@ -43,6 +37,11 @@
 from optimum.onnxruntime.modeling_decoder import ORTModelForCausalLM
 from optimum.onnxruntime.modeling_seq2seq import ORTModelForSeq2SeqLM, ORTModelForSpeechSeq2Seq
 from optimum.utils.testing_utils import grid_parameters
+from parameterized import parameterized
+from transformers import AutoTokenizer
+from transformers.onnx.utils import get_preprocessor
+from transformers.testing_utils import require_torch_gpu
+from utils_onnxruntime_tests import MODEL_NAMES
```
zachmayer, line 44: I ran black— I can undo this part

optimum/utils/normalized_config.py
```diff
@@ -102,6 +102,30 @@ class NormalizedVisionConfig(NormalizedConfig):
...
+            return attr_value
+
+        elif attr_name == "num_attention_heads":
+            attr_value = getattr(self.config, self.NUM_ATTENTION_HEADS, None)
+            if attr_value is None:
+                raise AttributeError(f"Attribute {self.NUM_ATTENTION_HEADS} not found in config")
+            if isinstance(attr_value, list):
+                return max(attr_value)
```
zachmayer, line 123: I went with max here, but if needed I can try to figure out how to implement list support

optimum/utils/normalized_config.py
zachmayer: actually `decoder_hidden_size` may not be a list— if its just an int I can remove the list handling here.

tests/onnxruntime/test_optimization.py
IlyasMoutawwakil: use the library's styling with `make style` after `pip install -e .[quality]`

tests/onnxruntime/test_optimization.py
```diff
@@ -171,6 +170,7 @@ def test_compare_original_seq2seq_model_with_optimized_model(self, model_cls, mo
 
     # Contribution note: Please add test models in alphabetical order. Find test models here: https://huggingface.co/hf-internal-testing.
     SUPPORTED_IMAGE_ARCHITECTURES_WITH_MODEL_ID = (
+        (ORTModelForImageClassification, "hf-internal-testing/tiny-random-segformer"),
```
IlyasMoutawwakil, line 173: there's no `tiny-random-segformer` on the hub, there's however `hf-internal-testing/tiny-random-SegformerForImageClassification`.

optimum/utils/normalized_config.py
IlyasMoutawwakil: I think it makes more sense to return the list as is. and it's the exporter's job to handle it. 
@mht-sharma do you know what `hidden_size` or `num_attention_heads` can be used for with image classification models ?

IlyasMoutawwakil (changes requested):
Thanks for the PR! I left few comments.

tests/onnxruntime/test_optimization.py
zachmayer: I am able to use both the processor and the model for `"hf-internal-testing/tiny-random-segformer"`:

```python
import torch
from transformers import AutoModel, AutoImageProcessor
model_name = "hf-internal-testing/tiny-random-segformer"
processor = AutoImageProcessor.from_pretrained(model_name)
model = AutoModel.from_pretrained(model_name)
dummy_image = torch.rand(1, 3, 224, 224)
inputs = processor(images=dummy_image, return_tensors="pt")
with torch.no_grad():
    outputs = model(**inputs)
print(outputs)
```

Maybe I'm misunderstanding something?  But the model seems to work.

zachmayer: There's a page for the model too: https://huggingface.co/hf-internal-testing/tiny-random-segformer

optimum/utils/normalized_config.py
zachmayer: Removed the list bit for decoder_hidden_size

optimum/utils/normalized_config.py
zachmayer: So if I remove the `max` and run `pytest tests/onnxruntime/test_optimization.py -k test_compare_original_image_model_with_optimized_model -vv`, I get the following error:

```python
../../venvs/optimum/lib/python3.11/site-packages/onnxruntime/transformers/optimizer.py:178: in optimize_by_fusion
    optimizer = optimizer_class(model, num_heads, hidden_size)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <onnx_model_bert.BertOnnxModel object at 0x16fbebf50>, model = ir_version: 7
opset_import {
  version: 12
}
opset_import {
  domain: "com.microsoft.experimental"
  version: 1
}
opse...m_param: "batch_size"
          }
          dim {
            dim_value: 150
          }
        }
      }
    }
  }
}
, num_heads = [1, 2, 5, 8], hidden_size = 256

    def __init__(self, model: ModelProto, num_heads: int = 0, hidden_size: int = 0):
        """Initialize BERT ONNX Model.
    
        Args:
            model (ModelProto): the ONNX model
            num_heads (int, optional): number of attention heads. Defaults to 0 (detect the parameter automatically).
            hidden_size (int, optional): hidden dimension. Defaults to 0 (detect the parameter automatically).
        """
>       assert (num_heads == 0 and hidden_size == 0) or (num_heads > 0 and hidden_size % num_heads == 0)
E       TypeError: '>' not supported between instances of 'list' and 'int'
```

zachmayer: It looks like the `onnxruntime` optimizer expects `num_attention_heads` to be an int.

zachmayer: maybe https://github.com/microsoft/onnxruntime/issues/17254 is related

zachmayer: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/onnx_model_bert.py#L37

The `BertOnnxModel` class can't handle a list for the number of attention heads.  Is BertOnnxModel the right class to use for the segformer model?

tests/onnxruntime/test_optimization.py
IlyasMoutawwakil: okay weird I couldn't find it for some reason.

optimum/utils/normalized_config.py
mht-sharma: HI @zachmayer it seems you are trying to optimise the `Segformer` for `image-classification` task. The model class for [SegformerModelForImageClassification](https://github.com/huggingface/transformers/blob/v4.40.1/src/transformers/models/segformer/modeling_segformer.py#L573) does not use the `decoder_hidden_size`. Not sure how the optimiser worked with `decoder_hidden_size`. Could you check the before and after graph if you see any changes?

Additionally, I am not super familiar with the architecture but on a quick read, the model uses hierarchal Transformer encoders, which explains multiple `hidden_sizes` and `attention_heads`. So this model may not be supported by onnxruntime itself.

I would suggest passing correct values, of hidden_size and heads, for a particular encoder block to the optimiser and see changes in the corresponding part in the graph.

zachmayer: @mht-sharma — I'd actually like to optimise the `Segformer` for the semantic segmentation task! I was just following the patter for the test for the `vit` model.  I will update the test to use `ORTModelForSemanticSegmentation`.

The onnx runtime does support this model: I am able to:
- export it to onnx
- quantize it
- reload it and run inference

Here is an example of loading this model in onnx and quantizing it.  Everything seems to work great!
```python
from io import BytesIO
from pathlib import Path

import requests
from PIL import Image

from optimum.onnxruntime import ORTModelForSemanticSegmentation, ORTQuantizer
from optimum.onnxruntime.configuration import AutoQuantizationConfig


# Load a segformer
model_name = "hf-internal-testing/tiny-random-segformer"
model = ORTModelForSemanticSegmentation.from_pretrained(model_name, export=True)

# Quanzize
quantized_model_path = Path("~/Downloads/onnx_models/").expanduser() / model_name
quantizer = ORTQuantizer.from_pretrained(model)
quantizer.quantize(
    save_dir=quantized_model_path,
    quantization_config=AutoQuantizationConfig.avx2(is_static=False, per_channel=True),
)
model_quantized = ORTModelForSemanticSegmentation.from_pretrained(quantized_model_path)
del model

# Use the quantized model for inference on a test image
url = "https://upload.wikimedia.org/wikipedia/en/7/7d/Lenna_%28test_image%29.png"
response = requests.get(url)
image = Image.open(BytesIO(response.content))
processor = model_quantized.preprocessors[0]
inputs = processor.preprocess(images=image, return_tensors="np")
inputs = {"pixel_values": inputs["pixel_values"]}

# Run inference
outputs = model_quantized.model.run(None, inputs)[0]
print(outputs)
assert outputs.shape[1] == len(model_quantized.config.id2label)
```

zachmayer: Given that the onnx runtime already supports `ORTModelForSemanticSegmentation`, and given that `ORTModelForSemanticSegmentation` supports quantizing, I thought we should also be able to support optimizing for this model.

zachmayer: When I look at the config for the quantized model, I see it has a decoder_hidden_size:
```python
In [17]: print(model_quantized.config.decoder_hidden_size)
256
```

It also has a list of `hidden_sizes` and a list of `num_attention_heads`:
```python
In [18]: print(model_quantized.config.hidden_sizes)
[1, 2, 5, 8]

```

```python
In [19]: print(model_quantized.config.num_attention_heads)
[1, 2, 5, 8]
```

zachmayer: So Microsoft hasn't tested the optimizer with segformers yet: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/README.md#supported-models

They expect both num_heads and hidden_size to be ints:
```
num_heads (int, optional): number of attention heads. Defaults to 0.
        0 allows detect the parameter from graph automatically.
    hidden_size (int, optional): hidden size. Defaults to 0.
        0 allows detect the parameter from graph automatically.
```
        
If I convert `model_quantized.config.hidden_sizes` and `model_quantized.config.num_attention_heads` to integers by taking the max, the optimizer works and does seem to change the graph of the model

mht-sharma: @zachmayer,  curious to know which part of the model / encoder block is optimised. Since you gave max values I suppose the last one in the hierarchy? Or all encoder blocks are optimised?

If first, then it could be worth to try out having a loop on the optimiser for each pair of hidden size / attention head and see if all the blocks are optimised?

zachmayer: How do I tell which blocks get optimized?

I just pushed a commit that uses sum instead of max, and uses the `ORTModelForSemanticSegmentation` in the test, and the test passes.

I'll try some different parameters and see what changes.

zachmayer:
@mht-sharma @IlyasMoutawwakil the optimizer is definitely doing something to the model.  I tested it on the [vikp/surya_layout](https://huggingface.co/vikp/surya_layout) segformer, where I exported the original model to onnx, optimized it, and then quantized the optimized model and counted the number of nodes in the graph:

```python
Original model graph: 2900
Optimized model graph: 1263
Quantized model graph: 1709
```

The optimizer definitely prunes nodes from the graph and the resulting model is faster for inference when I test it.

zachmayer:
I tried 3 ways of handling the lists:
- Sum
- Max
- Replace the list with 0 (and let microsoft's optimizer infer based on the graph)

Sum/Max yield pretty similar results, so I went with sum.  0 did not seem to work well.

The tests pass on the PR now, and when I test this optimizer on a real segformer, it definitely makes changes to the model graph.

zachmayer:
@mht-sharma @IlyasMoutawwakil  what do you think?  The tests pass when I run them locally, and the optimizer seems to be able to reduce the size of the model a lot.  (Almost 60%)

zachmayer:
huh.  I also tried using 0, which the docs said would infer the number of heads based on the model graph.

I changed from `sum` to `max` and the values seem correct now: 8 for num_heads and 512 for hidden_size.

In my testing the optimize model definitely has a smaller graph and faster inference.  So the optimizer is doing something to the model.

optimum/utils/normalized_config.py
tianleiwu: num_heads and hidden_size is optional for onnxruntime transformers optimizer. If you are not sure, set them to 0.

You can use Netron to view the onnx model before/after optimization.

zachmayer: @tianleiwu — ok!  I just pushed a new commit where I set them to 0.

zachmayer:
@IlyasMoutawwakil I changed the normalized config to use 0 instead of the max.  How's it look now?

IlyasMoutawwakil (approved):

IlyasMoutawwakil:
LGTM ! sorry for the late reply, and thanks @zachmayer

zachmayer:
Awesome!  No worries and thank you!

zachmayer:
Hmm looks like I have some tests to fix

zachmayer:
Looks like at least on test failed because of dependencies which is odd.

Let me rebase off the main branch and see what happens

IlyasMoutawwakil:
the dep failures are not from this PR, pypi has some issues on some plateforms. merging 🤗

zachmayer:
awesome!  Thank you for the feedback and for helping me get this over the line!

PR #1825, opened 2024-04-23
fix falcon dummy input generator

eaidova:
# What does this PR do?

during enabling falcon40b export for openvino, I found that this model is not covered by dummy input generator configuration and failed with shape mismatch. This PR fixes that.

I propose these changed into optimum-intel, but the proper place for it here.

## Who can review?

### DIFF
--- a/optimum/utils/input_generators.py
+++ b/optimum/utils/input_generators.py
@@ -1090,11 +1090,12 @@ class FalconDummyPastKeyValuesGenerator(DummyPastKeyValuesGenerator):
             random_sequence_length_range=random_sequence_length_range,
             **kwargs,
         )
-        self.num_kv_heads = self.num_kv_heads = (
-            normalized_config.num_kv_heads
-            if (normalized_config.new_decoder_architecture or not normalized_config.multi_query)
-            else 1
-        )
+        if normalized_config.new_decoder_architecture and normalized_config.multi_query:
+            self.num_kv_heads = normalized_config.num_attention_heads
+        elif normalized_config.new_decoder_architecture and not normalized_config.multi_query:
+            self.num_kv_heads = normalized_config.num_kv_heads
+        else:
+            self.num_kv_heads = 1
         self.head_dim = self.hidden_size // self.num_attention_heads
 
     def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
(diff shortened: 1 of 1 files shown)


optimum/utils/input_generators.py
```diff
@@ -1090,11 +1090,12 @@ def __init__(
...
-            normalized_config.num_kv_heads
-            if (normalized_config.new_decoder_architecture or not normalized_config.multi_query)
-            else 1
-        )
+        if normalized_config.new_decoder_architecture and normalized_config.multi_query:
+            self.num_kv_heads = normalized_config.num_attention_heads
+        elif normalized_config.new_decoder_architecture and not normalized_config.multi_query:
+            self.num_kv_heads = normalized_config.num_kv_heads
+        else:
+            self.num_kv_heads = 1
```
michaelbenayoun, line 1098: This can be simplified as follows I think:
```suggestion
        if normalized_config.new_decoder_architecture and normalized_config.multi_query:
            self.num_kv_heads = normalized_config.num_attention_heads
        elif normalized_config.new_decoder_architecture:
            self.num_kv_heads = normalized_config.num_kv_heads
        else:
            self.num_kv_heads = 1
```

optimum/utils/input_generators.py
```diff
@@ -1090,11 +1090,12 @@ def __init__(
...
-            if (normalized_config.new_decoder_architecture or not normalized_config.multi_query)
-            else 1
-        )
+        if normalized_config.new_decoder_architecture and normalized_config.multi_query:
+            self.num_kv_heads = normalized_config.num_attention_heads
+        elif normalized_config.new_decoder_architecture:
+            self.num_kv_heads = normalized_config.num_kv_heads
+        else:
+            self.num_kv_heads = 1
         self.head_dim = self.hidden_size // self.num_attention_heads
```
echarlaix, line 1100: also should be udpated [here](https://github.com/huggingface/optimum/blob/3b5c486821628d1f1e4c34a59d097e5a5f4e1803/optimum/exporters/onnx/model_configs.py#L405) and [here](https://github.com/huggingface/optimum/blob/3b5c486821628d1f1e4c34a59d097e5a5f4e1803/optimum/onnxruntime/modeling_decoder.py#L852) as well

echarlaix: also why not have this value set in the `normalized_config` directly and use it in both `FalconDummyPastKeyValuesGenerator` and `ORTFalconForCausalLM` @fxmarty ? (removing the need to have this check in both places)

PR #1837, opened 2024-04-26
Deprecated `use_auth_token`

### LINKED ISSUE #1836: Pushing to the hub with a token
### System Info

```shell
optimum main
```

### Who can help?

@echarlaix @fxmarty 

### Information

- [ ] My own modified scripts

### Tasks

The current integration with the hub doesn't allow pushing to it with some tokens because it requires the user's `email` and `fullname` (not present with fine-grained tokens).
```python
user = api.whoami(huggingface_token)
self.git_config_username_and_email(git_email=user["email"], git_user=user["fullname"])
```
to reproduce, use a fine-grained token to push to a repo (that the token has write access to) or simply `whoami(token="fine-grained-token-you-created")`, the output won't contain the `email` field.

I'm deprecating `use_auth_token` in a PR and will fix this on my way.

### Expected behavior

to push to hub

IlyasMoutawwakil:
# What does this PR do?

This PR deprecates the use `use_auth_token` in favor of `token`.

Fixes #1836 

## Before submitting

## Who can review?

@fxmarty @echarlaix @mht-sharma


optimum/configuration_utils.py
```diff
@@ -205,6 +218,15 @@ def _get_config_dict(
         from_auto_class = kwargs.pop("_from_auto", False)
         commit_hash = kwargs.pop("_commit_hash", None)
 
+        if use_auth_token is not None:
+            warnings.warn(
```
fxmarty, line 222:
```suggestion
            logger.warning(
```

fxmarty: Interestingly Transformers also uses warnings.warn, not sure why the mix?

optimum/exporters/tasks.py
```diff
@@ -18,12 +18,14 @@
...
 import itertools
 import os
+import warnings
 from functools import partial
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Set, Tuple, Type, Union
 
 import huggingface_hub
 from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
+from huggingface_hub.utils import OfflineModeIsEnabled
```
fxmarty, line 28: Good catch, thank you for fixing. Maybe even move to `huggingface_hub.errors.OfflineModeIsEnabled`? https://github.com/huggingface/huggingface_hub/blob/v0.23.3/src/huggingface_hub/errors.py#L16

As I guess the `huggingface_hub.utils.OfflineModeIsEnabled` path is deprecated?

fxmarty (approved):
LGTM

optimum/configuration_utils.py
IlyasMoutawwakil: my guess is that this is specific for FutureWarning, should we use logger everywhere ?

IlyasMoutawwakil: I asked in https://huggingface.slack.com/archives/C01N44FJDHT/p1719313821448529
the answer is in https://stackoverflow.com/a/14762106 which refers to `logging` documentation

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

PR #1840, opened 2024-04-29
make stable diffusion unet and vae number of channels static

eaidova:
# What does this PR do?

Ported from https://github.com/huggingface/optimum-intel/pull/692
remove number of channels from unet and vae dynamic inputs and outputs. From model topology perspective, this value can not be different during inference from specified in config (changing number of channels requires to recreate pytorch model as number of channels in the first conv depends on it). Having them static will simplify model graph

## Before submitting

## Who can review?

 @echarlaix


echarlaix (approved):
LGTM !

PR #1866, opened 2024-05-22
Fix FX CI

IlyasMoutawwakil:
# What does this PR do?

Fixes CI failures

## Before submitting

### DIFF
--- a/optimum/fx/optimization/transformations.py
+++ b/optimum/fx/optimization/transformations.py
@@ -19,15 +19,11 @@ import itertools
 import operator
 import warnings
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, List
+from typing import List
 
 import torch
+from torch.fx import GraphModule, Node, Proxy
 from transformers.file_utils import add_end_docstrings
-from transformers.utils.fx import gen_constructor_wrapper
-
-
-if TYPE_CHECKING:
-    from torch.fx import GraphModule, Node
 
 
 _ATTRIBUTES_DOCSTRING = r"""
@@ -402,7 +398,7 @@ class FuseBiasInLinear(ReversibleTransformation):
     preserves_computation = True
 
     def transform(self, graph_module: "GraphModule") -> "GraphModule":
-        torch_ones = gen_constructor_wrapper(torch.ones)[0]
+        torch_ones = _gen_constructor_wrapper(torch.ones)[0]
 
         def insert_concat(linear_input):
             shape = linear_input.shape[:-1] + (1,)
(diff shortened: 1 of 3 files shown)


optimum/fx/optimization/transformations.py
```diff
@@ -803,3 +799,26 @@ def reverse(self, graph_module):
                 return ComposeTransformation._reverse_composition(graph_module)
 
     return ComposeTransformation()
+
+
+# was removed from transformers in favor of gen_constructor_wrapper which only works during tracing
+# TODO: should transformation be applied during tracing ? or by a torch.compile backend instead ?
+def _gen_constructor_wrapper(target):
```
echarlaix, line 806: do you think we can add `gen_constructor_wrapper` https://github.com/huggingface/transformers/blob/87a351818e6c680282d86a00b61fe4c55b41477d/src/transformers/utils/fx.py#L795 directly after v4.40.2 (included) ? cc @IlyasMoutawwakil @michaelbenayoun

echarlaix (approved):
Thanks a lot for the fix @IlyasMoutawwakil

optimum/fx/optimization/transformations.py
IlyasMoutawwakil: I went with copying the old function but apparently it's possible to force tracing mode with:
`torch.fx._symbolic_trace._is_fx_tracing_flag = True`
we can use both: 
- `_gen_constructor_wrapper` from transformers when it's available, 
- and a new one defined as:
```python
def _gen_constructor_wrapper(*args, **kwargs):
    original_flag = torch.fx._symbolic_trace._is_fx_tracing_flag
    torch.fx._symbolic_trace._is_fx_tracing_flag = True
    out = gen_constructor_wrapper(*args, **kwargs)
    torch.fx._symbolic_trace._is_fx_tracing_flag = original_flag 
    return out
```

michaelbenayoun: I think the above solution might not work no?
Because the test if we are tracing is inside the wrapper itself, so it needs to `is_fx_tracing` needs to be `True` when the wrapper is called.

IlyasMoutawwakil: I tested it and it works
````
import torch.fx._symbolic_trace
from torch.fx._symbolic_trace import is_fx_tracing

torch.fx._symbolic_trace._is_fx_tracing_flag = True
print(is_fx_tracing())
torch.fx._symbolic_trace._is_fx_tracing_flag = False
print(is_fx_tracing())
```
Returns True then False, ie inside the wrapper when `is_fx_tracing` will be called, it will return the flag value that we forced. I will implement the solution and test with the two versions of transformers.

IlyasMoutawwakil: I pushed a fix that wraps the wrapper output of `gen_constructor_wrapper`

michaelbenayoun (approved):

PR #1867, opened 2024-05-22
Fix Utils CI

IlyasMoutawwakil:
# What does this PR do?

Fixes tasks processor ci

## Before submitting


.github/workflows/test_utils.yml
```diff
@@ -0,0 +1,45 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04, macos-13]
+        python-version: [3.8, 3.9]
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
```
echarlaix, line 24: why is it needed ?

echarlaix (approved):

.github/workflows/test_utils.yml
```diff
@@ -0,0 +1,45 @@
...
+        if: matrix.os == 'ubuntu-20.04'
+        with:
+          tool-cache: true
```
michaelbenayoun, line 27: Why?

tests/utils/test_task_processors.py
```diff
@@ -157,11 +160,15 @@ def _test_load_dataset(
                 path,
                 data_keys=data_keys,
                 only_keep_necessary_columns=only_keep_necessary_columns,
+                load_smallest_split=True,
```
michaelbenayoun, line 163: Nice!

.github/workflows/test_utils.yml
echarlaix: also do you think we can reduce the model / dataset size used during testing https://github.com/huggingface/optimum/blob/main/tests/utils/test_dummpy_input_generators.py#L38 ? (using tiny random models for example)

IlyasMoutawwakil:
still fails after two datasets 🥲

.github/workflows/test_utils.yml
IlyasMoutawwakil: context https://huggingface.slack.com/archives/C05A8AV5J1M/p1716543338536269?thread_ts=1716404954.641349&cid=C05A8AV5J1M

IlyasMoutawwakil:
reducing the number of samples works (it was the tensors written to disk memory and not the dataset splits)

optimum/utils/preprocessing/task_processors_manager.py
```diff
@@ -35,7 +35,7 @@ class TaskProcessorsManager:
     }
 
     @classmethod
-    def get_task_processor_class_for_task(cls, task: str) -> Type:
+    def get_task_processor_class_for_task(cls, task: str) -> Type["TaskProcessor"]:
```
mht-sharma, line 38: Could we maintain the same style as in other funds?
```suggestion
    def get_task_processor_class_for_task(cls, task: str) ->"TaskProcessor":
```

mht-sharma (approved):
Thanks Ilyas for the fixes

optimum/utils/preprocessing/task_processors_manager.py
IlyasMoutawwakil: but one returns a constructor and the other returns an instance

mht-sharma: Got it, thanks for clarifying!

PR #1875, opened 2024-05-24
Fix ORT CI

IlyasMoutawwakil:
# What does this PR do?

Fixes ort ci failures due to 1.18

## Before submitting


IlyasMoutawwakil:
one test is still fail during generation with old model `optimum/gpt2`
https://github.com/huggingface/optimum/blob/main/tests/onnxruntime/test_modeling.py#L2277

.github/workflows/test_onnxruntime.yml
```diff
@@ -22,11 +22,9 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v2
-
-    - name: Free disk space
-      if: matrix.os == 'ubuntu-20.04'
-      run: |
+      - name: Free disk space
```
IlyasMoutawwakil, line 25: should we use an action to handle this instead
https://github.com/marketplace/actions/free-disk-space-ubuntu

IlyasMoutawwakil:
ort tests are passing locally, the issue with broadcasting was solved.
but for some reason output ids are matching locally and not on the runner (two tests with old onnx model)

echarlaix:
> but for some reason output ids are matching locally and not on the runner (two tests with old onnx model)

Do you know where this could come from ?

.github/workflows/test_onnxruntime.yml
michaelbenayoun: I do not have a strong opinion on this. Whatever is easier.

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -521,24 +519,47 @@ def _from_pretrained(
...
             input_dims["input_ids"][1] = "sequence_length"
-            output_dims = {
-                node.name: [dim.dim_value or dim.dim_param for dim in node.type.tensor_type.shape.dim]
-                for node in onnx_model.graph.output
-            }
             output_dims["logits"][1] = "sequence_length"
-            onnx_model = update_model_dims.update_inputs_outputs_dims(onnx_model, input_dims, output_dims)
+            override_dims = True
 
+        # Since https://github.com/huggingface/optimum/pull/871/files
```
michaelbenayoun, line 536: nit
```suggestion
        # Since https://github.com/huggingface/optimum/pull/871/
```

optimum/onnxruntime/quantization.py
```diff
@@ -356,62 +356,45 @@ def quantize(
                 )
 
         quantizer_factory = QDQQuantizer if use_qdq else ONNXQuantizer
+        # TODO: maybe this logic can be moved to a method in the configuration class (get_ort_quantizer_kwargs())
```
michaelbenayoun, line 359: But the config should not be aware of the ORTQuantizer class right?

IlyasMoutawwakil: yes, the quant config already contains everything and can infer which quantizer will use its kwargs (from format and is_static https://github.com/huggingface/optimum/blob/f3008651c6f674d4b89de66a2d21fc5e7cafaf84/optimum/onnxruntime/quantization.py#L309)

.github/workflows/test_onnxruntime.yml
```diff
@@ -22,62 +22,34 @@ jobs:
...
-
-    - name: Install dependencies
-      run: |
-        pip install .[tests,onnxruntime]
-
-    - name: Test with pytest
-      working-directory: tests
-      run: |
-        pytest -n auto -m "not run_in_series" --durations=0 -vs onnxruntime
-        pytest -m "run_in_series" --durations=0 onnxruntime
```
IlyasMoutawwakil, line 83: I changed the order here and started seeing new errors in windows related to input dtype that I didn't see before.
I just noticed but apparently depending on the os, errors propagate into the workflow differently;
in linux based runners (ubuntu), this will run the first command and exit with non-zero code 1 if it fails.
in windows based runners, this will run the first command, and then the second, whether the first succeeds or fails, and will only check the exit code of the last one.

instance:
- [this is a windows run that failed in first command, ran the second either ways and returned a success](https://github.com/huggingface/optimum/actions/runs/9268480991/job/25497227691#step:6:36481)
- [this is an ubuntu run (from the same event) that failed in the first and exited directly](https://github.com/huggingface/optimum/actions/runs/9268480991/job/25497227192#step:6:30184)

this is probably due to difference between bash and powershell
@echarlaix @michaelbenayoun @JingyaHuang @mht-sharma

.github/workflows/test_onnxruntime.yml
```diff
@@ -22,62 +22,34 @@ jobs:
...
+      - name: Free Disk Space (Ubuntu)
+        if: matrix.os == 'ubuntu-20.04'
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
+          swap-storage: false
+          large-packages: false
```
fxmarty, line 31: Why this change?

tests/onnxruntime/test_modeling.py
```diff
@@ -2274,21 +2274,25 @@ class ORTModelForCausalLMIntegrationTest(ORTModelTestMixin):
...
+
+        text = "The capital of France is"
+        tokens = tokenizer(text, return_tensors="pt")
+
+        onnx_outputs = onnx_model.generate(
+            **tokens, num_beams=1, do_sample=False, min_new_tokens=10, max_new_tokens=10
         )
-        outputs = model.generate(**tokens, num_beams=1, do_sample=False, min_new_tokens=30, max_new_tokens=30)
-        self.assertTrue(torch.allclose(outputs_onnx, outputs))
+        outputs = model.generate(**tokens, num_beams=1, do_sample=False, min_new_tokens=10, max_new_tokens=10)
```
fxmarty, line 2292: Why reducing the number of new tokens?

.github/workflows/test_onnxruntime.yml
IlyasMoutawwakil: I asked in https://github.com/huggingface/optimum/pull/1875?notification_referrer_id=NT_kwDOA2yBoLQxMDgwMTU5NTExMjo1NzQ0MjcyMA#discussion_r1613549361. 
It takes half the time (2-3 min -> 1 min) and shows how much space was freed in each step. Either way we won't need it when we start using intel-cpu runners.

tests/onnxruntime/test_modeling.py
IlyasMoutawwakil: was trying to figure out where the failing was coming from and forgot to reset it to 30. Will do that in the windows PR.

.github/workflows/test_onnxruntime.yml
fxmarty: nice

PR #1878, opened 2024-05-27
Fix GPTQ CI

IlyasMoutawwakil:
# What does this PR do?

Fixes failures from the non reproducibility and non determinism (or lack of deterministic kernels when activated) by checking expected perplexity instead of specific outputs.

## Before submitting


fxmarty (approved):
LGTM as long as it passes. Thank you for fixing

PR #1879, opened 2024-05-27
Fix quantize.py

mr-sarthakgupta:
There's a simple typo in the file quantize.py which comes up while converting onnx models to their quantized forms. Fixing that.


echarlaix (approved):
Thanks for the fix @mr-sarthakgupta

PR #1886, opened 2024-06-03
Fix Windows and onnx dtype compatibility

IlyasMoutawwakil:
# What does this PR do?

Fixes multiple windows specific issues:
- dtype mismatch: numpy uses a default int representation that depends on os. On ubuntu/macos it uses `int64` but on windows it uses `int32`. This results in tokenizers returning input ids and attention masks in a different format than torch's default which is int64/long.
- shutil.rmtree doesn't always work on windows, there's some random failures that we can see when a tempfile is being closed (cleaned up) or when directly trying to shutil.rmtree a folder.

Also seq2seq decoder used to return past kv cache in torch format all the time.

## Before submitting

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -1845,25 +1845,25 @@ class ORTModelForAudioClassification(ORTModel):
     ):
         if self.input_name == "input_features":
             assert input_features is not None, "input_features must be provided for this model"
-            main_input = input_features
+            model_input = input_features
         elif self.input_name == "input_values":
             assert input_values is not None, "input_values must be provided for this model"
-            main_input = input_values
+            model_input = input_values
         else:
             raise ValueError(f"Input {self.input_name} not supported for Audio Classification")
 
-        use_torch = isinstance(main_input, torch.Tensor)
+        use_torch = isinstance(model_input, torch.Tensor)
         self.raise_on_numpy_input_io_binding(use_torch)
 
         if attention_mask is None:
             if use_torch:
-                attention_mask = torch.ones_like(input_values)
+                attention_mask = torch.ones_like(model_input)
             else:
-                attention_mask = np.ones_like(input_values)
+                attention_mask = np.ones_like(model_input)
 
         if self.device.type == "cuda" and self.use_io_binding:
             io_binding, output_shapes, output_buffers = self.prepare_io_binding(
-                main_input,
+                model_input,
                 attention_mask,
                 ordered_input_names=self._ordered_input_names,
             )
@@ -1875,7 +1875,7 @@ class ORTModelForAudioClassification(ORTModel):
 
             logits = output_buffers["logits"].view(output_shapes["logits"])
         else:
-            model_inputs = {self.input_name: main_input, "attention_mask": attention_mask}
+            model_inputs = {self.input_name: model_input, "attention_mask": attention_mask}
 
             onnx_inputs = self._prepare_onnx_inputs(use_torch, **model_inputs)
             onnx_outputs = self.model.run(None, onnx_inputs)
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -1727,7 +1727,6 @@ class ORTModelForSemanticSegmentation(ORTModel):
             checkpoint="optimum/segformer-b0-finetuned-ade-512-512",
         )
     )
-
     def forward(
         self,
         pixel_values: Union[torch.Tensor, np.ndarray],
@@ -1737,13 +1736,12 @@ class ORTModelForSemanticSegmentation(ORTModel):
         self.raise_on_numpy_input_io_binding(use_torch)
 
         if self.device.type == "cuda" and self.use_io_binding:
-            io_binding = IOBindingHelper.prepare_io_binding(
-                self,
+            io_binding, output_shapes, output_buffers = self.prepare_io_binding(
                 pixel_values,
                 ordered_input_names=self._ordered_input_names,
             )
 
-            # run inference with binding
+            # run inference with binding & synchronize in case of multiple CUDA streams
             io_binding.synchronize_inputs()
             self.model.run_with_iobinding(io_binding)
             io_binding.synchronize_outputs()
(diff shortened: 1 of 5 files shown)


tests/onnxruntime/test_modeling.py
```diff
@@ -186,7 +186,10 @@ def test_load_model_from_empty_cache(self):
         dirpath = os.path.join(default_cache_path, "models--" + self.TINY_ONNX_MODEL_ID.replace("/", "--"))
 
         if os.path.exists(dirpath) and os.path.isdir(dirpath):
-            shutil.rmtree(dirpath)
+            if os.name == "nt":
+                os.system(f"rmdir /S /Q {dirpath}")
+            else:
+                shutil.rmtree(dirpath)
```
tengomucho, line 192: why do you need this on windows? Also, you might want to consider isolating this code and make a function, to avoid repeating it.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -267,10 +267,14 @@ def __init__(
             **kwargs,
         )
 
-        self.inputs_names = {input_key.name: idx for idx, input_key in enumerate(model.get_inputs())}
+        # why are these not lists ?
+        self.input_names = {input_key.name: idx for idx, input_key in enumerate(model.get_inputs())}
```
fxmarty, line 271: Changing the variable names makes the diff for this PR bigger, which makes it a bit harder to review

tests/onnxruntime/test_modeling.py
IlyasMoutawwakil: it's this error https://github.com/huggingface/optimum/actions/runs/9366908110/job/25785437479#step:7:30120
some refs: https://github.com/python/cpython/issues/107408

optimum/onnxruntime/modeling_ort.py
```diff
@@ -1131,45 +1161,34 @@ def forward(
         use_torch = isinstance(input_ids, torch.Tensor)
         self.raise_on_numpy_input_io_binding(use_torch)
 
+        if attention_mask is None:
+            if use_torch:
+                attention_mask = torch.ones_like(input_ids)
+            else:
+                attention_mask = np.ones_like(input_ids)
```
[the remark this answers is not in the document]
IlyasMoutawwakil: but that makes the behavior different than transformers, and we do create them on the fly for the feature extraction class for example:
https://github.com/huggingface/optimum/blob/ac951ca0ae49983b1515b84342fea15c3b7ec35c/optimum/onnxruntime/modeling_ort.py#L995
I have already tried to do that and discussed it in https://github.com/huggingface/optimum/pull/1045 but the PR was merged too early with only one class updated.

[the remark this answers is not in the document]
fxmarty: This does not work for batched inputs.

optimum/onnxruntime/modeling_ort.py
IlyasMoutawwakil: yeah the issue is that there were naming discrepancies which made little sense to me, especially between `ORTModel` and `ORTModelPart`
https://github.com/huggingface/optimum/blob/ac951ca0ae49983b1515b84342fea15c3b7ec35c/optimum/onnxruntime/modeling_ort.py#L270
https://github.com/huggingface/optimum/blob/ac951ca0ae49983b1515b84342fea15c3b7ec35c/optimum/onnxruntime/base.py#L54
is that intended ?
I think there might be a way to remove these naming changes, which is by fixing dtype in the seq2seq modeling classes instead of model parts. tell me if that makes more sense to you.

fxmarty: most likely not

optimum/onnxruntime/modeling_ort.py
```diff
@@ -969,51 +1013,37 @@ def forward(
...
-                onnx_inputs["token_type_ids"] = token_type_ids
+            onnx_inputs = self._prepare_onnx_inputs(use_torch, **model_inputs)
+            onnx_outputs = self.model.run(None, onnx_inputs)
+            model_outputs = self._prepare_onnx_outputs(use_torch, *onnx_outputs)
 
-            outputs = self.model.run(None, onnx_inputs)
+            # why do we only return last_hidden_state? why not all outputs?
+            # that way, there will be less need for ORTModelForCustomTask in cases where
+            # we just want to extend model outputs with attentions, hidden_states, etc.
+            last_hidden_state = model_outputs["last_hidden_state"]
```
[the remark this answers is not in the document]
IlyasMoutawwakil: I added a todo to it ! I think it could be cool to support arbitrary i/o.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -852,6 +863,39 @@ def raise_on_numpy_input_io_binding(self, use_torch: bool):
                 " with model.use_io_binding = False, or pass torch.Tensor inputs instead."
             )
 
+    def _prepare_onnx_inputs(
```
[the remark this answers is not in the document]
IlyasMoutawwakil: it only applies when needed so for performance I think there's no added overhead vs what used to be done.
where things can be optimized and i think can be viewed as the optimization oriented path, is i/o binding:
- pre-allocation of output buffers, not during forward but before that, and dynamically augmenting the size of output buffers when batch size changes.
- decoder models with i/o binding where static cache implementation can be used as output buffers to reduce the overhead of their creation.
- decoder models input/output synchronization which can be before and after the generation loop instead of each forward call (if that's possible).

tests/onnxruntime/test_modeling.py
tengomucho: So if it is a workaround for a python issue, I believe it's even more important to isolate it in a function that should be well explained, so if one day we go back to this code we do not ask ourselves "why it is done this way"?

IlyasMoutawwakil: done !

optimum/onnxruntime/modeling_ort.py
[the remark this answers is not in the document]
fxmarty: I kind of agree, just not sure what each ORTModelForXXX is for then

tengomucho (approved):

fxmarty (approved):

tengomucho:
FWIW, numpy 2.0 has just been released, and int size in windows is now 64 bit! https://numpy.org/devdocs/numpy_2_0_migration_guide.html#windows-default-integer

PR #1887, opened 2024-06-03
Fix incorrect names for usage blenderbot for causallm

eaidova:
# What does this PR do?

in general case, number of encoder and decoder layers in seq2seq models like blenderbot, m2m, bart can be different from encoder.
As example blenderbot-3b - https://huggingface.co/facebook/blenderbot-3B/blob/main/config.json
If we use these models for causal lm case, where used only decoder, output names can be incorrect due to unequal number of layers in decoder and encoder (that is used for preparing inputs if model has past_key_values).

This PR fixes this issue

## Before submitting

## Who can review?


michaelbenayoun:
Can you run `make style` and push please. It is necessary to format the code.

fxmarty (approved):

eaidova:
@fxmarty  @michaelbenayoun @echarlaix  is it possible to merge these changes? We are blocked to convert and optimize this model on openvino side, thanks!

fxmarty:
@eaidova Done thank you, apology for the delay!

PR #1894, opened 2024-06-05
Allow optimum to discover and load subpackages

dacorvo:
# What does this PR do?

This introduces explicitly the concept of `optimum` subpackages that are loaded "dynamically" by the `optimum-cli` (and possibly in the future by other high-level abstraction modules that would require backend-specific implementations).

An `optimum` subpackage is primarily a package under the `optimum` namespace, but it can be also a module inside the `optimum` package itself, even if that module does not have its own `pyproject.toml` or `setup.py` (as a real namespace package would).

By convention, loading a subpackage means importing a `subpackage` module available at the root of the package.

This pull-request first adds a `load_subpackages` helper that will go through the installed namespace packages and try to "load" them.

It then adds a decorator to declare custom subcommands, using the same convention as the one already used in the `register` files.

This is by using this decorator that each subpackage can add its own subcommands, as it will be called by the `optimum-cli` when evaluating its commands.

In order to validate the new helpers, the `onnxruntime` commands are moved inside an `onnxruntime/subpackage` directory that is explicitly loaded by `load_subpackages`, but only it the `onnxruntime` package is available.

This improves the UX when using the `optimum-cli`:
- commands declared as subpackages will not disappear if the content of the register directory is wiped-out by a new `optimum` installation,
- the `onnxruntime` commands will only be available if the package is actually installed.

### DIFF
--- a/optimum/commands/__init__.py
+++ b/optimum/commands/__init__.py
@@ -16,4 +16,4 @@ from .base import BaseOptimumCLICommand, CommandInfo, RootOptimumCLICommand
 from .env import EnvironmentCommand
 from .export import ExportCommand, ONNXExportCommand, TFLiteExportCommand
 from .onnxruntime import ONNXRuntimeCommand, ONNXRuntimeOptimizeCommand, ONNXRuntimeQuantizeCommand
-from .optimum_cli import register_optimum_cli_subcommand
+from .optimum_cli import optimum_cli_subcommand
--- a/optimum/commands/__init__.py
+++ b/optimum/commands/__init__.py
@@ -15,5 +15,4 @@
 from .base import BaseOptimumCLICommand, CommandInfo, RootOptimumCLICommand
 from .env import EnvironmentCommand
 from .export import ExportCommand, ONNXExportCommand, TFLiteExportCommand
-from .onnxruntime import ONNXRuntimeCommand, ONNXRuntimeOptimizeCommand, ONNXRuntimeQuantizeCommand
 from .optimum_cli import optimum_cli_subcommand
--- a/optimum/onnxruntime/subpackage/commands/quantize.py
+++ b/optimum/onnxruntime/subpackage/commands/quantize.py
@@ -17,7 +17,7 @@
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from .. import BaseOptimumCLICommand
+from optimum.commands import BaseOptimumCLICommand
 
 
 if TYPE_CHECKING:
@@ -69,8 +69,8 @@ class ONNXRuntimeQuantizeCommand(BaseOptimumCLICommand):
         return parse_args_onnxruntime_quantize(parser)
 
     def run(self):
-        from ...onnxruntime.configuration import AutoQuantizationConfig, ORTConfig
-        from ...onnxruntime.quantization import ORTQuantizer
+        from ...configuration import AutoQuantizationConfig, ORTConfig
+        from ...quantization import ORTQuantizer
 
         if self.args.output == self.args.onnx_model:
             raise ValueError("The output directory must be different than the directory hosting the ONNX model.")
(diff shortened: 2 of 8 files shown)


optimum/onnxruntime/subpackage/commands/quantize.py
```diff
@@ -17,7 +17,7 @@
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from .. import BaseOptimumCLICommand
+from optimum.commands import BaseOptimumCLICommand
```
fxmarty, line 20: why absolute path?

optimum/subpackages.py
```diff
@@ -0,0 +1,81 @@
...
+
+    This method goes through packages inside the `optimum` namespace and loads the `subpackage` module if it exists.
+
+    This module is then in charge of registering the subpackage commands.
+    """
+    SUBPACKAGE_LOADER = "subpackage"
+    load_namespace_modules("optimum", SUBPACKAGE_LOADER)
+
+    # Load subpackages from internal modules not explicitly defined as namespace packages
+    loader_name = "." + SUBPACKAGE_LOADER
```
fxmarty, line 79: What is `SUBPACKAGE_LOADER`?

optimum/commands/__init__.py
```diff
@@ -15,5 +15,4 @@
 from .base import BaseOptimumCLICommand, CommandInfo, RootOptimumCLICommand
 from .env import EnvironmentCommand
 from .export import ExportCommand, ONNXExportCommand, TFLiteExportCommand
-from .onnxruntime import ONNXRuntimeCommand, ONNXRuntimeOptimizeCommand, ONNXRuntimeQuantizeCommand
-from .optimum_cli import register_optimum_cli_subcommand
```
fxmarty, line 19: Would need to make sure `register_optimum_cli_subcommand` is not used in subpackages

fxmarty (approved):
LGTM, apart that `optimum-cli` has a 1-3s delay to start running (already on main) on my laptop, which should be ideally improved in a future PR

optimum/onnxruntime/subpackage/commands/quantize.py
dacorvo: That's mostly because the relative path would have been complicated, but also to serve as a model for actual subpackages that can live outside the optimum tree.

optimum/subpackages.py
dacorvo: That's the name of the module that is imported in each subpackage to "load" them. I used a constant because I use it twice and was expecting maybe some discussion about the naming.

optimum/commands/__init__.py
dacorvo: Yes, and I was very much tempted to make it private to `optimum-cli`, actually.
There are no calls in `optimum` and I quickly checked the `optimum-XX` to verify that none of them was calling it either, but I was not 100 % sure.

optimum/subpackages.py
```diff
@@ -0,0 +1,81 @@
...
+
+import importlib
+import logging
+import sys
+
+
+if sys.version_info >= (3, 8):
+    from importlib import metadata as importlib_metadata
+else:
+    import importlib_metadata
```
michaelbenayoun, line 24: That raises the question for later: what is the minimal python version we want to support. I have seen different things accross differents huggingface libraries.

optimum/commands/__init__.py
dacorvo: Existing subpackages are using the "register" method at the moment.

michaelbenayoun (approved):
That is super cool!
Thanks for this improvement @dacorvo

optimum/subpackages.py
tengomucho: Python 3.8 is the oldest supported, EOL is october 2024: https://devguide.python.org/versions/
I guess we could only support 3.8 and newer.

tengomucho (approved):

PR #1899, opened 2024-06-07
feat(ci): add trufflehog secrets detector

McPatate:
### What does this PR do?

Adding a GH action to scan for leaked secrets on each commit.


fxmarty (approved):
Does this scan HF Hub token as well?

McPatate:
Yes! Full list of "detectors" here: https://github.com/trufflesecurity/trufflehog/tree/main/pkg/detectors

PR #1900, opened 2024-06-07
Table transformer

zachmayer:
# What does this PR do?
Add support for the table transformer in the optimizer class: https://github.com/huggingface/optimum/issues/1761

The table transformer does not support `ORTModelForImageClassification` and there's no `ORTModelForObjectDetection`, so I just used ORTModelForCustomTasks.

Maybe a good follow-up task would be to add `ORTModelForObjectDetection` and possibly also `ORTModelForImageFeatureExtraction` (as `ORTModelForFeatureExtraction` only supports text models)

I ran this on `microsoft/table-transformer-detection` and confirmed it optimizes the model:

Onnx model graph: 2247
Opt model graph: 1136

I see a very small speedup in inference time.  
Onnx Model Time: 65.42 reps per second
Optimized Onnx Model Time: 67.34 reps per second

The optimized model binary is slightly smaller than the original model. (115.0mb vs 115.1mb)

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/onnxruntime/utils.py
+++ b/optimum/onnxruntime/utils.py
@@ -129,6 +129,7 @@ class ORTConfigManager:
         "pegasus": "bert",
         "roberta": "bert",
         "segformer": "vit",
+        "table-transformer": "vit",
         "t5": "bert",
         "vit": "vit",
         "whisper": "bart",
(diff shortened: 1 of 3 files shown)


zachmayer:
@mht-sharma — if you have a minute to review my PR, I would appreciate it!

mht-sharma (changes requested):
@zachmayer  thanks for the PR. I suppose having an `ORTModelForObjectDetection` would be better than using `ORTModelForCustomTasks` since this is more generic. Could you please add a class for this?

PR #1910, opened 2024-06-17
Remove dataset with restrictive license

echarlaix:
cc @lhoestq


SunMarc (approved):
SGTM ! Thx for updating the datasets !

SunMarc:
Thanks for the quick review @SunMarc, let me know if you want me to open a PR on transformers as well

> Yes, that would be great !

PR #1917, opened 2024-06-21
Read use_external_data_format from ORTConfig file

### LINKED ISSUE #1916: Quantization by optimum-cli ignores 'use_external_data_format' flag in ORTConfig and fails to quantize models >2GB
### System Info

```shell
optimum 1.20.0
python 3.8
onnxruntime 1.16.3
Linux
```

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

Launching a command:
```
optimum-cli onnxruntime quantize -c ort_config.json --onnx_model ../mistral --output .
```
where `../mistral` is where an exported ONNX Mistral model (14Gb) is located and the content of ort_config.json is
```
{
    "use_external_data_format": true,
    "quantization": {
        "is_static": false,
        "mode": "IntegerOps",
        "per_channel": true,
        "format": "QDQ"
    }
}
```
Since the size of ONNX model >2Gb I set use_external_data_format=true to prevent 
```
ValueError: Message onnx.ModelProto exceeds maximum protobuf size of 2GB
```
failure. In fact, the passed-in parameter `use_external_data_format` of `def quantize()` in optumum's quantization.py is `False` as if the value in the configuration file is ignored. The result is that the quantization fails due to model size >2Gb!

### Expected behavior

The value of `use_external_data_format` argument of `quantize()` function must be as in the configuration file.

idruker-cerence:
When quantizing the models >2Gb, it's important to set the flag use_external_data_format to 'true',
since otherwise the quantization will fail due to

`ValueError: Message onnx.ModelProto exceeds maximum protobuf size of 2GB`

However, currently there is no way to set the parameter when using optimum-cli because there is no such
command option. Theoretically, it could be set when using ORTConfig file with -c comman flag, because one of the configuration parameters in it is use_external_data_format. In fact, the optimum code ignores it and does not pass it in quantize() function.

The goal of this change is to close this gap.

# What does this PR do?

Fixes #1916

## Before submitting

### DIFF
--- a/optimum/onnxruntime/subpackage/commands/quantize.py
+++ b/optimum/onnxruntime/subpackage/commands/quantize.py
@@ -77,6 +77,7 @@ class ONNXRuntimeQuantizeCommand(BaseOptimumCLICommand):
 
         save_dir = self.args.output
         quantizers = []
+        use_external_data_format = False
 
         quantizers = [
             ORTQuantizer.from_pretrained(self.args.onnx_model, file_name=model.name)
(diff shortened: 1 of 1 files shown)


fxmarty (approved):
LGTM thank you!

IlyasMoutawwakil:
Thanks @idruker-cerence 
can you please run `make style` (after `pip install -e .[quality]`)

PR #1933, opened 2024-07-01
Automatic Model Parallelism Through FX

zhenglongjiepheonix:
# What does this PR do?
This PR tries to add an automatic parallelization backend for torch dynamo, which takes the dynamo-captured fx graph, runs a few passes to automatically identify parts that can be parallelized and transforms the graph into its parallelized version. For simplicity it focuses on models supporting dynamo tracing in transformers library right now and might not support custom models because of the tricky parts in parallel pattern matching.

For now it only supports parallelization of linears in the graph, in the context of transformers they would be attention layers and mlp layers, with the following milestones left:

- [ ] support tensor parallelism on loss layers(need models support training-mode trace)
- [x] support weights loading from disk/hub
- [ ] support fusion of parallel linears(qkv fusion, mlp fusion)
- [ ] try supporting sequence parallelism

~~Please feel free to review and provide suggestions even if it's still in progress and not covering all features.~~
According to @michaelbenayoun , we should try merging the first version and iterations will be coming in following PRs


optimum/fx/parallelization/utils.py
```diff
@@ -0,0 +1,211 @@
...
+    if node.op != 'call_module':
+        return False
+    mod = node.graph.owning_module
+    return isinstance(mod.get_submodule(node.target), nn.Linear)
+
+def is_shape_consumer(node: Node) -> bool:
+    if node.op == 'call_method':
+        return node.target in {'view', 'reshape', 'expand', 'resize', 'resize_'}
+    elif node.op == 'call_function':
+        return node.target in {torch.reshape}
```
michaelbenayoun, line 28:
```suggestion
        return node.target in {torch.reshape}
    else:
        return False
```

optimum/fx/parallelization/utils.py
```diff
@@ -0,0 +1,211 @@
...
+    return False
+
+def is_getitem(node: Node) -> bool:
+    return node.op == 'call_function' and node.target is operator.getitem
+
+def is_output(node: Node) -> bool:
+    return node.op == 'output'
+
+def is_shape_generator(node: Node) -> bool:
+    return node.op == 'call_method' and node.target == 'size'
```
michaelbenayoun, line 51: But `tensor.shape` would not get caught here no?

optimum/fx/parallelization/utils.py
```diff
@@ -0,0 +1,211 @@
...
+
+def initialize_parameter_mapping(model: nn.Module, ctx: ParallelExecutionCtx) -> None:
+    mapping = ctx.parameter_mapping
+
+    for name, tensor in chain(model.named_parameters(), model.named_buffers()):
+        mapping[id(tensor)] = ParameterMapping(id = id(tensor), meta = ParameterMeta(source=name))
+
+def move_model_to_device(model: nn.Module, device: Union[torch.device, str]):
+    # move everything except tensors on meta devices on current device
+    # this function should be called before `intialize_parameter_mapping`
```
michaelbenayoun, line 196: Use docstring instead

optimum/fx/parallelization/distributed/dist_ops.py
```diff
@@ -0,0 +1,113 @@
...
+    return tensor
+
+def all_gather(group: dist.ProcessGroup, tensor: torch.Tensor, gather_dim: int = -1) -> torch.Tensor:
+    world_size = dist.get_world_size(group)
+    if world_size == 1:
+        return tensor
+    rank = dist.get_rank(group = group)
+
+    tensor = tensor.contiguous()
+    tensors = [torch.empty_like(tensor) for _ in range(world_size)]
```
fxmarty, line 19: I have not tried it, but may be https://pytorch.org/docs/stable/distributed.html#torch.distributed.all_gather_into_tensor is more efficient with a single `empty` call, like https://github.com/huggingface/text-generation-inference/blob/d0225b10156320f294647ac676c130d03626473d/server/text_generation_server/layers/tensor_parallel.py#L98

optimum/fx/parallelization/distributed/dist_ops.py
```diff
@@ -0,0 +1,113 @@
...
+def split(group: dist.ProcessGroup, tensor: torch.Tensor, split_dim: int = -1) -> torch.Tensor:
+    world_size = dist.get_world_size(group)
+    if world_size == 1:
+        return tensor
+
+    rank = dist.get_rank(group)
+    size = tensor.size()
+    assert size[split_dim] % world_size == 0
+    tensors = torch.split(tensor, size[split_dim] // world_size, dim = split_dim)
+    tensor = tensors[rank].contiguous()
```
fxmarty, line 34: why contiguous?

optimum/fx/parallelization/parallel_layers/linear.py
```diff
@@ -0,0 +1,191 @@
...
+                ),
+                parallel_dim=0
+            )
+
+            parameter_mapping.pop(key)
+            parameter_mapping[linear_bias_mapping.id] = linear_bias_mapping
+            self.bias.zero_()
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        input = differentiable_identity(input, self.process_group)
```
fxmarty, line 100: why do we need an identity here?

optimum/fx/parallelization/passes.py
```diff
@@ -0,0 +1,449 @@
...
+        if key in node.meta:
+            node.meta.pop(key)
+
+    def clean_all(self, graph_module: GraphModule) -> None:
+        g : Graph = graph_module.graph
+        for node in g.nodes:
+            self.clear_marker_per_node(node)
+
+
+class ParallelLinearAnnotatePass(AnalyzeBase):
```
fxmarty, line 108: I don't get the name parallel here? Isn't it more like successive?

optimum/fx/parallelization/passes.py
```diff
@@ -0,0 +1,449 @@
...
+                    self.place_marker_per_node(node, {'parallel_axis' : parallel_axis})
+        return graph_module
+
+
+class ParallelLinearReplacePass(PassBase):
+    """
+    A pass which modifies graph according to information provided by previous analytical passes,
+    in general it does two things for now:
+        1. replace linears with their parallel counterparts.
+        2. modify hard-coded arguments like the number of attenton heads in the graph by dividing it by parallelism level.
```
fxmarty, line 330:
```suggestion
        2. modify hard-coded arguments like the number of attention heads in the graph by dividing it by parallelism level.
```

tests/fx/parallelization/test_tensor_parallel.py
```diff
@@ -0,0 +1,190 @@
...
+
+    model = torch.compile(model, fullgraph=True, backend=partial(parallelize_backend, ctx=ctx, config=cfg))
+    logits = model(**inputs)[0]
+    tensors = gather_at_main_process(tensor=logits, group=tp_group, rank=rank, world_size=world_size)
+
+    # check results at main worker process
+    if rank == 0:
+        assert len(tensors) == world_size
+        for i in range(1, world_size):
+            torch.testing.assert_close(tensors[i - 1].cpu(), tensors[i].cpu(), rtol=1e-4, atol=1e-4)
```
fxmarty, line 73: it should probably be checked on all ranks

tests/fx/parallelization/test_tensor_parallel.py
```diff
@@ -0,0 +1,190 @@
...
+    ctx = ParallelExecutionCtx(tp_group=tp_group, current_device=device)
+    with MetaAwareMethodsPatcher():
+        model = model_cls(model_config)
+        model.eval()
+
+    # move model to current device, with linears still on meta
+    move_model_to_device(model, device=device)
+    initialize_parameter_mapping(model, ctx=ctx)
+
+    model = torch.compile(model, fullgraph=True, backend=partial(parallelize_backend, ctx=ctx, config=cfg))
```
fxmarty, line 164: can we compose with inductor?

fxmarty (commented):
this is super cool!

optimum/fx/parallelization/parallel_layers/linear.py
zhenglongjiepheonix: to take care of gradient reduce in backward

optimum/fx/parallelization/utils.py
zhenglongjiepheonix: `.shape` will be traced as `size`

tests/fx/parallelization/test_tensor_parallel.py
zhenglongjiepheonix: check at the main process should be enough, because it gathers results from other ranks at main process and does comparison

optimum/fx/parallelization/passes.py
```diff
@@ -0,0 +1,449 @@
...
+        if field not in info:
+            raise KeyError(f"Invalid query field {field} for {cls.__name__}, valid fields are {list(info.keys())}")
+
+        return info[field]
+    
+    @classmethod
+    def already_executed_per_node(cls, node: Node) -> None:
+        return cls.meta_key() in node.meta
+    
+    def place_marker_per_node(self, node: Node, info: Dict[Any, Any]) -> None:
```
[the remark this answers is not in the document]
zhenglongjiepheonix: this function does not return anything

optimum/fx/parallelization/passes.py
```diff
@@ -0,0 +1,449 @@
...
+        for node in upstream_nodes:
+            self.place_marker_per_node(node, {'axis' : 'column', 'gather_output' : False if downstream_nodes else True})
+
+        for node in downstream_nodes:
+            self.place_marker_per_node(node, {'axis' : 'row', 'input_is_parallel' : True})
+
+
+    def run(self, graph_module: GraphModule, ctx: ParallelExecutionCtx, config: Config) -> GraphModule:
+        graph: Graph = graph_module.graph
+        stable_topological_sort(graph)
```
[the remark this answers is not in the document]
zhenglongjiepheonix: yeah, just be precautious, it won't take much time anyway

optimum/fx/parallelization/__init__.py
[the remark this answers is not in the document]
zhenglongjiepheonix: Sure

optimum/fx/parallelization/distributed/dist_ops.py
zhenglongjiepheonix: yes

optimum/fx/parallelization/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+from .core import ParallelExecutionCtx, Config
+from .passes import build_parallel_pass_pipeline
+
+
+def parallelize_backend(graph_module: GraphModule, example_inputs: List[torch.Tensor], ctx: ParallelExecutionCtx, config: Config):
+    ctx.example_inputs = example_inputs
+    pass_pipeline = build_parallel_pass_pipeline()
+    graph_module = pass_pipeline(graph_module=graph_module, ctx=ctx, config=config)
+    ctx.compile_times += 1
+    return graph_module
```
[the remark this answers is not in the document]
zhenglongjiepheonix: actually I don't know, where normally should APIs be exposed?

optimum/fx/parallelization/distributed/dist_ops.py
zhenglongjiepheonix: tensors after `split` may not be contiguous, I think it's better be contiguous

optimum/fx/parallelization/passes.py
```diff
@@ -0,0 +1,449 @@
...
+
+        if len(node.args) == 3:
+            if parallel_axis not in node.args and parallel_axis - dims not in node.args:
+                return False
+            for arg in node.args:
+                if isinstance(arg, int) and (arg + dims) % dims != parallel_axis:
+                    self.place_marker_per_node(node, {'parallel_axis' : (arg + dims) % dims})
+                    return True
+        
+        return False
```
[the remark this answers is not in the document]
zhenglongjiepheonix: Sure

optimum/fx/parallelization/passes.py
```diff
@@ -0,0 +1,449 @@
...
+                return True
+            return False
+
+        if len(node.args) == 3:
+            if parallel_axis not in node.args and parallel_axis - dims not in node.args:
+                return False
+            for arg in node.args:
+                if isinstance(arg, int) and (arg + dims) % dims != parallel_axis:
+                    self.place_marker_per_node(node, {'parallel_axis' : (arg + dims) % dims})
+                    return True
```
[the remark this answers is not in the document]
zhenglongjiepheonix: already merged logic

optimum/fx/parallelization/passes.py
zhenglongjiepheonix: it actually means annotate some layers to be their parallel counterparts

optimum/fx/parallelization/distributed/dist_ops.py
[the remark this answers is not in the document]
IlyasMoutawwakil: That would be great ! maybe even the torch native parallelism layers.

optimum/fx/parallelization/__init__.py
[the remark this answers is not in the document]
michaelbenayoun: I think you can put that in `core.py` and just expose it here.

tests/fx/parallelization/test_tensor_parallel.py
michaelbenayoun: I would say that is the hope for the future.

optimum/fx/parallelization/parallel_layers/embedding.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class VocabParallelEmbedding(nn.Module):
+    """
+    Embedding layer parallelized in vocabulary dimension.
+
+    Arguments:
+        ctx: parallel execution context which contains runtime information.
+        embedding: the original embedding module being replaced.
+        init_fn: weight initialization function.
```
michaelbenayoun, line 35: Can you add type annotations please.

optimum/fx/parallelization/api.py
```diff
@@ -0,0 +1,155 @@
...
+    local_files_only: bool = False,
+    skip_load_weights: bool = False,
+    **kwargs,
+):
+    """
+    API for automatic model parallelism through Pytorch FX.
+
+    Args:
+        model (Union[torch.nn.Module, str]):
+            Model to parallelize, could either be a module or a model id in huggingface space.
```
michaelbenayoun, line 62:
```suggestion
        model (Union[torch.nn.Module, str]):
            Model to parallelize, could either be a module or a model id on the Hugging Face Hub.
```

optimum/fx/parallelization/api.py
```diff
@@ -0,0 +1,155 @@
...
+        parallel_ctx (ParallelExecutionCtx):
+            Parallel execution context containing process groups the current process belongs to.
+        model_args (additional postional arguments, optional):
```
michaelbenayoun, line 65:
```suggestion
        *model_args (Any):
```
Should we add also `model_kwargs`?

optimum/fx/parallelization/api.py
```diff
@@ -0,0 +1,155 @@
...
+            Additional postional arguments for intializing the model if a model id is passed.
+        revision (`str`, defaults to `main`):
+            Model revision for weights downloading if a model id is passed.
+        cache_dir (`Optional[str]`, defaults to `None`):
+            Cache directory to store downloaded weights. Defaults to None.
+        local_files_only (`bool`, defaults to `False`):
+            Whether to use local files only, will avoid downloading from remote if set to `True`.
+        skip_load_weights (`bool`, defaults to `False`):
+            Whether to skip loading weights from disk to model.
+        kwargs (additional keyword arguments, optional):
```
michaelbenayoun, line 75:
```suggestion
        **kwargs (Dict[str, Any]):
```

optimum/fx/parallelization/api.py
```diff
@@ -0,0 +1,155 @@
...
+                cache_dir=cache_dir,
+                allow_patterns=allow_patterns,
+                revision=revision,
+                local_files_only=local_files_only,
+                skip_download_weights=skip_load_weights,
+            )
+        else:
+            hf_folder = model
+
+        # should be able to load config using only local files
```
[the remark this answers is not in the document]
zhenglongjiepheonix: here I move all the dowload logic including config and index files into `download_model_from_hf`

optimum/fx/parallelization/api.py
```diff
@@ -0,0 +1,155 @@
...
+            # try directly construct weight_map from weight files, should have safetensors file on disk in any case
+            if not parallel_ctx.weight_map:
+                from safetensors import safe_open
+
+                weight_map, weight_files = {}, glob.glob(os.path.join(hf_folder, "*.safetensors"))
+                for weight_file in weight_files:
+                    with safe_open(filename=weight_file, framework="pt") as f:
+                        for key in f.keys():
+                            weight_map[key] = weight_file
+                parallel_ctx.weight_map = weight_map
```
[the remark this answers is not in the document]
zhenglongjiepheonix: I move the logic into `utils` so it looks cleaner in `api`, but the logic itself is indeed complex, because we need to take care of situations where a local directory is passed, then the only thing we can do is to try to peek inside the folder and see if there are safetensors/bin files, and if there are only bin files, we need to convert them into safetensors, and if there is an index file, we load the `weight_map` directly from it, otherwise we try scanning all the weight files in the folder and assemble a `weight_map` out of them.

michaelbenayoun (approved):
Not sure I am getting everything because it is a very long and complex PR but LGTM!
Let's iterate on smaller PRs from now on.
Thanks @zhenglongjiepheonix !

zhenglongjiepheonix:
Merge this as experimental first version, more fix-ups and features coming in following PRs!

.github/workflows/test_fx_automatic_parallel.yml
```diff
@@ -0,0 +1,65 @@
...
+    name: ${{ matrix.config.name }}
+    runs-on:
+      group: "${{matrix.gpu_target}}"
+
+    container:
+      image: ${{ matrix.config.image }}
+      options: --mount type=tmpfs,destination=/tmp --shm-size 64gb --gpus all --ipc host -v /mnt/hf_cache:/mnt/cache/
+      env:
+        NCCL_DEBUG: INFO
+        HF_TOKEN: ${{ secrets.HF_HUB_READ_TOKEN }}
```
echarlaix, line 38: @zhenglongjiepheonix @michaelbenayoun is `HF_TOKEN` used for the tests (can't see where) or can we remove ?

zhenglongjiepheonix: it's not used, you can remove it

echarlaix: removed in https://github.com/huggingface/optimum/pull/2061

PR #1935, opened 2024-07-02
Update optimum intel extra

echarlaix:


IlyasMoutawwakil:
just noticed that we still don't have the `ipex` extras, is it time we add them ?

IlyasMoutawwakil (approved):

PR #1939, opened 2024-07-03
Deprecate ORTModel class

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/onnxruntime/model.py
+++ b/optimum/onnxruntime/model.py
@@ -49,6 +49,11 @@ class ORTModel:
             label_names (`List[str]`, `optional`):
                 The list of keys in your dictionary of inputs that correspond to the labels.
         """
+
+        logger.warning(
+            "The class `optimum.onnxruntime.model.ORTModel` is deprecated and will be removed in the next release."
+        )
+
         self.compute_metrics = compute_metrics
         self.label_names = ["labels"] if label_names is None else label_names
         self.session = InferenceSession(str(model_path), providers=[execution_provider])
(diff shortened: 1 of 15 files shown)


optimum/onnxruntime/utils.py
```diff
@@ -341,3 +351,53 @@ class ORTQuantizableOperator(Enum):
     Resize = "Resize"
     AveragePool = "AveragePool"
     Concat = "Concat"
+
+
+def evaluation_loop(
```
echarlaix, line 356: should we also update the examples to directly use `evaluation_loop`

https://github.com/huggingface/optimum/blob/f755a58e56597f690be4a0c4bdb549ce0ffd4e03/examples/onnxruntime/optimization/question-answering/run_qa.py#L487

echarlaix (approved):
Thanks a lot @IlyasMoutawwakil !!

optimum/onnxruntime/model.py
```diff
@@ -49,6 +49,11 @@ def __init__(
             label_names (`List[str]`, `optional`):
                 The list of keys in your dictionary of inputs that correspond to the labels.
         """
+
+        logger.warning(
+            "The class `optimum.onnxruntime.model.ORTModel` is deprecated and will be removed in the next release."
```
echarlaix, line 54: It shouldn't be used anywhere so no strong opinion on this, but we can remove it in v2.0 just to be safe @fxmarty wdyt ?

fxmarty: No strong opinion on this either, it sounds fine to me to merge it now as well.

PR #1941, opened 2024-07-04
Fix forward bug in ORTModelForFeatureExtraction

moria97:
# What does this PR do?
Fix forward bug in ORTModelForFeatureExtraction when using CUDAExecutionProvider.
Current code will result in crash when using cuda.

Fixes # (issue)

## Before submitting

## Who can review?


echarlaix (approved):
Thanks for the fix !

PR #1945, opened 2024-07-05
Remove warning

echarlaix:


optimum/modeling_base.py
```diff
@@ -415,13 +415,6 @@ def from_pretrained(
                 trust_remote_code=trust_remote_code,
             )
 
-        if not export and trust_remote_code:
-            logger.warning(
-                "The argument `trust_remote_code` is to be used along with export=True. It will be ignored."
```
echarlaix, line 420: `trust_remote_code` will be used to load the configuration even when `export=False`

optimum/modeling_base.py
```diff
@@ -415,13 +415,6 @@ def from_pretrained(
...
-            )
-        elif export and trust_remote_code is None:
```
echarlaix, line 422: `trust_remote_code` default value is `False` so not sure why this is needed

PR #1947, opened 2024-07-08
Refactor diffusers tasks

IlyasMoutawwakil:
# What does this PR do?

This PR: 
- [x] Refactors "stable-diffusion" and "stable-diffusion-xl" into "text-to-image", "image-to-image" and "inpainting"
- [ ] Adds auto diffusion pipelines `ORTPipelineForText2Image`, `ORTPipelineForImage2Image` and `ORTPipelineForInpainting`

## Before submitting

### DIFF
--- a/optimum/exporters/onnx/__main__.py
+++ b/optimum/exporters/onnx/__main__.py
@@ -321,9 +321,7 @@ def main_export(
                 )
             model.config.pad_token_id = pad_token_id
 
-    if "stable-diffusion" in task:
-        model_type = "stable-diffusion"
-    elif hasattr(model.config, "export_model_type"):
+    if hasattr(model.config, "export_model_type"):
         model_type = model.config.export_model_type.replace("_", "-")
     else:
         model_type = model.config.model_type.replace("_", "-")
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -32,12 +32,11 @@ from transformers import AutoConfig, PretrainedConfig, is_tf_available, is_torch
 from transformers.utils import SAFE_WEIGHTS_NAME, TF2_WEIGHTS_NAME, WEIGHTS_NAME, logging
 
 from ..utils import CONFIG_NAME
-from ..utils.import_utils import is_onnx_available
+from ..utils.import_utils import is_diffusers_available, is_onnx_available
 
 
 if TYPE_CHECKING:
     from .base import ExportConfig
-
 
 logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
 
@@ -53,6 +52,14 @@ if is_torch_available():
 
 if is_tf_available():
     from transformers import TFPreTrainedModel
+
+if is_diffusers_available():
+    from diffusers import DiffusionPipeline
+    from diffusers.pipelines.auto_pipeline import (
+        AUTO_IMAGE2IMAGE_PIPELINES_MAPPING,
+        AUTO_INPAINT_PIPELINES_MAPPING,
+        AUTO_TEXT2IMAGE_PIPELINES_MAPPING,
+    )
 
 ExportConfigConstructor = Callable[[PretrainedConfig], "ExportConfig"]
 TaskNameToExportConfigDict = Dict[str, ExportConfigConstructor]
@@ -192,6 +199,9 @@ class TasksManager:
         _DIFFUSERS_TASKS_TO_MODEL_LOADERS = {
             "stable-diffusion": "StableDiffusionPipeline",
             "stable-diffusion-xl": "StableDiffusionXLImg2ImgPipeline",
+            "image-to-image": "AutoPipelineForImage2Image",
+            "text-to-image": "AutoPipelineForText2Image",
+            "inpainting": "AutoPipelineForInpainting",
         }
 
         _TIMM_TASKS_TO_MODEL_LOADERS = {
@@ -1534,7 +1544,7 @@ class TasksManager:
     @classmethod
     def _infer_task_from_model_or_model_class(
         cls,
-        model: Optional[Union["PreTrainedModel", "TFPreTrainedModel"]] = None,
+        model: Optional[Union["PreTrainedModel", "TFPreTrainedModel", "DiffusionPipeline"]] = None,
         model_class: Optional[Type] = None,
     ) -> str:
         if model is not None and model_class is not None:
@@ -1606,17 +1616,7 @@ class TasksManager:
                 )
             library_name = TasksManager.infer_library_from_model(model_name_or_path, subfolder, revision)
 
-            if library_name == "diffusers":
-                if model_info.config["diffusers"].get("class_name", None):
-                    class_name = model_info.config["diffusers"]["class_name"]
-                elif model_info.config["diffusers"].get("_class_name", None):
-                    class_name = model_info.config["diffusers"]["_class_name"]
-                else:
-                    raise ValueError(
-                        f"Could not automatically infer the class name for {model_name_or_path}. Please open an issue at https://github.com/huggingface/optimum/issues."
-                    )
-                inferred_task_name = "stable-diffusion-xl" if "StableDiffusionXL" in class_name else "stable-diffusion"
-            elif library_name == "timm":
+            if library_name == "timm":
                 inferred_task_name = "image-classification"
             else:
                 pipeline_tag = getattr(model_info, "pipeline_tag", None)
@@ -1626,7 +1626,7 @@ class TasksManager:
                 # zero-shot-object-detection or object-detection.
                 if pipeline_tag is not None and pipeline_tag not in ["conversational", "object-detection"]:
                     inferred_task_name = TasksManager.map_from_synonym(model_info.pipeline_tag)
-                else:
+                elif library_name == "transformers":
                     transformers_info = model_info.transformersInfo
                     if transformers_info is not None and transformers_info.get("pipeline_tag") is not None:
                         inferred_task_name = TasksManager.map_from_synonym(transformers_info["pipeline_tag"])
@@ -1649,12 +1649,13 @@ class TasksManager:
 
         if inferred_task_name is None:
             raise KeyError(f"Could not find the proper task name for {auto_model_class_name}.")
+
         return inferred_task_name
 
     @classmethod
     def infer_task_from_model(
         cls,
-        model: Union[str, "PreTrainedModel", "TFPreTrainedModel", Type],
+        model: Union[str, "PreTrainedModel", "TFPreTrainedModel", "DiffusionPipeline"],
         subfolder: str = "",
         revision: Optional[str] = None,
     ) -> str:
@@ -1690,7 +1691,8 @@ class TasksManager:
 
     @staticmethod
     def _infer_library_from_model(
-        model: Union["PreTrainedModel", "TFPreTrainedModel"], library_name: Optional[str] = None
+        model: Union["PreTrainedModel", "TFPreTrainedModel", "DiffusionPipeline"],
+        library_name: Optional[str] = None,
     ):
         if library_name is not None:
             return library_name
@@ -1800,7 +1802,7 @@ class TasksManager:
     @classmethod
     def standardize_model_attributes(
         cls,
-        model: Union["PreTrainedModel", "TFPreTrainedModel"],
+        model: Union["PreTrainedModel", "TFPreTrainedModel", "DiffusionPipeline"],
         library_name: Optional[str] = None,
     ):
         """
@@ -1808,25 +1810,25 @@ class TasksManager:
         libraries to follow transformers style.
 
         Args:
-            model_name_or_path (`Union[str, Path]`):
-                Can be either the model id of a model repo on the Hugging Face Hub, or a path to a local directory
-                containing a model.
-            model (`Union[PreTrainedModel, TFPreTrainedModel]`):
+            model (`Union[PreTrainedModel, TFPreTrainedModel, DiffusionPipeline]`):
                 The instance of the model.
-            subfolder (`str`, defaults to `""`):
-                In case the model files are located inside a subfolder of the model directory / repo on the Hugging
-                Face Hub, you can specify the subfolder name here.
-            revision (`Optional[str]`, *optional*, defaults to `None`):
-                Revision is the specific model version to use. It can be a branch name, a tag name, or a commit id.
-            cache_dir (`Optional[str]`, *optional*):
-                Path to a directory in which a downloaded pretrained model weights have been cached if the standard cache should not be used.
+
             library_name (`Optional[str]`, *optional*)::
                 The library name of the model. Can be any of "transformers", "timm", "diffusers", "sentence_transformers".
         """
         library_name = TasksManager._infer_library_from_model(model, library_name)
 
         if library_name == "diffusers":
-            model.config.export_model_type = "stable-diffusion"
+            diffusers_auto_mapping = {}
+            for mapping in [
+                AUTO_INPAINT_PIPELINES_MAPPING,
+                AUTO_IMAGE2IMAGE_PIPELINES_MAPPING,
+                AUTO_TEXT2IMAGE_PIPELINES_MAPPING,
+            ]:
+                diffusers_auto_mapping.update({v: k for k, v in mapping.items()})
+
+            model.config.export_model_type = diffusers_auto_mapping[model.__class__]
+
         elif library_name == "timm":
             # Retrieve model config
             model_config = PretrainedConfig.from_dict(model.pretrained_cfg)
--- a/optimum/exporters/onnx/convert.py
+++ b/optimum/exporters/onnx/convert.py
@@ -264,7 +264,7 @@ def _run_validation(
         atol = config.ATOL_FOR_VALIDATION
 
     if "diffusers" in str(reference_model.__class__) and not is_diffusers_available():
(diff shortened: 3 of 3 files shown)


optimum/exporters/onnx/__main__.py
```diff
@@ -221,6 +221,12 @@ def main_export(
             " and passing it is not required anymore."
         )
 
+    if task in ["stable-diffusion", "stable-diffusion-xl"]:
+        logger.warning(
+            f"The task `{task}` is deprecated and will be removed in a future release of Optimum. "
+            "Please use one of the following tasks instead: `text-to-image`, `image-to-image`, `inpainting`."
+        )
+
```
IlyasMoutawwakil, line 229: We warn and then it gets replaced with `text-to-image` by default. This is the same as the previous behavior for `stable-diffusion` but not `stable-diffusion-xl` as the model loader in that case used to be an `image-to-image` pipeline.

optimum/exporters/tasks.py
```diff
@@ -260,6 +280,9 @@ class TasksManager:
         "vision2seq-lm": "image-to-text",
         "zero-shot-classification": "text-classification",
         "image-feature-extraction": "feature-extraction",
+        # for backward compatibility
+        "stable-diffusion": "text-to-image",
+        "stable-diffusion-xl": "text-to-image",
```
IlyasMoutawwakil, line 285: should this be `image-to-image` for `stable-diffusion-xl` ? I don't think it matters

optimum/exporters/tasks.py
```diff
@@ -1132,7 +1154,6 @@ class TasksManager:
         "vae-decoder",
         "clip-text-model",
         "clip-text-with-projection",
-        "trocr",  # TODO: why?
```
IlyasMoutawwakil, line 1135: tried export and inference and it works

optimum/exporters/tasks.py
```diff
@@ -153,6 +158,13 @@ class TasksManager:
     _TRANSFORMERS_TASKS_TO_TF_MODEL_LOADERS = {}
     _LIBRARY_TO_TF_TASKS_TO_MODEL_LOADER_MAP = {}
 
+    # Torch model mappings
+    _TRANSFORMERS_TASKS_TO_MODEL_MAPPINGS = {}
+    _DIFFUSERS_TASKS_TO_PIPELINE_MAPPINGS = {}
+
+    # TF model mappings
+    _TRANSFORMERS_TASKS_TO_TF_MODEL_MAPPINGS = {}
```
IlyasMoutawwakil, line 166: uses the mappings from diffusers and transformers to construct a `task->model_type->model_class`

optimum/exporters/tasks.py
```diff
@@ -240,6 +268,37 @@ class TasksManager:
...
+            "text-to-image": AUTO_TEXT2IMAGE_PIPELINES_MAPPING,
+            "image-to-image": AUTO_IMAGE2IMAGE_PIPELINES_MAPPING,
+            "inpainting": AUTO_INPAINT_PIPELINES_MAPPING,
+        }
+
+        for task_name in _DIFFUSERS_TASKS_TO_PIPELINE_MAPPINGS:
+            pipeline_mapping = _DIFFUSERS_TASKS_TO_PIPELINE_MAPPINGS[task_name]
+            _DIFFUSERS_TASKS_TO_PIPELINE_MAPPINGS[task_name] = {
+                pipeline_name: pipeline_class.__name__ for pipeline_name, pipeline_class in pipeline_mapping.items()
+            }
```
IlyasMoutawwakil, line 300: transforming classes into class names to conform with transformers mappings

optimum/exporters/utils.py
```diff
@@ -72,23 +85,40 @@
...
-    if isinstance(pipeline, StableDiffusionXLImg2ImgPipeline):
+    is_stable_diffusion = isinstance(
+        pipeline, (StableDiffusionPipeline, StableDiffusionImg2ImgPipeline, StableDiffusionInpaintPipeline)
+    )
+    is_stable_diffusion_xl = isinstance(
+        pipeline, (StableDiffusionXLImg2ImgPipeline, StableDiffusionXLInpaintPipeline, StableDiffusionXLPipeline)
+    )
+    is_latent_consistency_model = isinstance(
+        pipeline, (LatentConsistencyModelPipeline, LatentConsistencyModelImg2ImgPipeline)
+    )
```
IlyasMoutawwakil, line 103: could be simplified using an `_infer_model_type_from_model_or_model_class` method and `_DIFFUSERS_TASKS_TO_PIPELINE_MAPPINGS`

optimum/exporters/tasks.py
```diff
@@ -1534,58 +1618,83 @@ def determine_framework(
...
+
+        if target_class_name.startswith("AutoModel"):
+            # transfromers models (auto)
+            for task_name, model_loader_class_names in cls._TRANSFORMERS_TASKS_TO_MODEL_LOADERS.items():
+                if isinstance(model_loader_class_names, str):
+                    model_loader_class_names = (model_loader_class_names,)
+                for model_loader_class_name in model_loader_class_names:
+                    if target_class_name == model_loader_class_name:
+                        inferred_task_name = task_name
+                        break
```
fxmarty, line 1642: Here we break out of a single loop only. I would return directly.

optimum/exporters/tasks.py
```diff
@@ -1534,58 +1618,83 @@ def determine_framework(
...
+        elif target_class_name.startswith("TFAutoModel"):
+            # transformers models (tf auto)
+            for task_name, model_loader_class_name in cls._TRANSFORMERS_TASKS_TO_TF_MODEL_LOADERS.items():
+                if isinstance(model_loader_class_names, str):
+                    model_loader_class_names = (model_loader_class_names,)
+                for model_loader_class_name in model_loader_class_names:
+                    if target_class_name == model_loader_class_name:
+                        inferred_task_name = task_name
+                        break
```
fxmarty, line 1651: same

optimum/exporters/tasks.py
```diff
@@ -1534,58 +1618,83 @@ def determine_framework(
...
+        elif target_class_name.startswith("AutoPipeline"):
+            # diffusers pipelines (auto)
+            for task_name, model_loader_class_name in cls._DIFFUSERS_TASKS_TO_MODEL_LOADERS.items():
+                if target_class_name == model_loader_class_name:
+                    inferred_task_name = task_name
+                    break
```
fxmarty, line 1657: same

optimum/exporters/tasks.py
```diff
@@ -1534,58 +1618,83 @@ def determine_framework(
...
+        elif target_class_module.startswith("diffusers"):
+            # diffusers pipelines
+            for task_name, pipeline_mapping in cls._DIFFUSERS_TASKS_TO_PIPELINE_MAPPINGS.items():
+                for pipeline_type, pipeline_class_name in pipeline_mapping.items():
+                    if target_class_name == pipeline_class_name:
+                        inferred_task_name = task_name
+                        break
```
fxmarty, line 1664: same

optimum/exporters/tasks.py
```diff
@@ -1534,58 +1618,83 @@ def determine_framework(
...
+            else:
+                task_name_to_model_mappings = cls._TRANSFORMERS_TASKS_TO_MODEL_MAPPINGS
+            for task_name, model_mappings in task_name_to_model_mappings.items():
+                if isinstance(model_mappings, dict):
+                    model_mappings = (model_mappings,)
+                for model_mapping in model_mappings:
+                    for model_type, model_class_name in model_mapping.items():
+                        if target_class_name == model_class_name:
+                            inferred_task_name = task_name
+                            break
```
fxmarty, line 1678: same

optimum/exporters/tasks.py
```diff
@@ -1626,7 +1731,7 @@ def _infer_task_from_model_name_or_path(
                 # zero-shot-object-detection or object-detection.
                 if pipeline_tag is not None and pipeline_tag not in ["conversational", "object-detection"]:
                     inferred_task_name = TasksManager.map_from_synonym(model_info.pipeline_tag)
-                else:
+                elif library_name == "transformers":
```
fxmarty, line 1734: So diffusers is handled by `inferred_task_name = TasksManager.map_from_synonym(model_info.pipeline_tag)`?

optimum/exporters/tasks.py
```diff
@@ -1642,72 +1747,106 @@ def _infer_task_from_model_name_or_path(
...
-        is_torch_pretrained_model = is_torch_available() and isinstance(model, PreTrainedModel)
-        is_tf_pretrained_model = is_tf_available() and isinstance(model, TFPreTrainedModel)
-        task = None
+        inferred_task_name = None
+
+        if use_auth_token is not None:
+            warnings.warn(
+                "The `use_auth_token` argument is deprecated and will be removed soon. Please use the `token` argument instead.",
+                FutureWarning,
+            )
```
fxmarty, line 1799: We should not add it if it is deprecated

optimum/exporters/tasks.py
```diff
@@ -1642,72 +1747,106 @@ def _infer_task_from_model_name_or_path(
...
 
     @classmethod
     def infer_task_from_model(
         cls,
-        model: Union[str, "PreTrainedModel", "TFPreTrainedModel", Type],
+        model: Union[str, "PreTrainedModel", "TFPreTrainedModel", "DiffusionPipeline", Type],
         subfolder: str = "",
         revision: Optional[str] = None,
+        cache_dir: str = HUGGINGFACE_HUB_CACHE,
+        use_auth_token: Optional[Union[bool, str]] = None,
```
fxmarty, line 1767: same, no need to add `use_auth_token`

optimum/exporters/tasks.py
```diff
@@ -1642,72 +1747,106 @@ def _infer_task_from_model_name_or_path(
...
+        inferred_task_name = None
+
+        if use_auth_token is not None:
+            warnings.warn(
+                "The `use_auth_token` argument is deprecated and will be removed soon. Please use the `token` argument instead.",
+                FutureWarning,
+            )
+            if token is not None:
+                raise ValueError("You cannot use both `use_auth_token` and `token` arguments at the same time.")
+            token = use_auth_token
```
fxmarty, line 1802: remove

optimum/exporters/tasks.py
```diff
@@ -1717,9 +1856,9 @@ def infer_library_from_model(
         subfolder: str = "",
         revision: Optional[str] = None,
         cache_dir: str = HUGGINGFACE_HUB_CACHE,
-        library_name: Optional[str] = None,
         use_auth_token: Optional[Union[bool, str]] = None,
         token: Optional[Union[bool, str]] = None,
+        library_name: Optional[str] = None,
```
fxmarty, line 1861: Why?

optimum/exporters/tasks.py
```diff
@@ -1800,33 +1942,38 @@ def infer_library_from_model(
...
         library_name = TasksManager._infer_library_from_model(model, library_name)
 
         if library_name == "diffusers":
-            model.config.export_model_type = "stable-diffusion"
+            inferred_model_type = None
+            for task_name, pipeline_mapping in cls._DIFFUSERS_TASKS_TO_PIPELINE_MAPPINGS.items():
+                for pipeline_type, pipeline_class_name in pipeline_mapping.items():
+                    if model.__class__.__name__ == pipeline_class_name:
+                        inferred_model_type = pipeline_type
+                        break
```
fxmarty, line 1967: only the last for is broken

optimum/exporters/tasks.py
```diff
@@ -222,7 +251,6 @@ class TasksManager:
             "text-classification": "TFAutoModelForSequenceClassification",
             "token-classification": "TFAutoModelForTokenClassification",
             "multiple-choice": "TFAutoModelForMultipleChoice",
-            "object-detection": "TFAutoModelForObjectDetection",
```
fxmarty, line 225: why removed?

optimum/exporters/tasks.py
```diff
@@ -210,6 +223,22 @@ class TasksManager:
...
+                    model_loader_class = getattr(auto_modeling_module, model_loader_name, None)
+                    if model_loader_class is not None:
+                        _TRANSFORMERS_TASKS_TO_MODEL_MAPPINGS[task_name].append(
+                            model_loader_class._model_mapping._model_mapping
+                        )
+            elif isinstance(_TRANSFORMERS_TASKS_TO_MODEL_LOADERS[task_name], str):
+                model_loader_name = _TRANSFORMERS_TASKS_TO_MODEL_LOADERS[task_name]
+                model_loader_class = getattr(auto_modeling_module, model_loader_name, None)
+                if model_loader_class is not None:
+                    _TRANSFORMERS_TASKS_TO_MODEL_MAPPINGS[task_name] = model_loader_class._model_mapping._model_mapping
```
fxmarty, line 240: I am not a big fan of defining class attributes like this, here I believe `auto_modeling_module`, `model_loader_class`, `model_loader_name`, etc. remain class attributes and are not treated as temporary variables.

fxmarty (commented):
I left a few comments, great work!

optimum/exporters/tasks.py
IlyasMoutawwakil: a lot of the tasks and their `TFAutoModel` actually don't exist for TF like:
`TFAutoModelForObjectDetection`, `TFAutoModelForAudioFrameClassification`, `TFAutoModelForAudioXVector` are nowhere to be found in [modeling_tf_auto](https://github.com/huggingface/transformers/blob/main/src/transformers/models/auto/modeling_tf_auto.py)

optimum/exporters/tasks.py
IlyasMoutawwakil: Yes when pipeline tag is not None. will add another case when `diffusers` is the lib but pipeline tag is not defined.

optimum/exporters/tasks.py
IlyasMoutawwakil: I just followed the pattern in the other `_infer_library_*` methods which is different than `_infer_task_*` for some reason. Another example is how the entrypoint `infer_task_from_model` can take model name/path/instance/class, but `infer_library_from_model` only takes name or path.
Would it make sense to standardize them ? the only hassle would be some changes on `optimum-subpackages` side.

optimum/exporters/tasks.py
IlyasMoutawwakil: oh I see what you mean, I'll put this part in some function to limit the scope.

optimum/exporters/onnx/__main__.py
echarlaix: `_optional_components` was introduced in diffusers v0.22.0 https://github.com/huggingface/diffusers/blob/v0.22.0/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L144 for `StableDiffusionXLPipeline` (not the case for `StableDiffusionXLImg2ImgPipeline`). I think we might need to upgrade [DIFFUSERS_MINIMUM_VERSION](https://github.com/huggingface/optimum/blob/dedb852eaea3d3d899920f4017b9fd899eb2756b/optimum/utils/import_utils.py#L53) as well to avoid any potential issue when loading a SDXL model with missing optional components

optimum/exporters/tasks.py
```diff
@@ -1642,103 +1693,116 @@ def _infer_task_from_model_name_or_path(
...
 
     @classmethod
     def infer_task_from_model(
         cls,
-        model: Union[str, "PreTrainedModel", "TFPreTrainedModel", Type],
+        model: Union[str, "PreTrainedModel", "TFPreTrainedModel", "DiffusionPipeline", Type],
         subfolder: str = "",
         revision: Optional[str] = None,
+        cache_dir: str = HUGGINGFACE_HUB_CACHE,
+        token: Optional[Union[bool, str]] = None,
```
echarlaix, line 1713: very nice!

optimum/exporters/tasks.py
```diff
@@ -1798,35 +1852,80 @@ def infer_library_from_model(
...
+        libraries to follow transformers style.
+
+        Args:
+            model (`Union[PreTrainedModel, TFPreTrainedModel, DiffusionPipeline]`):
+                The instance of the model.
+
         """
-        library_name = TasksManager._infer_library_from_model(model, library_name)
+
+        library_name = TasksManager.infer_library_from_model(model)
```
echarlaix, line 1908: user might specify the library name (for example whether to load the model with `sentence-transformers` or `transformers`)  in which case there could be a mismatch with the inferred `library_name`

echarlaix (commented):
Thanks a lot @IlyasMoutawwakil

optimum/exporters/tasks.py
echarlaix: actually here the model should already be loaded so shouldn't be an issue

optimum/exporters/onnx/__main__.py
```diff
@@ -221,13 +221,24 @@ def main_export(
             " and passing it is not required anymore."
         )
```
echarlaix, line 222: also could you update the documentation in https://github.com/huggingface/optimum/blob/b865809b76fff886df5f23e23b1da8f6f24a78b0/docs/source/exporters/onnx/usage_guides/export_a_model.mdx?plain=1#L90 and  https://github.com/huggingface/optimum/blob/b865809b76fff886df5f23e23b1da8f6f24a78b0/docs/source/exporters/tflite/usage_guides/export_a_model.mdx#L62

PR #1950, opened 2024-07-10
Stable diffusion controlnet optimum

TianmengChen:


optimum/version.py
```diff
@@ -12,4 +12,4 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-__version__ = "1.22.0.dev0"
+__version__ = "1.21.2"
```
sammysun0711, line 15: Please not change dev version

PR #1960, opened 2024-07-16
adding `ORTPipelineForxxx` entrypoints

IlyasMoutawwakil:
# What does this PR do?

Continuing the effort in #1947
This PR adds the Task specific entrypoints.
It also standardize and make the testing suite more compact, by grouping model tests by task, which revealed some problems in the current modeling, I fixed some, like accepting torch generators for compatibility with diffusers. But we observe many errors in reproducibility of outputs and comparison with diffusers.
Another PR that updates the export/modeling (mixins) fixes these issues https://github.com/huggingface/optimum/pull/2021.

## Before submitting


optimum/modeling_base.py
```diff
@@ -85,7 +85,6 @@ class PreTrainedModel(ABC):  # noqa: F811
 
 class OptimizedModel(PreTrainedModel):
     config_class = AutoConfig
-    load_tf_weights = None
```
echarlaix, line 88: Could this result in an issue for previous transformers version like v4.29.0 ?

optimum/onnxruntime/__init__.py
```diff
@@ -146,7 +150,11 @@
         )
     else:
         from .modeling_diffusion import (
+            ORTDiffusionPipeline,
             ORTLatentConsistencyModelPipeline,
+            ORTPipelineForImage2Image,
```
echarlaix, line 155: could you include it in `dummy_diffusers_objects.py` for cases when diffusers is not installed

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -69,23 +80,23 @@
 logger = logging.getLogger(__name__)
 
 
-class ORTStableDiffusionPipelineBase(ORTModel):
-    auto_model_class = StableDiffusionPipeline
-    main_input_name = "input_ids"
-    base_model_prefix = "onnx_model"
+class ORTPipeline(ORTModel):
+    auto_model_class = None
+    model_type = "onnx_pipeline"
```
echarlaix, line 85: why is it needed ?

optimum/pipelines/diffusers/pipeline_stable_diffusion_img2img.py
```diff
@@ -225,12 +262,8 @@ def __call__(
...
         timesteps = np.array([timesteps] * batch_size * num_images_per_prompt)
 
-        # add noise to latents using the timesteps
-        noise = generator.randn(*init_latents.shape).astype(latents_dtype)
-        init_latents = self.scheduler.add_noise(
-            torch.from_numpy(init_latents), torch.from_numpy(noise), torch.from_numpy(timesteps)
-        )
-        init_latents = init_latents.numpy()
+        # 5. Prepare latent variables
+        latents = self.prepare_latents(image, timesteps, batch_size, num_images_per_prompt, latents_dtype, generator)
```
echarlaix, line 266:
```suggestion
        init_latents = self.prepare_latents(image, timesteps, batch_size, num_images_per_prompt, latents_dtype, generator)
```

or we need to remove 

```python
        latents = init_latents
```

echarlaix (approved):

tests/onnxruntime/test_diffusion.py
```diff
@@ -0,0 +1,793 @@
...
+        model_args = {"test_name": model_arch, "model_arch": model_arch}
+        self._setup(model_args)
+
+        auto_pipeline = self.AUTOMODEL_CLASS.from_pretrained(MODEL_NAMES[model_arch])
+        ort_pipeline = self.ORTMODEL_CLASS.from_pretrained(self.onnx_model_dirs[model_arch])
+
+        self.assertEqual(ort_pipeline.auto_model_class, auto_pipeline.__class__)
+
+        # auto_pipeline = DiffusionPipeline.from_pretrained(MODEL_NAMES[model_arch])
+        # ort_pipeline = ORTDiffusionPipeline.from_pretrained(self.onnx_model_dirs[model_arch])
```
echarlaix, line 587: why commented ?

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -666,4 +696,140 @@ class ORTStableDiffusionXLImg2ImgPipeline(ORTStableDiffusionXLPipelineBase, Stab
...
+    ORTStableDiffusionImg2ImgPipeline,
+    ORTStableDiffusionInpaintPipeline,
+    ORTLatentConsistencyModelPipeline,
+    ORTStableDiffusionXLPipeline,
+    ORTStableDiffusionXLImg2ImgPipeline,
+]
+
+
+def _get_pipeline_class(pipeline_class_name: str, throw_error_if_not_exist: bool = True):
+    for ort_pipeline_class in SUPPORTED_ORT_PIPELINES:
```
echarlaix, line 716: for the case where export=False, we should look for diffusers classes name instead no ? (removing the OV prefix)

echarlaix (approved):
Looks good thanks @IlyasMoutawwakil

optimum/onnxruntime/modeling_diffusion.py
IlyasMoutawwakil: i'm not sure I understand, why when export=False ? the `ORTDiffusionPipeline` only dispatches to one of the classes based on `_class_name`

echarlaix: In the case where the model needs to be exported (so when export=True not export=False my bad) we should look for diffusers classes instead of the ORT ones

optimum/modeling_base.py
IlyasMoutawwakil: we already updated min transformers version to 4.36

optimum/onnxruntime/__init__.py
IlyasMoutawwakil: yes good catch

optimum/onnxruntime/modeling_diffusion.py
IlyasMoutawwakil: the class or `model_type` ?

optimum/modeling_base.py
echarlaix: I see v4.29.0 in https://github.com/huggingface/optimum/blob/v1.21.4/setup.py#L18, in any case asking to make sure that also works for the min supported transformers version

optimum/onnxruntime/modeling_diffusion.py
IlyasMoutawwakil: The underlying ORT class takes care of the export, these classes only dispatch to one of them (to one of the task and architecture specific ORT classes).

echarlaix: ok actually I see :
```
or ort_pipeline_class.auto_model_class.__name__ == pipeline_class_name
```
which is for the export=True case, so should work indeed!

tests/onnxruntime/test_diffusion.py
IlyasMoutawwakil: `DiffusionPipeline.from_pretrained(MODEL_NAMES[model_arch])` will always return the `_class_name` from the hub, but `ort_pipeline = ORTDiffusionPipeline.from_pretrained(self.onnx_model_dirs[model_arch])` will return the class used to save the model (architecture and task specific).

optimum/modeling_base.py
IlyasMoutawwakil: oh I must have been looking at optimum-intel 😅. Let me run a test on older versions and check.

IlyasMoutawwakil: tested and it doesn't seem to impact anything.

optimum/onnxruntime/base.py
```diff
@@ -153,11 +161,7 @@ def __init__(
 
         self.use_past_in_outputs = len(self.key_value_output_names) > 0
         self.use_past_in_inputs = len(self.key_value_input_names) > 0
-        self.use_fp16 = False
-        for inp in session.get_inputs():
-            if "past_key_values" in inp.name and inp.type == "tensor(float16)":
-                self.use_fp16 = True
-                break
+        self.use_fp16 = self.dtype == torch.float16
```
IlyasMoutawwakil, line 164: we use the `dtype` property we added in the base class

optimum/onnxruntime/base.py
```diff
@@ -41,17 +41,11 @@ class ORTModelPart:
...
-        self,
-        session: InferenceSession,
-        parent_model: "ORTModel",
-    ):
+    def __init__(self, session: InferenceSession, parent_model: "ORTModel"):
         self.session = session
         self.parent_model = parent_model
-        self.normalized_config = NormalizedConfigManager.get_normalized_config_class(
-            self.parent_model.config.model_type
-        )(self.parent_model.config)
```
IlyasMoutawwakil, line 53: there's no need for a model part specific normalized config in most cases, except for seq2seq models, we move this there.

optimum/modeling_base.py
echarlaix: thanks @IlyasMoutawwakil !

echarlaix (approved):
Looks great :fire:

optimum/utils/dummy_diffusers_objects.py
```diff
@@ -79,3 +79,91 @@ def __init__(self, *args, **kwargs):
...
+
+class ORTModelUnet(metaclass=DummyObject):
+    _backends = ["diffusers"]
+
+    def __init__(self, *args, **kwargs):
+        requires_backends(self, ["diffusers"])
+
+    @classmethod
+    def from_pretrained(cls, *args, **kwargs):
+        requires_backends(cls, ["diffusers"])
```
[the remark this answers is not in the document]
IlyasMoutawwakil: yeah sure, I only added them because we import them in testing, and if diffusers is not available, you can't run `test_modeling.py`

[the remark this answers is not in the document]
IlyasMoutawwakil: I removed them from optimum.onnxruntime and also made sure they're only imported in `test_modeling.py` if diffusers is available.

PR #1961, opened 2024-07-16
Fix optimizer's command line reading

idruker-cerence:
The command line optimum-cli onnxruntime optimize ... is incorrectly parsed. With -c flag specified, it creates an empty ORTConfig object without loading a passed-in configuration file.

This change:

fixes the issue
fixes the typo "from_pretained" --> "from_pretrained"
adds handling of the error case when none of -O1, -O2, -O3, -O4 or -c options is provided


fxmarty (approved):
Thanks a lot

PR #1965, opened 2024-07-18
Fixes detection of CuPy installed with pre-built wheels

### LINKED ISSUE #1925: is_cupy_available() can't find CuPy wheel distributions
### System Info

```shell
- Optimum 1.20.0 (also tested with 1.19.2)
- Python 3.11.7
- Amazon Linux

Trying to use CuPy pre-compiled wheel package (`cupy-cuda12x`)
```

### Who can help?

@JingyaHuang, @echarlai

### Information

- [X] My own modified scripts

### Tasks

- [X] My own task or dataset (give details below)

In the case that I'm using the `cupy-cuda12x` pre-compiled wheel package (rather than the standard `cupy` package that requires building from source) the `is_cupy_available` function (inside `optimum.onnxruntime.utils`) reports that it's not available. This is despite the fact that the `cupy` module is importable.

This prevents IO Binding from working.

This results in the message `Unable to access output memory in CUDA, will offload to CPU` appearing in logs.

Having looked a little deeper I can tell you that inside the `_is_package_available` function (which is used by `is_cupy_available`), it was able to find the module using `importlib.util.find_spec`, but when it tries to get the version using `importlib.metadata.version`, it fails. (BTW, the version isn't required for checking if CuPy is available, as far as I can tell).

The reason this seems to be happening is that while the `cupy-cuda12x` package (and I assume `cupy-cuda11x`, etc.) install the `cupy` module with the same name as the source package (which is what `importlib.util.find_spec` checks for), the version information is inside the `.dist-info` directory, which has a different name.

A hacky solution I've tested to work-around this issue is to trick `importlib.metadata.version` into working my manually copying the `.dist-info` directory for CuPy to the name the regular `cupy` package would have.

# Minimal example

Poetry project deps:
```
optimum = {extras = ["onnxruntime-gpu"], version = "^1.20.0"}
cupy-cuda12x = "^13.2.0"
```

Test script:
```python
import importlib.util

from optimum.onnxruntime.utils import is_cupy_available
from optimum.utils.import_utils import _is_package_available

print(f"is_cupy_available()              = {is_cupy_available()}")
print(f"_is_package_available('cupy')    = {_is_package_available('cupy')}")
print(f"importlib.util.find_spec('cupy') = {bool(importlib.util.find_spec('cupy'))}")
print(f"importlib.metadata.version('cupy') = {importlib.metadata.version('cupy')}")
```

Printed output:
```
is_cupy_available()              = False
_is_package_available('cupy')    = False
importlib.util.find_spec('cupy') = True
  File "/home/ec2-user/.asdf/installs/python/3.11.7/lib/python3.11/importlib/metadata/__init__.py", line 563, in from_name
    return next(cls.discover(name=name))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
StopIteration

During handling of the above exception, another exception occurred:

  File "/home/ec2-user/test-optimum-ort-bug/test.py", line 9, in <module>
    print(f"importlib.metadata.version('cupy') = {importlib.metadata.version('cupy')}")
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ec2-user/.asdf/installs/python/3.11.7/lib/python3.11/importlib/metadata/__init__.py", line 1008, in version
    return distribution(distribution_name).version
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ec2-user/.asdf/installs/python/3.11.7/lib/python3.11/importlib/metadata/__init__.py", line 981, in distribution
    return Distribution.from_name(distribution_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ec2-user/.asdf/installs/python/3.11.7/lib/python3.11/importlib/metadata/__init__.py", line 565, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: No package metadata was found for cupy
```

### Expected behavior

Optimum should not require use of the source distribution of `cupy` to enable IO Binding and should work fine with the appropriate wheel distribution for the targeted CUDA version (`cupy-cuda12x`, `cupy-cuda11x`, and presumably `cupy-rocm-5-0`, `cupy-rocm-4-3`)

tcsavage:
# What does this PR do?

This PR fixes discovery of CuPy when it's installed using one of the distributions of pre-built wheels.

The CuPy library ships both a source distribution (`cupy`) as well as versions containing pre-built wheels (`cupy-cuda11x`, `cupy-cuda12x`, `cupy-rocm-5-0`, `cupy-rocm-4-3`). Use of `_is_package_available` to detect CuPy only works for the source distribution of CuPy and fails when using the pre-built wheels versions.

This is because the `_is_package_available` will always attempt to resolve version information (even if it's not required) and in doing so assumes that the _importable_ package name matches the _installed_ distribution name. While this is usually the case, it doesn't work for CuPy and several other libraries. ONNX Runtime for example might be installed as `onnxruntime` or `onnxruntime-gpu` and thus Optimum just uses `importlib.util.find_spec` to work around the same problem. This commit replicates the same solution for CuPy.

Fixes #1925

## Who can review?

### DIFF
--- a/optimum/onnxruntime/utils.py
+++ b/optimum/onnxruntime/utils.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 """Utility functions, classes and constants for ONNX Runtime."""
 
+import importlib
 import os
 import re
 from enum import Enum
@@ -31,7 +32,6 @@ from transformers.utils import logging
 import onnxruntime as ort
 
 from ..exporters.onnx import OnnxConfig, OnnxConfigWithLoss
-from ..utils.import_utils import _is_package_available
 
 
 if TYPE_CHECKING:
@@ -91,9 +91,11 @@ def is_onnxruntime_training_available():
 
 def is_cupy_available():
     """
-    Checks if onnxruntime-training is available.
+    Checks if CuPy is available.
     """
-    return _is_package_available("cupy")
+    # Don't use _is_package_available as it doesn't work with CuPy installed
+    # with `cupy-cuda*` and `cupy-rocm-*` package name (prebuilt wheels).
+    return importlib.util.find_spec("cupy") is not None
 
 
 class ORTConfigManager:


IlyasMoutawwakil:
Thank you for the fix !

PR #1929, opened 2024-06-28
Support transformers 4.42

fxmarty:

PR #1968, opened 2024-07-23
Support transformers 4.43

liammcinroy:
# What does this PR do?
Allows `transformers<4.44` as a dependency, since with the release of 4.43.0 (and 4.43.1 today), then `optimum` can't be installed with these versions.

Looking at the release notes of 4.43.0, I didn't see anything that would obviously break here (the biggest standout was https://github.com/huggingface/transformers/pull/31696, but it seems that was handled in https://github.com/huggingface/optimum/pull/1929 already). If there are any that stand out to anyone else, I'd be happy to take a look at the necessary fixes.

## Before submitting

## Who can review?

@fxmarty @IlyasMoutawwakil @echarlaix 

based off of https://github.com/huggingface/optimum/pull/1929

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ except Exception as error:
 REQUIRED_PKGS = [
     "coloredlogs",
     "sympy",
-    "transformers[sentencepiece]>=4.26.0,<4.43.0",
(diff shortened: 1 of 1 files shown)


IlyasMoutawwakil:
there seems to be quite some failures

PR #1971, opened 2024-07-25
Support transformers 4.43

IlyasMoutawwakil:
# What does this PR do?

This PR adds support for transformers 4.43 with which the following issues emerge:
- `clip` models using `sdpa` attention.
- `pipeline` calling `model.dtype` to convert inputs to the correct floating point precision
- `bark` models can't be saved due to shared tensors (track in https://github.com/huggingface/transformers/issues/32224)
- whisper introducing a new forward input argument `cache_position`.
## Before submitting

### DIFF
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -30,6 +30,7 @@ from ...utils import (
     DummySeq2SeqPastKeyValuesGenerator,
     DummyTextInputGenerator,
     DummyVisionInputGenerator,
+    check_if_transformers_greater,
     is_diffusers_available,
     logging,
 )
@@ -288,9 +289,13 @@ class AudioToTextOnnxConfig(OnnxSeq2SeqConfigWithPast):
 
         if self._behavior is not ConfigBehavior.ENCODER:
             if self.use_past_in_inputs:
-                common_inputs["cache_position"] = {0: "1"}  # it's 1 when past is used
                 common_inputs["decoder_input_ids"] = {0: "batch_size"}
                 self.add_past_key_values(common_inputs, direction="inputs")
+
+                if check_if_transformers_greater("4.43.0"):
+                    # shape is [1] when using cache and [sequence_length] when not using it
+                    common_inputs["cache_position"] = {0: "1"}
+
             else:
                 common_inputs["decoder_input_ids"] = {0: "batch_size", 1: "decoder_sequence_length"}
 
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -30,7 +30,6 @@ from ...utils import (
     DummySeq2SeqPastKeyValuesGenerator,
     DummyTextInputGenerator,
     DummyVisionInputGenerator,
-    check_if_transformers_greater,
     is_diffusers_available,
     logging,
 )
@@ -291,11 +290,6 @@ class AudioToTextOnnxConfig(OnnxSeq2SeqConfigWithPast):
             if self.use_past_in_inputs:
                 common_inputs["decoder_input_ids"] = {0: "batch_size"}
                 self.add_past_key_values(common_inputs, direction="inputs")
-
-                if check_if_transformers_greater("4.43.0"):
-                    # shape is [1] when using cache and [sequence_length] when not using it
-                    common_inputs["cache_position"] = {0: "1"}
-
             else:
                 common_inputs["decoder_input_ids"] = {0: "batch_size", 1: "decoder_sequence_length"}
 
--- a/optimum/exporters/onnx/config.py
+++ b/optimum/exporters/onnx/config.py
@@ -288,12 +288,11 @@ class AudioToTextOnnxConfig(OnnxSeq2SeqConfigWithPast):
 
         if self._behavior is not ConfigBehavior.ENCODER:
             if self.use_past_in_inputs:
+                common_inputs["cache_position"] = {0: "1"}  # it's 1 when past is used
                 common_inputs["decoder_input_ids"] = {0: "batch_size"}
+                self.add_past_key_values(common_inputs, direction="inputs")
             else:
                 common_inputs["decoder_input_ids"] = {0: "batch_size", 1: "decoder_sequence_length"}
-
-            if self.use_past_in_inputs:
-                self.add_past_key_values(common_inputs, direction="inputs")
 
         if self._behavior is ConfigBehavior.DECODER:
             common_inputs["encoder_outputs"] = {0: "batch_size", 1: "encoder_sequence_length"}
--- a/optimum/exporters/utils.py
+++ b/optimum/exporters/utils.py
@@ -96,7 +96,7 @@ def _get_submodels_for_export_diffusion(
         pipeline, (StableDiffusionPipeline, StableDiffusionImg2ImgPipeline, StableDiffusionInpaintPipeline)
     )
     is_stable_diffusion_xl = isinstance(
-        pipeline, (StableDiffusionXLImg2ImgPipeline, StableDiffusionXLInpaintPipeline, StableDiffusionXLPipeline)
+        pipeline, (StableDiffusionXLPipeline, StableDiffusionXLImg2ImgPipeline, StableDiffusionXLInpaintPipeline)
     )
     is_latent_consistency_model = isinstance(
         pipeline, (LatentConsistencyModelPipeline, LatentConsistencyModelImg2ImgPipeline)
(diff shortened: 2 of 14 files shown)


IlyasMoutawwakil:
OwlViT and OwlV2 need an upgrade to their onnx config (opset=11->12 for the einsum operator), I will push it once the onnxruntime tests finish (to see if there's anything left).

optimum/exporters/utils.py
```diff
@@ -150,7 +150,6 @@ def _get_submodels_for_export_diffusion(
 
     text_encoder_2 = getattr(pipeline, "text_encoder_2", None)
     if text_encoder_2 is not None:
-        text_encoder_2.config.output_hidden_states = True
```
echarlaix, line 153: won't this be causing an issue here https://github.com/huggingface/optimum/blob/5ece6e88156de6a0de848ad6c1af381c81cfbcc3/optimum/exporters/onnx/model_configs.py#L975 ?

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -490,7 +515,8 @@ def __init__(self, session: ort.InferenceSession, parent_model: ORTModel):
         self.output_names = {output_key.name: idx for idx, output_key in enumerate(self.session.get_outputs())}
         config_path = Path(session._model_path).parent / self.CONFIG_NAME
         self.config = self.parent_model._dict_from_json_file(config_path) if config_path.is_file() else {}
-        self.input_dtype = {inputs.name: _ORT_TO_NP_TYPE[inputs.type] for inputs in self.session.get_inputs()}
+        self.input_dtypes = {inputs.name: inputs.type for inputs in self.session.get_inputs()}
```
echarlaix, line 499: all pipelines needs to be updated no https://github.com/huggingface/optimum/blob/5ece6e88156de6a0de848ad6c1af381c81cfbcc3/optimum/pipelines/diffusers/pipeline_stable_diffusion_xl.py#L107C66-L107C77

echarlaix: not sure it's worth it but if we want to change the attribute name we should have something like that for a couple of release :
```python
 @property
  def input_dtype(self):
        # add warning that this will be deprecated in optimum vX.X.X
        return self.input_dtypes
```

echarlaix (approved):
Thanks a lot for taking care of this @IlyasMoutawwakil

optimum/onnxruntime/modeling_diffusion.py
IlyasMoutawwakil: The `dtype` for diffusion is not really needed here and generally not used (should be tbh, to cast input ids to the right dtype), I changed it here to conform with the base model part (in base.py).
I'll revert it for now and we can add that later on.

optimum/exporters/utils.py
IlyasMoutawwakil: it actually worked without, the problem is that clip text model with projection doesn't output the hidden states anymore, I finally found where it's coming from and it's fixed now (gotta set it in model.text_model as well)

IlyasMoutawwakil: fixed

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -33,14 +32,14 @@
...
+    EncoderDecoderCache,
     GenerationConfig,
     Pix2StructForConditionalGeneration,  # Pix2struct does not support AutoModel
 )
 from transformers.file_utils import add_end_docstrings, add_start_docstrings_to_model_forward
-from transformers.generation.logits_process import WhisperTimeStampLogitsProcessor
 from transformers.modeling_outputs import BaseModelOutput, Seq2SeqLMOutput
 from transformers.models.auto.modeling_auto import MODEL_FOR_SPEECH_SEQ_2_SEQ_MAPPING_NAMES
-from transformers.models.whisper.tokenization_whisper import TASK_IDS, TO_LANGUAGE_CODE
+from transformers.models.whisper.generation_whisper import WhisperGenerationMixin
```
echarlaix, line 42: this will break for transformers < v4.37.0 no ?

IlyasMoutawwakil: should we update the minimum version for whisper ? because this way we are using the transformers custom generation loop, which solves many discrepancies with transformers transcription (e.g. 2 init tokens)

IlyasMoutawwakil: for example if I run this code on main optimum:
```python
transformers_outputs = model.generate(
    **inputs,
    num_beams=1,
    use_cache=True,
    max_new_tokens=3,
    min_new_tokens=3,
)
print(transformers_outputs)

ort_outputs = ort_model.generate(
    **inputs,
    num_beams=1,
    use_cache=True,
    max_new_tokens=3,
    min_new_tokens=3,
)
print(ort_outputs)
```
I get:
```
tensor([[50257, 50362,   345,   345,   345],
        [50257, 50362,   345,   345,   345]])

tensor([[50257, 50362,   345,   345],
        [50257, 50362,   345,   345]])
```
and there other stuff that were added to the generation loop like short vs long form generation (probably related to https://github.com/huggingface/optimum/issues/1972)

IlyasMoutawwakil:
should be all fixed now

optimum/exporters/onnx/config.py
```diff
@@ -288,13 +288,12 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
 
         if self._behavior is not ConfigBehavior.ENCODER:
             if self.use_past_in_inputs:
+                common_inputs["cache_position"] = {0: "1"}  # it's 1 when past is used
```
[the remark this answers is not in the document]
echarlaix: could we add a test for lower transformers version just to make sure this doesn't break anything ?

echarlaix (approved):
Looks great thanks a lot @IlyasMoutawwakil

optimum/utils/input_generators.py
```diff
@@ -658,27 +659,38 @@ def __init__(
...
+
+        elif input_name == "cache_position":
+            return self.random_int_tensor(
+                shape=[1],
+                max_value=self.sequence_length,
+                framework=framework,
+                dtype=int_dtype,
+            )
+
+        return super().generate(input_name, framework=framework, int_dtype=int_dtype)
```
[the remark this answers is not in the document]
IlyasMoutawwakil: normally never, should I remove it ?

optimum/onnxruntime/modeling_seq2seq.py
echarlaix: I see, would make sense to add it as done for decoder models
https://github.com/huggingface/optimum/blob/2a6d8573301b164155e6151ee60efd9ccf57554c/tests/onnxruntime/test_modeling.py#L2425

optimum/exporters/onnx/config.py
```diff
@@ -289,12 +290,15 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
         if self._behavior is not ConfigBehavior.ENCODER:
             if self.use_past_in_inputs:
                 common_inputs["decoder_input_ids"] = {0: "batch_size"}
+                self.add_past_key_values(common_inputs, direction="inputs")
+
+                if check_if_transformers_greater("4.43.0"):
+                    # shape is [1] when using cache and [sequence_length] when not using it
+                    common_inputs["cache_position"] = {0: "1"}
```
echarlaix, line 297: should it be  ?
```suggestion
                    # shape is [1] when using cache and [sequence_length] when not using it
                    common_inputs["cache_position"] = {0: "sequence_length"}
```

IlyasMoutawwakil: I guess you're right !

IlyasMoutawwakil: I also moved to whisper config

sreenivasulureddysura:
@IlyasMoutawwakil Is this PR ready to support transformers 4.43.3 version?

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -1270,6 +1281,7 @@ def get_encoder(self) -> ORTEncoder:
     # Copied from transformers.models.bart.modeling_bart.BartForConditionalGeneration._reorder_cache
     @staticmethod
     def _reorder_cache(past, beam_idx) -> Tuple[Tuple[torch.FloatTensor]]:
+        print("REORDER CACHE CALLED")
```
IlyasMoutawwakil, line 1284: just saw this one, will remove it

IlyasMoutawwakil:
@sreenivasulureddysura yep it's ready

PR #1973, opened 2024-07-25
fix gpt bigcode ONNX export for transformers<4.39.0

echarlaix:
In https://github.com/huggingface/optimum/pull/1963 `unmask_unattended_patched`  signature was modified to fix compatibility with transformers >= v4.39.0, here we fix compatibility for  < v4.39.0

### DIFF
--- a/optimum/exporters/onnx/model_patcher.py
+++ b/optimum/exporters/onnx/model_patcher.py
@@ -276,12 +276,13 @@ class VisionEncoderDecoderPatcher(Seq2SeqModelPatcher):
             model.decoder.model.decoder.config.use_cache = True
 
 
-def _unmask_unattended_patched(
-    expanded_mask: torch.Tensor,
-    min_dtype: float,
+def _unmask_unattended_patched_legacy(
+    expanded_mask: torch.Tensor, attention_mask: torch.Tensor, unmasked_value: Union[bool, float]
 ):
     return expanded_mask
 
+def _unmask_unattended_patched(expanded_mask: torch.Tensor, min_dtype: float):
+    return expanded_mask
 
 def _make_causal_mask_patched(
     input_ids_shape: torch.Size,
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -15,7 +15,7 @@
 REQUIRED_PKGS = [
     "coloredlogs",
     "sympy",
-    "transformers[sentencepiece]>=4.26.0,<4.43.0",
+    "transformers[sentencepiece]>=4.33.0,<4.43.0",
```
echarlaix, line 18: we don't support 4.26.0 since a long time https://github.com/huggingface/optimum/blob/5ece6e88156de6a0de848ad6c1af381c81cfbcc3/optimum/exporters/onnx/model_patcher.py#L25 at least for `optimum.exporters` and looks like it won't impact any other subpackage : 
amd : https://github.com/huggingface/optimum-amd/blob/main/setup.py#L17
intel : https://github.com/huggingface/optimum-intel/blob/v1.18.1/setup.py
habana : https://github.com/huggingface/optimum-habana/blob/v1.12.1/setup.py#L32
neuron : https://github.com/huggingface/optimum-neuron/blob/v0.0.23/setup.py

optimum/exporters/onnx/model_patcher.py
```diff
@@ -316,7 +319,11 @@ def _make_causal_mask_patched(
 
 
 _make_causal_mask_patched_staticmethod = staticmethod(_make_causal_mask_patched)
-_unmask_unattended_patched_staticmethod = staticmethod(_unmask_unattended_patched)
+
+if _transformers_version >= version.parse("4.39.0"):
```
echarlaix, line 323: we need to have test for the lowest transformers version supported

setup.py
echarlaix: will impact furiosa that with "transformers>=4.20.0"  https://github.com/huggingface/optimum-furiosa/blob/main/setup.py#L16 cc @mht-sharma

echarlaix: should be done in an other PR with tests, reverting

IlyasMoutawwakil (approved):
LGTM

PR #1984, opened 2024-08-07
HF token in CLI for private and gated repos

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting


IlyasMoutawwakil:
can simply be done with `HF_TOKEN="hf_..." optimum export ...`

lekhnath:
> can simply be done with `HF_TOKEN="hf_..." optimum export ...`

Does not work. getting following error even after setting env variable.


```bash


Please make sure you specified the correct `repo_id` and `repo_type`.
If you are trying to access a private or gated repo, make sure you are authenticated.
Invalid username or password.

```

PR #1988, opened 2024-08-14
Modify Parallelization Strategy to Make it More General

zhenglongjiepheonix:
As per title, this PR tries a more general approach rather than relying purely on human heuristics, basically it uses the following steps to search a possible parallelization strategy for a transformer model

- Use dynamo for graph tracing so that we get the graph to operate on
- Decompose and functionalize the traced graph so that we get a smaller op set to work with
- Apply parallel axis analysis and do a constrained backtracking search on the whole graph to get a possible solution(not necessarily optimal)
- Replace ops the original traced graph with their parallelized version(Linear -> ColumnLinear/RowLinear)

And for the API design, we disable the support of passing custom modules and only focus on models in transformers because supporting custom models is not the priority for now.

### DIFF
--- a/optimum/fx/parallelization/api.py
+++ b/optimum/fx/parallelization/api.py
@@ -15,7 +15,7 @@
 import importlib
 import os
 from functools import partial
-from typing import List
+from typing import Callable, List
 
 import torch
 from torch.fx import GraphModule
@@ -48,7 +48,7 @@ def parallelize_model(
     parallel_ctx: ParallelExecutionCtx,
     *model_args,
     **kwargs,
-):
+) -> Callable:
     """
     API for automatic model parallelism through Pytorch FX.
 
--- a/optimum/fx/parallelization/api.py
+++ b/optimum/fx/parallelization/api.py
@@ -54,7 +54,8 @@ def parallelize_model(
 
     Args:
         model (str):
-            Model to parallelize, a model id on the Huggingface Hub.
+            Model to parallelize, a model id on the Huggingface Hub or path to a local directory containing config and weights
+            of the model.
         parallel_ctx (ParallelExecutionCtx):
             Parallel execution context containing process groups the current process belongs to.
         *model_args (Any):
--- a/optimum/fx/parallelization/api.py
+++ b/optimum/fx/parallelization/api.py
@@ -53,22 +53,22 @@ def parallelize_model(
     API for automatic model parallelism through Pytorch FX.
 
     Args:
-        model (str):
+        model (`str`):
             Model to parallelize, a model id on the Huggingface Hub or path to a local directory containing config and weights
             of the model.
-        parallel_ctx (ParallelExecutionCtx):
+        parallel_ctx (`ParallelExecutionCtx`):
             Parallel execution context containing process groups the current process belongs to.
-        *model_args (Any):
+        *model_args (`Any`):
             Additional postional arguments for intializing the model if a model id is passed.
-        revision (str, defaults to `main`):
+        revision (`str`, defaults to `main`):
             Model revision for weights downloading if a model id is passed.
-        cache_dir (Optional[str], defaults to `None`):
+        cache_dir (`Optional[str]`, defaults to `None`):
             Cache directory to store downloaded weights. Defaults to None.
-        local_files_only (bool, defaults to `False`):
+        local_files_only (`bool`, defaults to `False`):
             Whether to use local files only, will avoid downloading from remote if set to `True`.
-        skip_load_weights (bool, defaults to `False`):
+        skip_load_weights (`bool`, defaults to `False`):
             Whether to skip loading weights from disk to model.
-        **kwargs (Dict[str, Any]):
+        **kwargs (`Dict[str, Any]`):
             Addtional keyword arguments for overriding fields in parallel config, model config and `Model.__init__`.
     """
     revision = kwargs.pop("revision", "main")
--- a/optimum/fx/parallelization/api.py
+++ b/optimum/fx/parallelization/api.py
@@ -15,10 +15,11 @@
 import importlib
 import os
 from functools import partial
-from typing import List, Union
+from typing import List
 
 import torch
 from torch.fx import GraphModule
+from transformers import AutoConfig
 
 from .core import Config, ParallelExecutionCtx
 from .passes import build_parallel_pass_pipeline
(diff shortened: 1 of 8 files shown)


optimum/fx/parallelization/decomp.py
```diff
@@ -0,0 +1,215 @@
...
+                return super().run(*args, **kwargs)
+
+
+def decompose_and_functionalize(
+    graph_module: GraphModule,
+    decomposition_table: Dict[torch._ops.OperatorBase, Callable] = core_aten_decompositions(),
+    leaf_function_targets: List[Callable] = [F.scaled_dot_product_attention],
+) -> Callable:
+    """
+    API to decompose and funcitonalize a high-level graph module.
```
michaelbenayoun, line 193:
```suggestion
    API to decompose and functionalize a high-level graph module.
```

optimum/fx/parallelization/decomp.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Args:
+        graph_module (GraphModule):
+            The high-level graph module to be decomposed and functionalized.
+        decomposition_table (Dict[torch._ops.OperatorBase, Callable], defaults to `core_aten_decompostions()`):
+            The lookup table which maps high-level torch op to their equivalent low-level implementation.
+        leaf_function_targets (List[Callable], defaults to `[F.scaled_dot_product_attention]`):
```
michaelbenayoun, line 200:
```suggestion
        graph_module (`GraphModule`):
            The high-level graph module to be decomposed and functionalized.
        decomposition_table (`Dict[torch._ops.OperatorBase, Callable]`, defaults to `core_aten_decompostions()`):
            The lookup table which maps high-level torch op to their equivalent low-level implementation.
        leaf_function_targets (`List[Callable]`, defaults to `[F.scaled_dot_product_attention]`):
```

optimum/fx/parallelization/op_registry/op_handlers.py
```diff
@@ -0,0 +1,445 @@
...
+                for op_type in op_types:
+                    self.mapping[op_type] = cls
+            else:
+                self.mapping[op_types] = cls
+            return cls
+
+        return wrapper
+
+    def is_supported(self, op_type) -> bool:
+        return op_type in self.mapping
```
michaelbenayoun, line 41: What is this registry used for?

optimum/fx/parallelization/op_registry/op_handlers.py
```diff
@@ -0,0 +1,445 @@
...
+        torch.ops.aten.bitwise_and.Scalar,
+        torch.ops.aten.bitwise_or.Scalar,
+        torch.ops.aten.bitwise_xor.Scalar,
+    ]
+)
+class UnaryOpParallelAxisPropagateHandler(OpParallelAxisPropagateHandler):
+    def propagate(self) -> bool:
+        arg = self.node.all_input_nodes[0]
+        axis = self.extract_axis(arg)
+        return [axis]
```
michaelbenayoun, line 130: nit it's not returning a bool.
If I understand properly it returns the axis that is supposed to be parallel?

optimum/fx/parallelization/api.py
```diff
@@ -43,17 +44,18 @@ def parallelize_backend(
 
 
 def parallelize_model(
-    model: Union[torch.nn.Module, str],
+    model: str,
     parallel_ctx: ParallelExecutionCtx,
     *model_args,
     **kwargs,
-):
+) -> Callable:
```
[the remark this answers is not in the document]
zhenglongjiepheonix: Not exactly，`torch.compile` returns a handle patched around the original module, and will trigger tracing and compilation when firstly called with data

optimum/fx/parallelization/op_registry/op_handlers.py
zhenglongjiepheonix: Yes, you are right, it looks up the axis of all the inputs and try deducting the axis of the output

optimum/fx/parallelization/op_registry/op_handlers.py
zhenglongjiepheonix: This is for registration of parallel axis policy of different aten ops

optimum/fx/parallelization/decomp.py
```diff
@@ -0,0 +1,215 @@
...
+    def create_node_(*args, **kwargs):
+        node = creat_node(*args, **kwargs)
+        node.meta["traced_from"] = traceback.get_current_meta()["from_node"]
+        return node
+
+    try:
+        Graph.create_node = create_node_
+        yield
+    finally:
+        Graph.create_node = creat_node
```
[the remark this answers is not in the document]
zhenglongjiepheonix: Yes, it records the original high-level node from which it comes from

optimum/fx/parallelization/op_registry/op_handlers.py
michaelbenayoun: Can you add a docstring to explain that please?

optimum/fx/parallelization/passes.py
```diff
@@ -135,238 +134,151 @@ def clean_all(self, graph_module: GraphModule) -> None:
...
-            elif is_embedding(node):
-                # directly mark `nn.Embedding` layers
-                self.place_marker_per_node(node, {"axis": "vocab"})
 
-        return graph_module
+        nodes = list(graph.nodes)
 
+        def search(idx: int):
+            if idx == len(nodes):
+                return True
```
michaelbenayoun, line 201:
```suggestion
        def search(idx: int) -> bool:
            return idx == len(nodes)     
```

zhenglongjiepheonix: the `search` here actually is a backtracking search function entailing more logic following, so we can only return True if we have reached the very last op

optimum/fx/parallelization/op_registry/op_handlers.py
```diff
@@ -0,0 +1,450 @@
...
+from torch.fx import Node
+
+from ..core import Config
+from ..utils import is_activation, is_embedding, is_linear
+
+
+class Registry:
+    """
+    Registry class handles registration of parallel axis propagation handlers of different aten ops, to support a new
+    aten op, you need to register the corresponding handler class by decorating it with `register` function.
```
michaelbenayoun, line 28: nit:
 ```suggestion
    Registry class handles registration of parallel axis propagation handlers of different aten ops.
    To support a new aten op, you need to register the corresponding handler class by decorating 
    it with the `register` function.
```

michaelbenayoun (approved):
LGTM!

zhenglongjiepheonix:
merge this for irrelevant failures

PR #1996, opened 2024-08-22
Transformers 4.44 support

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ except Exception as error:
 REQUIRED_PKGS = [
     "coloredlogs",
     "sympy",
-    "transformers[sentencepiece]>=4.29.0,<4.44.0",
+    "transformers[sentencepiece]>=4.29.0,<4.45.0",
     "torch>=1.11",
     "packaging",
     "numpy<2.0",  # transformers requires numpy<2.0 https://github.com/huggingface/transformers/pull/31569
@@ -80,9 +80,9 @@ EXTRAS_REQUIRE = {
     "nncf": "optimum-intel[nncf]>=1.18.0",
     "neural-compressor": "optimum-intel[neural-compressor]>=1.18.0",
     "ipex": "optimum-intel[ipex]>=1.18.0",
-    "habana": ["optimum-habana", "transformers >= 4.43.0, < 4.44.0"],
-    "neuron": ["optimum-neuron[neuron]>=0.0.20", "transformers >= 4.36.2, < 4.42.0"],
-    "neuronx": ["optimum-neuron[neuronx]>=0.0.20", "transformers >= 4.36.2, < 4.42.0"],
+    "habana": ["optimum-habana", "transformers>=4.43.0,<4.44.0"],
+    "neuron": ["optimum-neuron[neuron]>=0.0.20", "transformers>=4.36.2,<4.42.0"],
+    "neuronx": ["optimum-neuron[neuronx]>=0.0.20", "transformers>=4.36.2,<4.42.0"],
     "graphcore": "optimum-graphcore",
     "furiosa": "optimum-furiosa",
     "amd": "optimum-amd",


setup.py
```diff
@@ -15,7 +15,7 @@
 REQUIRED_PKGS = [
     "coloredlogs",
     "sympy",
-    "transformers[sentencepiece]>=4.29.0,<4.44.0",
+    "transformers[sentencepiece]>=4.29,<=4.44",
```
IlyasMoutawwakil, line 18: I think using ">=" and "<=" for min and max versions makes them more explicit

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -336,8 +336,7 @@ def prepare_past_key_values(
             dtype = constructor.float16 if self.use_fp16 else constructor.float32
 
             # TODO: find a way to better handle this controlflow, this is EXTREMELY UGLY.
-            # "1" is the dummy sequence length
-            if self.model_type == "bloom":
+            if self.model_type == "bloom" and not check_if_transformers_greater("4.44"):
```
IlyasMoutawwakil, line 339: in modeling, idealy we should use the transformers version used for export, but idk if there's anyway to access that from an onnx model.

optimum/utils/preprocessing/token_classification.py
```diff
@@ -28,7 +28,7 @@
 
 class TokenClassificationProcessing(TaskProcessor):
     ACCEPTED_PREPROCESSOR_CLASSES = (PreTrainedTokenizerBase,)
-    DEFAULT_DATASET_ARGS = "conll2003"
+    DEFAULT_DATASET_ARGS = {"path": "conll2003", "trust_remote_code": True}
```
echarlaix, line 31: we shouldn't set `trust_remote_code` to True by default here, where is this needed ?

optimum/onnxruntime/modeling_decoder.py
echarlaix: what if a model was previously exported with a lower version ?

IlyasMoutawwakil: it will fail, we need something similar to https://github.com/huggingface/optimum-intel/pull/877#discussion_r1732369193

optimum/utils/preprocessing/token_classification.py
IlyasMoutawwakil: this dataset was being used with "remote code" (it generates the dataset) for as long as it existed, but the `datasets` package introduced the new argument `trust_remote_code` lately, this preprocessing is used in in tflite tests, we can either find a similar dataset that doesn't use code or allow remote code (that was already being executed).
https://huggingface.co/datasets/eriktks/conll2003/blob/main/conll2003.py

michaelbenayoun (approved):
Looks good to me. Not very knowledgeable on the bloom bettertransformer part though.

optimum/onnxruntime/modeling_decoder.py
echarlaix: Could this be added (can be done in a following PR) to make sure we don't lose compatibility with previously exported model ? Can be done like optimum-intel, an other option could be to check the graph to check what input shape is expected

IlyasMoutawwakil: using input shapes makes sense, I'm testing it rn

IlyasMoutawwakil: done!

echarlaix (approved):
LGTM, thanks @IlyasMoutawwakil

optimum/utils/preprocessing/token_classification.py
echarlaix: I think we need to change the dataset to something that doesn't require `trust_remote_code` to be set to True by default then (something that should be explicitly set by users)

PR #2000, opened 2024-08-28
Add Param Cache For Recompilation

zhenglongjiepheonix:
The parameter cache instance is needed to handle recompilation where we need to make sure the parameters we created in the first run are used, currently the use case does not fall into error even without param cache because we directly replace layers in layer cache in recompilation(parameters are replaced automatically because layers are replaced), but there are still some parameters which is not traced within a standalone module(like layernorm weight), it still works fine for now because we directly use the original parameter instead of creating new ones for initialization and weights loading if it is already on the current rank device, however, in cases where we need to support third-party backends like nanotron which has its own implementation of `NanotronParameter`, we do need to track all the newly created parameters so that no new parameter is created in recompilation.

### DIFF
--- a/optimum/fx/parallelization/core.py
+++ b/optimum/fx/parallelization/core.py
@@ -125,6 +125,11 @@ class ParallelExecutionCtx:
             because we have to make sure we don't initiate new parameters and replace original ones when
             recompilation happens in training process.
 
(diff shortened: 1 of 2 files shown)


michaelbenayoun (approved):
LGTM

PR #2017, opened 2024-09-06
Add Parallel Cross Entropy

zhenglongjiepheonix:
# What does this PR do?

- [x] add parallel cross-entropy implementation
- [x] add parallel axis propagation rules for cross-entropy
- [x] add parallel cross-entropy replacement logic in pass
- [x] modify tests to trigger the critical path


michaelbenayoun (approved):
LGTM!
Is it for the default backend?

zhenglongjiepheonix:
> LGTM! Is it for the default backend?

Nanotron backend will also use it because there is a `label_mask` input in Nanotron which is not compatible with transformers current implementation, the current implementation of shared cross entropy is compatible with transformers implementation, and it does not matter because cross entropy layer does not contain any backend-specific stuffs, it does not even contain a single parameter

zhenglongjiepheonix:
If everything is good, can we merge this, I don't have write access to the main branch now :(

PR #2021, opened 2024-09-10
Update/Fix Pipeline Mixins and ORT Pipelines

IlyasMoutawwakil:
# What does this PR do?

This PR allows for using the same modeling in diffusers for ORT diffusion pipelines without maintaining custom mixins.
It also fixes the issues in output reproducibility and numeric consistency vs diffusers observed in https://github.com/huggingface/optimum/pull/1960.
Breaking changes: 
- We export the vae encoder by outputting its latent distribution parameters instead of sampling during export, this way we can solve the above mentioned issues.

We also conduct a benchmark to prove this change doesn't hurt perf: https://github.com/huggingface/optimum/pull/2021#issuecomment-2349068900

We also show how reproducibility is impossible with an onnx model that performs random numbers generation at export-time: https://github.com/huggingface/optimum/pull/2021#issuecomment-2348905666

## Before submitting


optimum/pipelines/diffusers/pipeline_stable_diffusion_img2img.py
```diff
@@ -72,222 +88,434 @@ def check_inputs(
...
+            elif isinstance(generator, list):
+                if image.shape[0] < batch_size and batch_size % image.shape[0] == 0:
+                    image = torch.cat([image] * (batch_size // image.shape[0]), dim=0)
+                elif image.shape[0] < batch_size and batch_size % image.shape[0] != 0:
+                    raise ValueError(
+                        f"Cannot duplicate `image` of batch size {image.shape[0]} to effective batch_size {batch_size} "
+                    )
+
+                init_latents = [
+                    retrieve_latents(self.vae_encoder(image[i : i + 1]), generator=generator[i])
```
echarlaix, line 135: we would like to keep as much numpy as possible (no need to have np -> torch -> np) sot not sure why we need `retrieve_latents` + `DiagonalGaussianDistribution`, could you extend ?

IlyasMoutawwakil: `retrieve_latents` will sample from the `DiagonalGaussianDistribution` distribution that's parametrized by the outputs of the VAE encoder using the generator. Exporting the encoder with its final sampled output results in inconsistency with the diffusers output and inability to even have deterministic outputs (with the same generator/seed and outputs) in ORT pipelines for Img2Img and Inpainting.

optimum/exporters/onnx/model_configs.py
```diff
@@ -1111,8 +1111,8 @@ def ordered_inputs(self, model) -> Dict[str, Dict[int, str]]:
 
 
 class VaeEncoderOnnxConfig(VisionOnnxConfig):
-    ATOL_FOR_VALIDATION = 1e-2
-    # The ONNX export of a CLIPText architecture, an other Stable Diffusion component, needs the Trilu
+    ATOL_FOR_VALIDATION = 1e-4
+    # The ONNX export of a VaeEncoder architecture, an other Stable Diffusion component, needs the Trilu
```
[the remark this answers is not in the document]
IlyasMoutawwakil: so we set `opset >= 14` for `VaeEncoder` because another part of the pipeline requires it ?

optimum/pipelines/diffusers/pipeline_stable_diffusion_img2img.py
IlyasMoutawwakil: I think that (np -> torch -> np) is inevitable if we want drop-in replacement compatibility with diffusers, the whole logic in diffusers is dependent on torch functionalities (like distributions and generators) that can't be translated perfectly to numpy (the underlying implementations and random numbers generators are not the same).
I would argue that this doesn't add a lot of overhead in this case, especially on CPUs, e.g. [torch.from_numpy](https://pytorch.org/docs/stable/generated/torch.from_numpy.html) doesn't create a new tensor but rather points to the same block of memory. [Tensor.numpy](https://pytorch.org/docs/stable/generated/torch.Tensor.numpy.html) work similarly when force is False (default), so there's almost no copying or allocation of tensors/ndarrays (if I understand correctly).
With GPUs it allows for better perf because with torch we can easily implement io binding for cuda 😁.

IlyasMoutawwakil:
@echarlaix, as discussed, one reason why we need to export the distribution parameters instead of performing sampling at export time, is because we can't control the randomness otherwise, which results in some issues, one of which is not being able to reproduce the same output even with the same inputs and seeds. Something as simple as the following snippet does not reproduce the same output with ort.

```python
import random
import torch
import onnxruntime as ort
import numpy as np

def set_all_seeds(seed):
    random.seed(seed)
    np.random.seed(seed)
    torch.manual_seed(seed)
    ort.set_seed(seed)

class RandomModule(torch.nn.Module):
    def __init__(self):
        super(RandomModule, self).__init__()

    def forward(self, x):
        return x + torch.rand(x.size())
    
model = RandomModule()
x = torch.randn(1, 3, 224, 224)

set_all_seeds(0)
torch_output_1 = model(x)
set_all_seeds(0)
torch_output_2 = model(x)
print(np.testing.assert_allclose(torch_output_1, torch_output_2))

set_all_seeds(0)
torch.onnx.export(model, x, "random.onnx")
ort_session = ort.InferenceSession("random.onnx")
input_name = ort_session.get_inputs()[0].name
output_name = ort_session.get_outputs()[0].name

set_all_seeds(0)
ort_output_1 = ort_session.run([output_name], {input_name: x.numpy()})[0]

set_all_seeds(0)
ort_output_2 = ort_session.run([output_name], {input_name: x.numpy()})[0]

print(np.testing.assert_allclose(ort_output_1, ort_output_2))
```
PS: I just found out abourt ort.set_seed and it doesn't seem to make any difference.

IlyasMoutawwakil:
the overhead from using `from_numpy()` and `tensor.numpy` is minimal because these methods only copy the pointer to the block of memory that contains the array (on cpu). An end to end benchmark of the tiny model (where if there's an overhead, it should be most apparent) demonstrates that the perf is almost the same:
- on main:
```
 [MAIN-PROCESS][2024-09-13 13:55:22,816][process][INFO] -        + Received report from isolated process
[MAIN-PROCESS][2024-09-13 13:55:22,817][memory][INFO] -                 + load memory:
[MAIN-PROCESS][2024-09-13 13:55:22,818][memory][INFO] -                         - max RAM: 878.497792 (MB)
[MAIN-PROCESS][2024-09-13 13:55:22,818][latency][INFO] -                + load latency:
[MAIN-PROCESS][2024-09-13 13:55:22,818][latency][INFO] -                        - count: 1
[MAIN-PROCESS][2024-09-13 13:55:22,818][latency][INFO] -                        - total: 7.848160 s
[MAIN-PROCESS][2024-09-13 13:55:22,818][latency][INFO] -                        - mean: 7.848160 s
[MAIN-PROCESS][2024-09-13 13:55:22,818][latency][INFO] -                        - stdev: 0.000000 s (0.00%)
[MAIN-PROCESS][2024-09-13 13:55:22,818][latency][INFO] -                        - p50: 7.848160 s
[MAIN-PROCESS][2024-09-13 13:55:22,818][latency][INFO] -                        - p90: 7.848160 s
[MAIN-PROCESS][2024-09-13 13:55:22,818][latency][INFO] -                        - p95: 7.848160 s
[MAIN-PROCESS][2024-09-13 13:55:22,819][latency][INFO] -                        - p99: 7.848160 s
[MAIN-PROCESS][2024-09-13 13:55:22,819][memory][INFO] -                 + call memory:
[MAIN-PROCESS][2024-09-13 13:55:22,819][memory][INFO] -                         - max RAM: 1038.311424 (MB)
[MAIN-PROCESS][2024-09-13 13:55:22,819][latency][INFO] -                + call latency:
[MAIN-PROCESS][2024-09-13 13:55:22,819][latency][INFO] -                        - count: 10
[MAIN-PROCESS][2024-09-13 13:55:22,819][latency][INFO] -                        - total: 39.163408 s
[MAIN-PROCESS][2024-09-13 13:55:22,819][latency][INFO] -                        - mean: 3.916341 s
[MAIN-PROCESS][2024-09-13 13:55:22,819][latency][INFO] -                        - stdev: 0.175419 s (4.48%)
[MAIN-PROCESS][2024-09-13 13:55:22,819][latency][INFO] -                        - p50: 3.862534 s
[MAIN-PROCESS][2024-09-13 13:55:22,820][latency][INFO] -                        - p90: 4.104698 s
[MAIN-PROCESS][2024-09-13 13:55:22,820][latency][INFO] -                        - p95: 4.234256 s
[MAIN-PROCESS][2024-09-13 13:55:22,820][latency][INFO] -                        - p99: 4.337903 s
[MAIN-PROCESS][2024-09-13 13:55:22,820][latency][INFO] -                + call throughput: 0.255340 images/s
```
- on pr:
```
[MAIN-PROCESS][2024-09-13 13:57:36,918][process][INFO] -        + Received report from isolated process
[MAIN-PROCESS][2024-09-13 13:57:36,919][memory][INFO] -                 + load memory:
[MAIN-PROCESS][2024-09-13 13:57:36,919][memory][INFO] -                         - max RAM: 612.913152 (MB)
[MAIN-PROCESS][2024-09-13 13:57:36,920][latency][INFO] -                + load latency:
[MAIN-PROCESS][2024-09-13 13:57:36,920][latency][INFO] -                        - count: 1
[MAIN-PROCESS][2024-09-13 13:57:36,920][latency][INFO] -                        - total: 7.528061 s
[MAIN-PROCESS][2024-09-13 13:57:36,920][latency][INFO] -                        - mean: 7.528061 s
[MAIN-PROCESS][2024-09-13 13:57:36,920][latency][INFO] -                        - stdev: 0.000000 s (0.00%)
[MAIN-PROCESS][2024-09-13 13:57:36,920][latency][INFO] -                        - p50: 7.528061 s
[MAIN-PROCESS][2024-09-13 13:57:36,920][latency][INFO] -                        - p90: 7.528061 s
[MAIN-PROCESS][2024-09-13 13:57:36,920][latency][INFO] -                        - p95: 7.528061 s
[MAIN-PROCESS][2024-09-13 13:57:36,920][latency][INFO] -                        - p99: 7.528061 s
[MAIN-PROCESS][2024-09-13 13:57:36,921][memory][INFO] -                 + call memory:
[MAIN-PROCESS][2024-09-13 13:57:36,921][memory][INFO] -                         - max RAM: 1033.846784 (MB)
[MAIN-PROCESS][2024-09-13 13:57:36,921][latency][INFO] -                + call latency:
[MAIN-PROCESS][2024-09-13 13:57:36,921][latency][INFO] -                        - count: 10
[MAIN-PROCESS][2024-09-13 13:57:36,921][latency][INFO] -                        - total: 38.591033 s
[MAIN-PROCESS][2024-09-13 13:57:36,921][latency][INFO] -                        - mean: 3.859103 s
[MAIN-PROCESS][2024-09-13 13:57:36,921][latency][INFO] -                        - stdev: 0.174273 s (4.52%)
[MAIN-PROCESS][2024-09-13 13:57:36,921][latency][INFO] -                        - p50: 3.814653 s
[MAIN-PROCESS][2024-09-13 13:57:36,921][latency][INFO] -                        - p90: 4.081044 s
[MAIN-PROCESS][2024-09-13 13:57:36,921][latency][INFO] -                        - p95: 4.187505 s
[MAIN-PROCESS][2024-09-13 13:57:36,922][latency][INFO] -                        - p99: 4.272674 s
[MAIN-PROCESS][2024-09-13 13:57:36,922][latency][INFO] -                + call throughput: 0.259128 images/s
```
On actual/bigger pipelines the overhead is beyond minimal as one inference step takes a couple of seconds, while the conversion torch-np-torch takes microseconds:
```python
import time

import torch

with torch.no_grad():
    torch_tensor = torch.randn(1000, 1000)

    start = time.time()
    for _ in range(30):
        numpy_array = torch_tensor.numpy()
        torch_tensor = torch.from_numpy(numpy_array)
    end = time.time()

    print("Time taken for conversion: ", (end - start) / 30)
```

```
$ python bench.py 
Time taken for conversion:  7.883707682291667e-06
```

optimum/exporters/onnx/model_configs.py
```diff
@@ -1131,13 +1131,13 @@ def inputs(self) -> Dict[str, Dict[int, str]]:
...
-            "latent_sample": {0: "batch_size", 2: "height_latent", 3: "width_latent"},
+            "latent_parameters": {0: "batch_size", 2: "height_latent", 3: "width_latent"},
         }
 
 
 class VaeDecoderOnnxConfig(VisionOnnxConfig):
-    ATOL_FOR_VALIDATION = 1e-3
-    # The ONNX export of a CLIPText architecture, an other Stable Diffusion component, needs the Trilu
+    ATOL_FOR_VALIDATION = 1e-4
+    # The ONNX export of a VaeDecoder architecture, an other Stable Diffusion component, needs the Trilu
```
IlyasMoutawwakil, line 1140:
```suggestion
    # The ONNX export of a CLIPText architecture, an other Stable Diffusion component, needs the Trilu
```

optimum/pipelines/diffusers/pipeline_stable_diffusion_img2img.py
IlyasMoutawwakil: I just noticed that there's also already `.from_numpy` `.numpy()` calls in the unet loop:
https://github.com/huggingface/optimum/blob/ca36fc4f66577cd4ac2e6cedcc204d830a1f4985/optimum/pipelines/diffusers/pipeline_latent_consistency.py#L169-L173

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -538,177 +551,291 @@ def forward(
...
+        if "latent_sample" in model_outputs:
+            model_outputs["latents"] = model_outputs.pop("latent_sample")
+
+        if return_dict:
+            return model_outputs
+
         return ModelOutput(**model_outputs)
 
 
+class ORTVaeWrapper(ORTPipelinePart):
```
JingyaHuang, line 631: Love it!

tests/onnxruntime/test_modeling.py
```diff
@@ -320,6 +318,7 @@ def test_load_stable_diffusion_model_from_hub(self):
         self.assertIsInstance(model.vae_encoder, ORTModelVaeEncoder)
         self.assertIsInstance(model.unet, ORTModelUnet)
         self.assertIsInstance(model.config, Dict)
+        model(prompt="cat", num_inference_steps=2)
```
IlyasMoutawwakil, line 321: this to check inferencing and not only loading

tests/onnxruntime/test_modeling.py
```diff
@@ -144,7 +144,7 @@ def __init__(self, *args, **kwargs):
         self.ONNX_SEQ2SEQ_MODEL_ID = "optimum/t5-small"
         self.LARGE_ONNX_SEQ2SEQ_MODEL_ID = "facebook/mbart-large-en-ro"
         self.TINY_ONNX_SEQ2SEQ_MODEL_ID = "fxmarty/sshleifer-tiny-mbart-onnx"
-        self.TINY_ONNX_STABLE_DIFFUSION_MODEL_ID = "hf-internal-testing/tiny-random-OnnxStableDiffusionPipeline"
+        self.TINY_ONNX_STABLE_DIFFUSION_MODEL_ID = "IlyasMoutawwakil/tiny-stable-diffusion-onnx"
```
IlyasMoutawwakil, line 147: this is because this pipeline doesn't even have configs, it used to load but sisnce inferencing with it wasn't tested, it didn't raise any issues, but there is an issue, which is that configs are needed for inference.

IlyasMoutawwakil:
@echarlaix

tests/onnxruntime/test_modeling.py
echarlaix: can you move this model to https://huggingface.co/optimum-internal-testing (we try to progressively move our model for testing there like done for https://huggingface.co/optimum-internal-testing/tiny-random-llama)

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -495,41 +441,169 @@ def _load_config(cls, config_name_or_path: Union[str, os.PathLike], **kwargs):
...
+            model_outputs[output_name] = onnx_outputs[idx]
+
+            if use_torch:
+                model_outputs[output_name] = torch.from_numpy(model_outputs[output_name]).to(self.device)
+
+        return model_outputs
+
+    @abstractmethod
+    def forward(self, *args, **kwargs):
+        pass
```
echarlaix, line 563:
```suggestion
        raise NotImplementedError
```

echarlaix (approved):

optimum/onnxruntime/modeling_seq2seq.py
```diff
@@ -68,7 +68,7 @@
 if check_if_transformers_greater("4.25.0"):
     from transformers.generation import GenerationMixin
 else:
-    from transformers.generation_utils import GenerationMixin
+    from transformers.generation_utils import GenerationMixin  # type: ignore
```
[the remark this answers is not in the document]
IlyasMoutawwakil: it's a conditional import where the object might or might not exist in the version installed locally, resulting in IDE linters flagging it. That's a comment to avoid flagging by linters.

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -538,181 +612,239 @@ def forward(
...
-        model_outputs = self._prepare_onnx_outputs(use_torch, *onnx_outputs)
+        model_outputs = self.prepare_onnx_outputs(use_torch, *onnx_outputs)
+
+        if return_dict:
+            return model_outputs
+
+        return ModelOutput(**model_outputs)
+
+    @property
+    def add_embedding(self):
```
JingyaHuang, line 645: why not just override `_get_add_time_ids` and skip that dimension check part, feel it could be a bit hard-coded here.

IlyasMoutawwakil: yes that's what I did at first, then I thought this is more generalizable, but yeh it is too hacky, reverting it.

IlyasMoutawwakil:
I made sure the model also accepts `safety_chacker` if passed to `from_pretrained`.
and also if some sessions are passed directly to `from_pretrained` like in `diffusers`.

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -495,36 +383,156 @@ def _load_config(cls, config_name_or_path: Union[str, os.PathLike], **kwargs):
...
+    @property
+    def components(self) -> Dict[str, Any]:
+        components = {
+            "vae": self.vae,
+            "unet": self.unet,
+            "text_encoder": self.text_encoder,
+            "text_encoder_2": self.text_encoder_2,
+            "safety_checker": self.safety_checker,
+            "image_encoder": self.image_encoder,
+        }
```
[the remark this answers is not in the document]
JingyaHuang: And for making this to work you might need to allow passing ORTModelXXX to `__init__()` instead of just bare ort session. Never mind, it's a small nit, we can put it as todo anyways!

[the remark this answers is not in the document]
IlyasMoutawwakil: yes this one will need a bit more work for compatibility, I can do it in another PR.

IlyasMoutawwakil:
I uploaded a model that was exported using optimum 1.22 :
- https://huggingface.co/optimum-internal-testing/tiny-stable-diffusion-onnx

I also added the same checks in optimum-intel for missing attributes in model configs, that way even older versions can still work (e.g. `hf-internal-testing/tiny-stable-diffusion-openvino` was exported using `"_diffusers_version": "0.14.0"`).

I also removed the need for separate `ORTPipeline` "base" class and `ORTDiffusionPipeline`, now it's just `ORTDiffusionPipeline` which works as an entrypoint and parent class for other pipelines. This is more like what's done in `diffusers` as well.

echarlaix (approved):

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -278,14 +284,26 @@ def _from_pretrained(
                 else:
                     submodels[submodel] = load_method(model_save_path)
 
-        return cls(
+        # same as DiffusionPipeline.from_pretraoned, if called directly, it loads the class in the config
```
JingyaHuang, line 287: smol typo here

PR #2024, opened 2024-09-14
Update ort CIs (slow, gpu, train)

IlyasMoutawwakil:
# What does this PR do?

This is a PR that fixes:
- Onnxruntime Training testing and examples.
- Onnruntime GPU testing (CUDA & TRT EPs), and io binding implementation.

I tried my best to make things work with the least change, in some cases that was hard, e.g. io binding was dependant on the input order in the forward signature, which made some model types incompatible with the ORTModel signature.

## Before submitting


optimum/onnxruntime/modeling_ort.py
```diff
@@ -835,36 +810,37 @@ def _prepare_io_binding(
...
+
+            tensor_dtype = model_inputs[input_name].dtype
+            expected_dtype = TypeHelper.ort_type_to_torch_type(self.input_dtypes[input_name])
+            if tensor_dtype != expected_dtype:
+                model_inputs[input_name] = model_inputs[input_name].to(expected_dtype)
 
-            data_ptr = tensor.data_ptr()
-            if "past" in name and data_ptr == 0:
-                # During first generation, sequence_length can be 0 when use_cache=True, which results in data_ptr to also be 0.
-                # To keep compatibility with IO binding, we pass the data pointer of input_ids instead. This will have no impact because past_key_values will not be used during the first generation.
```
echarlaix, line 851: should we keep this comment ?

optimum/utils/import_utils.py
```diff
@@ -64,12 +84,9 @@ def _is_package_available(pkg_name: str, return_version: bool = False) -> Union[
...
 _torch_available, _torch_version = _is_package_available("torch", return_version=True)
-
-# importlib.metadata.version seem to not be robust with the ONNX Runtime extensions (`onnxruntime-gpu`, etc.)
-_onnxruntime_available = _is_package_available("onnxruntime", return_version=False)
-
-# TODO : Remove
-torch_version = version.parse(importlib.metadata.version("torch")) if _torch_available else None
+_onnxruntime_available, _onnxruntime_version = _is_package_available(
+    "onnxruntime", return_version=True, alt_pkg_names=["onnxruntime-gpu", "onnxruntime-training"]
+)
```
echarlaix, line 89: might be interesting to do the same for tensorflow

echarlaix (approved):
Thanks a lot for the great PR @IlyasMoutawwakil !

optimum/utils/import_utils.py
IlyasMoutawwakil: yes the same method should work for tf distributions.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -835,36 +810,37 @@ def _prepare_io_binding(
...
-                data_ptr = model_inputs[0].data_ptr()
+            data_ptr = model_inputs[input_name].data_ptr()
+            if data_ptr == 0:
+                data_ptr = model_inputs["input_ids"].data_ptr()
```
IlyasMoutawwakil, line 827:
```suggestion
            if data_ptr == 0:
                # During first generation, sequence_length can be 0 when use_cache=True, which results in data_ptr to also be 0.
                # To keep compatibility with IO binding, we pass the data pointer of input_ids instead. This will have no impact because past_key_values will not be used during the first generation.
                data_ptr = model_inputs["input_ids"].data_ptr()
```

tests/onnxruntime/test_diffusion.py
```diff
@@ -759,16 +763,18 @@ def test_image_reproducibility(self, model_arch: str):
         grid_parameters(
             {
                 "model_arch": SUPPORTED_ARCHITECTURES,
-                "provider": ["CUDAExecutionProvider", "ROCMExecutionProvider", "TensorrtExecutionProvider"],
```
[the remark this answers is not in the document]
IlyasMoutawwakil: We never had amd runners here and using them will require some work in `hf-workflows`, also there's nothing intrinsic to the `ROCMExecutionProvider`, quite the opposite since it doesn't support io binding.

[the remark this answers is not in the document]
IlyasMoutawwakil: For now I only removed it from diffusion tests

setup.py
```diff
@@ -84,7 +84,14 @@
         "datasets<=2.16",
         "transformers>=4.36,<4.38",
     ],
-    # "executorch": "optimum-executorch",
+    "exporters-executorch": [
+        "executorch>=0.4.0",
+        "transformers>=4.46",
+    ],
```
IlyasMoutawwakil, line 90:
```suggestion
```

optimum/utils/import_utils.py
```diff
@@ -64,12 +84,9 @@ def _is_package_available(pkg_name: str, return_version: bool = False) -> Union[
...
 _transformers_available, _transformers_version = _is_package_available("transformers", return_version=True)
 _torch_available, _torch_version = _is_package_available("torch", return_version=True)
-
-# importlib.metadata.version seem to not be robust with the ONNX Runtime extensions (`onnxruntime-gpu`, etc.)
-_onnxruntime_available = _is_package_available("onnxruntime", return_version=False)
-
-# TODO : Remove
-torch_version = version.parse(importlib.metadata.version("torch")) if _torch_available else None
+_onnxruntime_available, _onnxruntime_version = _is_package_available(
+    "onnxruntime", return_version=True, alt_pkg_names=["onnxruntime-gpu", "onnxruntime-training"]
```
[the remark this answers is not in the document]
IlyasMoutawwakil: done !

setup.py
```diff
@@ -58,7 +58,15 @@
         "datasets>=1.2.1",
         "evaluate",
         "protobuf>=3.20.1",
-        "accelerate",  # ORTTrainer requires it.
+        "transformers>=4.36,<4.49.0",
+    ],
+    "onnxruntime-training": [
```
IlyasMoutawwakil, line 63: The readme and docs should reflect this change

echarlaix:
Thanks a lot @IlyasMoutawwakil for adding many fixes !

setup.py
echarlaix: added in https://github.com/huggingface/optimum/pull/2173

PR #2012, opened 2024-09-03
update transformers imports for `deepspeed` and `is_torch_xla_available`

Rohan138:
# What does this PR do?
Minor fixes corresponding to https://github.com/huggingface/transformers/pull/25599 and https://github.com/huggingface/transformers/pull/29334

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #2028, opened 2024-09-18
Fix `is_torch_tpu_available` in ORT Trainer

regisss:
# What does this PR do?

Following https://github.com/huggingface/optimum/pull/2012. There is one occurrence of `is_torch_tpu_available` in the code of the `ORTTrainer` class, which will fail for `transformers >= 4.39.0`.

## Before submitting


IlyasMoutawwakil (approved):
LGTM thanks for the fix !

PR #2048, opened 2024-10-10
remove upper transformers version limit

echarlaix:
To not block a release from optimum-intel, optimum-neuron, optimum-habana and others to support the latest transformers release (removing the need to dsync it with an optimum release)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ except Exception as error:
 REQUIRED_PKGS = [
     "coloredlogs",
     "sympy",
-    "transformers[sentencepiece]>=4.29,<4.46.0",
+    "transformers[sentencepiece]>=4.29",
     "torch>=1.11",
     "packaging",
     "numpy",
@@ -54,6 +54,7 @@ EXTRAS_REQUIRE = {
         "datasets>=1.2.1",
         "evaluate",
         "protobuf>=3.20.1",
+        "transformers<4.46.0",
     ],
     "onnxruntime-gpu": [
         "onnx",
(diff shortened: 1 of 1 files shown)


JingyaHuang (approved):
LGTM, neuron is pinned to trfrs 4.43.2 anyway so...

setup.py
```diff
@@ -62,9 +63,10 @@
...
         "accelerate",  # ORTTrainer requires it.
+        "transformers<4.46.0",
     ],
-    "exporters": ["onnx", "onnxruntime", "timm"],
-    "exporters-gpu": ["onnx", "onnxruntime-gpu", "timm"],
+    "exporters": ["onnx", "onnxruntime", "timm", "transformers<4.46.0"],
+    "exporters-gpu": ["onnx", "onnxruntime-gpu", "timm", "transformers<4.46.0"],
     "exporters-tf": [
         "tensorflow>=2.4,<=2.12.1",
         "tf2onnx",
```
dacorvo, line 72: Maybe we can also lift the explicit restrictions here also, since `optimum-neuron` will enforce a pinned version anyway (cc @JingyaHuang ):

```python
    "neuron": ["optimum-neuron[neuron]>=0.0.20", "transformers>=4.36.2,<4.42.0"],
    "neuronx": ["optimum-neuron[neuronx]>=0.0.20", "transformers>=4.36.2,<4.42.0"],
```

echarlaix: Not related to this PR specifically so merging, let me know if you'd like me to include this change in today's release!

JingyaHuang: @dacorvo you mean something like:
```
    "neuron": ["optimum-neuron[neuron]>=0.0.20", "transformers>=4.36.2,<=4.43.2"],
    "neuronx": ["optimum-neuron[neuronx]>=0.0.20", "transformers>=4.36.2,<=4.43.2"],
``` 
? Or you would that we remove transformers from the extra? if so wouldn't we install once again a super old optimum-neuron if we do
```
pip install optimum[neuronx]
```

echarlaix: merging this PR as not directly related to this discussion, let me know if you want me to include any modification related to neuron in todays release!

PR #2050, opened 2024-10-10
Fix doc build

regisss:
# What does this PR do?

Recent doc builds failed:
- for main and releases: https://github.com/huggingface/optimum/actions/workflows/build_main_documentation.yml
- for PRs: https://github.com/huggingface/optimum/actions/workflows/build_pr_documentation.yml

It seems Github runners are now configured differently, hence these errors. This PR fixes that.

## Before submitting

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build_documentation:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


tengomucho:
The optimum-tpu installation fails because torch xla does not support python 3.12 yet, you need to use an earlier version (perhaps the simplest would be to use the `ubuntu:22.04` image).

echarlaix (approved):
Thanks a lot for the fix @regisss !

PR #2056, opened 2024-10-13
`ORTDiffusionPipeline`s with IO Binding

IlyasMoutawwakil:
# What does this PR do?

This is also my attempt to create a generalizable io binding framework, the idea is to always have `output_shapes = fn(input_shapes, known_shapes)` where `known_shapes` is mostly stuff we find in the config, we the use this information at runtime with a simple symbolic resolver, keeping the shape inference time minimal, to create output tensors in torch and thus accelerate inference without the need to pass by ort values / cupy / numpy.

## Before submitting

### DIFF
--- a/optimum/onnxruntime/modeling_diffusion.py
+++ b/optimum/onnxruntime/modeling_diffusion.py
@@ -17,7 +17,6 @@ import inspect
 import logging
 import os
 import shutil
-import traceback
 from abc import abstractmethod
 from collections import OrderedDict
 from pathlib import Path
@@ -64,7 +63,7 @@ from ..utils import (
     DIFFUSION_MODEL_VAE_DECODER_SUBFOLDER,
     DIFFUSION_MODEL_VAE_ENCODER_SUBFOLDER,
 )
-from .io_binding import IOBindingHelper, TypeHelper
+from .io_binding import TypeHelper
 from .modeling_ort import ONNX_MODEL_END_DOCSTRING, ORTModel
 from .utils import (
     ONNX_WEIGHTS_NAME,
@@ -192,7 +191,10 @@ class ORTDiffusionPipeline(ORTModel, DiffusionPipeline):
             if hasattr(component, attribute)
         }
 
-        if len(set(attribute_values.values())) > 1:
+        if len(attribute_values) == 0:
+            raise ValueError(f"Attribute {attribute} is not defined for any component.")
+        # make sure there is exactly one value for the attribute (bypass unhashable types)
+        elif len(set(map(str, attribute_values.values()))) > 1:
             raise ValueError(f"Attribute {attribute} is not the same across components: {attribute_values}.")
 
         return next(iter(attribute_values.values()))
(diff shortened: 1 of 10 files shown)


IlyasMoutawwakil:
You notice that no io binding code was added to specific models, and that's because this is also my attempt to create a generalisable io binding framework, the idea is to always have output_shapes = f(input_shapes, known_shapes) where known_shapes is mostly stuff we find in the config / used to construct the pytorch model, we the use this information at runtime with a simple symbolic resolver, keeping the shape inference time minimal, to create output tensors and thus eliminate the need to pass by ort values / cupy / numpy.

optimum/onnxruntime/modeling_ort.py
```diff
@@ -297,59 +286,72 @@ def dtype(self) -> torch.dtype:
...
 
+        device, provider_options = parse_device(device)
         provider = get_provider_for_device(device)
-        validate_provider_availability(provider)  # raise error if the provider is not available
+        validate_provider_availability(provider)
 
-        # IOBinding is only supported for CPU and CUDA Execution Providers.
-        if device.type == "cuda" and self._use_io_binding is False and provider == "CUDAExecutionProvider":
+        if self.use_io_binding is False and provider == "CUDAExecutionProvider":
             self.use_io_binding = True
```
tianleiwu, line 331: This overrides use_io_binding choice from user. What if user want to run performance test with io binding disabled?

I suggest that:
    if use_io_binding is None: change it to True
    if not use_io_binding and it is cuda provider, log a warning.

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -146,13 +163,93 @@ def __init__(
...
+    def provider(self) -> str:
+        return self._validate_same_attribute_value_across_components("provider")
+
+    @property
+    def providers_options(self) -> Dict[str, Dict[str, Any]]:
+        return self._validate_same_attribute_value_across_components("providers_options")
+
+    @property
+    def provider_options(self) -> Dict[str, Any]:
+        return self._validate_same_attribute_value_across_components("provider_options")
```
tianleiwu, line 224: It is not necessary to validate same value across components.

I think it is feasible to use different provider and different provider options for components. For example, we can run text_encoder with CPU, and unet with CUDA provider. Or we want to enable cuda graph in one component but not the other in provider option.

May add some comments and loose the constraint later.

optimum/onnxruntime/modeling_ort.py
IlyasMoutawwakil: This is the default behavior in other ORTModels, I kept it for consistency (I'm not a fan of it)

optimum/onnxruntime/modeling_diffusion.py
IlyasMoutawwakil: there's a comment in `_validate_same_attribute_value_across_components` definition explaining the reasoning behind these checks, which is exactly what you said. Pipeline attributes can be accessed but they only make sense when they're consistent, for now this is my proposition for multi model parts pipelines, an alternative would be to return that of the main component (unet/transformer) or not supporting these attributes at all for the main pipeline.

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -465,35 +538,133 @@ def dtype(self):
...
+
+        if unresolved_output_shapes:
+            raise ValueError(
+                f"Failed to resolve the symbolic dimensions for the model {self.__class__.__name__}. "
+                f"Unresolved output shapes: {unresolved_output_shapes}"
+            )
+
+        return resolved_output_shapes
+
+    def _prepare_io_binding(self, model_inputs: torch.Tensor) -> Tuple[ort.IOBinding, Dict[str, torch.Tensor]]:
```
tianleiwu, line 623: model_inputs data type is Dict[str, torch.Tensor]

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -465,35 +538,133 @@ def dtype(self):
...
+            io_binding.bind_output(
+                name=output_name,
+                device_type=self.device.type,
+                device_id=self.device.index if self.device.index is not None else -1,
+                element_type=TypeHelper.ort_type_to_numpy_type(self.output_dtypes[output_name]),
+                buffer_ptr=self._output_buffers[output_name].data_ptr(),
+                shape=tuple(self._output_buffers[output_name].size()),
+            )
+
+        return io_binding, model_inputs, self._output_buffers
```
tianleiwu, line 659: model_inputs are not used by caller. Not need to return here.

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -465,35 +538,133 @@ def dtype(self):
...
+        for input_name in self.input_names.keys():
+            input_dtype = TypeHelper.ort_type_to_torch_type(self.input_dtypes[input_name])
+
+            if model_inputs[input_name].dtype != input_dtype:
+                model_inputs[input_name] = model_inputs[input_name].to(input_dtype)
+
+            io_binding.bind_input(
+                name=input_name,
+                device_type=self.device.type,
+                device_id=self.device.index if self.device.index is not None else -1,
```
tianleiwu, line 635: Suggest to `assert self.device.index is not None`. 
ORT does not handle device id -1

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -465,35 +538,133 @@ def dtype(self):
...
+                f"Failed to set the device to {device}. "
+                f"Requested provider {provider} with options: {provider_options}, "
+                f"but got provider {self.provider} with options: {self.provider_options}."
+            )
+
+        self._device = device
+
+        return self
+
+    def _get_output_shapes(self, **model_inputs: torch.Tensor) -> Dict[str, int]:
```
tianleiwu, line 577: This function is very slow (in milliseconds). Any way to improve the performance?

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -465,35 +538,133 @@ def dtype(self):
...
+            input_dtype = TypeHelper.ort_type_to_torch_type(self.input_dtypes[input_name])
+
+            if model_inputs[input_name].dtype != input_dtype:
+                model_inputs[input_name] = model_inputs[input_name].to(input_dtype)
+
+            io_binding.bind_input(
+                name=input_name,
+                device_type=self.device.type,
+                device_id=self.device.index if self.device.index is not None else -1,
+                element_type=TypeHelper.ort_type_to_numpy_type(self.input_dtypes[input_name]),
```
tianleiwu, line 636: For onnxruntime 1.20 or later, recommend using onnx type instead of numpy type here. It is because numpy does not support bfloat16, float8; but onnx type supports it.

The mapping from ort type to onnx type is like:
{
   "tensor(float)":  onnx.TensorProto.FLOAT,
   "tensor(float16)":  onnx.TensorProto.FLOAT16,
   ...
}

PR #2065, opened 2024-10-16
Clean up ORT documentation

echarlaix:
Clarify that on-the-fly conversion + model saving + pushing your model on the hub work for all models (not just transformers models). 
Also replace `ORTXxxxPipeline` with `ORTDiffusionPipeline`, couple of code snippets were removed to not overburden the documentation, but could be moved in a tutorial page like done for openvino https://huggingface.co/docs/optimum/main/en/intel/openvino/tutorials/diffusers


regisss (approved):
Awesome @echarlaix, that looks much clearer :smile:

PR #2066, opened 2024-10-16
Enable ONNX export of CLIP models with sdpa

echarlaix:


IlyasMoutawwakil:
Should work now as of pytorch 2.5 release, also with this release, no need to override AttnProcessor in diffusion pipelines.

IlyasMoutawwakil:
> Should work now as of pytorch 2.5 release.

can we condition on torch version 2.5

PR #2074, opened 2024-10-21
Add support to export ColPali Model to ONNX

akshayballal95:
# What does this PR do?

This PR adds support for exporting the ColPali merged model to ONNX format. The model is based on the "pali gemma" model type, and thus, I have added it under the "feature-extraction" task. Do suggest if there is a better way to integrate this. If this looks fine with a few modifications, I can add support for the Paligemma text-generation task as well. 

## Before submitting
- I have tested the model by exporting the https://huggingface.co/vidore/colpali-v1.2-merged model and tested it by providing both document inputs and text query inputs. 

## Who can review?


akshayballal95:

Are you open to merging this?

echarlaix:
Apologies for the delay @akshayballal95, could you add a test with a tiny random model like https://huggingface.co/optimum-internal-testing/tiny-random-PaliGemmaForConditionalGeneration, can be added here https://github.com/huggingface/optimum/blob/main/tests/exporters/exporters_utils.py#L37

optimum/exporters/onnx/model_patcher.py
```diff
@@ -509,6 +509,24 @@ def patched_forward(*args, **kwargs):
...
+    ):
+        super().__init__(config, model, model_kwargs)
+
+        def patched_forward(input_ids=None, pixel_values=None, attention_mask=None, **kwargs):
+            outputs = self.orig_forward(
+                input_ids=input_ids, pixel_values=pixel_values, attention_mask=attention_mask, **kwargs
+            )
+            return outputs
+
+        self.patched_forward = patched_forward
```
echarlaix, line 527: why is it needed ?

optimum/exporters/onnx/model_patcher.py
```diff
@@ -1154,3 +1172,18 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        model: Union["PreTrainedModel", "TFPreTrainedModel"],
+        model_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        super().__init__(config, model, model_kwargs)
+        def patched_forward(input_ids=None, pixel_values=None, attention_mask=None, **kwargs):
+            outputs = self.orig_forward(
+                input_ids=input_ids, pixel_values=pixel_values, attention_mask=attention_mask, **kwargs
+            )
+            return outputs
+        self.patched_forward = patched_forward
```
echarlaix, line 1189: already added above 
```suggestion
```

optimum/exporters/tasks.py
```diff
@@ -949,6 +949,10 @@ class TasksManager:
             "text-generation-with-past",
             onnx="GraniteOnnxConfig",
         ),
+        "paligemma": supported_tasks_mapping(
+            "feature-extraction",
```
echarlaix, line 953: looks like the task here should be `"image-text-to-text"` https://huggingface.co/google/paligemma2-3b-pt-224/blob/main/README.md?code=true#L4


In which case should also be added [here]( https://github.com/huggingface/optimum/blob/4a7cb298140ee9bed968d98a780a950d15bb2935/optimum/exporters/tasks.py#L203)
```
"image-text-to-text": "AutoModelForImageTextToText"
```
 
and here 

https://github.com/huggingface/optimum/blob/4a7cb298140ee9bed968d98a780a950d15bb2935/optimum/exporters/onnx/base.py#L144

optimum/exporters/onnx/model_configs.py
```diff
@@ -2474,3 +2475,48 @@ class EncoderDecoderOnnxConfig(EncoderDecoderBaseOnnxConfig):
...
+            if self.task == "feature-extraction":
+                generator = self.DUMMY_INPUT_GENERATOR_CLASSES[0](self.task, self._normalized_config)
+                prefix_tensor = generator.constant_tensor(
+                    shape=[dummy_inputs["input_ids"].shape[0], 1024],
+                    value=self._normalized_config.image_token_index,
+                    framework=framework,
+                )
+                dummy_inputs["input_ids"] = generator.concat_inputs([prefix_tensor, dummy_inputs["input_ids"]], dim=1)
+                dummy_inputs["attention_mask"] = generator.random_mask_tensor(
+                    shape=[generator.batch_size, generator.sequence_length + 1024],
```
echarlaix, line 2510: where does the value 1024 comes from? shouldn't it depend from the models config ?

optimum/exporters/onnx/model_patcher.py
akshayballal95: its needed because the original ColPali Model only takes **kwargs and no named arguments. This resulted in an error. This fixes that error

akshayballal95:
> Apologies for the delay @akshayballal95, could you add a test with a tiny random model like https://huggingface.co/hf-internal-testing/tiny-random-PaliGemmaForConditionalGeneration, can be added here https://github.com/huggingface/optimum/blob/main/tests/exporters/exporters_utils.py#L37

I have added the conversion test. It works fine locally.

PR #2085, opened 2024-10-30
add transformers 4.36 tests

echarlaix:


IlyasMoutawwakil:
only failing tests are with granite which is not supported on transformers 4.36 https://github.com/huggingface/optimum/actions/runs/11859647666/job/33053100596?pr=2085

PR #2086, opened 2024-10-30
Drop python 3.8

echarlaix:


.github/workflows/dev_test_optimum_common.yml
```diff
@@ -12,10 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version:
-        - 3.7
-        - 3.8
-        - 3.9
+        python-version: ['3.9', '3.11']
```
michaelbenayoun, line 15: Why not 3.10?

setup.py
```diff
@@ -123,9 +123,9 @@
         "Intended Audience :: Education",
         "Intended Audience :: Science/Research",
         "Operating System :: OS Independent",
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
```
michaelbenayoun, line 128: Why not 3.11?

.github/workflows/dev_test_optimum_common.yml
echarlaix: 3.11 also has the security status : https://devguide.python.org/versions/ so should be tested

setup.py
echarlaix: added in https://github.com/huggingface/optimum/pull/2086/commits/44059cc1fa785ff96fbc01be0be00d64fc55e136

michaelbenayoun (approved):
LGTM thanks!

echarlaix:
also adding every subpackages maintainer to make sure this works for everyone before we do any release @regisss @IlyasMoutawwakil @JingyaHuang @mht-sharma @

PR #2089, opened 2024-11-05
Update Dockerfile base image

echarlaix:
Fixes doc generation now that python 3.8 was dropped in https://github.com/huggingface/optimum/pull/2086

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -1,4 +1,4 @@
-FROM nikolaik/python-nodejs:python3.8-nodejs18
+FROM nikolaik/python-nodejs:python3.9-nodejs18
 
 ARG commit_sha
 ARG clone_url


docs/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM nikolaik/python-nodejs:python3.8-nodejs18
+FROM nikolaik/python-nodejs:python3.9-nodejs18
```
echarlaix, line 1: could make sense to replace it with 

```suggestion
FROM node:18
```

cc @regisss

regisss: If the workflow works with this image, sure let's do it!

echarlaix: merging to fix main broken since https://github.com/huggingface/optimum/pull/2086, will open a following PR

PR #1578, opened 2023-12-08
Add Timm support in ORTModelForImageClassification

mht-sharma:
# What does this PR do?

Adds the inference of Timm models in `Optimum` using `ORTModelForImageClassification`

```python
from urllib.request import urlopen

import timm
import torch
from PIL import Image

from optimum.onnxruntime import ORTModelForImageClassification

model_id = "timm/resnext101_64x4d.c1_in1k"
model = ORTModelForImageClassification.from_pretrained(model_id, export=True)

# get model specific transforms (normalization, resize)
data_config = timm.data.resolve_data_config(pretrained_cfg=model.config.pretrained_cfg)
transforms = timm.data.create_transform(**data_config, is_training=False)

img = Image.open(
    urlopen("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/beignets-task-guide.png")
)
output = model(transforms(img).unsqueeze(0)).logits  # unsqueeze single image into batch of 1
top5_probabilities, top5_class_indices = torch.topk(torch.softmax(output, dim=1) * 100, k=5)

```

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #2090, opened 2024-11-06
Export to ExecuTorch: Code Skeleton

guangy10:
# What does this PR do?

This PR is the first initiative to create an e2e path for ["Export to ExecuTorch"](https://github.com/huggingface/transformers/issues/32253). 

In this very first revision, I'm focusing outline the skeleton of the integration work:

`./setup.py`: Specify new dependency in order to "Export to ExecuTorch". In this case, it's adding a new pip package `executorch`(Beta version) and require the latest released `transformfers` (models from older versions may not work with ExecuTorch)

### Env setup
`./optimum/commands/export/executorch.py|__init__.py`: To support running export with ExecuTorch backend via cli.

### Export AOT(Ahead-of-time)
`./optimum/exporters/executorch/`: Main entry point to export via ExecuTorch. 
```
optimum/exporters/executorch/
├── __main__.py
├── causal_lm.py
├── convert.py
├── recipe_registry.py
├── task_registry.py
└── xnnpack.py
```
 - It contains the `coverter.py` which defines the common workflow to export a `transformers` model from 🤗 to ExecuTorch. 
 - The export workflow can be called with different recipes (e.g. quantize and delegate to XNNPACK, Core ML, QNN, MPS, etc.). 
 - For each model that performs different tasks may require different patches, this is handled via registered tasks. For example, `CausalLM` w/ cache must be loaded with `generation_config` and exported via `transformers.integrations.executorch`.

### Run with exported ExecuTorch model
`optimum/executorchruntime/modeling_executorch.py` defines the python class that wrap the Transformers `AutoModelForXXX` classes. We start with `ExecuTorchModelForCausalLM` in this file, which inherits the base `OptimizedModel` and overrides all abstract methods. This is where the ExecuTorch pybinding and the runtime gets integrated. 

`optimum/pipelines/pipelines_base.py` defines a pipeline for ExecuTorch where everything is put together, deciding which model/task/recipe to load and run. It also consists of tokenizers that paired with the model for encoding/decoding.

## Before submitting

## Who can review?

@michaelbenayoun @echarlaix

### DIFF
--- a/optimum/commands/export/executorch.py
+++ b/optimum/commands/export/executorch.py
@@ -1,0 +1,54 @@
+"""Defines the command line for the export with ExecuTorch."""
+
+import argparse
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from ...exporters import TasksManager
+from ..base import BaseOptimumCLICommand
+
+
+if TYPE_CHECKING:
+    from argparse import ArgumentParser
+
+
+def parse_args_executorch(parser):
+    required_group = parser.add_argument_group("Required arguments")
+    required_group.add_argument(
+        "-m", "--model", type=str, required=True, help="Model ID on huggingface.co or path on disk to load model from."
+    )
+    required_group.add_argument(
+        "output", type=Path, help="Path indicating the directory where to store the generated ExecuTorch model."
+    )
+
+    optional_group = parser.add_argument_group("Optional arguments")
+    optional_group.add_argument(
+        "--task",
+        default="auto",
+        help=(
+            "The task to export the model for. If not specified, the task will be auto-inferred based on the model. Available tasks depend on the model, but are among:"
+            f" {str(TasksManager.get_all_tasks())}. For decoder models, use `xxx-with-past` to export the model using past key values in the decoder."
+        ),
+    )
+    optional_group.add_argument(
+        "--recipe",
+        type=str,
+        default="xnnpack_fp32",
+        help='Pre-defined recipes for export to ExecuTorch. Defaults to "xnnpack_fp32".',
+    )
+
+
+class ExecuTorchExportCommand(BaseOptimumCLICommand):
+    @staticmethod
+    def parse_args(parser: "ArgumentParser"):
+        return parse_args_executorch(parser)
+
+    def run(self):
+        from ...exporters.executorch import main_export
+
+        main_export(
+            model_name_or_path=self.args.model,
+            output=self.args.output,
+            task=self.args.task,
+            recipe=self.args.recipe,
+        )
--- a/optimum/executorchruntime/modeling_executorch.py
+++ b/optimum/executorchruntime/modeling_executorch.py
@@ -1,0 +1,158 @@
+"""ExecuTorchModelForXXX classes, allowing to run ExecuTorch Models with ExecuTorch Runtime using the same API as Transformers."""
+
+import logging
+import os
+import warnings
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
+
+import torch
+from executorch.extension.pybindings.portable_lib import _load_for_executorch
+from huggingface_hub import hf_hub_download
+from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
+from huggingface_hub.utils import EntryNotFoundError
+from transformers import (
+    AutoConfig,
+    AutoModel,
+    GenerationMixin,
+    AutoModelForCausalLM,
+    GenerationConfig,
+)
+from transformers.integrations.executorch import TorchExportableModuleWithStaticCache
+from transformers.modeling_outputs import (
+    BaseModelOutput,
+    CausalLMOutput,
+    CausalLMOutputWithPast,
+    ModelOutput,
+)
+
+from ..exporters import TasksManager
+from ..exporters.executorch import main_export
+from ..modeling_base import FROM_PRETRAINED_START_DOCSTRING, OptimizedModel
+
+if TYPE_CHECKING:
+    from transformers import PretrainedConfig
+
+
+logger = logging.getLogger(__name__)
+
+
+class ExecuTorchModelForCausalLM(OptimizedModel):
+    """
+    ExecuTorch model with a causal language modeling head for ExecuTorch Runtime inference.
+    """
+
+    auto_model_class = AutoModelForCausalLM
+
+    def __init__(
+        self,
+        model: "ExecuTorchModule",  # Similar to ort.InferenceSession for ORTModel
+        config: "PretrainedConfig",
+    ):
+        super().__init__(model, config)
+        # TODO: Remove once figure out how make `ExecuTorchModule` inherit from `PreTrainedModel`
+        self.et_model = model
+
+    def forward(self, input_ids: torch.Tensor, cache_position: torch.Tensor) -> torch.Tensor:
+        return self.et_model.forward(input_ids, cache_position)[0]
+
+    @classmethod
+    def _from_pretrained(
+        cls,
+        model_dir_path: Union[str, Path],
+        task: str,
+        recipe: str,
+        config: "PretrainedConfig",
+        use_auth_token: Optional[Union[bool, str]] = None,
+        token: Optional[Union[bool, str]] = None,
+        revision: Optional[str] = None,
+        force_download: bool = False,
+        cache_dir: str = HUGGINGFACE_HUB_CACHE,
+        subfolder: str = "",
+        local_files_only: bool = False,
+    ) -> "ExecuTorchModelForCausalLM":
+        if use_auth_token is not None:
+            warnings.warn(
+                "The `use_auth_token` argument is deprecated and will be removed soon. Please use the `token` argument instead.",
+                FutureWarning,
+            )
+            if token is not None:
+                raise ValueError("You cannot use both `use_auth_token` and `token` arguments at the same time.")
+            token = use_auth_token
+
+        full_path = os.path.join(f"{model_dir_path}", "model.pte")
+        model = _load_for_executorch(full_path)
+
+        return cls(
+            model=model,
+            config=config,
+        )
+
+    @classmethod
+    def _export(
+        cls,
+        model_id: str,
+        task: str,
+        recipe: str,
+        config: "PretrainedConfig",
+        use_auth_token: Optional[Union[bool, str]] = None,
+        token: Optional[Union[bool, str]] = None,
+        revision: Optional[str] = None,
+        force_download: bool = False,
+        cache_dir: str = HUGGINGFACE_HUB_CACHE,
+        subfolder: str = "",
+        local_files_only: bool = False,
+        trust_remote_code: bool = False,
+    ):
+        if use_auth_token is not None:
+            warnings.warn(
+                "The `use_auth_token` argument is deprecated and will be removed soon. Please use the `token` argument instead.",
+                FutureWarning,
+            )
+            if token is not None:
+                raise ValueError("You cannot use both `use_auth_token` and `token` arguments at the same time.")
+            token = use_auth_token
+
+        save_dir = TemporaryDirectory()
+        save_dir_path = Path(save_dir.name)
+
+        # Export to ExecuTorch and save the pte file to the temporary directory
+        main_export(
+            model_name_or_path=model_id,
+            output=save_dir_path,
+            task=task,
+            recipe=recipe,
+            subfolder=subfolder,
+            revision=revision,
+            cache_dir=cache_dir,
+            token=token,
+            local_files_only=local_files_only,
+            force_download=force_download,
+            trust_remote_code=trust_remote_code,
+        )
+
+        return cls._from_pretrained(
+            model_dir_path=save_dir_path,
+            task=task,
+            recipe=recipe,
+            config=config,
+            use_auth_token=use_auth_token,
+            subfolder=subfolder,
+            revision=revision,
+            cache_dir=cache_dir,
+            token=token,
+            local_files_only=local_files_only,
+            force_download=force_download,
+        )
+
+    def generate(
+        self,
+        prompt_tokens: List[int],
+    ) -> List[int]:
+        """
+        Generate a sequence of token ids using the ExecuTorchModule.
+
+        `pipeline()` is where everything puts together. It consists of the tokenizer for encoding the inputs and decoding the model generated outputs.
+        """
+        pass
--- a/optimum/exporters/executorch/convert.py
+++ b/optimum/exporters/executorch/convert.py
@@ -1,0 +1,54 @@
+"""ExecuTorch model check and export functions."""
+
+import os
+from pathlib import Path
+from typing import Union
+
+from transformers.utils import is_torch_available
+
+from ...utils import (
+    TORCH_MINIMUM_VERSION,
+    check_if_transformers_greater,
+    is_diffusers_available,
+    logging,
+)
+from ..tasks import TasksManager
+from .recipe_registry import recipe_registry
+
+
+if is_torch_available():
+    import torch
+    from transformers.modeling_utils import PreTrainedModel
+
+
+def export_to_executorch(
+    model: Union["PreTrainedModel", "TorchExportableModuleWithStaticCache"],
+    task: str,
+    recipe: str,
+    output_dir: Union[str, Path],
+    **kwargs,
+):
+    """
+    Full-suite ExecuTorch export function, exporting **from a pre-trained PyTorch model**. This function is especially useful in case one needs to do modifications on the model, as overriding a forward call, before exporting to ExecuTorch.
+
+    Args:
+        model (`Union["PreTrainedModel", "TorchExportableModuleWithStaticCache"]`):
+            PyTorch model to export to ExecuTorch.
+        recipe (`str`, defaults to `"xnnpack"`):
+            The recipe to use to do the export. Defaults to "xnnpack".
+        output (`Union[str, Path]`):
+            Path indicating the directory where to store the generated ExecuTorch model.
+    """
+    try:
+        recipe_func = recipe_registry.get(recipe)
+    except KeyError as e:
+        raise RuntimeError(f"The recipe '{recipe}' isn't registered. Detailed error: {e}")
+
+    executorch_prog = recipe_func(model, task, kwargs)
+
+    full_path = os.path.join(f"{output_dir}", "model.pte")
+    with open(full_path, "wb") as f:
+        executorch_prog.write_to_file(f)
+        print(f"Saved exported program to {full_path}")
+
+    return executorch_prog
--- a/optimum/pipelines/pipelines_base.py
+++ b/optimum/pipelines/pipelines_base.py
@@ -293,9 +293,28 @@ def load_ort_pipeline(
     return model, model_id, tokenizer, feature_extractor
 
 
+def load_executorch_pipeline(
+    model,
+    targeted_task,
+    load_tokenizer,
+    tokenizer,
+    feature_extractor,
+    load_feature_extractor,
+    SUPPORTED_TASKS,
+    subfolder: str = "",
+    token: Optional[Union[bool, str]] = None,
+    revision: str = "main",
+    model_kwargs: Optional[Dict[str, Any]] = None,
+    config: AutoConfig = None,
+    **kwargs,
+):
+    raise NotImplementedError("Executorch pipeline is not implemented yet.")
+
+
 MAPPING_LOADING_FUNC = {
     "ort": load_ort_pipeline,
     "bettertransformer": load_bettertransformer,
+    "executorch": load_executorch_pipeline,
 }
 
 
--- a/setup.py
+++ b/setup.py
@@ -88,6 +88,10 @@ EXTRAS_REQUIRE = {
         "datasets<=2.16",
         "transformers>=4.26,<4.38",
     ],
+    "exporters-executorch": [
+        "executorch>=0.4.0",
+        "transformers>=4.46",
+    ],
     "diffusers": ["diffusers"],
     "intel": "optimum-intel>=1.18.0",
     "openvino": "optimum-intel[openvino]>=1.18.0",
--- a/optimum/commands/export/__init__.py
+++ b/optimum/commands/export/__init__.py
@@ -14,5 +14,6 @@
 
 
 from .base import ExportCommand
+from .executorch import ExecuTorchExportCommand
 from .onnx import ONNXExportCommand
 from .tflite import TFLiteExportCommand
--- a/optimum/exporters/executorch/__main__.py
+++ b/optimum/exporters/executorch/__main__.py
@@ -1,0 +1,138 @@
+"""Entry point to the optimum.exporters.executorch command line."""
+
+import argparse
+import warnings
+from pathlib import Path
+
+from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
+from requests.exceptions import ConnectionError as RequestsConnectionError
+from transformers import AutoConfig, AutoTokenizer
+from transformers.utils import is_torch_available
+
+from ...commands.export.executorch import parse_args_executorch
+from ...utils import logging
+from ..tasks import TasksManager
+from .convert import export_to_executorch
+from .task_registry import task_registry
+
+if is_torch_available():
+    import torch
+
+from typing import Optional, Union
+
+
+logger = logging.get_logger()
+
+
+def main_export(
+    model_name_or_path: str,
+    output: Union[str, Path],
+    task: str,
+    recipe: str = "xnnpack",
+    cache_dir: str = HUGGINGFACE_HUB_CACHE,
+    trust_remote_code: bool = False,
+    pad_token_id: Optional[int] = None,
+    subfolder: str = "",
+    revision: str = "main",
+    force_download: bool = False,
+    local_files_only: bool = False,
+    use_auth_token: Optional[Union[bool, str]] = None,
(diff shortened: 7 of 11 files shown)


optimum/executorchruntime/modeling_executorch.py
```diff
@@ -0,0 +1,158 @@
...
+        self,
+        model: "ExecuTorchModule",  # Similar to ort.InferenceSession for ORTModel
+        config: "PretrainedConfig",
+    ):
+        super().__init__(model, config)
+        # TODO: Remove once figure out how make `ExecuTorchModule` inherit from `PreTrainedModel`
+        self.et_model = model
+
+    def forward(self, input_ids: torch.Tensor, cache_position: torch.Tensor) -> torch.Tensor:
+        return self.et_model.forward(input_ids, cache_position)[0]
```
guangy10, line 58: @michaelbenayoun I had trouble to figure out how the runtime is integrated. It seems like ONNX is wrapping the exported module into a `ort.InferenceSession` but I can't find where the `ort.InferenceSession` is defined and how it could be passed down to the parent class and recognized as a `PreTrainedModel` that can be used in the pipeline later. Would you mind explaining a bit?

A bit context, for ExecuTorch, `ExecuTorchModule` is the ExecuTorch exported format that gets loaded from the local file system. The `.pte` file is sort of equivalent to the `.onnx` file from ONNX export.

optimum/executorchruntime/modeling_executorch.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @classmethod
+    def _from_pretrained(
```
guangy10, line 61: If I understand the example here https://github.com/huggingface/optimum/tree/7e8d857d1ed6be32046324bf8f424690f116b4e9?tab=readme-ov-file#run-the-exported-model-using-onnx-runtime correctly, this method is to load a exported model from local file system, correct?

optimum/executorchruntime/modeling_executorch.py
```diff
@@ -0,0 +1,158 @@
...
+                raise ValueError("You cannot use both `use_auth_token` and `token` arguments at the same time.")
+            token = use_auth_token
+
+        full_path = os.path.join(f"{model_dir_path}", "model.pte")
+        model = _load_for_executorch(full_path)
+
+        return cls(
+            model=model,
+            config=config,
+        )
```
guangy10, line 90: I'm not 100% sure about the returning model here: It should be an exported model, not the eager one to be exported, correct?

optimum/executorchruntime/modeling_executorch.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @classmethod
+    def _export(
```
guangy10, line 93: I'm not 100% sure why this method is needed and how it will be used. By looking at the onnxruntime code, this method is only used in a deprecated method (https://github.com/guangy10/optimum/blob/main/optimum/onnxruntime/modeling_ort.py#L570-L635), which doesn't make sense to me.

optimum/executorchruntime/modeling_executorch.py
```diff
@@ -0,0 +1,158 @@
...
+            recipe=recipe,
+            config=config,
+            use_auth_token=use_auth_token,
+            subfolder=subfolder,
+            revision=revision,
+            cache_dir=cache_dir,
+            token=token,
+            local_files_only=local_files_only,
+            force_download=force_download,
+        )
```
guangy10, line 147: Here I assume this method will first fetch the model from transformers, export it and save it to a local path, then load it back via `from_pretrained`. Let me know if this how this method should be implemented. 

Please guide me how this method is used in the upper stack? As mentioned, ONNX is using this API only in a deprecated code `_from_transformers`.

optimum/executorchruntime/modeling_executorch.py
```diff
@@ -0,0 +1,158 @@
...
+    def generate(
+        self,
+        prompt_tokens: List[int],
+    ) -> List[int]:
+        """
+        Generate a sequence of token ids using the ExecuTorchModule.
+
+        `pipeline()` is where everything puts together. It consists of the tokenizer for encoding the inputs and decoding the model generated outputs.
+        """
+        pass
```
guangy10, line 158: Idk it this method belongs to this class. This should be the place where we execute the model with delegated kernel libraries via pybinding. For ONNX, it seems to be the `ort.InferenceSession` (can't find its definition anywhere)

optimum/pipelines/pipelines_base.py
```diff
@@ -293,9 +293,28 @@ def load_ort_pipeline(
...
+    **kwargs,
+):
+    raise NotImplementedError("Executorch pipeline is not implemented yet.")
+
+
 MAPPING_LOADING_FUNC = {
     "ort": load_ort_pipeline,
     "bettertransformer": load_bettertransformer,
+    "executorch": load_executorch_pipeline,
 }
```
guangy10, line 318: Is this pipeline ever used? If yes, can you show me examples?

The example in the landing page is using the transformers pipeline directly: https://github.com/huggingface/optimum?tab=readme-ov-file#run-the-exported-model-using-onnx-runtime. I fail to understand how it is possible to run the exported model with transformers pipeline. If the exported model may have a modified signature that is different from eager, which won't work with the eager `generate` I think?

guangy10:
Fixed the export path. Now `optimum-cli export executorch --model "meta-llama/Llama-3.2-1B" --task "text-generation" --recipe "xnnpack" --output_dir="meta_llama3_2_1b"` can work as expected

optimum/executorchruntime/modeling_executorch.py
guangy10: NVM, figured out a way to integrate and unblock

michaelbenayoun (commented):
Left a few comments. I have to say it is really great work what you have done here since it is not easy to go from our main ONNX codebase and generalize!

Do you think we could also write little tests in the same spirit as what's done for ONNX, where we validate the exported model against a vanilla PyTorch model?

optimum/executorchruntime/modeling_executorch.py
```diff
@@ -0,0 +1,234 @@
...
+        self.dtype = self.et_model.run_method("get_dtype")[0]
+        self.bos_token_id = self.et_model.run_method("get_bos_id")[0]
+        self.eos_token_id = self.et_model.run_method("get_eos_id")[0]
+        self.vocab_size = self.et_model.run_method("get_vocab_size")[0]
+
+    def forward(self, input_ids: torch.Tensor, cache_position: torch.Tensor) -> torch.Tensor:
+        return self.et_model.forward((input_ids, cache_position))[0]
+
+    @classmethod
+    def from_pretrained(
```
echarlaix, line 68: here should be `_from_pretrained` which is used to load the model once exported 


```suggestion
    def _from_pretrained(
```

https://github.com/huggingface/optimum/blob/a7a807c9e712fd9669865358e34c1de072b78d8e/optimum/modeling_base.py#L436

optimum/executorchruntime/modeling_executorch.py
echarlaix: yes `_export` was added in https://github.com/huggingface/optimum/pull/1578 but  `_from_transformers` is still used for some reason https://github.com/huggingface/optimum/blob/a7a807c9e712fd9669865358e34c1de072b78d8e/optimum/modeling_base.py#L436 we need to change that at some point cc @mht-sharma

optimum/executorchruntime/modeling_executorch.py
```diff
@@ -0,0 +1,234 @@
...
+        cls,
+        model_dir_path: Union[str, Path],
+        task: str,
```
echarlaix, line 71: I think we can remove, it will always be `"text-generation"`in this case

```suggestion
```

optimum/executorchruntime/modeling_executorch.py
```diff
@@ -0,0 +1,234 @@
...
+        recipe: str,
```
echarlaix, line 72: can be removed as well I think as will only be used for export 

```suggestion
```

optimum/pipelines/pipelines_base.py
echarlaix: yes with the current integration executorch models will not be compatible with transformers pipelines, which is not a big deal as I don't think many people uses pipelines with optimum models

guangy10:
Thanks for the feedback @michaelbenayoun @echarlaix. I was on PTO last week right after the Hackathon. I will prioritize and continue to iterating this PR this week.

optimum/pipelines/pipelines_base.py
guangy10: > yes with the current integration executorch models will not be compatible with transformers pipelines, which is not a big deal as I don't think many people uses pipelines with optimum models

Okay, I think we can add pipeline support later in a separate PR. Would focus to add minimal backbone for ExecuTorch in this initial integration.

optimum/executorchruntime/modeling_executorch.py
guangy10: There will be other tasks like "fill-mask" or "semantic-segmentation", etc. Instead of having the smart of inferring the task, I will prefer start with asking users to specific the task explicitly.

optimum/executorchruntime/modeling_executorch.py
guangy10: There will be other recipes, users will need to specify which recipe to use in order to export to ExecuTorch when fetching the model from hub.

test_executorch.py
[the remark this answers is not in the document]
guangy10: Will reverted pipeline related changes as this file will be removed.

optimum/executorchruntime/modeling_executorch.py
guangy10: @echarlaix I understand that in the base class `_from_pretrained` is the virtual one that needs to be implemented and called in `from_pretrained`. However, I think currently the `from_pretrained` has lots of things that can't be generalized and applied to executorch yet, for example [these code snippet](https://github.com/guangy10/optimum/blob/35eebfe62bf721bbab365f569bd0c73057239732/optimum/modeling_base.py#L383-L434). We want to bypass these code (at least for the initial integration). And `_from_transformers` is going to be deprecated, so essentially I won't need anything in `from_pretrained ` and decide to override it directly in the derived class instead.

guangy10:
> Do you think we could also write little tests in the same spirit as what's done for ONNX, where we validate the exported model against a vanilla PyTorch model?

I should definitely add tests and make sure it runs in CI, for both export cli and runtime. However, note that the generated contents from exported model may not be identical as the eager pytorch model due to specialization during export. We can goal on testing the functionality first in this PR, i.g. the ability to generate using the exported model, then guard on correctness later though some lm_eval or something similar.

optimum/exporters/executorch/convert.py
```diff
@@ -0,0 +1,58 @@
...
+        output_dir (`Union[str, Path]`):
+            Path indicating the directory where to store the generated ExecuTorch model.
+    """
+    try:
+        recipe_func = recipe_registry.get(recipe)
+    except KeyError as e:
+        raise RuntimeError(f"The recipe '{recipe}' isn't registered. Detailed error: {e}")
+
+    executorch_prog = recipe_func(model, task, **kwargs)
+    # print(f"Exported program: {executorch_prog.exported_program().graph}")
```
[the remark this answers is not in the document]
guangy10: Addressed

optimum/exporters/executorch/convert.py
```diff
@@ -0,0 +1,58 @@
...
+
+    full_path = os.path.join(f"{output_dir}", "model.pte")
+    with open(full_path, "wb") as f:
+        executorch_prog.write_to_file(f)
+        print(f"Saved exported program to {full_path}")
```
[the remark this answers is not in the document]
guangy10: Addressed

guangy10:
@michaelbenayoun @echarlaix Could you guide me where I should add documentations for the new backend ExecuTorch?

optimum/executorchruntime/modeling_executorch.py
guangy10: Updated the code by overriding both `_from_pretrained` and `from_pretrained` to support load from local path and hub (with export).

guangy10:
Several new updates:
  - Support dynamic task & recipe discovery and import
  - Support load `from_pretrained` with `export=True` which will fetch `model_id` from hub and export to ExecuTorch to a temp dir and load the exported model back
  - Clean up and fix code API documentations
  - Add unit tests for `optimum-cli export executorch` and runtime using `ExecuTorchModelForCausalLM`.

optimum/commands/export/executorch.py
```diff
@@ -0,0 +1,53 @@
...
+        help=(
+            "The task to export the model for. If not specified, the task will be auto-inferred based on the model. Available tasks depend on the model, but are among:"
+            f" {str(TasksManager.get_all_tasks())}. For decoder models, use `xxx-with-past` to export the model using past key values in the decoder."
+        ),
+    )
+    optional_group.add_argument(
+        "--recipe",
+        type=str,
+        default="xnnpack",
+        help='Pre-defined recipes for export to ExecuTorch. Defaults to "xnnpack".',
```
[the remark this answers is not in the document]
guangy10: Now all recipes and tasks can be dynamically discovered and imported

tests/cli/test_cli.py
```diff
@@ -51,15 +52,29 @@ def test_helps_no_raise(self):
     def test_env_commands(self):
         subprocess.run("optimum-cli env", shell=True, check=True)
 
-    def test_export_commands(self):
```
guangy10, line 54: FYI, I have to split this because tflite is broken. Also it's not optimal to combine independent backends in a single test where failure in one would block testing other backends.

michaelbenayoun:
> @michaelbenayoun @echarlaix Could you guide me where I should add documentations for the new backend ExecuTorch?

You can add a subsection "ExecuTorch" under this section: https://huggingface.co/docs/optimum/exporters/overview.

To do that add the relevant files under `docs/source/exporters` and edit the `docs/source/_toctree.yml` file to integrate them in the menu.

tests/cli/test_cli.py
michaelbenayoun: You are totally right, we should have independent tests.

setup.py
```diff
@@ -88,6 +88,10 @@
         "datasets<=2.16",
         "transformers>=4.26,<4.38",
     ],
+    "exporters-executorch": [
+        "executorch>=0.4.0",
+        "transformers>=4.46",
+    ],
```
guangy10, line 94: @michaelbenayoun I looked into one CI failure related to ExecuTorch https://github.com/huggingface/optimum/actions/runs/12171925051/job/33984832982?pr=2090#step:5:50
```
ValueError: The minimum Transformers version compatible with ExecuTorch is 4.46.0. Please upgrade to Transformers 4.46.0 or later.
```
The failure is because the required version is not installed in the runner. Here is where I specify the dep packages and versions. Do I need to modify somewhere else to force the install the required package?

guangy10:
> > > Do you think we could also write little tests in the same spirit as what's done for ONNX, where we validate the exported model against a vanilla PyTorch model?
> > 
> > 
> > I should definitely add tests and make sure it runs in CI, for both export cli and runtime. However, note that the generated contents from exported model may not be identical as the eager pytorch model due to specialization during export. We can goal on testing the functionality first in this PR, i.g. the ability to generate using the exported model, then guard on correctness later though some lm_eval or something similar.
> > Tests are added. `pytest tests/cli/test_cli.py -k test_export_commands -v`
> > ```
> > tests/cli/test_cli.py::TestCLI::test_export_commands PASSED                                                            [100%]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `pytest tests/executorch/test_modeling.py -k test_text_generation_with_xnnpack -v`
> > ```
> > tests/executorch/test_modeling.py::ExecuTorchModelIntegrationTest::test_text_generation_with_xnnpack PASSED                          [100%]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Don't see any CI is running. @michaelbenayoun @echarlaix How can I make sure these tests are running in CI?
> 
> The CI must be triggered by a HF member, that is why it is not running. I just triggered it.

@michaelbenayoun @echarlaix The tests can run well locally so we are in the stage to tuning the tests in the CI. It won't be efficient if every single update would require manually triggering the CI by a HF member.  Is there a way to trigger the CI automatically (at least on this PR) through something like tagging, or any other way to ensure the test can be triggered automatically when the PR gets updated?

guangy10:
@michaelbenayoun @echarlaix I noticed the python version in the CI is 3.9. ExecuTorch is requiring minimal of 3.10 however. Not saying things will not work with python 3.9, but there is a potential to run into unnecessary issues. I'm wondering if I can I should create a new workflow for ExecuTorch that is configured to run with python 3.10 or higher? Happy to hear you suggestions as well

guangy10:
Added two new CI workflows for ExecuTorch AOT (export) and Runtime (inference) respectively, with the expected python versions.

@michaelbenayoun could you help re-run the CI?

tests/cli/test_cli.py
guangy10: Ok, given the required python versions are different. I decided to create new workflows for ExecuTorch with exact required setup, to avoid unintentionally bump the python version for other backends.

guangy10:
- Fixed the loading from local path usign `from_pretrained()`
- Added documentation for the new ExecuTorch backend

michaelbenayoun:
> @michaelbenayoun @echarlaix I noticed the python version in the CI is 3.9. ExecuTorch is requiring minimal of 3.10 however. Not saying things will not work with python 3.9, but there is a potential to run into unnecessary issues. I'm wondering if I can I should create a new workflow for ExecuTorch that is configured to run with python 3.10 or higher? Happy to hear you suggestions as well

For me it is perfectly okay to create a new workflow to run on Python 3.10. Maybe let's raise an error when the Python version is below that in the exporting function, wdyt?

guangy10:
- Fixed the print version issue in newly added ExecuTorch jobs
- Fixed import and `[[autodoc]]` doc issue in `configuration.mdx`

The Code Quality failure is irrelevant. It's raising from  `would reformat /home/runner/work/optimum/optimum/optimum/onnxruntime/runs/__init__.py` which is not touched in this PR. Run `make style` locally won't generate any new changes.

@michaelbenayoun @echarlaix Would you mind re-running the CI on my behalf?

docs/source/exporters/executorch/package_reference/configuration.mdx
```diff
@@ -0,0 +1,54 @@
...
+- Specify whether and how to **quantize** the model.
+- Delegate computation to various accelerators, such as **CPU**, **GPU**, **NPU**, **DSP**, and others.
+- Define **custom transformation passes**.
+- Implement advanced techniques like memory planning algorithms to optimize resource utilization.
+
+[[autodoc]] exporters.executorch.recipe_registry.discover_recipes
+
+[[autodoc]] exporters.executorch.recipe_registry.register_recipe
+
+[[autodoc]] exporters.executorch.recipes.xnnpack.export_to_executorch_with_xnnpack
```
guangy10, line 51: @michaelbenayoun The doc build still complains about this line:
```
ValueError: There was an error when converting /optimum/docs/source/exporters/executorch/package_reference/configuration.mdx to the MDX format.
Unable to find exporters.executorch.recipes.xnnpack.export_to_executorch_with_xnnpack in optimum. Make sure the path to that object is correct.
```
I verified that the package is valid. Any pointer what I missed?

michaelbenayoun: I also observed the same thing, working on fixing this.

guangy10:
Thanks @michaelbenayoun for making the tests runnable on the CI. For the "ExecuTorch Runtime" tests, it seems like the failure is due to OOM. Per the chart, M1 runner only has 7GB RAM, inference with 1B fp32 mdoel should be fine, but won't work for 2B+ models. 

We can either switch to use a M1 with larger RAM (16GB would be sufficient for all models in this PR), or we can add a TODO and disable runtime tests temporarily for 2B+ models and re-enable it when we get runners with larger RAMs.

Let me know your thoughts

tests/executorch/runtime/test_modeling.py
```diff
@@ -0,0 +1,212 @@
...
+        generated_text = model.text_generation(
+            tokenizer=tokenizer,
+            prompt="Simply put, the theory of relativity states that",
+            max_seq_len=len(tokenizer.encode(EXPECTED_GENERATED_TEXT)),
+        )
+        self.assertEqual(generated_text, EXPECTED_GENERATED_TEXT)
+
+    @slow
+    @pytest.mark.run_slow
+    def test_llama3_2_3b_text_generation_with_xnnpack(self):
```
guangy10, line 98: @michaelbenayoun If there is no M1 runner with larger RAM, we can skip this test for 3B model

tests/executorch/runtime/test_modeling.py
```diff
@@ -0,0 +1,212 @@
...
+        generated_text = model.text_generation(
+            tokenizer=tokenizer,
+            prompt="My favourite condiment is ",
+            max_seq_len=len(tokenizer.encode(EXPECTED_GENERATED_TEXT)),
+        )
+        self.assertEqual(generated_text, EXPECTED_GENERATED_TEXT)
+
+    @slow
+    @pytest.mark.run_slow
+    def test_gemma2_text_generation_with_xnnpack(self):
```
guangy10, line 146: @michaelbenayoun If there is no M1 runner with larger RAM, we can skip this test for 2B model

tests/executorch/runtime/test_modeling.py
```diff
@@ -0,0 +1,212 @@
...
+        generated_text = model.text_generation(
+            tokenizer=tokenizer,
+            prompt="Hello I am doing a project for my school",
+            max_seq_len=len(tokenizer.encode(EXPECTED_GENERATED_TEXT)),
+        )
+        self.assertEqual(generated_text, EXPECTED_GENERATED_TEXT)
+
+    @slow
+    @pytest.mark.run_slow
+    def test_gemma_text_generation_with_xnnpack(self):
```
guangy10, line 169: @michaelbenayoun If there is no M1 runner with larger RAM, we can skip this test for 2B model

guangy10: No need to disable 2B if 15GB RAM

guangy10:
- Updated expected texts for llama3.2 3b, gemma, gemma2
- Fixed linter error

guangy10:
I see `executorch/runtime/test_modeling.py::ExecuTorchModelIntegrationTest::test_gemma2_text_generation_with_xnnpack PASSED` but 
```
/Users/runner/work/_temp/e6f2ff22-a7d4-4ea4-b0a1-8b1d5a386502.sh: line 1:  4688 Killed: 9               RUN_SLOW=1 pytest executorch/runtime/test_*.py -s -vvvv --durations=0
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
executorch/runtime/test_modeling.py::ExecuTorchModelIntegrationTest::test_gemma_text_generation_with_xnnpack
```
In ExecuTorch Runtime jobs.

Trying to figure out if it's because previous tests is taking pretty much all SSD on the runner. I'm splitting the tests for each model into a separate test file, so each job will only run test for a single model, which is more scalable as we adding more recipes, optimizations, and tests for a model. cc: @michaelbenayoun

docs/source/exporters/executorch/package_reference/configuration.mdx
guangy10: This is still failing, but I guess we can ignore it and merge

guangy10:
Weird doc-build error but doesn't look relevant to this PR:
```
69.99 ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device
69.99 
------
Dockerfile:11
--------------------
   9 |     
  10 |     RUN git clone $clone_url && cd optimum && git checkout $commit_sha
  11 | >>> RUN python3 -m pip install --no-cache-dir ./optimum[onnxruntime,benchmark,quality,exporters-tf,doc-build,diffusers]
  12 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c python3 -m pip install --no-cache-dir ./optimum[onnxruntime,benchmark,quality,exporters-tf,doc-build,diffusers]" did not complete successfully: exit code: 1
```
@michaelbenayoun safe to merge?

.github/workflows/test_executorch_export.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies for ExecuTorch
+        run: |
+          pip install .[tests,exporters-executorch]
```
guangy10, line 30: cc: @michaelbenayoun Compatible python versions and package installation.

echarlaix:
Thanks for the great PR @guangy10 ! We discussed about having everything executorch related in a dedicated `optimum-executorch` repo, is this still happening @michaelbenayoun ?

michaelbenayoun:
I guess it can be done.
@guangy10 you want to do it, or I do it when I have a the bandwidth?

It requires:

- Move all the code from this PR in a new repo
- Register the export commands so that the Optimum CLI can find it when `optimum-executorch` is installed.
- Update the `setup.py`
- Update the way the doc is being built

PR #2101, opened 2024-11-21
Add ONNX export support for `PatchTST`

xenova:
# What does this PR do?

This PR adds ONNX export support for PatchTST models, as well as fixes other models that contain the `torch.unfold` operator, which is not supported by `torch.onnx.export`. This is done by applying a set of default patching specs.

Fixes https://github.com/huggingface/transformers.js/issues/1047

## Before submitting

## Who can review?
@michaelbenayoun, @echarlaix


optimum/exporters/onnx/model_patcher.py
```diff
@@ -113,6 +113,20 @@ class PatchingSpec:
...
 
+# An ONNX-export-compatible version of `tensor.unfold`. Without this, we get:
+# torch.onnx.errors.SymbolicValueError: Unsupported: ONNX export of operator Unfold, input size not accessible.
+# See https://github.com/pytorch/pytorch/issues/81871 for more information
+def onnx_compatible_unfold(self, dimension, size, step):
+    num_patches = (self.size(dimension) - size) // step + 1
+    return torch.stack(
+        [self[:, i : i + size, :] for i in range(0, num_patches * step, step)],
+        dim=1,
+    ).transpose(3, 2)
```
xenova, line 124: I ran into some issues with my naive attempt (only seems to work for 4D. Here's a better version I found:

```suggestion
def onnx_compatible_unfold(input_tensor, dimension, size, step):
    """
    Custom implementation of torch.unfold without using torch.unfold.
    
    Args:
        input_tensor (torch.Tensor): The input tensor.
        dimension (int): The dimension to unfold.
        size (int): The size of each slice.
        step (int): The step size between slices.

    Returns:
        torch.Tensor: The unfolded tensor.
    """
    # Compute the shape of the unfolded output
    input_size = input_tensor.size(dimension)
    num_slices = (input_size - size) // step + 1

    # Permute dimension to the end for easier indexing
    input_tensor = input_tensor.transpose(dimension, -1)

    # Extract slices
    slices = []
    for i in range(num_slices):
        start = i * step
        end = start + size
        slices.append(input_tensor[..., start:end])

    # Stack slices and permute dimensions back
    result = torch.stack(slices, dim=-2).transpose(dimension, -2)
    return result
```

michaelbenayoun (approved):
LGTM!

Let's wait for @echarlaix 's review before merging.

tests/exporters/exporters_utils.py
```diff
@@ -255,6 +257,8 @@
     "nystromformer": "hf-internal-testing/tiny-random-NystromformerModel",
     "owlv2": "google/owlv2-base-patch16",
     "owlvit": "google/owlvit-base-patch32",
+    "patchtst": "ibm/test-patchtst",
+    "patchtsmixer": "ibm/test-patchtsmixer",
```
echarlaix, line 261: already included in `PYTORCH_EXPORT_MODELS_TINY` so can be removed 
```suggestion
```

optimum/exporters/tasks.py
```diff
@@ -314,6 +314,10 @@ class TasksManager:
     }
 
     _CUSTOM_CLASSES = {
+        ("pt", "patchtsmixer", "feature-extraction"): ("transformers", "PatchTSMixerModel"),
+        ("pt", "patchtsmixer", "time-series-forecasting"): ("transformers", "PatchTSMixerForPrediction"),
+        ("pt", "patchtst", "feature-extraction"): ("transformers", "PatchTSTModel"),
+        ("pt", "patchtst", "time-series-forecasting"): ("transformers", "PatchTSTForPrediction"),
```
echarlaix, line 320: shouldn't `AutoModel` load the models as expected https://huggingface.co/datasets/huggingface/transformers-metadata/blob/main/pipeline_tags.json#L666-L667 ?
```suggestion
        ("pt", "patchtsmixer", "time-series-forecasting"): ("transformers", "PatchTSMixerForPrediction"),
        ("pt", "patchtst", "time-series-forecasting"): ("transformers", "PatchTSTForPrediction"),
```

optimum/exporters/onnx/base.py
```diff
@@ -177,6 +177,7 @@ class OnnxConfig(ExportConfig, ABC):
         "text2text-generation": OrderedDict({"logits": {0: "batch_size", 1: "decoder_sequence_length"}}),
         "text-classification": OrderedDict({"logits": {0: "batch_size"}}),
         "text-generation": OrderedDict({"logits": {0: "batch_size", 1: "sequence_length"}}),
+        "time-series-forecasting": OrderedDict({"prediction_outputs": {0: "batch_size"}}),
```
echarlaix, line 180: I can't find "time-series-forecasting" in https://huggingface.co/datasets/huggingface/transformers-metadata/blob/main/pipeline_tags.json, do you know if this specific to PatchTST models ? (-> `PatchTSTXxxForPrediction`)

optimum/exporters/onnx/model_configs.py
```diff
@@ -2445,3 +2446,51 @@ class EncoderDecoderOnnxConfig(EncoderDecoderBaseOnnxConfig):
     NORMALIZED_CONFIG_CLASS = NormalizedEncoderDecoderConfig
 
     DEFAULT_ONNX_OPSET = 14  # uses SDPA in Transformers, hence opset>=14.
+
+
+class PatchTSTDummyInputGenerator(DummyInputGenerator):
```
echarlaix, line 2451: would move to https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/model_patcher.py

echarlaix (approved):
Looks great, thank you so much @xenova for all the great work!

optimum/exporters/tasks.py
xenova: You're right! Fixed.

optimum/exporters/onnx/base.py
xenova: Indeed, this is not a transformers pipeline id, but it is a pipeline tag on the hub: https://huggingface.co/models?pipeline_tag=time-series-forecasting

(Not specific to patchtst)

optimum/exporters/onnx/model_configs.py
xenova: Moved to input_generators.py! (assuming that's what you meant)

xenova:
- Applied review suggestions ✅ 
- Fixed onnx compatible unfold for negative indices ✅ 

Ready to merge, I'd say! wdyt @echarlaix?

echarlaix:
Yes sounds good, thanks @xenova !

PR #2105, opened 2024-11-27
Pass the revision to SentenceTransformer models

bndos:
# What does this PR do?

Ensures that the `revision` parameter is passed to `SentenceTransformer` models in `get_model_from_task`.


echarlaix (approved):
Thanks for the fix @bndos !

PR #2107, opened 2024-11-28
Unbundle inputs generated by `DummyTimestepInputGenerator`

JingyaHuang:
# What does this PR do?

As per title, when using `DummyTimestepInputGenerator`, we don't necessarily need to generate all inputs managed by this generator. E.g. Pixart alpha, no "text_embeds" / "timestep_cond," no attribute `text_encoder_projection_dim`/`time_cond_proj_dim` then, but they are currently mandatory, making it diffisult to use the generator just to generate `timestep` / `time_ids.`


echarlaix (approved):
LGTM!

PR #2109, opened 2024-12-01
fix bug `ModelPatcher` returns empty outputs

LoSealL:
When model's output is tuple or list, `filtered_outputs` doesn't get assigned and hence always a empty dict

# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/exporters/onnx/model_patcher.py
+++ b/optimum/exporters/onnx/model_patcher.py
@@ -168,7 +168,7 @@ class ModelPatcher:
                         filterd_outputs[name] = value
             elif isinstance(outputs, (list, tuple)):
                 outputs_list = list(config.outputs.keys())
-                filtered_outputs = dict(zip(outputs_list, outputs))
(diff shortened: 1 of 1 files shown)


IlyasMoutawwakil:
there's also a typo in the lines that come after (and the function's return) where `filtered_outputs` is written as `filterd_outputs`

optimum/exporters/onnx/model_patcher.py
```diff
@@ -168,7 +168,7 @@ def patched_forward(*args, **kwargs):
                         filterd_outputs[name] = value
             elif isinstance(outputs, (list, tuple)):
                 outputs_list = list(config.outputs.keys())
-                dict(zip(outputs_list, outputs))
+                filtered_outputs = dict(zip(outputs_list, outputs))
```
echarlaix, line 171:
```suggestion
                filterd_outputs = dict(zip(outputs_list, outputs))
```

echarlaix (approved):
Thanks for the fix @LoSealL !

PR #2110, opened 2024-12-03
Fix workflow to mark issues as stale

echarlaix:
https://github.com/huggingface/optimum/actions/runs/12131603639

```
Error: [#2083] Error when creating a comment: Resource not accessible by integration
Error: [#2083] Error when adding a label: Resource not accessible by integration
```

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -6,9 +6,12 @@ on:
 jobs:
   stale:
     runs-on: ubuntu-latest
+    permissions:
+      issues: write
     steps:
-      - uses: actions/stale@v8
+      - uses: actions/stale@v9
         with:
           stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -6,9 +6,12 @@ on:
...
     runs-on: ubuntu-latest
+    permissions:
+      issues: write
     steps:
-      - uses: actions/stale@v8
+      - uses: actions/stale@v9
         with:
-          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          stale-issue-message: 'This issue has been marked as stale because it has been open for 30 days with no activity. This thread will be automatically closed in 5 days if no further activity occurs.'
+          exempt-issue-labels: 'bug,exporters,good first issue,onnx,onnxruntime,quantization'
```
echarlaix, line 15: also added  `exempt-issue-labels` so that issues with the following labels will not be marked as stale for now, can be changed in the future

regisss (approved):
LGTM!

PR #2117, opened 2024-12-05
Allow sending directly torch/tf/np dtype for generating dummy inputs

JingyaHuang:
# What does this PR do?

Currently always via git push `DTYPE_MAPPER`:

https://github.com/huggingface/optimum/blob/4a7cb298140ee9bed968d98a780a950d15bb2935/optimum/utils/input_generators.py#L73

It would be nice to directly accept the different dtypes of frameworks. Eg. I could load a model in `torch.bfloat16` in this case, I would like to define the dummy inputs dtype with `model.dtype` instead of going through a reversed dict of the `DTYPE_MAPPER`.

### DIFF
--- a/optimum/utils/input_generators.py
+++ b/optimum/utils/input_generators.py
@@ -672,7 +672,7 @@ class DummySeq2SeqDecoderTextInputGenerator(DummyDecoderTextInputGenerator):
                 None,
             )
 
-        return super().generate(input_name, framework=framework, int_dtype=int_dtype)
+        return super().generate(input_name, int_dtype=int_dtype, float_dtype=float_dtype, framework=framework)
 
 
 class DummyPastKeyValuesGenerator(DummyInputGenerator):
@@ -1670,7 +1670,7 @@ class DummyTransformerTimestepInputGenerator(DummyTimestepInputGenerator):
             shape = [self.batch_size]  # With transformer diffusers, timestep is a 1D tensor
             return self.random_float_tensor(shape, max_value=self.vocab_size, framework=framework, dtype=float_dtype)
 
-        return super().generate(input_name, framework, int_dtype, float_dtype)
+        return super().generate(input_name, int_dtype=int_dtype, float_dtype=float_dtype, framework=framework)
 
 
 class DummyTransformerVisionInputGenerator(DummyVisionInputGenerator):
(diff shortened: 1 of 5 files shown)


optimum/utils/input_generators.py
```diff
@@ -132,7 +133,13 @@ def supports_input(self, input_name: str) -> bool:
         return any(input_name.startswith(supported_input_name) for supported_input_name in self.SUPPORTED_INPUT_NAMES)
 
     @abstractmethod
-    def generate(self, input_name: str, framework: str = "pt", int_dtype: str = "int64", float_dtype: str = "fp32"):
+    def generate(
+        self,
+        input_name: str,
+        framework: str = "pt",
```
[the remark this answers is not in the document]
JingyaHuang: Yeah, make sense, I actually in this PR prioritize the framework indicated by the dtype to the given framework, kinda worried that these are needed for subpkg for specific reasons... 
Let me put on a deprecation for it.

optimum/utils/input_generators.py
```diff
@@ -306,6 +338,40 @@ def _infer_framework_from_input(input_) -> str:
...
+        else:
+            return np.int64
+
+    @staticmethod
+    def infer_framework_from_dtype(dtype):
+        if is_torch_available() and isinstance(dtype, torch.dtype):
+            framework = "pt"
+        elif is_tf_available() and isinstance(dtype, tf.dtypes.DType):
+            framework = "tf"
+        elif isinstance(dtype, type):
```
IlyasMoutawwakil, line 367: why not np.dtype ?

optimum/utils/input_generators.py
```diff
@@ -306,6 +338,40 @@ def _infer_framework_from_input(input_) -> str:
...
+
+    @staticmethod
+    def _set_default_float_dtype():
+        "Default to float32 of available framework."
+        if is_torch_available():
+            return torch.int64
+        elif is_tf_available():
+            return tf.int64
+        else:
+            return np.int64
```
IlyasMoutawwakil, line 359: should be floats I guess

IlyasMoutawwakil (commented):
Thanks ! This looks better and more flexible ! I left a couple of comments

IlyasMoutawwakil:
tests are failing, seems like some string inputs are still being passed around
```
dtype = 'int64'

    @staticmethod
    def infer_framework_from_dtype(dtype):
        if is_torch_available() and isinstance(dtype, torch.dtype):
            framework = "pt"
        elif is_tf_available() and isinstance(dtype, tf.dtypes.DType):
            framework = "tf"
        elif isinstance(dtype, type):
            framework = "np"
        else:
>           raise ValueError(
                f"Unable to create a tensor/array with dtype({dtype}). Please ensure the corresponding framework is installed."
            )
E           ValueError: Unable to create a tensor/array with dtype(int64). Please ensure the corresponding framework is installed.
```

optimum/utils/input_generators.py
JingyaHuang: good catch!

optimum/utils/input_generators.py
JingyaHuang:
```
>>> import numpy as np
>>> isinstance(np.float32, np.dtype)
False
>>> isinstance(np.float32, type)
True
```

PR #2119, opened 2024-12-06
Enable transformers v4.47 support

echarlaix:


tests/onnx/test_onnx_export_custom_module.py
```diff
@@ -22,7 +22,7 @@
 
 if is_torch_available():
     import torch
-    from transformers.models.deberta import modeling_deberta
+    from transformers.models.sew_d import modeling_sew_d
```
echarlaix, line 25: `StableDropout` was removed from modeling_deberta in https://github.com/huggingface/transformers/pull/22105

echarlaix:
Tests failing coming from mps not supported by arm64 macOS runners https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#limitations-for-arm64-macos-runners and from mps being now set default when device not specified and mps backend available since transformers v4.47 and [huggingface/transformers#34026](https://github.com/huggingface/transformers/pull/34026/files). Issue not related to optimum so merging to not block transformers v4.47 support

PR #2122, opened 2024-12-10
add ipex to quicktour

jiqing-feng:
Hi @echarlaix @IlyasMoutawwakil . As ipex reactor is done in optimum-intel, it's time to update the readme and add ipex to quicktour.

### DIFF
--- a/docs/source/quicktour.mdx
+++ b/docs/source/quicktour.mdx
@@ -40,6 +40,26 @@ If you want to load a PyTorch checkpoint, set `export=True` to convert your mode
 You can find more examples in the [documentation](https://huggingface.co/docs/optimum/intel/inference) and in the [examples](https://github.com/huggingface/optimum-intel/tree/main/examples/openvino).
 
 
+#### IPEX
+To load a model and run inference with IPEX optimization, you can just replace your `AutoModelForXxx` class with the corresponding `IPEXModelForXxx` class.
+
+```diff
+- from transformers import AutoModelForSequenceClassification
++ from optimum.intel import IPEXModelForSequenceClassification
+  from transformers import AutoTokenizer, pipeline
+
+  # Download a tokenizer and model from the Hub and convert to IPEX format
+  model_id = "distilbert-base-uncased-finetuned-sst-2-english"
+  tokenizer = AutoTokenizer.from_pretrained(model_id)
+- model = AutoModelForSequenceClassification.from_pretrained(model_id)
++ model = IPEXModelForSequenceClassification.from_pretrained(model_id)
+
+  # Run inference!
+  classifier = pipeline("text-classification", model=model, tokenizer=tokenizer)
+  results = classifier("He's a dreadful magician.")
+```
+
+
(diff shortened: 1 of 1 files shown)


echarlaix (commented):
Thanks @jiqing-feng but the quicktour looks a bit crowded already, I don't think we want to add more code snippet for now : would make more sense to have a description explaning when to use ipex vs openvino vs neural-compressor (+ a link towards a benchmark could be very nice also) as it's currently a bit confusing for users what should be used when + what are the integrations for each. I think it also makes sense to keep all code snippet related to optimum-intel in the dedicated section to not duplicate them multiple times (same applied for others sections) we can just add links in the quickstour instead

jiqing-feng:
> Thanks @jiqing-feng but the quicktour looks a bit crowded already, I don't think we want to add more code snippet for now : would make more sense to have a description explaning when to use ipex vs openvino vs neural-compressor (+ a link towards a benchmark could be very nice also) as it's currently a bit confusing for users what should be used when + what are the integrations for each. I think it also makes sense to keep all code snippet related to optimum-intel in the dedicated section to not duplicate them multiple times (same applied for others sections) we can just add links in the quickstour instead

Agree, do you think we should use optimum-intel pipeline instead? In this case, users can pass `accelerator=ipex;ov;neural-compressor` to choose which kind of backend they want to use. The descriptions should be:
```
To apply optimum-intel optimization, you just need to import pipeline from optimum-intel instead of transformers, and choose **accelerator** from ov(link), ipex(link), neural-compressor(link) to get different acceleration. Following codes:
```
```diff
- from transformers import pipeline
+ from optimum.intel.pipelines import pipeline

model_id = "distilbert-base-uncased-finetuned-sst-2-english"

  # Run inference!
- classifier = pipeline("text-classification", model=model_id)
+ classifier = pipeline("text-classification", model=model_id, accelerator=accelerator)
results = classifier("He's a dreadful magician.")
```

yao-matrix:
> Thanks @jiqing-feng but the quicktour looks a bit crowded already, I don't think we want to add more code snippet for now : would make more sense to have a description explaning when to use ipex vs openvino vs neural-compressor (+ a link towards a benchmark could be very nice also) as it's currently a bit confusing for users what should be used when + what are the integrations for each. I think it also makes sense to keep all code snippet related to optimum-intel in the dedicated section to not duplicate them multiple times (same applied for others sections) we can just add links in the quickstour instead

@echarlaix , we're fine w/ the "add link" idea. 
For positioning, do you think it helps to position `ipex + inc` as intel's HF HW-accelerated PyTorch-native runtime, no need for proprietary format/conversion/... ?  For `inc`, we position it as intel's quantization lib(like autoawq etc. in ecosystem), but have intel enhancement algorithms like autoround, and we are planning to deprecate its runtime role and move runtime role back to `ipex` backend. So, IPEXModel can run both models quantized by ecosystem libraries(autoawq, gptqmodel etc.) and models quantized by `inc`. 
It needs time, but it's what we will do in 2025.

PR #2123, opened 2024-12-10
Fix ORT Trainer

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -804,7 +804,9 @@ class ORTTrainer(Trainer):
                     self.state.epoch = epoch + (step + 1 + steps_skipped) / steps_in_epoch
                     self.control = self.callback_handler.on_step_end(args, self.state, self.control)
 
-                    self._maybe_log_save_evaluate(tr_loss, grad_norm, model, trial, epoch, ignore_keys_for_eval)
+                    self._maybe_log_save_evaluate(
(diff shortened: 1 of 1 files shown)


IlyasMoutawwakil:
tests are passing in https://github.com/huggingface/optimum-benchmark/actions/runs/12257503594/job/34195790218?pr=306

optimum/onnxruntime/trainer.py
```diff
@@ -818,7 +821,7 @@ def get_dataloader_sampler(dataloader):
                 self.control.should_training_stop = True
 
             self.control = self.callback_handler.on_epoch_end(args, self.state, self.control)
-            self._maybe_log_save_evaluate(tr_loss, grad_norm, model, trial, epoch, ignore_keys_for_eval)
+            self._maybe_log_save_evaluate(tr_loss, grad_norm, model, trial, epoch, ignore_keys_for_eval, start_time)
```
echarlaix, line 824: `start_time` was introduced in transformers v4.47 https://github.com/huggingface/transformers/blob/v4.47.0/src/transformers/trainer.py#L3021 so maybe we should add a check for the transformers version (min / max supported version) + tests to make sure we keep compatibility with the min/max versions, wdyt @IlyasMoutawwakil ?

echarlaix (commented):
Thanks @IlyasMoutawwakil ! Not sure if we should deprecate the `ORTTrainer` at some point as usage seems low, also cc @JingyaHuang to have her opinion on this

PR #2130, opened 2024-12-19
Skip private model loading for external contributors

echarlaix:
https://github.com/huggingface/optimum/actions/runs/12406440330/job/34639615964?pr=2114

```
FAILED tests/onnxruntime/test_modeling.py::ORTModelIntegrationTest::test_load_model_from_hub_private - huggingface_hub.errors.RepositoryNotFoundError: 401 Client Error. (Request ID: Root=1-6763cff5-38ad3a1d75a3598d6b742f30;13c34049-a28a-43d1-a22b-e79829428ccf)

Repository Not Found for url: https://huggingface.co/api/models/optimum-internal-testing/tiny-random-phi-private/tree/onnx?recursive=True&expand=False.
Please make sure you specified the correct `repo_id` and `repo_type`.
If you are trying to access a private or gated repo, make sure you are authenticated.
Invalid username or password.
```

### DIFF
--- a/tests/onnxruntime/test_modeling.py
+++ b/tests/onnxruntime/test_modeling.py
@@ -974,7 +974,7 @@ class ORTModelIntegrationTest(unittest.TestCase):
     def test_load_model_from_hub_private(self):
         token = os.environ.get("HF_HUB_READ_TOKEN", None)
 
-        if token is None:
+        if not token:
             self.skipTest(
                 "Test requires a read access token for optimum-internal-testing in the environment variable `HF_HUB_READ_TOKEN`."
             )


tests/onnxruntime/test_modeling.py
```diff
@@ -974,7 +974,7 @@ def test_stable_diffusion_model_on_rocm_ep_str(self):
     def test_load_model_from_hub_private(self):
         token = os.environ.get("HF_HUB_READ_TOKEN", None)
 
-        if token is None:
+        if not token:
```
IlyasMoutawwakil, line 977: There's a `require_hf_token` decorator in https://github.com/huggingface/optimum/blob/main/optimum/utils/testing_utils.py#L88
it's used in some places but it checks for `HF_AUTH_TOKEN` instead of `HF_HUB_READ_TOKEN` 🤔 should we use that decorator instead ?

echarlaix: Here the issue seems that the token is set as an empty string instead of being set to `None` (so the same issue will happen for `require_hf_token`). Let's see if this fixes it, in which case I'll update `require_hf_token`

PR #2131, opened 2024-12-19
Add ONNX export support for ModernBERT

xenova:
# What does this PR do?

This PR adds ONNX export support for ModernBERT, added to transformers in https://github.com/huggingface/transformers/pull/35158. It uses `torch.compile` decorators for certain functions, which is not supported in `torch.onnx.export` (with dynamo=False), so I introduced a model loader context manager to disable this when the model is being loaded. Note that this cannot happen in a model patcher since by then, the model class has already been created.

## Before submitting


xenova:
Now works with sentence-transformers & transformers models 👍

PR #2133, opened 2024-12-20
fix sdxl refiner export

eaidova:
# What does this PR do?

after some refactoring export part, sdxl-refiner model become broken for exporting.
https://huggingface.co/stabilityai/stable-diffusion-xl-refiner-1.0
This model does not have text_encoder and only text_encoder_2, now export failed with AtributeError: 'NoneType' object has no attribute 'config'

see for reference https://github.com/huggingface/optimum-intel/actions/runs/12426942003/job/34696091118?pr=1085


eaidova:
@echarlaix  @IlyasMoutawwakil could you please take a look?

echarlaix (approved):
Thanks a lot for the fix @eaidova !

echarlaix:
It should have been caught by the CI, we need to add a test for this (looks like `optimum-intel` is using https://huggingface.co/echarlaix/tiny-random-stable-diffusion-xl-refiner which is not the case of `optimum`). @IlyasMoutawwakil let me know if you have some time for this otherwise I'll take care of it

IlyasMoutawwakil:
Yes the refiner architecture totally went over my head, thank for the fix @eaidova. Feel free to add a test in this PR or I'll add it in another one.

IlyasMoutawwakil:
Merging for the transformers 4.47 support

PR #2134, opened 2024-12-21
fix auto_gptq layer error device

ZX-ModelCloud:
Fix the device error: `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0!` mentioned by [MekkCyber commented](https://github.com/huggingface/transformers/pull/35012#issuecomment-2556930206).

[test_quantization](https://github.com/huggingface/optimum/blob/main/tests/gptq/test_quantization.py) 
`RUN_SLOW=1 pytest tests/gptq/test_quantization.py`
* [x]  cpu tests
* [x]  cuda tests

### DIFF
--- a/optimum/gptq/quantizer.py
+++ b/optimum/gptq/quantizer.py
@@ -519,8 +519,6 @@ class GPTQQuantizer(object):
         blocks = recurse_getattr(model, self.block_name_to_quantize)
 
         cur_layer_device = get_device(blocks[0])
-        if not is_gptqmodel_available():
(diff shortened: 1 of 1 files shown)


Qubitium:
@ZX-ModelCloud Move this PR to draft. May not be needed. This is actually related to deficiency in autogptq unable to pass cpu tests as optimum force move layer to gpu. 

Gptqmodel has no such restrictions. We may bypass this by disabling cpu only tests for AutoGPTQ.

PR #2135, opened 2024-12-22
TimesFM

kashif:
# What does this PR do?

Add support for TimesFM model


optimum/exporters/onnx/model_configs.py
```diff
@@ -2499,3 +2500,46 @@ class EncoderDecoderOnnxConfig(EncoderDecoderBaseOnnxConfig):
...
+    DUMMY_INPUT_GENERATOR_CLASSES = (TimesFMDummyInputGenerator,)
+
+
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        return {"inputs": {0: "batch_size", 1: "sequence_length"}}
+
+    @property
+    def outputs(self) -> Dict[str, Dict[int, str]]:
+        return super().outputs
```
xenova, line 2545: I think the error you are facing is because you haven't specified the names + shapes for the expected outputs:
image

i.e., `mean_predictions` and `full_predictions` (you can skip `loss`)

echarlaix: added in https://github.com/huggingface/optimum/pull/2135/commits/5c2fee550b4c3d7826fc25b9c53173342c80f66f, thanks @xenova !

optimum/exporters/onnx/model_configs.py
```diff
@@ -2629,6 +2629,51 @@ class EncoderDecoderOnnxConfig(EncoderDecoderBaseOnnxConfig):
     DEFAULT_ONNX_OPSET = 14  # uses SDPA in Transformers, hence opset>=14.
 
 
+class TimesFMDummyInputGenerator(DummyInputGenerator):
```
echarlaix, line 2632: could make sense to re-use `DummyPatchTSTInputGenerator` instead (with some extra modifications), let's wait for https://github.com/huggingface/transformers/pull/34082/ to be merged

kashif: @echarlaix one question i had about the https://github.com/huggingface/transformers/pull/34082  in the timesfm_config i do not think my `TimesFmOnnxConfig` is correct... do you agree?

echarlaix: Yes it shouldn't inherit from `OnnxSeq2SeqConfigWithPast` (which should be used for the export of encoder-decoder models). I think you can directly remove `TimesFmOnnxConfig` in https://github.com/huggingface/transformers/pull/34082/ as it shouldn't be used anywhere from what I know (also I'm expecting anything onnx related to be deprecated in v5)

kashif: ok let me remove it... it would be good to then get it working via transformers.js... let me remove it first and then get back to you!

echarlaix: sure thanks, also cc @michaelbenayoun @JingyaHuang @IlyasMoutawwakil who can take a look while I'm ooo

kashif:
thanks!

optimum/exporters/onnx/model_configs.py
```diff
@@ -2641,6 +2642,25 @@ class EncoderDecoderOnnxConfig(EncoderDecoderBaseOnnxConfig):
...
 
+class TimesFMOnnxConfig(OnnxConfig):
+    NORMALIZED_CONFIG_CLASS = NormalizedTimeSeriesForecastingConfig
+    MIN_TRANSFORMERS_VERSION = version.parse("4.52.0")
+    DUMMY_INPUT_GENERATOR_CLASSES = (TimesFMDummyInputGenerator,)
+    DEFAULT_ONNX_OPSET = 14  # uses SDPA in Transformers, needs opset>=14
+
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        return {"past_values": {0: "batch_size", 1: "sequence_length"}}
```
echarlaix, line 2653: an issue is remaining before we can merge : it seems that the resulting onnx graph doesn't considered the `batch_size` as dynamic (won't accept one different than the one given during export) likely coming from :

https://github.com/huggingface/transformers/blob/79d4bc761d9da7a125b65a3b8c09441608d05436/src/transformers/models/timesfm/modeling_timesfm.py#L791

@kashif any chance we can add the modification directly in the transformers modeling ? otherwise we need to patch the model before export

xenova: Indeed, that does look to be the source of the loss of dynamic shapes. torch.onnx.export doesn't play well with python list operations, so we'd need to refactor it to use tensor operations, if possible.

PR #2138, opened 2024-12-25
Allow GPTQModel to auto select Marlin or faster kernels for inference only ops

LRL-ModelCloud:


Qubitium:
This PR addresses the issue where GPTQModel is unable to auto-select the fastest quant linear (Marlin) due to missing information regarding if this model is loaded via `pre-trained` or `from_quantized`. Add `pack=True` requirement  to `hf_select_quant_linear` (for pretrained quantization will solve this. Without the PR, user need to manually set `backend="marlin"` to use Marlin. This PR allows auto selection of `Marlin`, if compatible, in default `auto` mode.

Qubitium:
@SunMarc  @IlyasMoutawwakil  Ready for review.  Unlike AutoGPTQ, GPTQModel has pretty robust internal code to select the best/fastet quant linear available for model inference. However, for this to happen, we need to receive signal if `pack: bool` (packing) will be used. For loading a post-quantized model, packing is not needed so we can auto select `Marlin` on Nivida `CUDA` available platform or next-best kernels for the current env. Some fast inference kernels are inference only and cannot perform packing.

SunMarc (approved):
LGTM !

PR #2151, opened 2025-01-08
fix infer task from model_name if model from sentence transformer

eaidova:
# What does this PR do?
issue found during attempt to export https://huggingface.co/BAAI/bge-small-zh-v1.5 via optimum-intel without --task specification
The root cause is infer_library_from_model detects task as sentence-transformers even if --library transformers selected in cli and there is no additional condition for specification task for sentence transformers, as the result method raises error
```
KeyError: "The task could not be automatically inferred. Please provide the argument --task with the relevant task from image-text-to-text, audio-classification, zero-shot-image-classification, masked-im, audio-xvector, image-to-text, audio-frame-classification, depth-estimation, visual-question-answering, semantic-segmentation, image-to-image, image-classification, feature-extraction, image-segmentation, inpainting, sentence-similarity, object-detection, automatic-speech-recognition, multiple-choice, text-classification, fill-mask, text-generation, zero-shot-object-detection, text-to-image, reinforcement-learning, text2text-generation, token-classification, question-answering, mask-generation, text-to-audio. Detailed error: 'Could not find the proper task name for the model BAAI/bge-small-zh-v1.5.'
```

provided default task for sentence-transformers and added mechanism to explicitly set library_name (in case if user force --library_name in cli, need additional changes for enabling this behaviour in optimum-intel and other integrations that allow to select library_name) for avoiding mismatch between auto-detected and selected library (sentence transformers model can be exported with both transformers and sentence transformers library name and export configuration and exported model may be different for this case)

### DIFF
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -1856,6 +1856,7 @@ class TasksManager:
         revision: Optional[str] = None,
         cache_dir: str = HUGGINGFACE_HUB_CACHE,
         token: Optional[Union[bool, str]] = None,
+        library_name: Optional[str] = None,
     ) -> str:
         """
         Infers the task from the model repo, model instance, or model class.
@@ -1874,7 +1875,9 @@ class TasksManager:
             token (`Optional[Union[bool,str]]`, defaults to `None`):
                 The token to use as HTTP bearer authorization for remote files. If `True`, will use the token generated
                 when running `huggingface-cli login` (stored in `huggingface_hub.constants.HF_TOKEN_PATH`).
-
+            library_name (`Optional[str]`, defaults to `None`):
+                The library name of the model. Can be any of "transformers", "timm", "diffusers", "sentence_transformers". See `TasksManager.infer_library_from_model` for the priority should
+                none be provided.
         Returns:
             `str`: The task name automatically detected from the HF hub repo, model instance, or model class.
         """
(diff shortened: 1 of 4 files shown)


eaidova:
@IlyasMoutawwakil  @echarlaix  could you please take a look?

optimum/exporters/tasks.py
```diff
@@ -1770,6 +1770,7 @@ def _infer_task_from_model_name_or_path(
         revision: Optional[str] = None,
         cache_dir: str = HUGGINGFACE_HUB_CACHE,
         token: Optional[Union[bool, str]] = None,
+        library_name: Optional[str] = None,
```
IlyasMoutawwakil, line 1773: I see you're adding `library_name` to the signature but shouldn't this argument be specified in the method's calls as well ?

eaidova: I just want to get confirmation that it is reliable approach before doing that.
In case of optimum itself it also need to provide it to infer_task_from_model too https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/__main__.py#L259

IlyasMoutawwakil: yeah you can go ahead and add it where needed, it seems reliable.

eaidova: ok, added.

Only not sure about tflite exported, as I can see for getting export config, it uses library_name="transformers" explicitly, so it means that other library are not supported, right? I also added library_name="transformers" for avoid model type mismatch (e.g. in case of sentence transformers described above, if it will not be provided, model will be loaded as sentence transformer and raises error in some steps later

eaidova:
failed test is not related to my changes, modernbert is available only on transformers master branch, there is no transformers package that contains this model code yet

PR #2152, opened 2025-01-08
Replace check_if_xxx_greater with is_xxx_version

echarlaix:
Replace `check_if_xxx_greater` with `is_xxx_version` :  the later gives more freedom (we can check for a specific version with "==" for example instead of https://github.com/huggingface/optimum/blob/53240c3f680409226231f90450511fc321618248/optimum/exporters/onnx/model_configs.py#L374) also I find it easier to understand (I always need to check whether `check_if_xxx_greater` is >= or >)


IlyasMoutawwakil (approved):
LGTM ! this is definitely more readable !

PR #2153, opened 2025-01-09
Unpin diffusers and pass onnx exporters tests

IlyasMoutawwakil:
# What does this PR do?

There are also problems with modernbert making the CI red because it's still released in transformers.

## Before submitting


.github/workflows/test_export_onnx.yml
```diff
@@ -27,13 +27,15 @@ jobs:
...
-          pip install .[tests,exporters]
+          pip install .[tests,exporters,diffusers]
+          pip install transformers@git+https://github.com/huggingface/transformers.git
       - name: Test with unittest
         working-directory: tests
         run: |
           pytest exporters/onnx/test_onnx_*.py -s -n auto -m "not tensorflow_test and not timm_test" --durations=0
       - name: Install dependencies for tensorflow export
         run: |
+          pip uninstall diffusers -y
```
IlyasMoutawwakil, line 38: because diffusers no longer supports the transformers version exporters-tf uses

optimum/exporters/tasks.py
```diff
@@ -51,6 +56,12 @@
     from transformers import TFPreTrainedModel
 
 if is_diffusers_available():
+    if is_diffusers_version(">=", "0.32.0") and is_transformers_version("<", "4.41.2"):
+        raise ImportError(
+            "The current version of `diffusers` requires `transformers>=4.41.2`."
+            " Please upgrade to the latest version of Transformers."
+        )
+
```
IlyasMoutawwakil, line 64: I added this in case a user have incompatible versions but it's not necessary.

echarlaix (approved):
LGTM, thanks a lot @IlyasMoutawwakil

xenova:
Thanks @IlyasMoutawwakil 👍  For modernbert, is this something we can fix with
```py
MIN_TRANSFORMERS_VERSION = version.parse("4.48.0")
```
?

IlyasMoutawwakil:
oh yes apparently we can thanks to
https://github.com/huggingface/optimum/blob/c46de055049acc550eca1ff96414b2aba3a95dee/tests/exporters/onnx/test_onnx_export.py#L212-L216C
thanks for the catch, fixing it in a second

PR #2157, opened 2025-01-15
ExporterConfig refactorization

echarlaix:
The different exporters configs ([`ONNXConfig`](https://github.com/huggingface/optimum/blob/512d5c6ab25927e52a4b7fd2ef14361eb8f4d753/optimum/exporters/onnx/base.py#L105), [`TFLiteConfig`](https://github.com/huggingface/optimum/blob/512d5c6ab25927e52a4b7fd2ef14361eb8f4d753/optimum/exporters/tflite/base.py#L118), [`NeuronExportConfig`](https://github.com/huggingface/optimum-neuron/blob/9c02cbdf68827c6aec9eae3fea2c0ff5648f7cdc/optimum/exporters/neuron/base.py#L79)) which all are subclass of `ExportersConfig` share a lot of different attributes and methods


optimum/exporters/onnx/base.py
```diff
@@ -213,53 +177,12 @@ def __init__(
...
-        self.task = task
-        self.int_dtype = int_dtype
-        self.float_dtype = float_dtype
+        super().__init__(config=config, task=task, int_dtype=int_dtype, float_dtype=float_dtype)
 
-        self._config = config
-        self._preprocessors = preprocessors
-        self._normalized_config = self.NORMALIZED_CONFIG_CLASS(self._config)
         self.variant = "default"
+        self._preprocessors = preprocessors
```
IlyasMoutawwakil, line 183: could be moved to the base config as well (the way i see it is that information about the model is sometimes scattered across config and preprocessors)

optimum/exporters/base.py
```diff
@@ -14,8 +14,238 @@
...
+                The number of frames to use in the dummpy inputs for audio tasks in case the input is not raw audio.
+            audio_sequence_length (`int`, defaults to {audio_sequence_length}):
+                The number of frames to use in the dummpy inputs for audio tasks in case the input is raw audio.
+
+        Returns:
+            `Dict[str, [tf.Tensor, torch.Tensor]]`: A dictionary mapping the input names to dummy tensors in the proper framework format.
+"""
 
 
 class ExportConfig(ABC):
```
IlyasMoutawwakil, line 71: Should we keep the name as ExportConfig or rename it ExportersConfig ?

PR #2165, opened 2025-01-21
Move executorch to optimum-executorch

echarlaix:
ExecuTorch export (introduced in https://github.com/huggingface/optimum/pull/2090) is moved in `optimum-executorch` see https://github.com/huggingface/optimum-executorch/pull/1

Following https://github.com/huggingface/optimum/pull/2090#issuecomment-2566266913


setup.py
```diff
@@ -84,10 +84,7 @@
         "datasets<=2.16",
         "transformers>=4.36,<4.38",
     ],
-    "exporters-executorch": [
-        "executorch>=0.4.0",
-        "transformers>=4.46",
-    ],
+    # "executorch": "optimum-executorch",
```
michaelbenayoun, line 87: Why is it commented for now?

echarlaix: package doesn't exist yet, we could install it from source after https://github.com/huggingface/optimum-executorch/pull/1 is merged

guangy10:
@echarlaix It looks like all docs are deleted unintentionally. Can you take a look?

echarlaix:
@guangy10 we are planning to add the documentation to https://github.com/huggingface/optimum-executorch directly, let's continue the discussion in https://github.com/huggingface/optimum-executorch/issues/2

PR #2190, opened 2025-02-14
Update optimum neuron extra

dacorvo:
# What does this PR do?

This updates the `neuronx` extra to remove the `transformers` version.
Installing `optimum-neuron` through `optimum` is anyway not the preferred installation mode: the documentation now recommends to install `optimum-neuron` directly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,6 @@ EXTRAS_REQUIRE = {
     "neural-compressor": "optimum-intel[neural-compressor]>=1.18.0",
     "ipex": "optimum-intel[ipex]>=1.18.0",
     "habana": ["optimum-habana", "transformers>=4.45.0,<4.46.0"],
-    "neuron": ["optimum-neuron[neuron]>=0.0.20", "transformers>=4.36.2,<4.42.0"],
     "neuronx": ["optimum-neuron[neuronx]>=0.0.20", "transformers>=4.36.2,<4.42.0"],
     "graphcore": "optimum-graphcore",
     "furiosa": "optimum-furiosa",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -99,8 +99,7 @@
     "neural-compressor": "optimum-intel[neural-compressor]>=1.18.0",
     "ipex": "optimum-intel[ipex]>=1.18.0",
     "habana": ["optimum-habana", "transformers>=4.45.0,<4.46.0"],
-    "neuron": ["optimum-neuron[neuron]>=0.0.20", "transformers>=4.36.2,<4.42.0"],
```
echarlaix, line 102: why remove the neuron extra (looks like still mentioned in the documentation https://huggingface.co/docs/optimum-neuron/en/installation, even though it's using the extra from optimum-neuron directly)

setup.py
```diff
@@ -99,8 +99,7 @@
...
-    "neuronx": ["optimum-neuron[neuronx]>=0.0.20", "transformers>=4.36.2,<4.42.0"],
+    "neuronx": ["optimum-neuron[neuronx]>=0.0.28"],
```
echarlaix, line 102: Could this create incompatibilities for some transformers version ? I don't think it should but I have a vague memory of this mentioned by @JingyaHuang in some specific cases (it would also explain why it was added in the first place)

echarlaix (approved):

setup.py
dacorvo: We don't really support neuron anymore (at least we don't run the CI), I think eventually we will completely remove it from `optimum-neuron`.

setup.py
dacorvo: I think it was added as a tentative fix for the installation through `optimum[neuronx]`, but this did not really solve anything because there were now potentially three competing versions (one from `optimum`, one from this extra, one from `optimum-neuron` itself).

setup.py
echarlaix: ok very clear, thanks for the explanation @dacorvo

PR #2193, opened 2025-02-20
Add executorch documentation section

echarlaix:
Can be merged once executorch documentation is generated https://github.com/huggingface/optimum-executorch/pull/21

Also a section can be added in https://github.com/huggingface/optimum/blob/v1.24.0/docs/source/index.mdx?plain=1#L60 after the first release of `optimum-executorch`

Fixes https://github.com/huggingface/optimum-executorch/issues/2

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -186,7 +186,7 @@ jobs:
       - name: Combine subpackage documentation
         run: |
           cd optimum
-          sudo python docs/combine_docs.py --subpackages nvidia amd intel neuron tpu habana furiosa --version ${{ env.VERSION }}
+          sudo python docs/combine_docs.py --subpackages nvidia amd intel neuron tpu habana furiosa executorch --version ${{ env.VERSION }}
           cd ..
 
       - name: Push to repositories
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -120,7 +120,7 @@ jobs:
       - name: Combine subpackage documentation
         run: |
           cd optimum
-          sudo python docs/combine_docs.py --subpackages nvidia amd intel neuron tpu habana furiosa --version pr_$PR_NUMBER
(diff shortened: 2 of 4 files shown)


guangy10:
@echarlaix Thanks for bringing the doc back. Lots of users were asking about this! 

I'm also very excited about the first release! What is the timeline and do you have particular things in mind that we should include in the first release? To me I feel we should probably include implementation of `save_pretrained`, `push_to_hub`, etc. so that users will have a complete story of distribute the pte artifacts through HF. Besides that, I'd like to get more models covered, that's something I'm looking into, but this can be optional. wdyt?

guangy10 (approved):
LGTM

guangy10:
@echarlaix Where are we going to list the supported models? Don't see that page is ported over from the original PR: https://github.com/huggingface/optimum/commit/d21256c2964945fc3fe4623f7befb21082b69a25#diff-203c8cf6d5e8aa8a958f2fbe209e15c840dbbedf1d8f6afb2f31e55046d28d75. We could include Deepseek-R1 (Llama based), and SmolLM2

regisss (approved):
LGTM!

guangy10:
@echarlaix Is there a time to merge this PR? We are currently hold sharing it broadly because of the user experience. With the documentation, users won't know explicit what is working and what is not, and they have been asking this in our discord channel. I cc'd you on some thoughts above about features we need to have in the 1st release. wdyt?

echarlaix:
Sorry for the delay @guangy10, documentation should soon be made available

echarlaix:
@guangy10 the executorch documentation can be found in [here](https://huggingface.co/docs/optimum-executorch/index), will also be part of the optimum documentation under the ExecuTorch section (currently main documentation is not pushed correctly but this is not related)

image

guangy10:
> @guangy10 the executorch documentation can be found in [here](https://huggingface.co/docs/optimum-executorch/index), will also be part of the optimum documentation under the ExecuTorch section (currently main documentation is not pushed correctly but this is not related)
> 
> image

Thanks @echarlaix I added this to the bugbash as well for folks to test and review

PR #2196, opened 2025-02-21
Remove instance-level forward on unpatch

AlexanderDokuchaev:
# What does this PR do?

Prevents unexpected instance-level overrides of the forward function after the original method is restored in ModelPatcher

```python
assert "forward" not in model.__dict__
patcher = ModelPatcher(export_config, model)
with patcher:
    pass
assert "forward" not in model.__dict__  # AssertionError
```

## Before submitting
- This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- Did you make sure to update the documentation with your changes?

## Who can review?

### DIFF
--- a/optimum/exporters/onnx/model_patcher.py
+++ b/optimum/exporters/onnx/model_patcher.py
@@ -225,6 +225,7 @@ class ModelPatcher:
             self._patching_specs.append(final_spec)
 
         self.orig_forward_name = "forward" if hasattr(self._model, "forward") else "call"
+        self.orig_forward_is_instance_attr = self.orig_forward_name in model.__dict__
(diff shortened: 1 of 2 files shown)


IlyasMoutawwakil:
where does this create an issue ? ModelPatcher is only used in the context of model export so I don't see how it could impact anything outside of that context.

AlexanderDokuchaev:
@IlyasMoutawwakil 
```
1. load model
2. export
3. train as DataParallel <- fail 
```

ModelPatcher in __exit__ function does not restore the model to its original state, "forward" restored as instance attribute (collectes in model.__dict__), but in original it's class method.

In DataParallel case, this leads to the fact that each replicate of the model has a `forward` which is a bound method that links to original model.

PR #2203, opened 2025-02-28
fix the onnx export custom model example; fix repo name; fix opset version; remove deprecated arg;

janak2:
# What does this PR do?
Fix the example for custom model export 

Fixes # (issue)

## Before submitting
- [ x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ x] Did you make sure to update the documentation with your changes?

## Who can review?
@fxmarty


echarlaix (approved):
Thanks for the fix @janak2

PR #2213, opened 2025-03-11
Remove FP16_Optimizer patch for DeepSpeed

Rohan138:
Deepspeed (since 2022) already includes the FusedAdam FP16_Optimizer originally from NVIDIA/apex here (and it's better maintained wrt deepspeed grad_norm, MOE training, etc): https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/runtime/fp16/fused_optimizer.py

Currently this line gives a warning saying:
```
/opt/conda/envs/py_3.10/lib/python3.10/site-packages/onnxruntime/training/optim/_modifier_registry.py:56: UserWarning: Skip modifying optimizer because of optimizer name not found in the registry: accelerate.utils.deepspeed.DeepSpeedOptimizerWrapper
```

which is actually caused by `get_full_qualified_type_name(optimizer.optimizer) = 'deepspeed.runtime.fp16.fused_optimizer.FP16_Optimizer'` not being in the onnxruntime registry: https://github.com/microsoft/onnxruntime/blob/main/orttraining/orttraining/python/training/optim/_modifier_registry.py

i.e. the FP16 Optimizer from onnxruntime (https://github.com/microsoft/onnxruntime/blob/main/orttraining/orttraining/python/training/optim/fp16_optimizer.py) is not actually doing anything, just falling back to the DeepSpeed fused Adam optimizer anyway, so this line and the warning it causes are redundant.

# What does this PR do?

Fixes # (issue)

## Before submitting

## Who can review?
@JingyaHuang

### DIFF
--- a/optimum/onnxruntime/trainer.py
+++ b/optimum/onnxruntime/trainer.py
@@ -572,12 +572,6 @@ class ORTTrainer(Trainer):
                 )
             self.model = unwrap_model(model)
 
-        # ORT optimized FP16 optimizer for Deepspeed training
-        if self.is_deepspeed_enabled and args.fp16:
-            from onnxruntime.training.optim.fp16_optimizer import FP16_Optimizer
-
-            self.optimizer = FP16_Optimizer(self.optimizer)
-
         if self.is_fsdp_enabled:
             self.model = model
 


IlyasMoutawwakil (approved):
LGTM, tests are passing !
is onnxruntime-training still useful for you ? we are discussing deprecating it since usage is so low.

PR #2001, opened 2024-08-29
Add ONNX export support for Hiera models

xenova:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #2227, opened 2025-04-09
add dinov2 onnx optimizer support

EFord36:
# What does this PR do?

Following the addition of Dinov2 Onnx export support in https://github.com/huggingface/optimum/pull/2001 , this PR seeks to enable optimiser support for DinoV2.

I thought I'd have a go at this since I could see ViT was supported, and DinoV2 is just a ViT backbone.

I think this is working - when I run the test locally, it actually fails, but the output is visually identical, and the test for vit fails for me in the same way. I suspect this is something about my environment - opening the PR to see how it behaves on CI, I can debug further when I get a chance.

Fixes # (issue)

## Before submitting
- [X] Did you write any new necessary tests?

## Who can review?

Onnx runtime related, so:


IlyasMoutawwakil:
hi, can you please rebase your branch, I will retrigger the tests afterwards.

EFord36:
Looks to me like the two failing checks fail with the same errors in the latest commit to main

IlyasMoutawwakil:
Thanks @EFord36 🤗

PR #2234, opened 2025-04-19
Introduce ORTSessionMixin and enable general io binding (works for diffusers as well)

IlyasMoutawwakil:
# What does this PR do?

This PR tries to further extend the io binding feature to multi part pipelines and enables it more generally through two classes:

- `ORTSessionMixin`: should be used whenever a class has an underlying inference session, it enables inference with and without io/binding, and handles the torch/onnxruntime interface like provider/device/dtype.

- `ORTSessionsWrapper`: should be used when a class has many models/parts (like encoder-decoders or diffusion pipelines), it usinfies the interface to these parts without compromising flexibility, for example one on can set `pipeline.use_io_binding` to True/False which will set it to all parts, but can also target a specific part `pipeline.unet.use_io_binding` differently. Parts can also be on different devices.

I also added a special feature to diffusion pipelines and especially Unet/Tranformers component so that it reuses the same input tensor as output buffer (instead of creating a new one), this makes the diffusion operation more of an in-place operation, which should make it faster (theoretically), or at least uses less memory.

## Before submitting


tests/onnxruntime/test_diffusion.py
```diff
@@ -72,6 +89,116 @@ def _generate_images(height=128, width=128, batch_size=1, channel=3, input_type=
     return [image] * batch_size
 
 
+class ORTDiffusionPipelineTest(TestCase):
```
IlyasMoutawwakil, line 92: moved the ORTDiffusionPipeline integration tests here

optimum/onnxruntime/base.py
```diff
@@ -11,453 +11,592 @@
...
-        self.output_shapes = {output_key.name: output_key.shape for output_key in session.get_outputs()}
+        self.input_names = {input.name: idx for idx, input in enumerate(session.get_inputs())}
+        self.output_names = {output.name: idx for idx, output in enumerate(session.get_outputs())}
+        self.input_shapes = {input.name: input.shape for input in session.get_inputs()}
+        self.output_shapes = {output.name: output.shape for output in session.get_outputs()}
+        self.input_dtypes = {input.name: input.type for input in session.get_inputs()}
+        self.output_dtypes = {output.name: output.type for output in session.get_outputs()}
+
+        self.model_path = Path(session._model_path)
+        self.model_name = self.model_path.name
```
echarlaix, line 87: doesn't seem to be used anywhere so we could potentially remove it in a future release (adding a warning in the meantime)

optimum/onnxruntime/base.py
```diff
@@ -11,453 +11,592 @@
...
+            ValueError: If the device is not supported or if the provider is not available.
+        """
+
+        dtype = None
+        device = None
 
-    def to(self, *args, device: Optional[Union[torch.device, str, int]] = None, dtype: Optional[torch.dtype] = None):
         for arg in args:
-            if isinstance(arg, torch.device):
+            if isinstance(arg, str):
```
echarlaix, line 175:
```suggestion
            if isinstance(arg, (str, torch.device)):
```

optimum/onnxruntime/base.py
```diff
@@ -11,453 +11,592 @@
...
+            output_name (`str`):
+                The name of the output for which to prepare the buffer.
+            output_shape (`Tuple[int]`):
+                The shape of the output buffer.
+
+        Returns:
+            `torch.Tensor`: The output buffer.
+
+        """
+        assert len(output_shape) > 0, "`output_shape` should not be empty."
```
echarlaix, line 308: would replace with if-raise

optimum/onnxruntime/base.py
```diff
@@ -11,453 +11,592 @@
...
+            return model.provider_option
+
+    @property
+    def device(self):
+        """
+        Returns the `torch.device` associated with the ONNX Runtime session.
+        This device is inferred from the provider and provider options.
+        """
+        for model in self.sessions:
+            return model.device
```
echarlaix, line 575: why not 
```suggestion
         return self.sessions[0].device
```

optimum/onnxruntime/modeling_ort.py
```diff
@@ -152,68 +136,47 @@ class ORTModel(OptimizedModel):
...
-        Get the task corresponding to a class (for example AutoModelForXXX in transformers).
-        """
-        return TasksManager.infer_task_from_model(auto_model_class)
-
-    def shared_attributes_init(
+    def __init__(
         self,
-        model: ort.InferenceSession,
+        config: "PretrainedConfig",
+        session: "InferenceSession",
```
echarlaix, line 155: not against but breaking change so not sure that's worth it (could result in issues for users who don't use .from_pretrained)

echarlaix (approved):
Excellent work, thanks @IlyasMoutawwakil

optimum/exporters/onnx/__main__.py
```diff
@@ -265,6 +272,17 @@ def main_export(
                 f"The task could not be automatically inferred as this is available only for models hosted on the Hugging Face Hub. Please provide the argument --task with the relevant task from {', '.join(TasksManager.get_all_tasks())}. Detailed error: {e}"
             )
 
+    if (
+        task == "feature-extraction"
```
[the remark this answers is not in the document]
IlyasMoutawwakil: no need ! nice catch !

optimum/exporters/onnx/__main__.py
```diff
@@ -265,6 +272,17 @@ def main_export(
...
+        and library_name == "sentence_transformers"
+        and not is_sentence_transformers_available()
+    ):
+        logger.warning(
+            "The library name was inferred as `sentence_transformers` with the task `feature-extraction`, "
+            "but the library is not installed. Defaulting to `transformers` to avoid breaking the export."
+        )
+        library_name = "transformers"
```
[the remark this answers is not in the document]
IlyasMoutawwakil: Yes will generalize it !

optimum/onnxruntime/modeling_ort.py
IlyasMoutawwakil: I agree, this is the change I'm hesitant about the most 😅
I think we can do it slowly by using *args and warning users from passing positional arguments to the ORTModel class.

IlyasMoutawwakil: I pushed a "solution" that (hopefully) doesn't break BC, tell me what you think.
I see that some libs are using ORTModelForCausalLM and instantiating it directly (mostly intel/onnx neural-compressor).
We can keep this until optimum 2.0

optimum/exporters/onnx/model_configs.py
```diff
@@ -1337,45 +1338,51 @@ class VaeEncoderOnnxConfig(VisionOnnxConfig):
...
+                0: "batch_size",
+                2: f"sample_height / {down_sampling_factor}",
+                3: f"sample_width / {down_sampling_factor}",
+            },
         }
 
 
 class VaeDecoderOnnxConfig(VisionOnnxConfig):
-    ATOL_FOR_VALIDATION = 1e-4
+    ATOL_FOR_VALIDATION = 3e-4
```
echarlaix, line 1363: why is it needed ? (flaky test?)

optimum/onnxruntime/base.py
echarlaix: yes talking about `model_name` which seems to be used in the tests (and can be easily extracted from `model_path`) so we can remove imo, not specific to this PR though so feel free to close!

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -596,50 +639,34 @@ def _from_pretrained(
...
 
     @classmethod
-    def _from_transformers(
+    def _export(
         cls,
-        model_id: str,
+        model_id: Union[str, Path],
         config: "PretrainedConfig",
+        # export options
+        task: Optional[str] = None,
```
echarlaix, line 657:
```suggestion
```
what do you think about deprecating `task` completely ? (raisin an error if provided)

optimum/onnxruntime/modeling_diffusion.py
```diff
@@ -161,145 +191,184 @@ def __init__(
...
-            (self.unet, save_directory / DIFFUSION_MODEL_UNET_SUBFOLDER),
-            (self.transformer, save_directory / DIFFUSION_MODEL_TRANSFORMER_SUBFOLDER),
-            (self.vae_decoder, save_directory / DIFFUSION_MODEL_VAE_DECODER_SUBFOLDER),
-            (self.vae_encoder, save_directory / DIFFUSION_MODEL_VAE_ENCODER_SUBFOLDER),
-            (self.text_encoder, save_directory / DIFFUSION_MODEL_TEXT_ENCODER_SUBFOLDER),
-            (self.text_encoder_2, save_directory / DIFFUSION_MODEL_TEXT_ENCODER_2_SUBFOLDER),
-            (self.text_encoder_3, save_directory / DIFFUSION_MODEL_TEXT_ENCODER_3_SUBFOLDER),
+    @property
+    def components(self) -> Dict[str, Optional[Union[ORTSessionMixin, torch.nn.Module]]]:
+        # TODO: all compoenents should be ORTSessionMixin's at some point
```
echarlaix, line 206: yes good point! 

also small typo : 
```suggestion
        # TODO: all components should be ORTSessionMixin's at some point
```

echarlaix (approved):
LGTM! Thanks a lot for the great PR @IlyasMoutawwakil  :fire:

optimum/exporters/onnx/model_configs.py
IlyasMoutawwakil: yeah it's the flaky onnx export (fails once in while, also matches atol of encoder), I think the encoding/decoding process to/from a compressed representation with convolutions is a very flaky in general (convolution kernels are generally non-deterministic).

PR #2247, opened 2025-04-29
Respect custom file_name in ORTModelForCausalLM.from_pretrained

### LINKED ISSUE #2246: Custom file_name in ORTModelForCausalLM.from_pretrained not respected when multiple files ONNX files found
### System Info

```shell
optimum v1.25.0-dev
Python 3.11
```

### Who can help?

_No response_

### Information

- [x] My own modified scripts

### Tasks

```
from transformers import AutoTokenizer, pipeline
import onnxruntime as ort

# Load model with custom ONNX filename
    "HuggingFaceTB/SmolLM2-135M-Instruct",
    export=True,
    file_name="model_fp16.onnx",
    provider="CPUExecutionProvider"
)
```

A warning will be rendered  that a `file_name` should be used to specify which file to load and `Loading the file model_quantized.onnx in the folder onnx`.

### Expected behavior

The method should prioritize the user's explicit choice over auto-detection mechanisms so that users have precise control over which model file gets loaded.

dpfens:
Fixes an issue (#2246 ) in the `_from_pretrained` method where the `file_name` parameter was not being respected when multiple ONNX files were found. The main changes:

1. When a user explicitly provides a `file_name`, we now respect that choice rather than potentially overriding it with a discovered file
2. Only when `file_name` is `None` do we attempt to find a default file (first trying "model.onnx", then falling back to the first file found), and emit warning suggesting a file_name` be specified.
3. The discovered filename is only assigned to `file_name` if no value was initially provided

### Steps to verify correction
```python

# Load model with custom ONNX filename
    "HuggingFaceTB/SmolLM2-135M-Instruct",  # this model has multiple eligible ONNX files
    export=True,
    file_name="model_fp16.onnx", # previously would default to model_quantized.onnx instead of the specified file
    provider="CPUExecutionProvider"
)
```

### DIFF
--- a/optimum/onnxruntime/modeling_decoder.py
+++ b/optimum/onnxruntime/modeling_decoder.py
@@ -459,21 +459,24 @@ class ORTModelForCausalLM(ORTModel, GenerationMixin):
             _file_name = model_files[0].name
             subfolder = model_files[0].parent
 
-            defaut_file_name = file_name or "model.onnx"
-            for file in model_files:
-                if file.name == defaut_file_name:
-                    _file_name = file.name
-                    subfolder = file.parent
-                    break
+            if file_name is not None:
+                _file_name = file_name
+            else:
+                _file_name = model_files[0].name
+                for file in model_files:
+                    if file.name == "model.onnx":
+                        _file_name = file.name
+                        subfolder = file.parent
+                        break
 
-            file_name = _file_name
+                file_name = _file_name
 
-            if len(model_files) > 1:
-                logger.warning(
-                    f"Too many ONNX model files were found in {' ,'.join(map(str, model_files))}. "
-                    "specify which one to load by using the `file_name` and/or the `subfolder` arguments. "
-                    f"Loading the file {file_name} in the subfolder {subfolder}."
-                )
+                if len(model_files) > 1:
+                    logger.warning(
+                        f"Too many ONNX model files were found in {' ,'.join(map(str, model_files))}. "
+                        "specify which one to load by using the `file_name` and/or the `subfolder` arguments. "
+                        f"Loading the file {file_name} in the subfolder {subfolder}."
+                    )
 
         if os.path.isdir(model_id):
             model_id = subfolder


optimum/onnxruntime/modeling_decoder.py
```diff
@@ -459,21 +459,24 @@ def _from_pretrained(
...
-            file_name = _file_name
-
-            if len(model_files) > 1:
-                logger.warning(
-                    f"Too many ONNX model files were found in {' ,'.join(map(str, model_files))}. "
-                    "specify which one to load by using the `file_name` and/or the `subfolder` arguments. "
-                    f"Loading the file {file_name} in the subfolder {subfolder}."
-                )
+            if file_name is not None:
+                _file_name = file_name
```
IlyasMoutawwakil, line 463: no need for `_file_name` (or its assignment) in this case

IlyasMoutawwakil (approved):
LGTM

echarlaix:
Hi @dpfens  here the issue is that you set `export=True` which will force the model to be re-exported, you shouldn't explicitly set export to True if you don't want to force the model to be re-exported (which could be needed if you want the model to be exported with a different opset for example). Outside this situation it's recommended not to set export apart from that as it will be automatically inferred :  if an onnx model is found in the repo it will be detected and loaded, if no onnx model is found then it will look for a pytorch model to export. If you remove `export=True` : the onnx model will be loaded as expected (using the `file_name` you specified).

```python

    "HuggingFaceTB/SmolLM2-135M-Instruct",
    file_name="model_fp16.onnx",
    provider="CPUExecutionProvider"
)
```

optimum/onnxruntime/modeling_decoder.py
echarlaix, line 463: this part of the code is used to find the subfolder associated to the `file_name`, with your modifications loading will fail if there is multiple subfolder with onnx models and if the later is not correctly specified, for example with this modification this loading will fail :

```python
    "optimum-internal-testing/tiny-random-llama",
    revision="merged-onnx",
    file_name="decoder_model_with_past.onnx",
)
```

echarlaix (changes requested):

optimum/onnxruntime/modeling_decoder.py
echarlaix: I don't think this code should be modified, please let me know if there is a model that is being currently being incorrectly loaded

IlyasMoutawwakil:
ooh I see what's happening thanks @echarlaix , should we raise an error when export=True and `file_name` is provided ?

dpfens:
@echarlaix I followed your instructions again today and it's behaving exactly as you said it should.

Sorry for the delay, and for opening a need-less PR.

PR #2249, opened 2025-04-30
Add ONNX export support for D-FINE

xenova:
# What does this PR do?
Adds ONNX export support for D-FINE models, added in https://github.com/huggingface/transformers/pull/36261

Fixes # (issue)

## Before submitting

## Who can review?

@echarlaix @qubvel


echarlaix (approved):
LGTM, thanks a lot @xenova

xenova:
Should we wait for a new transformers release (which includes d-fine, as it's currently in main), or is merging now fine?

We currently get this error due to an out-of-date transformers version:
```
FAILED tests/exporters/onnx/test_export_cli.py::OnnxCLIExportTestCase::test_exporters_cli_pytorch_cpu_127_d_fine_no_task_ustc_community_dfine_n_coco - ValueError: The checkpoint you are trying to load has model type `d_fine` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.
```

Merging is probably fine since it doesn't break anything for users, but it does make our CI red 😅

optimum/exporters/onnx/model_patcher.py
```diff
@@ -32,7 +32,7 @@
     from transformers.modeling_attn_mask_utils import AttentionMaskConverter
 if is_transformers_version(">=", "4.36"):
     from transformers.modeling_attn_mask_utils import _prepare_4d_causal_attention_mask_for_sdpa
-if is_transformers_version(">=", "4.43"):
+if is_transformers_version(">=", "4.43") and is_transformers_version("<", "4.51.999"):
```
IlyasMoutawwakil, line 35:
```suggestion
if is_transformers_version(">=", "4.43") and is_transformers_version("<", "4.48"):
```

tests/exporters/utils.py
```diff
@@ -64,6 +64,7 @@
     "convnextv2": "hf-internal-testing/tiny-random-ConvNextV2Model",
     "codegen": "hf-internal-testing/tiny-random-CodeGenModel",
     "cvt": "hf-internal-testing/tiny-random-CvTModel",
+    "d-fine": "ustc-community/dfine_n_coco",
```
IlyasMoutawwakil, line 67: do we not have a smaller variant ?

optimum/exporters/onnx/model_patcher.py
IlyasMoutawwakil, line 35: this is pretty tricky, it probably requires incrementing the min opset to 14 since the model defaults to sdpa

optimum/exporters/onnx/model_configs.py
```diff
@@ -2693,3 +2693,7 @@ def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGen
 
 class RTDetrV2OnnxConfig(RTDetrOnnxConfig):
     pass
+
+
+class DFineOnnxConfig(RTDetrOnnxConfig):
+    pass
```
IlyasMoutawwakil, line 2699: let's add a min transformers version so that the test is skipped

PR #2251, opened 2025-05-06
Add ONNX exporter support for ColPali model

Balladie:
# What does this PR do?

Adds ONNX exporter support for `ColPaliForRetrieval` model. The resulting ONNX has two variant options: one for image embedding extraction, and the other one for text embedding. Although previous PR exists (#2074), that work became stale and I just wanted to follow up and create a new one with some modifications.

The exported models, conversion script, and usages are all uploaded to the in my [collection](https://huggingface.co/collections/ssonpull519/colpali-v13-onnx-6819c4aee1bb72c4beaaba0b), so please refer to it.

Some notes fyi below:
- The reason for two variants (resulting in two separate ONNX models) stems from the difference between the forward of image and text. The image takes additional forward of SigLIP vision tower, and I can rather make separate ONNX for vision tower and take `input_embeds` as input. But I sticked to current implementation for consistency of input levels with other configs in optimum.
- In #2074 the variants for image and text are split by task, but I thought it would be better to use other separation rather than task and decided to use variant.
- It can be either implemented by first implementing an onnx config for PaliGemma and then inherting it, but this model is in purpose of embedding extraction (without need for past KV, etc.), so I just skipped it and left it as a class inheriting Gemma config directly. Can be implemented as well in the future.
- ~~`optimum-cli` is currently not supported. To support it, `ColPaliForRetrieval` should be mapped in `AutoModel` in transformers, not in `AutoModelForPreTraining` as of now.~~

Thanks for the review, and please let me know if there's any suggestion! Any better idea is much appreciated.

## Who can review?


optimum/exporters/onnx/model_configs.py
```diff
@@ -2693,3 +2693,65 @@ def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGen
...
+        allow_new=True,
+        text_config="text_config",
+        vision_config="vlm_config.vision_config",
+        vlm_config="vlm_config",
+    )
+    ATOL_FOR_VALIDATION = 1e-4
+
+    VARIANTS = {
+        "vision": "Embedding extraction for image.",
+        "text": "Embedding extraction for text.",
```
IlyasMoutawwakil, line 2710: wouldn't it make more sense to call them vision-language and laguage (or vision+text and text-only), etc
because we do pass text as well in the first no ?

optimum/exporters/tasks.py
```diff
@@ -525,6 +525,10 @@ class TasksManager:
             "text-generation-with-past",
             onnx="CodeGenOnnxConfig",
         ),
+        "colpali": supported_tasks_mapping(
+            "feature-extraction",
+            onnx="ColPaliOnnxConfig",
+        ),
```
IlyasMoutawwakil, line 531: > optimum-cli is currently not supported. To support it, ColPaliForRetrieval should be mapped in AutoModel in transformers, not in AutoModelForPreTraining as of now.

You should be able to bypass that using `_CUSTOM_CLASSES` by registering it as a custom class (same is done for pix2struct)

optimum/exporters/onnx/model_configs.py
Balladie: Yeah I understand your point. My thoughts on the naming was that it is more intuitive for end user to call it as the embedding of each modality regardless of the internal logic, considering the usage scenario of real users (the implementation of `ColPaliProcessor` also takes either image + prompt w/o custom text, or just a text). But in terms of internal inputs it is also true so if you still prefer it I will follow your opinion.

Balladie:
Confirmed it works with optimum-cli after the fix. Can now export it like below:

```bash
optimum-cli export onnx --model vidore/colpali-v1.3-hf ./onnx_output --task feature-extraction --variant vision
```

The dtype of fp16 is not working for now but it seems to be the problem with transformers, but works with fp32 so I just keep it as it is. Will probably fix in transformers in the future.

@IlyasMoutawwakil

optimum/exporters/onnx/model_configs.py
```diff
@@ -2693,3 +2693,65 @@ def _create_dummy_input_generator_classes(self, **kwargs) -> List["DummyInputGen
...
+    DEFAULT_VARIANT = "vision"
+
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        dynamic_axis = {0: "batch_size", 1: "sequence_length"}
+        if self.variant == "vision":
+            return {
+                "input_ids": dynamic_axis,
+                "attention_mask": dynamic_axis,
+                "pixel_values": {0: "batch_size"},
```
IlyasMoutawwakil, line 2721: are the width/height static here ?

cyan-channel-io: yea the official [usage](https://huggingface.co/vidore/colpali-v1.3-hf) preprocesses image to the fixed size of 448 and 3 channels, and as far as I know the model is also trained with 448 size images, so considering the usage I thought it would be fine to leave it as static. let me know if i missed

Balladie: think both can be fine but the official weight should take fixed image size of 448 as input and the guides also preprocess in that way so I left it as static, let me know if i missed

Balladie:
> Okay I took a deeper look and it seems to me that doing it this way with variants will require users to create two onnx models to solve the retrieval task (which is what colpali is for), the problem is that those two models share most of the same weights. I think a better way to do it is to follow the example in [optimum-intel](https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/model_configs.py#L1659) where for vlms we define a couple "behaviors" that are exported separately and used together during inference. So instead of having to load two VLMs, one that takes pixel values and another that doesn't, we can divide the forward call defined [here](https://github.com/huggingface/transformers/blob/a5c6172c81d69a6fa2c3b1340d72fc669b941dcd/src/transformers/models/paligemma/modeling_paligemma.py#L338) into "components"/"behaviors" that run depending on the input. @Balladie tell me if what I said makes sense 😄
> 
> @echarlaix what do you think ? I think we should have at least one clean VLM implementation (currently we have none) so that contributors can use it as a base recipe.

You've got the point and that's what I exactly thought and looking for (but I did not have much time to work on it atm), so I greatly appreciate that you pointed out the necessity for baseline VLM configs. I did not know there are some vlms already in optimum-intel, and I think it looks fine for the base for now.
So basically what I understand is that we can divide into parts (ex. vision tower, text embeds, lm decoder) for each behavior and export them into three onnx models like [this](https://huggingface.co/Xenova/nanoLLaVA/tree/main/onnx) which has compatible formats with transformers.js (that's also what I'd like to have). Would love to see optimum naturally supports that (which has been done quite manually as far as I know), and also it would also be great if optimum-cli supports that as an option for the users (ex. `--behavior vision`).

I think your team would have better understanding of how it should be implemented but if it's helpful to make a simple starting point at least in this model I can definitely work on it. Kindly request your suggestion on how we can move forward for that!

@IlyasMoutawwakil

optimum/exporters/onnx/model_configs.py
IlyasMoutawwakil: let's go with vision-language and language-only, as the vision language variant can work for both text+pixels and text alone (with empty pixels). cc @kstavro if you have any input on this !

IlyasMoutawwakil (approved):
LGTM will merge tomorrow !
The name of variants is a nit, let me know what you think.

optimum/exporters/onnx/model_configs.py
cyan-channel-io: yeah I think that makes sense, looks good to me :)

Balladie: yea I think that makes sense, looks good to me!

kstavro:
Hey @Balladie @IlyasMoutawwakil, I just saw the tag from the other issue, thanks for taking care of this.

The main thing to take from my notebook when exporting the model was that in order to export a single model instead of two, the ‘ColPaliProcessor’ needs to be changed to also return dummy pixel_values for text as well (maybe with an argument like return_dummy_pixel_values or so). And that is actually enough for everything to work, as this makes the expected input for the ONNX model uniform.

I think the dummy values don’t actually play any role when processing text, as there are no image tokens as input for the model (image tokens are only concatenatied for image and images are embedded only when model image tokens exist). I first tried to manually concatenate the dummy pixel values to the text tokens for the text cases (while properly setting the attention mask to 0 for the image tokens) and that had the exact same results. But with the additional computational cost of 1024 unnecessary image tokens, so not worth it. Then I realized that what I was doing was a modified ColPaliProcessor in its essence.

The trade off to such an approach is that the ColPaliProcessor is part of the transformers library, not optimum. Where as with the export of two models everything remains more self contained, but with double the memory consumption. That is why I asked in the other issue about any design preferences.

I hope this helps somehow, thank you for integrating this model to the library :)

Balladie:
@IlyasMoutawwakil have manually tested that separating into three parts works (codes and onnx models [here](https://github.com/Balladie/optimum/pull/1))
it's not a full reflection of the concept of "behavior" but the flow would be aligned so I just share it with current form, hope it helps

IlyasMoutawwakil:
@kstavro no need to modify the processor logic, dummy pixel values could be created rather in the inference class `ORTModelForRetrieval` for example if not passed.
We do something similar for LLMs where we export them with past key values and create dummy ones in the first forward pass. https://github.com/huggingface/optimum/blob/e15053d33e60f42bb87389a869c3a9d823ea972f/optimum/onnxruntime/modeling_decoder.py#L255

kstavro:
Hi @Balladie @IlyasMoutawwakil, I am just back from my vacation and I noticed that the PR is still open. Do you require any assistance from my side for the implementation? If desired and/or preferred, I could find some time on the weekend to assist with the PR.

Thanks again for taking care of this so far :)

PR #2252, opened 2025-05-06
Add ONNX export support for Qwen3

Abdennacer-Badaoui:
# What does this PR do?

## Before submitting


optimum/exporters/onnx/utils.py
```diff
@@ -85,7 +85,7 @@
     "phi",
     "phi3",
     "qwen2",
-    "granite",
+    "qwen3" "granite",
```
IlyasMoutawwakil, line 88:
```suggestion
    "qwen3",
    "granite",
```

Issue #2257, opened 2025-05-12
Add support for `token-classification` with LLMs


danielkorat:
### Feature request

Hello,
We would greatly appreciate adding support for exporting with `token-classification` tasks for LLMs such as `qwen2`:
https://github.com/huggingface/optimum/blob/cf5c6ecdf62e48b0070da953ff5089b85bd4519f/optimum/exporters/tasks.py#L991

### Motivation

Many downstream tasks rely on `token-classification` For example, Hugging Face's implementations of Process Reward Models rely on `token-classification` modeling.
We require this capability for a project we are working on.

### Your contribution

Can review and assist with PR after it is created

PR #2260, opened 2025-05-12
adding token-classification task for qwen2

Abdennacer-Badaoui:
# What does this PR do?

Add token-classification task for qwen2 

Fix Issue: https://github.com/huggingface/optimum/issues/2257


docs/source/notebooks.md
```diff
@@ -1 +0,0 @@
-../../notebooks/README.md
```
echarlaix, line 1: When adding a notebook or any other change we don't want to add the same modification each time to both https://github.com/huggingface/optimum/blob/main/docs/source/notebooks.md and https://github.com/huggingface/optimum/blob/main/notebooks/README.md, would revert

optimum/exporters/tasks.py
```diff
@@ -994,6 +994,7 @@ class TasksManager:
             "text-generation",
             "text-generation-with-past",
             "text-classification",
+            "token-classification",
```
echarlaix, line 997: thanks for adding!

Issue #2266, opened 2025-05-14
TypeError: ORTModelForFeatureExtraction.forward() got an unexpected keyword argument 'return_dict'


yxtay:
### System Info

```shell
optimum[onnxruntime]==1.25.0
python==3.13
os: debian bookworm slim
```

### Who can help?

Geting the following error when using `SentenceTransformers` with `backend="onnx"`. It was working fine with `optimum[onnxruntime]==1.24.0`.
```
Traceback (most recent call last):
  File "/app/test.py", line 6, in <module>
    embeddings = model.encode(sentences)
  File "/app/.venv/lib/python3.13/site-packages/sentence_transformers/SentenceTransformer.py", line 685, in encode
    out_features = self.forward(features, **kwargs)
  File "/app/.venv/lib/python3.13/site-packages/sentence_transformers/SentenceTransformer.py", line 758, in forward
    input = module(input, **module_kwargs)
  File "/app/.venv/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1762, in _call_impl
    return forward_call(*args, **kwargs)
  File "/app/.venv/lib/python3.13/site-packages/sentence_transformers/models/Transformer.py", line 442, in forward
    outputs = self.auto_model(**trans_features, **kwargs, return_dict=True)
  File "/app/.venv/lib/python3.13/site-packages/optimum/modeling_base.py", line 99, in __call__
    return self.forward(*args, **kwargs)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
TypeError: ORTModelForFeatureExtraction.forward() got an unexpected keyword argument 'return_dict'
```

I think the commit that introduced the issue was f93964c1c45cdc77563e24feb48517c0b64adb6d by @Abdennacer-Badaoui, which removed `**kwargs` from `ORTModelForFeatureExtraction.forward()`. Will it be ok to add back `**kwargs`?

ONNX Runtime: @JingyaHuang, @echarlaix

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

[...]

PR #2267, opened 2025-05-14
Fix bug ORTModelForFeatureExtraction

Abdennacer-Badaoui:
# What does this PR do?

This PR updates the `ORTModelForFeatureExtraction`'s `forward `method to accept `**kwargs`, ensuring better compatibility with Hugging Face Transformers' conventions. In particular, it adds proper support for the `return_dict `argument: when explicitly passed, the output is returned as a plain dictionary instead of being wrapped in a `ModelOutput`. Additionally, any unused keyword arguments are logged with a warning to help users identify unsupported inputs while maintaining backward compatibility.

Fixes issue : https://github.com/huggingface/optimum/issues/2266

### DIFF
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -1072,7 +1072,8 @@ class ORTModelForFeatureExtraction(ORTModel):
         if kwargs:
             logger.warning(
                 f"{self.__class__.__name__} received {', '.join(kwargs.keys())}, but do not handle those arguments. "
-                "Please use `ORTModelForCustomTasks` if your model takes arbitrary or custom tensor arguments."
+                "Please use `ORTModelForCustomTasks` if your model takes/returns arbitrary or custom tensor inputs/outputs. "
+                "Or open an issue/PR in optimum repository (https://github.com/huggingface/optimum) if this argument needs to be supported in this class."
             )
 
         # Determine the tensor type from any available tensor input
--- a/optimum/onnxruntime/modeling_ort.py
+++ b/optimum/onnxruntime/modeling_ort.py
@@ -1076,11 +1076,10 @@ class ORTModelForFeatureExtraction(ORTModel):
         pixel_values: Optional[Union[torch.Tensor, np.ndarray]] = None,
         input_features: Optional[Union[torch.Tensor, np.ndarray]] = None,
         input_values: Optional[Union[torch.Tensor, np.ndarray]] = None,
+        *,
+        return_dict: bool = False,
         **kwargs,
     ):
-        # Handle return_dict from kwargs with default value False
-        return_dict = kwargs.pop("return_dict", False)
-
         # Warn about any unexpected kwargs using the helper method
         self._warn_on_unhandled_inputs(kwargs)
 
@@ -1136,7 +1135,7 @@ class ORTModelForFeatureExtraction(ORTModel):
                 last_hidden_state = next(iter(model_outputs.values()))
 
         if return_dict:
-            return model_outputs
+            return {"last_hidden_state": last_hidden_state}
 
         # converts output to namedtuple for pipelines post-processing
         return BaseModelOutput(last_hidden_state=last_hidden_state)
@@ -1251,11 +1250,10 @@ class ORTModelForMaskedLM(ORTModel):
         input_ids: Optional[Union[torch.Tensor, np.ndarray]] = None,
         attention_mask: Optional[Union[torch.Tensor, np.ndarray]] = None,
         token_type_ids: Optional[Union[torch.Tensor, np.ndarray]] = None,
+        *,
+        return_dict: bool = False,
         **kwargs,
     ):
-        # Handle return_dict from kwargs with default value False
-        return_dict = kwargs.pop("return_dict", False)
-
         # Warn about any unexpected kwargs using the helper method
         self._warn_on_unhandled_inputs(kwargs)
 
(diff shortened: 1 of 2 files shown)


optimum/onnxruntime/modeling_ort.py
```diff
@@ -1063,7 +1063,17 @@ def forward(
...
+        **kwargs,
     ):
+        # Handle return_dict from kwargs with default value False
+        return_dict = kwargs.pop("return_dict", False)
+
+        # Raise error for any unexpected kwargs
+        if kwargs:
+            raise ValueError(
+                f"{self.__class__.__name__} received {', '.join(kwargs.keys())}, but do not accept those arguments."
+            )
```
IlyasMoutawwakil, line 1075: hmm what about warning the user and directing them to use ORTModelForCustomTask if their model takes arbitrary tensor arguments.
```suggestion
            logger.warning(
                f"{self.__class__.__name__} received {', '.join(kwargs.keys())}, but do not handle those arguments. "
                "Please use `ORTModelForCustomTasks` if your model takes arbitrary or custom tensor arguments."
            )
```

optimum/onnxruntime/modeling_ort.py
```diff
@@ -1115,6 +1125,9 @@ def forward(
                 # TODO: This allows to support sentence-transformers models (sentence embedding), but is not validated.
                 last_hidden_state = next(iter(model_outputs.values()))
 
+        if return_dict:
+            return model_outputs
```
IlyasMoutawwakil, line 1129: can you please add this to other ORModelForxxx, it's strange that we never handled this argument 🤔

optimum/onnxruntime/modeling_ort.py
```diff
@@ -1230,6 +1253,12 @@ def forward(
         token_type_ids: Optional[Union[torch.Tensor, np.ndarray]] = None,
         **kwargs,
     ):
+        # Handle return_dict from kwargs with default value False
+        return_dict = kwargs.pop("return_dict", False)
```
daip-yxtay, line 1257: Is there any reason why `return_dict` is not added as an optional argument in the `forward` function instead so that it can be taken directly instead of retrieved from `kwargs`? Is it because the arguments must be `torch.Tensor | np.ndarray | None`?

```suggestion
        return_dict: bool = False,
        **kwargs,
    ):
```

yxtay, line 1257: Is there any reason why `return_dict` is not added as an optional argument in the `forward` function instead so that it can be taken directly instead of retrieved from `kwargs`? Is it because the arguments must be `torch.Tensor | np.ndarray | None`?

```suggestion
        return_dict: bool = False,
        **kwargs,
    ):
```

IlyasMoutawwakil: yes thanks ! @Abdennacer-Badaoui it should be possible to enforce it being a keword argument (can't be passed as a positional) with this "*," trick 👌

optimum/onnxruntime/modeling_ort.py
```diff
@@ -2274,6 +2402,10 @@ def forward(
             onnx_outputs = self.model.run(None, onnx_inputs)
             model_outputs = self._prepare_onnx_outputs(use_torch, onnx_outputs)
             reconstruction = model_outputs["reconstruction"]
+
+        if return_dict:
```
echarlaix, line 2406: would you mind adding a test for this ?

echarlaix (approved):
Thanks a lot for the fix @Abdennacer-Badaoui !

optimum/onnxruntime/modeling_ort.py
Abdennacer-Badaoui: Yes, of course.

tests/onnxruntime/test_modeling.py
```diff
@@ -2138,11 +2138,18 @@ def test_compare_to_transformers(self, model_arch):
...
-
+            self.assertIsInstance(onnx_outputs, BaseModelOutput)
             self.assertIn("last_hidden_state", onnx_outputs)
             self.assertIsInstance(onnx_outputs.last_hidden_state, self.TENSOR_ALIAS_TO_TYPE[input_type])
 
+            # Test return_dict=True
+            onnx_outputs_dict = onnx_model(**tokens, return_dict=True)
+            self.assertIsInstance(onnx_outputs_dict, dict)
+            self.assertIn("last_hidden_state", onnx_outputs_dict)
+            self.assertIsInstance(onnx_outputs_dict["last_hidden_state"], self.TENSOR_ALIAS_TO_TYPE[input_type])
```
IlyasMoutawwakil, line 2151: nice ! we can generalise this to other classes after the patch, for now this is enough to unblock sentence transformers

IlyasMoutawwakil (approved):

malteos:
Is this really fixing the underlying issue? With this PR, native Python dicts are returned. However, transformers uses ModelOutput, or?

At least for me I have this issue now: When using a sentence-transformers ONNX model, inference fails at this point: https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/models/Transformer.py#L449-L450

```
# outputs (with return_dict=True)
{'last_hidden_state': tensor([[[-3.2455e-01, -5.7767e-01, -1.2708e+00,  ...,  8.5406e-01,
           3.3558...00,
           1.3960e+00, -8.4762e-01]]])}
```

IlyasMoutawwakil:
oh wow ! so transformers doesn't return a dict when return_dict is True, and returns a tuple when it's False 😵

PR #2270, opened 2025-05-15
compatibility with transformers 4.52

echarlaix:


optimum/pipelines/pipelines_base.py
```diff
@@ -366,16 +406,22 @@ def pipeline(
         **kwargs,
     )
 
+    use_fast = kwargs.get(use_fast, "True")
     if tokenizer is None and load_tokenizer:
-        tokenizer = get_preprocessor(model_id)
+        tokenizer = AutoTokenizer.from_pretrained(model_id, use_fast=use_fast, _from_pipeline=task, **hub_kwargs)
```
IlyasMoutawwakil, line 411: is the `_from_pipeline` necessary here ? asking because it might be dropped from transformers at any point

IlyasMoutawwakil (commented):
Thanks Ella, looks good !
it seems the tests are still not passing with onnxruntime

optimum/pipelines/pipelines_base.py
echarlaix: removed in https://github.com/huggingface/optimum/pull/2270/commits/28bba3c3eda7d539130c7d6235f0690d5660565d

echarlaix:
thanks for the review @IlyasMoutawwakil ! the loading of encoder-decoder models should now be fixed since https://github.com/huggingface/optimum/pull/2270/commits/350c8f9327775fd3dc5ba5157dffc1c77fa9fab8

tests/onnxruntime/test_modeling.py
```diff
@@ -4969,10 +4971,10 @@ def test_pipeline_image_to_text(self, test_name: str, model_arch: str, use_cache
             "image-to-text",
             model=onnx_model,
             tokenizer=tokenizer,
-            feature_extractor=feature_extractor,
+            image_processor=feature_extractor,
```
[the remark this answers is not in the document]
IlyasMoutawwakil: thanks, only asked to correct the naming `image_processor=feature_extractor`

PR #2276, opened 2025-05-20
Fix and uniformize hub kwargs

IlyasMoutawwakil:
# What does this PR do?

In my last PR I must have changed some defaults by mistake (e.g. force_download for CausalLM)
maybe it makes sense to start using something like this for the hub kwargs to guarantee they are the same across the lib
https://github.com/huggingface/transformers/blob/46a4b7c909fab58355936e1c7109bb5e2e558267/src/transformers/models/glm4/modular_glm4.py#L63

## Before submitting


echarlaix (approved):
LGTM, thanks for the fix!

PR #2279, opened 2025-05-25
No more forcing separators

IlyasMoutawwakil:
# What does this PR do?

There's a lot of separators shenanigans we have to do every time we deal with model_types (or is it model-types 😅).
This PR removes that by using the names models were given in transformers.
Motivation for this PR is that in #2278 I was able to add the officially named "qwen3_moe" and hardest part was figuring out where it should be "qwen3-moe" and where "qwen3_moe" should be used.

## Before submitting


echarlaix (approved):
LGTM, thanks @IlyasMoutawwakil

echarlaix:
thanks a lot @IlyasMoutawwakil !! if that works for you we can merge after the next release to have some time to make sure this doesn't introduce any issue for other repos (optimum-neuron / optimum-executorch and others)

echarlaix:
cc @JingyaHuang let us know if that works for you (might have an impact on optimum-neuron)

PR #2282, opened 2025-05-29
Add flags to allow dynamo usage & provide export pathway for onnx opset >= 21

### LINKED ISSUE #2281: Need a pathway to convert HF pytorch models to onnx when opset >= 21 is required
### Feature request

Optimum exporter for pytorch to onnx needs to allow dynamo to be used. When onnx opset >= 21 is requested, optimum should choose dynamo for conversion since torchscript is deprecated for use beyond opset 20.

### Motivation

When requesting target opset as >= 21, conversion fails because torchscript doesn't support it. The error sites
```
OnnxExporterWarning: Exporting to ONNX opset version 21 is not supported. by 'torch.onnx.export()'. The highest opset version supported is 20. To use a newer opset version, consider 'torch.onnx.export(..., dynamo=True)'
...
...
    raise ValueError(f"Unsupported ONNX opset version: {value}")
ValueError: Unsupported ONNX opset version: 21
```

As a result, optimum exporter should allow dynamo usage since torchscript is deprecated for use beyond opset 20. When onnx opset >= 21 is requested, optimum should choose dynamo for conversion automatically.

### Your contribution

[PR 2282](https://github.com/huggingface/optimum/pull/2282)

kkhode:
# What does this PR do?

Enable flag for comparing pytorch vs ORT output of original vs coverted model
Enable flag to dump fx graph
Dump debug info into converted model location by default

Fixes #2281

## Before submitting

## Who can review?

For faster review, we strongly recommend you to ping the following people:

### DIFF
--- a/optimum/commands/export/onnx.py
+++ b/optimum/commands/export/onnx.py
@@ -177,7 +177,8 @@ def parse_args_onnx(parser):
 
     optional_group.add_argument(
         "--use-dynamo",
-        action="store_true",
+        type=bool,
+        default=None,
         help="Use dynamo instead of torchscript for onnx model export.",
     )
 
--- a/optimum/commands/export/onnx.py
+++ b/optimum/commands/export/onnx.py
@@ -175,6 +175,24 @@ def parse_args_onnx(parser):
         help="Enables onnxslim optimization.",
     )
 
+    optional_group.add_argument(
+        "--use-dynamo",
+        action="store_true",
+        help="Use dynamo instead of torchscript for onnx model export.",
+    )
+
+    optional_group.add_argument(
+        "--verify-accuracy",
+        action="store_true",
+        help="Run torch and ORT model inference and compare outputs.",
+    )
+
+    optional_group.add_argument(
+        "--debug-reports",
+        action="store_true",
+        help="Provide debug reports for the conversion process.",
+    )
+
     input_group = parser.add_argument_group(
         "Input shapes (if necessary, this allows to override the shapes of the input given to the ONNX exporter, that requires an example input)."
     )
--- a/optimum/exporters/onnx/__main__.py
+++ b/optimum/exporters/onnx/__main__.py
@@ -23,13 +23,8 @@ from transformers import AutoConfig, AutoTokenizer
 from transformers.utils import is_torch_available
 
 from ...commands.export.onnx import parse_args_onnx
+from ...utils.import_utils import is_transformers_version
 from ...utils import DEFAULT_DUMMY_SHAPES, logging
-from ...utils.import_utils import (
-    is_diffusers_available,
-    is_sentence_transformers_available,
-    is_timm_available,
-    is_transformers_version,
-)
 from ...utils.save_utils import maybe_load_preprocessors
 from ..tasks import TasksManager
 from ..utils import DisableCompileContextManager
@@ -83,7 +78,6 @@ def main_export(
     legacy: bool = False,
     no_dynamic_axes: bool = False,
     do_constant_folding: bool = True,
-    slim: bool = False,
     **kwargs_shapes,
 ):
     """
@@ -431,7 +425,6 @@ def main_export(
         task=task,
         use_subprocess=use_subprocess,
         do_constant_folding=do_constant_folding,
-        slim=slim,
         **kwargs_shapes,
     )
 
@@ -448,6 +441,10 @@ def main():
     input_shapes = {}
     for input_name in DEFAULT_DUMMY_SHAPES.keys():
         input_shapes[input_name] = getattr(args, input_name)
+
+    if args.use_dynamo == False and args.opset and args.opset >= 21:
+        warnings.warn('Support for ONNX opset >= 21 will only be supported via dynamo going forward. Switching to dynamo.', FutureWarning)
+        args.use_dynamo = True
 
     main_export(
         model_name_or_path=args.model,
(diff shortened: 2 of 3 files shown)


xenova:
Thanks for the PR! :) Upgrading to the new dynamo exporter is definitely something I'm in favour of. The exporter has become far more stable over time, but I do think it would be a good idea to hook it into our CI in some capacity. @echarlaix do you have any ideas for this?

optimum/commands/export/onnx.py
```diff
@@ -175,6 +175,24 @@ def parse_args_onnx(parser):
         help="Enables onnxslim optimization.",
     )
 
+    optional_group.add_argument(
+        "--use-dynamo",
```
justinchuby, line 179: Is this user facing? The export team from PyTorch recommended something like `--use-pt2` to disambiguate, because `dynamo` is really a legacy name in the onnx export api and is not really what the exporter uses to capture the graph. (It uses torch.export non-strict mode and aot autograd) to get the fx graph/exported program.

xadupre:
I tried something similar in #2219 but I realized there were many patches implemented to make torchscript work which did not play well with the new exporter.

kkhode:
Since unlike #2219, we're not forcing dynamo as default pipeline, this PR shouldn't be a breaking change. It just allows dynamo or higher onnx opset to be chosen by user that specifically ask for it.

@echarlaix Is it possible to run the checks and try to merge this?

optimum/exporters/onnx/__main__.py
```diff
@@ -449,6 +451,10 @@ def main():
     for input_name in DEFAULT_DUMMY_SHAPES.keys():
         input_shapes[input_name] = getattr(args, input_name)
 
+    if args.use_dynamo == False and args.opset and args.opset >= 21:
+        warnings.warn('Support for ONNX opset >= 21 will only be supported via dynamo going forward. Switching to dynamo.', FutureWarning)
+        args.use_dynamo = True
```
IlyasMoutawwakil, line 456: shouldn't the opset be set to 18+ when using dynamo ?

optimum/exporters/onnx/convert.py
```diff
@@ -573,24 +576,35 @@ def remap(value):
                 dynamic_axes=dynamix_axes,
                 do_constant_folding=do_constant_folding,
                 opset_version=opset,
+                dynamo=use_dynamo,
+                verify=verify_accuracy,
+                artifacts_dir = Path(output.as_posix()).parent,
+                report=debug_reports,
+                dump_exported_program=debug_reports,
```
IlyasMoutawwakil, line 583: what's the min torch version for these apis ?

optimum/exporters/onnx/__main__.py
justinchuby: @kkhode yes that would be ideal

kkhode: Technically, export torchscript pipeline doesn't throw an error unless requested opset >= 21. There is disparity to documentation and code.

At this location: https://github.com/pytorch/pytorch/blob/457dd79927db6ae7f1c53f36768b7ac93b61e507/torch/onnx/utils.py#L1396, I get the following error
> OnnxExporterWarning: Exporting to ONNX opset version 21 is not supported. by 'torch.onnx.export()'. The highest opset version supported is 20. To use a newer opset version, consider 'torch.onnx.export(..., dynamo=True)'.

So I wouldn't recommend force switching for 18+. Aim of the PR is not to break existing downstream pipelines unless absolutely required.

justinchuby: > The highest opset version supported is 20

The message was inaccurate. We don't really "support" opset>=18 in the old exporter

kkhode: Then shouldn't this be changed? https://github.com/pytorch/pytorch/blob/457dd79927db6ae7f1c53f36768b7ac93b61e507/torch/onnx/_constants.py#L8

justinchuby: 18-20 is in the partial support range where new ops were added, but old ops were not tested (and we know the reduce_*) ops don't work. I suppose we can just keep it as is as we don't intend to change the old exporter much. Starting from torch 2.8 a warning will be created whenever dynamo=True is not set.

kkhode: I ended up simply moving the logic to torch. @justinchuby I'll create a small PR with this change: https://github.com/kkhode/pytorch/commit/b6c6a5e0dad84a43002468631417d72edaa747db. Agree? If torch wants to facilitate the shift, it can do it there.

optimum/exporters/onnx/convert.py
IlyasMoutawwakil: then let's either guard/dispatch based on version or group these in a dynamo_kwargs dict that's populated based on torch version.

IlyasMoutawwakil:
we still need some testing for this to be merged, simply adding a workflow similar to https://github.com/huggingface/optimum/blob/main/.github/workflows/test_exporters_onnx_cli.yml and adding a test file similar `tests/exporters/onnx/test_export_cli.py`

kkhode:
> we still need some testing for this to be merged, simply adding a workflow similar to https://github.com/huggingface/optimum/blob/main/.github/workflows/test_exporters_onnx_cli.yml and adding a test file similar `tests/exporters/onnx/test_export_cli.py`

I can add a few test cases to test_export_cli.py so it can test the latest changes. Is that the recommendation? I don't know if another workflow/check needs to be added. After that checks need to be kicked off by a maintainer.

kkhode:
@IlyasMoutawwakil Just wanted to re-ask the question from above. I can add a few test cases to test_export_cli.py so it can test the latest changes. However, I'm not aware of all the work that'll go into creating a new workflow. So should I simply add a test file and it will be hooked up into a workflow by someone else?

PR #2288, opened 2025-06-13
Fix

echarlaix:
Add back deprecated `_from_transformers` method to not break inference for other subpackages which still uses it like optimum-intel

### DIFF
--- a/optimum/modeling_base.py
+++ b/optimum/modeling_base.py
@@ -316,11 +316,6 @@ class OptimizedModel(PreTrainedModel):
         )
 
     @classmethod
-    def _from_transformers(cls, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


optimum/modeling_base.py
```diff
@@ -403,7 +403,21 @@ def from_pretrained(
...
 
-        from_pretrained_method = cls._export if export else cls._from_pretrained
+        if export:
+            if hasattr(cls, "_export"):
+                from_pretrained_method = cls._export
+            elif hasattr(cls, "_from_transformers"):
+                logger.warning(
+                    "The `export` argument is set to `True`, but the class does not implement `_export` methods. "
+                    "Using the `_from_transformers` method instead."
+                )
```
IlyasMoutawwakil, line 413:
```suggestion
```
no need

PR #2297, opened 2025-06-18
Add task tflite register

echarlaix:
same as https://github.com/huggingface/optimum/pull/2291


echarlaix:
tflite is being deprecated https://github.com/huggingface/optimum/pull/2316, closing

PR #2298, opened 2025-06-18
 Move onnx and ort related code to optimum-onnx 

echarlaix:


noelportillo (approved):

setup.py
```diff
@@ -94,6 +94,7 @@
         "tensorflow>=2.4,<=2.12.1",
         "transformers>=4.36,<4.38",
     ],
+    "onnx": "optimum-onnx @ git+https://github.com/huggingface/optimum-onnx.git@mv-onnx",
```
IlyasMoutawwakil, line 97: the exporters and onnxruntime(-gpu) will also need to install optimum-onnx

.github/workflows/test_exporters_common.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install --upgrade pip
-          pip install --no-cache-dir torch==2.1.2 torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
```
IlyasMoutawwakil, line 38: removing this pin result in longer installation time, because we install latest cpu wheels and then in the next line the resolver uninstalls it to install torch 2.1.2 from pypi (with cuda wheels)

setup.py
echarlaix: good point added in https://github.com/huggingface/optimum/pull/2298/commits/8806db59850fda63d4a5e5aaaf12d4d64bf54db8

PR #2303, opened 2025-06-24
Adding DummyInputGenerator for VisualBert

Abdennacer-Badaoui:
# What does this PR do?

Add the necessary input names for VisualBERT to the `DummyVisionInputGenerator`.

PR #2306, opened 2025-06-25
Revert "Adding DummyInputGenerator for VisualBert"

echarlaix:
Reverts huggingface/optimum#2303 which break the optimum cli


echarlaix:
visual_seq_length and visual_embedding_dim needs to be added to https://github.com/huggingface/optimum/blob/eb187facea7a0ba1ffbce619f5efa174d03b1ffe/optimum/commands/export/onnx.py#L32 and https://github.com/huggingface/optimum/blob/eb187facea7a0ba1ffbce619f5efa174d03b1ffe/optimum/commands/export/tflite.py#L32 to fix the broken CLI, in the meantime reverting

PR #2310, opened 2025-06-27
Fix tflite cli

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting


echarlaix:
Thanks a lot for checking this @IlyasMoutawwakil ! Looks like optimum TFLite export usage is pretty low, wondering if we should either deprecate if completely in v2 or move everything to optimum-tflite, what do you think?

IlyasMoutawwakil:
for example https://github.com/MPolaris/onnx2tflite

Issue #2300, opened 2025-06-20
Support for EuroBERT models


antonioloison:
### Feature request

I would like to export and optimize the [EuroBERT models](https://huggingface.co/collections/EuroBERT/eurobert-67ceb6c01804878b1f7999c6).

Currently, it doesn't seem to be possible. When I run :

```python
from optimum.onnxruntime import ORTModelForSequenceClassification

onnx_model = ORTModelForSequenceClassification.from_pretrained(
    "EuroBERT/EuroBERT-210m",
    export=True,
    trust_remote_code=True,
)
```

Here is the output I got:
```
ValueError: Trying to export a eurobert model, that is a custom or unsupported architecture, but no custom onnx configuration was passed as `custom_onnx_configs`. Please refer to https://huggingface.co/docs/optimum/main/en/exporters/onnx/usage_guides/export_a_model#custom-export-of-transformers-models for an example on how to export custom models. Please open an issue at https://github.com/huggingface/optimum/issues if you would like the model type eurobert to be supported natively in the ONNX export.
```

Environment Specs:
- Python Version: 3.11.10
- Optimum Version: 1.26.1

Are you planning to support these models? 

### Motivation

[EuroBERT models](https://huggingface.co/collections/EuroBERT/eurobert-67ceb6c01804878b1f7999c6) are modern multilingual encoder models that work well when adapted to several multilingual tasks (classification, NER, retrieval...).

### Your contribution

I can try to add them if you are not planning to do it.

DWarez:
Ciao! 

I drafted a PR here https://github.com/huggingface/optimum/pull/2321.

Please read the first comment, maybe you can help me with that 😄

PR #2321, opened 2025-07-15
[Draft] Add Eurobert configuration for ONNX export

DWarez:
# What does this PR do?
Adding configuration for Eurobert ONNX export support

Fixes # [(2300)](https://github.com/huggingface/optimum/issues/2300)

## Before submitting

## Who can review?

### DIFF
--- a/optimum/exporters/onnx/model_configs.py
+++ b/optimum/exporters/onnx/model_configs.py
@@ -281,6 +281,19 @@ class FlaubertOnnxConfig(BertOnnxConfig):
     DEFAULT_ONNX_OPSET = 11
 
 
+@register_tasks_manager_onnx("eurobert", *COMMON_TEXT_TASKS)
+class EurobertOnnxConfig(BertOnnxConfig):
+    @property
+    def inputs(self) -> Dict[str, Dict[int, str]]:
+        dynamic_axis = {0: "batch_size", 1: "sequence_length"}
+
+        return {
+            "input_ids": dynamic_axis,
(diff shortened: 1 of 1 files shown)


DWarez:
I created a configuration to support Eurobert, but I'm still unsure about its correctness. When exporting the model, I'm getting some trace warnings:

```
/python3.11/site-packages/transformers/modeling_attn_mask_utils.py:122: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if (input_shape[-1] > 1 or self.sliding_window is not None) and self.is_causal:
```
```
python3.11/site-packages/transformers/modeling_attn_mask_utils.py:196: TracerWarning: torch.tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  inverted_mask = torch.tensor(1.0, dtype=dtype) - expanded_mask
```

As you can see they come from some modelling methods in the transformers library (which are invoked by the modelling eurobert scripts). 

I investigated a bit and to me it's quite clear why the warnings are getting triggered, but I'm unsure about their impacts on the usability and correctness of the exported model.

Any suggestion about how to test this would be appreciated, thanks.

IlyasMoutawwakil:
Hi ! thanks for the contribution, please move it to `optimum-onnx` as we are moving the entire onnx/ort integrations there.

PR #2325, opened 2025-07-23
Add import sys.exit to optimum_cli.py

rasapala:
Add import sys.exit to optimum_cli.py.
optimum_cli.exe when run with no parameters throws error missing name: exit.


rasapala:
Yes, it fails on newer python version like 3.12.10.

PR #2326, opened 2025-07-24
Fix batched inference/generation, position_ids creation, falcon alibi, gpt_bigcode multi-query,..

IlyasMoutawwakil:
# What does this PR do?

The batched inference issue and numerical mismatch has persisted in the ort class since very long, and even with https://github.com/huggingface/optimum/pull/1381, the position ids were only created in the generation tests instead of creating them inside the inference class, not to mention that for generation, `prepare_input_for_generation` takes care of generating position ids. So for simple forward pass, it was still an issue.

This PR is the result of a rabbit hole I went into when I realized that my decoder testing refactorization removed [batched the only batched generation check we had](https://github.com/huggingface/optimum/pull/2278/files#diff-2fa018fa815e73870a475fab33ee1bd9a16210bb81d5f6bf3c66a077e83cb75dL2378) 🥲. So I enabled batched inference/generation by default and to my surprise, all models that require position ids were failing batched inference/generation. [But simply installing transformers==4.52 and it was passing](https://github.com/huggingface/optimum/pull/2326/commits/b7bec5e4d8aa220c38fc4ba672b8dd5a165e0d70), so the problem is obviously in something that happened in the transformers 4.53 refactorization.

Starting from transformers 4.53 the modeling code uses boolean 4D masks, which are not "officially" supported by the torch onnx export (it's not really a "masked operation"), the boolean mask is simply converted to 0 and -inf filled tensor https://github.com/pytorch/pytorch/blob/f8fafdc7a6d260cea6c145643f4cf73631c81460/torch/onnx/symbolic_opset14.py#L187
This, in the case of padded batched inputs, results in the softmax returning nans, which pollutes the entire sequence logits (the entire padded sequences return nans as logits). This behavior used to be avoided by not calling `_unmask_unattended` in the case of onnx export. Instead of going that path again for the new masking methods (whch also results in small numerical mismatches), we fix this by patching the torch onnx exporter directly, and overloading graph it uses to replace `aten::scaled_dot_product_attention`.

A lot of fixes and changes, but hey, we were able to un-bloat the model patcher, get matching logits even for the mask tokens and get better support and testing of batched inference/generation across transformers versions.

Ps: I tested every version of transformers from 4.36 to 4.53. and also tested these changes on torch 2.1.

## Before submitting


optimum/exporters/onnx/config.py
```diff
@@ -94,13 +95,23 @@ def __init__(
...
             }
+
+        # Decoders based on GPT2 require a position_ids input to avoid generating wrong position_ids in the model itself:
+        # https://github.com/huggingface/transformers/blob/v4.33.1/src/transformers/models/gpt2/modeling_gpt2.py#L802
+        if (
+            not self.legacy
+            and self.task in ["text-generation", "feature-extraction"]
+            and self._config.model_type in MODEL_TYPES_REQUIRING_POSITION_IDS
+        ):
+            common_inputs["position_ids"] = {0: "batch_size", 1: "sequence_length"}
```
IlyasMoutawwakil, line 113: the idea here is to have a single source of truth for whether a model uses/requires position ids, because otherwise it is possible to inherit from a class that inherits from `TextDecoderWithPositionIdsOnnxConfig` class, without adding the model to MODEL_TYPES_REQUIRING_POSITION_IDS, which is confusing.

optimum/exporters/onnx/model_patcher.py
```diff
@@ -228,12 +219,61 @@ def onnx_compatible_linalg_norm(x, ord=2, dim=None, keepdim=False, *, dtype=None
...
+        args[5] = args[5].item()
+    elif "is_causal" in kwargs and isinstance(kwargs["is_causal"], torch.Tensor):
+        kwargs["is_causal"] = kwargs["is_causal"].item()
+
+    return original_scaled_dot_product_attention(*args, **kwargs)
+
+
+# No-op bfloat16 casting to avoid issues with legacy ONNX export which cast to complex128
+def noop_bfloat16_casting(self):
+    return self
```
IlyasMoutawwakil, line 259: this is for falcon with alibi

optimum/onnxruntime/modeling_decoder.py
```diff
@@ -396,37 +426,48 @@ def prepare_inputs_for_generation(self, *args, **kwargs):
...
         input_ids,
-        attention_mask=None,
         past_key_values=None,
-        token_type_ids=None,
+        attention_mask=None,
+        inputs_embeds=None,
+        cache_position=None,
         position_ids=None,
-        use_cache=None,
+        use_cache=True,
```
IlyasMoutawwakil, line 438:
```suggestion
        use_cache=None,
```

tests/onnxruntime/test_diffusion.py
```diff
@@ -82,12 +82,18 @@ def generate_prompts(batch_size=1):
...
-            "https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main"
-            "/in_paint/overture-creations-5sI6fQgYIuo.png"
-        ).resize((width, height))
+        global IMAGE
+        if IMAGE is None:
+            # Load a sample image from the Hugging Face Hub
+            IMAGE = load_image(
+                "https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main/in_paint/overture-creations-5sI6fQgYIuo.png"
+            )
+        image = IMAGE.resize((width, height))
```
IlyasMoutawwakil, line 96: to avoid load_image failing because of multiple calls, the method itself is very error prone when many calls are made to the same url in parallel.

tests/onnxruntime/test_decoder.py
```diff
@@ -140,6 +195,19 @@ def test_find_untested_architectures(self):
...
+            model_type_requires_position_ids = model_type in MODEL_TYPES_REQUIRING_POSITION_IDS
+            onnx_config_class = TasksManager._SUPPORTED_MODEL_TYPE[model_type]["onnx"][self.TASK].func
+            onnx_config_class_with_position_ids = issubclass(onnx_config_class, TextDecoderWithPositionIdsOnnxConfig)
+
+            if model_type_requires_position_ids ^ onnx_config_class_with_position_ids:
+                raise ValueError(
+                    f"Model type {model_type} {'requires' if model_type_requires_position_ids else 'does not require'} position ids, "
+                    f"but the ONNX config class {onnx_config_class} {'is' if onnx_config_class_with_position_ids else 'is not'} "
+                    f"subclassed from TextDecoderWithPositionIdsOnnxConfig.\n"
+                )
```
IlyasMoutawwakil, line 209: so that they're always in sync (found a couple models that weren't added)

optimum/exporters/onnx/model_configs.py
```diff
@@ -634,10 +608,8 @@ def __init__(
     def inputs(self) -> Dict[str, Dict[int, str]]:
         common_inputs = super().inputs
 
-        if not self.legacy and not self._config.alibi and self.task in ["text-generation", "feature-extraction"]:
-            # When alibi is used, position_ids are not used in Falcon.
-            # Reference: https://github.com/huggingface/transformers/blob/v4.34.0/src/transformers/models/falcon/modeling_falcon.py#L1116
-            common_inputs["position_ids"] = {0: "batch_size", 1: "sequence_length"}
+        if self._config.alibi:
+            common_inputs.pop("position_ids", None)
```
echarlaix, line 612: not very important but here why can we remove the check for legacy models ? (I don't think legacy models are tested very well if not at all, also and we should definitely remove their support in v2 so not very important in all cases)

echarlaix (approved):
Huuuge work, thank you so much @IlyasMoutawwakil :fire::fire:

optimum/exporters/onnx/model_configs.py
IlyasMoutawwakil: legacy is tested in https://github.com/huggingface/optimum/blob/main/tests/onnxruntime/test_decoder.py#L635, we compare the legacy+non-merged with merged

echarlaix: ok very clear, thanks @IlyasMoutawwakil !!

PR #2335, opened 2025-08-07
[ONNX, OpenVINO]Update the dummy generator for Llama

openvino-dev-samples:
To support minicpm4v
https://huggingface.co/openbmb/MiniCPM-V-4/blob/main/config.json#L19


openvino-dev-samples:
@IlyasMoutawwakil could you help to check ?

PR #2336, opened 2025-08-08
Fix gpt_bigcode input generator for transformers 4.54

echarlaix:
Following https://github.com/huggingface/transformers/pull/38635/ so starting transformers v4.54 the pkv from gpt_bigcode will have size : [batch_size, [kv_heads](
https://github.com/huggingface/transformers/blob/v4.54.0/src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py#L134), seq_length, [head_dim](
https://github.com/huggingface/transformers/blob/v4.54.0/src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py#L133)], adapting `GPTBigCodeDummyPastKeyValuesGenerator` to match this modification


optimum/utils/input_generators.py
```diff
@@ -1139,27 +1139,42 @@ def __init__(
...
+                self.hidden_size // self.num_attention_heads,
             )
-            return [
-                self.random_float_tensor(shape, framework=framework, dtype=float_dtype) for _ in range(self.num_layers)
+            pkv = [
+            (
+                self.random_float_tensor(shape, framework=framework, dtype=float_dtype),
+                self.random_float_tensor(shape, framework=framework, dtype=float_dtype),
+            )
+            for _ in range(self.num_layers)
```
echarlaix, line 1173: resulting from bigcode modeling modifications from https://github.com/huggingface/transformers/pull/38635/files

PR #2348, opened 2025-09-04
Remove neuronx extra from setup.py

JingyaHuang:
Install with Optimum extra is not recommended for neuron for a while. Remove the extra to avoid confusion.


IlyasMoutawwakil:
I was actually wondering about this as well !
I wrote a long rant about optimum "extras"/"subpackages" and their removal as of optimum v2 😭

PR #2361, opened 2025-09-26
Native namespace packages (PEP 420)

IlyasMoutawwakil:
# What does this PR do?

This PR proposes the use of [native namespace packages (PEP 420)](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages) which only requires us to not have an __init__.py file in the namespace directories. I believe this is a very simple rule to follow and allows us to have all benefits of namespace packages while also supporting working in editable mode in multiple distributions.

It also updates how subcommands are loaded following the above mentioned changes.

As a proof of work, I made a couple changes (simply avoiding importing from namespaces) in optimum-onnx, and with that we can now test both in editable mode, cli works in editable mode as well.

## Before submitting

### DIFF
--- a/optimum/commands/optimum_cli.py
+++ b/optimum/commands/optimum_cli.py
@@ -111,57 +111,48 @@ def load_optimum_namespace_cli_commands() -> (
         subclass of `BaseOptimumCLICommand` or a `CommandInfo`. The second element corresponds to the parent command,
         where `None` means that the parent command is the root CLI command.
     """
-    command_registration_files = []
-    for dist in importlib.metadata.distributions():
-        if dist.metadata["Name"] is None:
-            continue
-        if dist.metadata["Name"] == "optimum-benchmark":
-            continue
-        if not dist.metadata["Name"].startswith("optimum"):
-            # it might be better (and more secure ?) to use an explicit list of optimum subpackages here
-            continue
 
-        if dist.metadata["Name"] == "optimum":
-            # optimum can't have an __init__.py file, so we use optimum.pipelines instead
-            dist_name = "optimum.pipelines"
-        else:
-            dist_name = dist.metadata["Name"].replace("-", ".")  # optimum.onnx
+    # Check if the commands.register namespace exists in any installed subpackage
+    commands_register_namespace = "optimum.commands.register"
+    commands_register_spec = importlib.util.find_spec(commands_register_namespace)
+    if commands_register_spec is None or commands_register_spec.submodule_search_locations is None:
+        # no installed subpackage is using the namespace optimum.commands.register
+        return []
 
-        dist_module = importlib.import_module(dist_name)  #  optimum.onnx
-        dist_module_path = Path(dist_module.__file__).parent.parent  # optimum/ (in optimum/onnx/)
-        commands_register_path = dist_module_path / "commands" / "register"  # optimum/commands/register
-
-        if not commands_register_path.is_dir():
-            # if distribution does not register any commands
-            continue
-
-        for file in commands_register_path.iterdir():
-            if file.name == "__init__.py":
-                # Following PEP 420, a namespace should not contain an __init__.py file.
-                raise ValueError(
-                    "The namespace package optimum.commands.register should not contain an `__init__.py` file."
-                )
-            elif file.suffix == ".py":
-                command_registration_files.append(file.stem)
-
+    # Find all registration files and load the commands to register
     commands_to_register = []
-    for registration_file in command_registration_files:
-        submodule = importlib.import_module(f"optimum.commands.register.{registration_file}")
-        commands_to_register_in_module = getattr(submodule, "REGISTER_COMMANDS", [])
-        for command_idx, command in enumerate(commands_to_register_in_module):
-            if isinstance(command, tuple):
-                command_or_command_info, parent_command_cls = command
-            else:
-                command_or_command_info = command
-                parent_command_cls = None
+    for register_path in commands_register_spec.submodule_search_locations:
+        register_path = Path(register_path)
+        if not register_path.is_dir():
+            # skip non-directory paths
+            continue
 
-            if not isinstance(command_or_command_info, CommandInfo) and not issubclass(
-                command_or_command_info, BaseOptimumCLICommand
-            ):
-                raise ValueError(
-                    f"The command at index {command_idx} in the `commands.register` module of {dist_name} is not of the right type: {type(command_or_command_info)}."
-                )
-            commands_to_register.append((command_or_command_info, parent_command_cls))
+        # Look for python files
+        for register_file in register_path.iterdir():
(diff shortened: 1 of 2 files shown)


IlyasMoutawwakil:
@regisss no not really it's just for readability (I can revert it and do it in a separate PR), 

The only condition here is that directories which we want to be extensible / to be a namespaces e.g. optimum, optimum.exporters, optimum.commands, etc should not have an `__init__.py` file, and that alone will tell the package manager that they are namespaces and not modules. It does introduce a breaking change, so for example `from optimum.exporters import TasksManager` will not work because `optimum.exporters` won't have its init file anymore, but imo that's a fair price to get things working +/- out of the box and following the "native" way of doing namespaces.

optimum/commands/optimum_cli.py
```diff
@@ -17,58 +17,17 @@
...
 logger = logging.get_logger()
 
 # The table below contains the optimum-cli root subcommands provided by the optimum package
 OPTIMUM_CLI_ROOT_SUBCOMMANDS = [ExportCommand, EnvironmentCommand]
 
-# The table below is dynamically populated when loading subpackages
-_OPTIMUM_CLI_SUBCOMMANDS = []
-
-
-def optimum_cli_subcommand(parent_command: Optional[Type[BaseOptimumCLICommand]] = None):
```
IlyasMoutawwakil, line 36: left only one way for command registration, wdyt ?

regisss:
> The only condition here is that directories which we want to be extensible / to be a namespaces e.g. optimum, optimum.exporters, optimum.commands, etc should not have an `__init__.py` file, and that alone will tell the package manager that they are namespaces and not modules. It does introduce a breaking change, so for example `from optimum.exporters import TasksManager` will not work because `optimum.exporters` won't have its init file anymore, you would need to do `from optimum.exporters.tasks import TasksManager`, but imo that's a fair price to get things working out of the box (e.g editable mode in both) and following the "native" way of managing namespace packages.

So that means that the same should be done for Optimum subpackages? E.g. this [__init__.py](https://github.com/huggingface/optimum-habana/blob/main/optimum/habana/__init__.py) in optimum-habana should be removed?

optimum/commands/optimum_cli.py
regisss: Using `load_optimum_namespace_cli_commands ` defined below right?

optimum/commands/optimum_cli.py
```diff
@@ -105,34 +64,57 @@ def resolve_command_to_command_instance(
...
+        if dist.metadata["Name"] == "optimum-benchmark":
+            continue
+        if not dist.metadata["Name"].startswith("optimum"):
+            # it might be better (and more secure ?) to use an explicit list of optimum subpackages here
+            continue
+
+        if dist.metadata["Name"] == "optimum":
+            # NOTE: optimum no longer registers commands in the main package, we do this for testing purposes only
+            # optimum can't have an __init__.py file, so we use optimum.pipelines instead
+            dist_name = "optimum.pipelines"
```
echarlaix, line 92: why do we need this ?

echarlaix (approved):
Great work, thanks a lot @IlyasMoutawwakil !

optimum/commands/optimum_cli.py
IlyasMoutawwakil: there's a test in tests/cli that puts a register_custom_command.py in optimum/commands/register and for it to work we need to the location of optimum but since the package sees 'optimum' as a namespace it doesn't really have a location (it has multiple locations, one in every distribution) so we use optimum.pipelines.
this hack and the test i'm talking about are not really needed at this point, because now we do test command registration by installing optimum and optimum-onnx. I can remove both the hack and the custom command test wdyt ?

optimum/commands/optimum_cli.py
IlyasMoutawwakil: it is used in optimum-quanto and optimum-nvidia

IlyasMoutawwakil: @dacorvo do you prefer we keep this mechanism ?

dacorvo: FYI, the legacy convention does **not** work with editable mode for subpackages, because it expects the aforementioned files to be present in a path rooted under `sites-packages`. I also suspect this legacy mechanism is why the `optimum-cli` is so slow to start (sometimes tenth of seconds to just display the help).

IlyasMoutawwakil: > not require files to be installed at a specific path.

I think this was not a problem in the registration mechanism but rather in how the "namespacing" of optimum was implemented. with native namespaces, everything works out of the box, as long as we follow pep 420 (no init in namespaces, ie. optimum. optimum.exporters. optimum.commands.).

dacorvo:
> > I disagree with removing the dynamic registration mechanism implemented in subpackages.py, as it will break the packages that were using it.
> 
> reverted the changes to that part, this PR now only concerns namespace packages that also use the optimum.commands registration mechanism.
> 
Thanks
> > First, I am not convinced that with that solution the optimum-cli extension mechanism works in editable mode, as in the CI the subpackages are installed twice: first in normal then in editable mode. This means that when using the optimum-cli, the files under sites-packages for optimum commands will be used, and not the ones under the source tree (this is exactly the issue we are having in optimum-neuron for instance).
> 
> When you install a package in editable mode after having installed it statically, the package under site-packages is removed. I added an explicit uninstallation command `pip uninstall optimum optimum-onnx -y` in my last commit to demonstrate that. We can also go as far as creating two clean envs in the CI if you are still concerned. You can also create a clean env locally and install optimum-onnx and optimum from this branch, both in editable mode and test it. I have opened this PR after having made sure that it works.
>
With these explanations, I understand it should work
> > Second, with that convention of "sharing" a path (optimum.commands), it is unknown what happens if you update optimum after having installed your package. We had users reporting that the neuron commands were not registered anymore.
> > If you don't want to keep subpackage.py, which I still think was an improvement over the solution implemented here, then consider using python importlib.metadata.entry_points, which can be declared in each subpackage pyproject.toml and discovered dynamically by optimum.
> 
> The convention of "sharing" (optimum.commands), is the definition of namespace packages, the same way they share optimum.exporters, I'm not sure why one is okay with you (sharing optimum.exporters) but not the other (optimum.commands). I'm confused because even in optimum-neuron, the registration mechanism used is through optimum.commands which is what we are fixing in this PR. Why haven't optimum-neuron switched to using `subpackages.py` ?
I am not at all at ease with using namespaces to share code between packages, either `exporters` and `commands`. To be honest I think the python namespaces are not a good fit for what we are doing in optimum-xxx packages, ie just extend exporters and commands: to me this looks very much like a plugin/backend use case. The reason why we did not switch to `subpackages.py` is lazyness 🫣.

PR #2398, opened 2025-12-23
Fix issue with c4 dataset creation

nikita-savelyevv:
# What does this PR do?

Condition needs to be updated because if `enc.input_ids.shape[1] == seqlen` then later the right random bound `enc.input_ids.shape[1] - seqlen - 1` equals `-1` which leads to an error.

## Before submitting


nikita-savelyevv:
@echarlaix @IlyasMoutawwakil please take a look, thanks!

PR #2407, opened 2026-02-09
Fix get_wikitext2 tokenization bug causing sequence length warning

### LINKED ISSUE #2020: get_wikitext2 has bug
### System Info

```shell
optimum version 1.21.4 (latest)
# Use the official Python image from the Docker Hub
FROM public.ecr.aws/docker/library/python:3.10-slim
```

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

```python
from optimum.gptq.data import get_wikitext2

tokenizer = AutoTokenizer.from_pretrained("TinyLlama/TinyLlama-1.1B-Chat-v1.0")
get_wikitext2(tokenizer=tokenizer, nsamples=128, seqlen=32, split="train")
```
Produce warning:

Token indices sequence length is longer than the specified maximum sequence length for this model (73218 > 2048). Running this sequence through the model will result in indexing errors

### Expected behavior

This is proposed fix:

```python


def get_wikitext2(tokenizer: Any, seqlen: int, nsamples: int, split: str = "train"):
    if split == "train":
        data = load_dataset("wikitext", "wikitext-2-raw-v1", split="train")
    elif split == "validation":
        data = load_dataset("wikitext", "wikitext-2-raw-v1", split="test")
    ## length of 288059 should be enough
    #text = "".join([" \n" if s == "" else s for s in data["text"][:1000]])

    dataset = []
    for _ in range(nsamples):
        while True:
            i = random.randint(0, len(data) - 1)
            text = data[i]["text"]
            if len(tokenizer.tokenize(text)) >= seqlen:
                enc = tokenizer(text, return_tensors="pt")
                break
        i = random.randint(0, enc.input_ids.shape[1] - seqlen - 1)
        j = i + seqlen
        inp = enc.input_ids[:, i:j]
        attention_mask = torch.ones_like(inp)
        dataset.append({"input_ids": inp, "attention_mask": attention_mask})
    return dataset
```    

This is inline with get_c4 and get_c4_new.

No warning is produced.

Mr-Neutr0n:
## Summary

Fixes #2020

`get_wikitext2` was concatenating up to 1000 dataset entries into a single string and tokenizing it all at once. This produced a 73K+ token sequence that exceeds the model's maximum sequence length, triggering the warning:

> Token indices sequence length is longer than the specified maximum sequence length for this model (73218 > 2048). Running this sequence through the model will result in indexing errors

This fix changes the function to tokenize individual samples instead, consistent with how `get_c4` and `get_c4_new` already work. Each sample is tokenized separately and retried with a different random sample if it is shorter than the requested `seqlen`.

## Changes

- Removed concatenation of 1000 text entries into a single string
- Removed bulk tokenization of the entire concatenated text
- Added per-sample tokenization with retry loop (matching `get_c4` / `get_c4_new` pattern)

## Test plan

- [x] Verify `get_wikitext2(tokenizer, nsamples=128, seqlen=32, split="train")` no longer produces the sequence length warning
- [ ] Verify the returned dataset has the correct number of samples and shape
- [ ] Verify quantization workflows using wikitext2 dataset still work correctly

### DIFF
--- a/optimum/gptq/data.py
+++ b/optimum/gptq/data.py
@@ -125,12 +125,14 @@ def get_wikitext2(tokenizer: Any, seqlen: int, nsamples: int, split: str = "trai
         data = load_dataset("wikitext", "wikitext-2-raw-v1", split="train")
     elif split == "validation":
         data = load_dataset("wikitext", "wikitext-2-raw-v1", split="test")
-    # length of 288059 should be enough
-    text = "".join([" \n" if s == "" else s for s in data["text"][:1000]])
 
-    enc = tokenizer(text, return_tensors="pt")
     dataset = []
     for _ in range(nsamples):
+        while True:
+            i = random.randint(0, len(data) - 1)
+            enc = tokenizer(data[i]["text"], return_tensors="pt")
+            if enc.input_ids.shape[1] >= seqlen:
+                break
         i = random.randint(0, enc.input_ids.shape[1] - seqlen - 1)
         j = i + seqlen
         inp = enc.input_ids[:, i:j]


Mr-Neutr0n:
Friendly bump! Let me know if there's anything I should update or improve to help move this forward.

PR #2408, opened 2026-02-17
Transformers v5

echarlaix:
AutoModelForVision2Seq deprecated and removed in Transformers v5

### DIFF
--- a/tests/pipelines/test_pipelines.py
+++ b/tests/pipelines/test_pipelines.py
@@ -17,12 +17,14 @@ import unittest
 from typing import Any, Dict
 
 import numpy as np
+import pytest
 from huggingface_hub.constants import HF_HUB_CACHE
 from PIL import Image
 from transformers import AutoTokenizer
 from transformers.pipelines import Pipeline
 
 from optimum.pipelines import pipeline as optimum_pipeline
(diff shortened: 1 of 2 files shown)


tests/pipelines/test_pipelines.py
```diff
@@ -192,6 +194,10 @@ def test_image_segmentation_pipeline(self):
             self.assertIn("score", result[0])
             self.assertIn("mask", result[0])
 
+    @pytest.mark.skipif(
+        is_transformers_version(">=", "5"),
+        reason="requires transformers < v5 since image-to-text pipelines is deprecated",
+    )
     def test_image_to_text_pipeline(self):
         """Test image to text ORT pipeline"""
         pipe = optimum_pipeline(task="image-to-text", accelerator="ort")
```
IlyasMoutawwakil, line 203: maybe we can use the image-text-to-text pipeline ?

IlyasMoutawwakil: hmm yeh maybe leave to abother pr as we don't really support that pipeline class here

IlyasMoutawwakil (approved):
LGTM ! thanks for the fix !

PR #2416, opened 2026-03-25
fix gptq quantization condition

jiqing-feng:
Same as https://github.com/huggingface/transformers/pull/44588. The quantization only works for original nn.Linear module, subclass has custom forward so quantized layer cannot handle it.

### DIFF
--- a/optimum/gptq/quantizer.py
+++ b/optimum/gptq/quantizer.py
@@ -310,7 +310,7 @@ class GPTQQuantizer(object):
             if name1 in names:
                 device = get_device(layer)
                 delattr(module, attr)
-                if isinstance(layer, nn.Linear):
+                if type(layer) is nn.Linear:
                     in_features = layer.in_features
                     out_features = layer.out_features
                 elif isinstance(layer, nn.Conv2d):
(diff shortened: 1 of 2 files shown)


jiqing-feng:
Hi @SunMarc . Please also review this PR. Thanks!

cc @Qubitium

Qubitium:
@jiqing-feng @SunMarc  Looks good to me for the short-term. 

Long term, since we have lack of information, we really don't know if this object which inherits nn.Linear but is not exactly nn.Linear is truely not-quantizable. For example, if we have a module that overrides nn.Linear and only wraps forward and the code inside just move tensors from disk to gpu pre-forward, and then after fwd, move the tensor back to disk, it would be black-listed by this logic but it is actually qualifiable for quantization. 

Like my comment in https://github.com/huggingface/transformers/pull/44588#issuecomment-4124453666, in the future, we need much more information to better decide.

In the current space with lack of info, any decision we make is going to be incomplete and will either target too wide or too narrow.

Qubitium (approved):

jiqing-feng:
Hi @IlyasMoutawwakil . Would you please review this PR? Thanks!

SunMarc (commented):
Thanks !

jiqing-feng:
Hi @SunMarc . Would you please help to merge the PR? Thanks!

PR #2445, opened 2026-06-03
fix tests preprocessor loading

echarlaix:


.github/workflows/test_fx_optimization.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
+          pip install "transformers<5"
```
echarlaix, line 38: torch.fx support removed in transformers v5 https://github.com/huggingface/transformers/commit/75da795d8f93296ae226706f7150bee2479b5bab

PR #2449, opened 2026-06-12
Fix #2407: Fix get_wikitext2 tokenization bug causing sequence length warning

Mr-Neutr0n:
Fixes #2407

Changed `get_wikitext2` in `optimum/gptq/data.py:120-141` to tokenize individual samples with a retry loop (matching `get_c4`/`get_c4_new`) instead of concatenating 1000 entries and bulk-tokenizing, which produced sequences exceeding the model's max length.

Local test infra unavailable in CI sandbox.

---
This change was prepared with AI assistance under human direction and review.


qgallouedec:
This contribution comes from a profile that opens a massive number of pull requests generated purely by AI; I'm closing it.

PR #2458, opened 2026-07-06
transformers v5.5

echarlaix:
TasksManager._infer_library_from_model_name_or_path infers the library that will be used to load the model by creating an instance of PretrainedConfig, since transformers 5.5 this will result in an error for models if their config still uses the legacy rope_scaling + rope_theta keys (for example openai/gpt-oss-20b))

Fix:  AttributeError: 'PreTrainedConfig' object has no attribute 'max_position_embeddings'


regisss (approved):
LGTM!

PR #2467, opened 2026-07-22
Fix dummy input docstring typos

cupkk:
# What does this PR do?

Fix four `dummpy inputs` typos in `GENERATE_DUMMY_DOCSTRING`. This template is attached to `ExporterConfig.generate_dummy_inputs`, so the misspellings surface in its generated API documentation.

This is a documentation-only change with no runtime behavior or dependency changes.

## Validation

- Parsed `optimum/exporters/base.py` with Python `ast` and compiled the source.
- Extracted the generated docstring and checked that `dummpy` is absent and the corrected parameter descriptions are present.
- Ran `git diff --check`.

## Before submitting

- [x] This PR fixes a typo or improves the docs.
- [x] No documentation or runtime tests need to be added for this text-only correction.


echarlaix (approved):
thanks for the fix!

PR #2469, opened 2026-07-28
fix(gptq): delegate native GPTQModel loads

ZX-ModelCloud:
# What does this PR do?

  This PR allows Optimum to delegate native GPTQModel checkpoint preparation and post-initialization to GPTQModel when its optional loading bridge is available.

  The existing GPTQ loading path discovers linear modules from the model structure and replaces them using a globally selected `QuantLinear` implementation. Native GPTQModel checkpoints may instead contain:

  - A mix of dense and quantized modules.
  - Per-module dynamic quantization settings.
  - Model-specific structural conversions.
  - Quantized modules determined by the serialized weight manifest.

  Replacing modules outside that manifest can incorrectly convert an intentionally dense layer or select a kernel that does not support its shape.

  This PR:

  - Feature-detects `hf_gptqmodel_prepare_model_for_load` and `hf_gptqmodel_post_init_for_load`.
  - Delegates model conversion to GPTQModel when it recognizes a native checkpoint.
  - Preserves the returned quantization configuration and selected kernel.
  - Runs GPTQModel post-initialization after checkpoint tensors are loaded.
  - Keeps the existing Optimum loading path unchanged when the bridge is unavailable or does not handle the checkpoint.

  There is no new hard dependency. GPTQModel versions that do not expose the bridge APIs continue using the existing Optimum behavior.

  ## Testing

  - `pytest -q tests/gptq/test_quantization.py -k "GPTQNativeLoadBridgeTest or GPTQPostInitTest"`
    - all passed
  - `ruff check optimum/gptq/quantizer.py tests/gptq/test_quantization.py`
  - `python -m py_compile optimum/gptq/quantizer.py tests/gptq/test_quantization.py`

  ## Before submitting

  - [x] Did you make sure to update the documentation with your changes? No user-facing documentation change is required.

  ## Who can review?

  GPTQ and quantization: @SunMarc, @IlyasMoutawwakil

### DIFF
--- a/optimum/gptq/quantizer.py
+++ b/optimum/gptq/quantizer.py
@@ -46,13 +46,29 @@ if is_accelerate_available():
     )
     from accelerate.hooks import remove_hook_from_module
 
+# The bridge is optional for GPTQModel releases that predate this API.
+hf_gptqmodel_post_init_for_load = None
+hf_gptqmodel_prepare_model_for_load = None
+
 if is_gptqmodel_available():
     from gptqmodel import BACKEND, QuantizeConfig
     from gptqmodel.quantization import FORMAT, GPTQ, METHOD
     from gptqmodel.utils.importer import hf_select_quant_linear_v2
-    from gptqmodel.utils.model import hf_convert_gptq_v1_to_v2_format, hf_convert_gptq_v2_to_v1_format
+    from gptqmodel.utils.model import (
+        hf_convert_gptq_v1_to_v2_format,
+        hf_convert_gptq_v2_to_v1_format,
+    )
     from gptqmodel.utils.model import hf_gptqmodel_post_init as gptq_post_init
     from gptqmodel.version import __version__ as gptqmodel_version
+
+    try:
+        from gptqmodel.utils.model import (
+            hf_gptqmodel_post_init_for_load,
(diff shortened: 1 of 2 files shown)


optimum/gptq/quantizer.py
```diff
@@ -46,14 +46,30 @@
     )
     from accelerate.hooks import remove_hook_from_module
 
+# The bridge is optional for GPTQModel releases that predate this API.
+hf_gptqmodel_post_init_for_load = None
+hf_gptqmodel_prepare_model_for_load = None
+
```
Qubitium, line 52: Rename for simplicity.

```py
 _gptqmodel_load_post_init
 _gptqmodel_load_prepare_model
````

Also we need to fix this mult-model loading. Right now this code is a global var, we need it per model so two models loading in the same script will have unique states.

Qubitium (changes requested):

Qubitium (changes requested):
@ZX-ModelCloud  Unit test is testing pr change but actually not testing why we want this PR upstreamed to transformers. For example, you need write a specific test using hf/modelcloud hosted small quantied moe model with dynamic bit/groups gptq and test that it only loads correctly with our delegate. 

This appears to be the last change I see before we can remove the wip tag.

PR #2470, opened 2026-07-30
Fix offline loading for huggingface_hub >= 1.22

echarlaix:
`huggingface_hub` >= 1.22 made `snapshot_download` raise `IncompleteSnapshotError` instead of silently returning a partial folder when the local cache is missing files.

https://github.com/huggingface/huggingface_hub/blob/v1.26.0/src/huggingface_hub/_snapshot_download.py#L576

`TasksManager.get_model_files` relies on `snapshot_download` as its offline fallback, so this surfaced as failures for any repo where only part of the tree was ever cached locally. This is expected for diffusers and encoder/decoder models for openvino as described in https://github.com/huggingface/optimum-intel/pull/1855

### DIFF
--- a/optimum/exporters/tasks.py
+++ b/optimum/exporters/tasks.py
@@ -35,7 +35,7 @@ from ..utils.logging import get_logger
 
 
 try:
-    # only raised by `snapshot_download` since huggingface_hub v1.22.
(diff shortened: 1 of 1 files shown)


regisss (approved):
LGTM!

PR #2491, opened 2026-09-10
Add XLM to NormalizedConfigManager

nicolasdmolina:
# What does this PR do?

`NormalizedConfigManager` is the current equivalent of `ORTConfigManager`. XLM is on the #351 list and already has `XLMOnnxConfig` in optimum-onnx, but it is still missing from `_conf`, so ONNX Runtime optimization cannot resolve its hidden size or attention-head count.

XLM does not use the BERT-style names. In `XLMConfig` they are:

- hidden size: `emb_dim`
- attention heads: `n_heads`
- layers: `n_layers`

This PR registers only `xlm`, mapping those names with `NormalizedTextConfig.with_args` the same way DistilBERT and MT5 do, and adds an offline regression test for the three normalized attributes. It does not download weights or touch other model types.

Addresses the XLM subtask in #351.

## Before submitting

## Validation
- `python -m pytest tests/utils/test_normalized_config.py -q`: 1 passed
- `make style` / `make style_check`: passed
- `git diff --check`: passed

## Who can review?

Exporters (ONNX/OpenVINO): @echarlaix, @JingyaHuang, @michaelbenayoun, @IlyasMoutawwakil

### DIFF
--- a/optimum/utils/normalized_config.py
+++ b/optimum/utils/normalized_config.py
@@ -201,6 +201,11 @@ DistilBertNormalizedTextConfig = NormalizedTextConfig.with_args(
     num_attention_heads="n_heads",
     hidden_size="dim",
 )
+XLMNormalizedTextConfig = NormalizedTextConfig.with_args(
+    num_attention_heads="n_heads",
+    hidden_size="emb_dim",
+    num_layers="n_layers",
+)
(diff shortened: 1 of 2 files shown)


nicolasdmolina:
Bumping for visibility. Happy to make changes if there's anything blocking this.

PR #2495, opened 2026-09-15
Use huggingface_hub's httpx re-export instead of requests

tengomucho:
`optimum/exporters/tasks.py` imported `requests` at module level, but `setup.py`
only declared it under the `tests` extra. Since `huggingface_hub` v1 dropped
`requests`, this broke downstream projects (e.g. optimum-neuron had to pin
`requests` themselves).

The `except` clauses were dead code too: hub v1 raises `httpx.ConnectError`, not
`requests.exceptions.ConnectionError`, so the `snapshot_download` cache fallback
never fired when the Hub was unreachable. They now catch `httpx.ConnectError` and
`httpx.TimeoutException`, mirroring what the hub does internally in
`file_download.py`.

`httpx` is imported via `huggingface_hub.utils` so optimum follows the hub to
httpx2 in its v2.x. **That re-export landed in hub v1.31.0, so this bumps the
floor** — happy to switch to a plain `import httpx` + an explicit `httpx`
dependency instead (allows `huggingface_hub>=1.0.0`) if that floor is too
aggressive.

`raise ConnectionError(...)` in `determine_framework` now resolves to the builtin,
which `OfflineModeIsEnabled` subclasses too.

### Verified

In a venv with hub 1.31.0 / transformers 5.17.0 and **`requests` not installed**:

- Before: `ModuleNotFoundError: requests`; with `requests` installed, an
  unreachable Hub leaked a raw `httpx.ConnectError`.
- After: online happy path unchanged (`infer_task_from_model`, `get_model_files`);
  with `HF_HUB_OFFLINE=1` or an unresolvable `HF_ENDPOINT`, `infer_task_from_model`
  raises the intended `RuntimeError` and `get_model_files` falls back to the cache.

`tests/exporters/common/test_tasks_manager.py` was not run — it fails at collection
on `No module named 'optimum.exporters.onnx'`, unrelated to this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


IlyasMoutawwakil (approved):
okay for me but not sure how this impacts optimum-intel

PR #2496, opened 2026-09-16
Compatibility `NormalizedConfig.with_args` with python 3.14

AlexanderDokuchaev:
# What does this PR do?

Change `NormalizedConfig.with_args` method to return a staticmethod.

On Python 3.14, `functools.partial` became a descriptor, as resutlts it's can not be use as class atributes.

[What’s New in Python 3.14](https://docs.python.org/3/whatsnew/3.14.html):
```
functools.partial is now a method descriptor. Wrap it in staticmethod() if you want to preserve the old behavior.
```

Example of usage in optimum-intel:

https://github.com/huggingface/optimum-intel/blob/6bd036e3410e8016086045d2532271cbdb47aa3e/optimum/exporters/openvino/model_configs.py#L411-L414

Reproducer

```python
from optimum.utils import NormalizedConfig

class Config:
    CFG = NormalizedConfig.with_args(allow_new=True)

NormalizedConfig.with_args(allow_new=True)({})
Config.CFG({})
Config().CFG({})  # python3.14: TypeError: NormalizedConfig.__init__() got multiple values for argument 'allow_new'
```

ref: https://github.com/python/cpython/issues/121027

## Before submitting

## Who can review?

@echarlaix, @JingyaHuang, @michaelbenayoun, @IlyasMoutawwakil


optimum/utils/normalized_config.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, config: Union["PretrainedConfig", Dict], allow_new: bool = Fa
 
     @classmethod
     def with_args(cls, allow_new: bool = False, **kwargs) -> Callable[["PretrainedConfig"], "NormalizedConfig"]:
-        return functools.partial(cls, allow_new=allow_new, **kwargs)
+        return staticmethod(functools.partial(cls, allow_new=allow_new, **kwargs))
```
IlyasMoutawwakil: is this backward compatible ?

AlexanderDokuchaev: @IlyasMoutawwakil yes

IlyasMoutawwakil: okay ci is green so i guess it is

